PR #5, opened 2016-04-01
added description of the hand position api

wcarr:
We have better information now on what the proposed API is.


anssiko:
LGTM, thanks!

PR #7, opened 2016-05-05
import spec from mozvr/webvr-spec

cvan:


README.md
```diff
@@ -1,6 +1,5 @@
-WebVR Community Group
-=====================
+# WebVR Specification
 
-This is the repository for the [WebVR Community Group](https://www.w3.org/community/webvr/).
+This repository houses the [WebVR specification](https://mozvr.github.io/webvr-spec/), in [Bikeshed](https://github.com/tabatkins/bikeshed) format.
```
anssiko: https://mozvr.github.io/webvr-spec/ should be https://w3c.github.io/webvr/ after the migration.

index.bs
```diff
@@ -0,0 +1,588 @@
+<pre class="metadata">
+Title: WebVR
+Status: ED
+ED: https://github.com/mozvr/webvr-spec/
```
anssiko: Ditto.

index.bs
```diff
@@ -0,0 +1,588 @@
...
+Shortname: webvr
+Level: 1
+Editor: Vladimir Vukicevic, Mozilla https://mozilla.org/, vladimir@mozilla.com
+Editor: Brandon Jones, Google http://google.com/, bajones@google.com
+Editor: Kearwood Gilbert, Mozilla https://mozilla.org/, kgilbert@mozilla.com
+Editor: Chris Van Wiemeersch, Mozilla https://mozilla.org/, cvan@mozilla.com
+Abstract: This specification describes support for accessing virtual reality devices, including sensors and head-mounted displays on the Web.
+Repository: mozvr/webvr-spec
```
anssiko: w3c/webvr

index.bs
```diff
@@ -0,0 +1,588 @@
...
+Mailing List: web-vr-discuss@mozilla.org
```
anssiko: Do you prefer to have spec discussions via GH issues? That is what most active specs use. If so, you want to point to w3c/webvr/issues here.

We can use use https://github.com/w3c/github-notify-ml-config to sync the W3C mailing list https://lists.w3.org/Archives/Public/public-webvr/ with GH issues.

index.bs
```diff
@@ -0,0 +1,588 @@
...
+Mailing List Archives: https://mail.mozilla.org/pipermail/web-vr-discuss/
```
anssiko: https://lists.w3.org/Archives/Public/public-webvr/

anssiko:
Thanks! LGTM with some comments.

README.md
cvan: good catch

index.bs
cvan: I would prefer, yes. oh and awesome re: tool. @tojiro, wfy?

toji: Definitely works for me!

cvan:
I had originally wanted to move https://github.com/mozvr/webvr-spec to https://github.com/w3c/webvr so we could preserve the commit history, but evidently I must have admin privileges in w3c to do that. In the past, we've emailed GitHub support to transfer between orgs for us.

instead, I have a branch that I've locally rebased the [mozvr/webvr-spec](https://github.com/mozvr/webvr-spec) commits onto the `gh-pages` branch. is that fine to open as a PR? I know it'll have to be a `force push`. or I could cherry-pick the commits. what do y'all prefer?

anssiko:
I'd be fine either way. Ping @dontcallmedom for comments.

@cvan I guess your preference would be to preserve the commit history as is?

cvan:
Correct. I think I'll go with cherry picking. It's least destructive.

anssiko:
@cvan feel free to submit a new PR with cherry-picked commits when you're ready.

From the W3C process perspective retaining history is not required when migrating a spec to a CG to be used as a starting point for standards incubation. That said, it is definitely nice to have commit history if it is meaningful, since it allows the wider community to better understand the design decisions that have been made prior to migration.

PR #49, opened 2016-06-10
Use FrozenArray rather than sequences for attribute values

dontcallmedom:


index.bs
```diff
@@ -518,7 +518,7 @@ Depth of the play-area bounds in meters. The bounds are defined as an axis-align
 <pre class="idl">
 partial interface Navigator {
   Promise&lt;sequence&lt;VRDisplay&gt;&gt; getVRDisplays();
-  readonly attribute sequence&lt;VRDisplay&gt; activeVRDisplays;
+  readonly attribute FrozenArray&lt;VRDisplay&gt; activeVRDisplays;
```
cvan: what about [`VRDisplay#getLayers`](https://github.com/dontcallmedom/webvr/blob/82032ee/index.bs#L153) too?

dontcallmedom: Since `getLayers` is an operation, not an attribute, it can return a `sequence` (array of values returned by value).

kearwood:
This looks good to me.

PR #59, opened 2016-07-03
Added VRDisplayEvent type and activated/deactivated events

toji:
Also added bikeshed error output to the makefile for my own sanity, and expanded documentation on existing events.

The new events are intended to communicate to the page author that something has happened that suggests that your page may want to enter (or pause/exit) VR presentation mode. First pass includes triggering proximity sensors (mount/unmount) and page navigation. I don't expect that this will be the only thing necessary to make VR links completely workable, we probably need a meta tag to suggest that the page is capable of VR presentation it so that UAs don't flash up non-VR content, but it seems like a piece that will be necessary regardless. These will be especially necessary if the UAs start requiring a user gesture to trigger VR mode, which I see as increasingly likely. These events will be treated as a user gesture for the purposes of VR presentation.

This pull request is not necessarily intended to be merged verbatim, though if the community decides that's what they want then great. I mostly want to use it to foster discussion about the feature. There's several aspects of it that I still question:

* The way I've got the reason attached to all `VRDisplayEvents` in this patch suggests that there should be `VR_DISPLAY_CONNECTED` and `VR_DISPLAY_DISCONNECTED` reasons as well. But if we're going to do that, then it almost seems better to drop the connected and disconnected events all together and just have a `vrdisplaystatuschange` event?  Then again, `connected` does not seem like an appropriate point to enter VR presentation.
* Maybe `VR_DISPLAY_MOUNTED` and `VR_DISPLAY_UNMOUNTED` are too specific? Is a generic `VR_DISPLAY_HARDWARE_TRIGGERED` reason better/more flexible?
* For GearVR-style systems there's a distinction between connecting the phone to the harness and triggering the proximity sensor. Do we need events for both? Seems like simply connecting the phone isn't worth responding to in that specific case because you can't see anything until the proximity sensor triggers anyway.
* I feel like there should be other reasons for activation and deactivation, but haven't quite figured out what. Oculus and OpenVR both have a concept of the VR App losing and gaining focus (for when you're viewing the various system's home screens) but it's not clear to me if that's appropriate to expose to the web.

Anyway, I'm itching to implement this feature in the Chrome builds now that I've got some infrastructure in place for it. I don't want to put out a speculative API only to remove it again two builds later, though, so I wanted community input first.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -548,16 +548,79 @@ navigator.getVRDisplays().then(function (displays) {
 </pre>
 </div>
 
+## VRDisplayEventReason ## {#interface-vrdisplayeventreason}
+
+<pre class="idl">
+enum VRDisplayEventReason {
+  "navigation",
+  "mounted",
+  "unmounted",
+};
+</pre>
+
+### Reasons ### {#vrdisplayeventreason-codes}
+
+<dfn enum-value for="VRDisplayEventReason">navigation</dfn>
+The page has been navigated to from a context that allows this page to begin presenting immediately, such as from another site that was already in VR presentation mode.
+
+<dfn enum-value for="VRDisplayEventReason">mounted</dfn>
+The {{VRDisplay}} has detected that the user has put it on.
+
+<dfn enum-value for="VRDisplayEventReason">unmounted</dfn>
+The {{VRDisplay}} has detected that the user has taken it off.
+
+## VRDisplayEvent ## {#interface-vrdisplayevent}
+
+<pre class="idl">
+[Constructor(VRDisplayEventInit eventInitDict)]
+interface VRDisplayEvent : Event {
+  readonly attribute VRDisplay display;
+  readonly attribute VRDisplayEventReason? reason;
+};
+
+dictionary VRDisplayEventInit : EventInit {
+    required VRDisplay display;
+    VRDisplayEventReason reason;
+};
+</pre>
+
+### Attributes ### {#vrdisplayevent-attributes}
+
+<dfn attribute for="VRDisplayEvent">display</dfn>
+The {{VRDisplay}} associated with this event.
+
+<dfn attribute for="VRDisplayEvent">reason</dfn>
+{{VRDisplayEventReason}} describing why this event has has been fired.
+
 
 ## Window Interface extension ## {#interface-window}
 
 <pre class="idl">
 partial interface Window {
-  attribute EventHandler onvrdisplayconnected;
-  attribute EventHandler onvrdisplaydisconnected;
+  attribute EventHandler onvrdisplayconnect;
+  attribute EventHandler onvrdisplaydisconnect;
+  attribute EventHandler onvrdisplayactivate;
+  attribute EventHandler onvrdisplaydeactivate;
   attribute EventHandler onvrdisplaypresentchange;
 };
 </pre>
+
+User agents implementing this specification MUST provide the following new DOM events. The corresponding events must be of type {{VRDisplayEvent}} and must fire on the window object. Registration for and firing of the events must follow the usual behavior of DOM4 Events.
+
+<dfn event for="Window" id="window-vrdisplayconnect-event">onvrdisplayconnect</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been connected.
+
+<dfn event for="Window" id="window-vrdisplaydisconnect-event">onvrdisplaydisconnect</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been disconnected.
+
+<dfn event for="Window" id="window-vrdisplayactivate-event">onvrdisplayactivate</dfn>
+A user agent MAY dispatch this event type to indicate that something has occured which suggests the {{VRDisplay}} should be presented to. For example, if the {{VRDisplay}} is capable of detecting when the user has put it on, this event SHOULD fire when they do so with the reason "mounted".
+
+<dfn event for="Window" id="window-vrdisplaydeactivate-event">onvrdisplaydeactivate</dfn>
+A user agent MAY dispatch this event type to indicate that something has occured which suggests the {{VRDisplay}} should exit presentation. For example, if the {{VRDisplay}} is capable of detecting when the user has taken it off, this event SHOULD fire when they do so with the reason "unmounted".
+
+<dfn event for="Window" id="window-vrdisplaypresentchange-event">onvrdisplaypresentchange</dfn>
+A user agent MUST dispatch this event type to indicate that a {{VRDisplay}} has begun or ended VR presentation. This event should not fire on subsequent calls to {{requestPresent()}} after the {{VRDisplay}} has already begun VR presentation.
 
 
 ## Gamepad Interface extension ## {#interface-gamepad}
--- a/index.bs
+++ b/index.bs
@@ -579,8 +579,8 @@ interface VRDisplayEvent : Event {
 };
 
 dictionary VRDisplayEventInit : EventInit {
-    required VRDisplay display;
-    VRDisplayEventReason reason;
+  required VRDisplay display;
+  VRDisplayEventReason reason;
 };
 </pre>
 
--- a/index.html
+++ b/index.html
@@ -1504,11 +1504,21 @@ pre.idl.highlight { color: #708090; }
        <ol class="toc">
         <li><a href="#navigator-attributes"><span class="secno">2.9.1</span> <span class="content">Attributes</span></a>
        </ol>
-      <li><a href="#interface-window"><span class="secno">2.10</span> <span class="content">Window Interface extension</span></a>
       <li>
-       <a href="#interface-gamepad"><span class="secno">2.11</span> <span class="content">Gamepad Interface extension</span></a>
+       <a href="#interface-vrdisplayeventreason"><span class="secno">2.10</span> <span class="content">VRDisplayEventReason</span></a>
        <ol class="toc">
-        <li><a href="#gamepad-attributes"><span class="secno">2.11.1</span> <span class="content">Attributes</span></a>
+        <li><a href="#vrdisplayeventreason-codes"><span class="secno">2.10.1</span> <span class="content">Reasons</span></a>
+       </ol>
+      <li>
+       <a href="#interface-vrdisplayevent"><span class="secno">2.11</span> <span class="content">VRDisplayEvent</span></a>
+       <ol class="toc">
+        <li><a href="#vrdisplayevent-attributes"><span class="secno">2.11.1</span> <span class="content">Attributes</span></a>
+       </ol>
+      <li><a href="#interface-window"><span class="secno">2.12</span> <span class="content">Window Interface extension</span></a>
+      <li>
+       <a href="#interface-gamepad"><span class="secno">2.13</span> <span class="content">Gamepad Interface extension</span></a>
+       <ol class="toc">
+        <li><a href="#gamepad-attributes"><span class="secno">2.13.1</span> <span class="content">Attributes</span></a>
        </ol>
      </ol>
     <li><a href="#security"><span class="secno">3</span> <span class="content">Security Considerations</span></a>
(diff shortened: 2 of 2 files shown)


capnmidnight:
So you have this display connected/disconnected event. Is harnessing the phone to the GearVR not just connecting the GearVR HMD and disconnecting the Google Cardboard HMD?

judax:
@capnmidnight At least while in developer mode, you can launch a Gear VR App as a regular Android app but if the device is not docked (that´s the term the Oculus Mobile SDK uses) there is no head tracking information. Then you dock it and you start receiving head tracking events. A Gear VR WebVR implementation should consider using `onvrdisplayconnected` / `onvrdisplaydisconnected` in this scenario. Your page could have been loaded without the device being docked. At least only in developer mode, I insist. Not the same behaviour in production as the Gear VR Service/Oculus Home takes control over if you try to execute an app without the device being docked.

These events are referring to the proximity sensor when you mount the headset on your head (or put a finger in front of it ;)).

index.bs
```diff
@@ -547,17 +547,85 @@ navigator.getVRDisplays().then(function (displays) {
...
+
+<pre class="idl">
+[Constructor(VRDisplayEventInit eventInitDict)]
+interface VRDisplayEvent : Event {
+  readonly attribute VRDisplay display;
+  readonly attribute VRDisplayEventReason? reason;
+};
+
+dictionary VRDisplayEventInit : EventInit {
+    required VRDisplay display;
```
cvan: use 2 spaces to be consistent

index.bs
```diff
@@ -547,17 +547,85 @@ navigator.getVRDisplays().then(function (displays) {
...
+  const unsigned short VR_DISPLAY_UNMOUNTED = 3;
+};
+</pre>
+
+### Attributes ### {#vrdisplayeventreason-attributes}
+
+<dfn attribute for="VRDisplayEventReason">code</dfn>
+The code describing the reason for this event, described below.
+
+### Reason Codes ### {#vrdisplayeventreason-codes}
```
cvan: can we use strings instead of integer enums?

I like how the Permissions API handles `status`:

https://www.w3.org/TR/permissions/#idl-PermissionStatus-status
https://www.w3.org/TR/permissions/#h-examples

index.bs
```diff
@@ -547,17 +547,85 @@ navigator.getVRDisplays().then(function (displays) {
...
+{{VRDisplayEventReason}} describing why this event has has been fired.
+
 
 ## Window Interface extension ## {#interface-window}
 
 <pre class="idl">
 partial interface Window {
   attribute EventHandler onvrdisplayconnected;
   attribute EventHandler onvrdisplaydisconnected;
+  attribute EventHandler onvrdisplayactivate;
```
cvan: it seems inconsistent to not have these be in past tense

toji: Good point. I'll update to "activated" and "deactivated". (Side note: I actually had a bug in my code over the weekend where I had assumed that it was past tense in one spot but not the other. Consistency for the win!)

index.bs
toji: I'm good with that! It's also consistent with how we handle the left and right eyes, now that I think about it.

I had modeled this after [MediaError ](https://dev.w3.org/html5/spec-author-view/video.html#error-codes) because it was the first thing that came to mind when looking for a pre-existing API that handled a similar situation. I have no particular love for that API, though.

toji:
Updated the pull request based on @cvan's feedback. Also took a moment to fix all the bikeshed errors. (Probably should have been a separate pull request, but it made getting the linking of the new elements right easier.)

jcarpenter:
[Copy/pasting my response from the WebVR Slack]:

This seems to nicely capture what we discussed a few weeks ago. I was mostly concerned with the site-to-site story while in VR mode. IIUC, this proposal provides a key enabler for that (maintaining WebVR presentation mode when moving between WebVR sites). To be fully realized w/ content-side meta tag or some such (tbd). And provides enabler for developer adapting page layout when the user switches display modes (phone in hand, dropped into headset, or vice versa)

> With the Gear VR, you wouldn't ever have the change in connection state while in-page, because you don't run the same actual app inside versus outside of the harness. 

We should plan for a future where the same app can support both 2D and VR modes. :) Per Iker, it sounds like it's already here to some degree, w/ GearVR developer modes. I expect it to increase. Strikes me as a good thing for VR if any desktop/mobile app can deploy VR support w/o creating entirely new stand alone apps for distribution through X marketplace channels. Lots of neat flows and use cases :)

index.bs
```diff
@@ -547,17 +547,80 @@ navigator.getVRDisplays().then(function (displays) {
...
+User agents implementing this specification MUST provide the following new DOM events. The corresponding events must be of type {{VRDisplayEvent}} and must fire on the window object. Registration for and firing of the events must follow the usual behavior of DOM4 Events.
+
+<dfn attribute for="Window" id="window-vrdisplayconnected-event">onvrdisplayconnected</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been connected.
+
+<dfn attribute for="Window" id="window-vrdisplaydisconnected-event">onvrdisplaydisconnected</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been disconnected.
+
+<dfn attribute for="Window" id="window-vrdisplayactivate-event">onvrdisplayactivated</dfn>
+A user agent MAY dispatch this event type to indicate that something has occured which suggests the {{VRDisplay}} should be presented to. For example, if the {{VRDisplay}} is capable of detecting when the user has put it on, this event SHOULD fire when they do so with the reason "mounted".
```
borismus: Are activated/deactivated events intended to fire if you go into the system menu, for example? Vive, Oculus, etc all have this mode and mostly the game pauses. How do we handle this with WebVR?

toji:
That's an excellent question. I imagine that in many cases content wouldn't want to stop VR presentation all together, but it would probably be nice to know that focus has been lost so that the app can pause it's logic loop if the situation calls for it. Then again, it depends on the API. Both the Oculus API and OpenVR refer to this as gaining and losing focus. In OpenVR the VR app is still visible but dimmed out behind the system overlay, so it could be nice to continue rendering. In the Oculus case the app is completely hidden by the system menu, so the optimal would be to actually stop the render loop all together. Not sure how to best communicate that.

Another question: Would that be a different event (`vrdisplayfocused/blurred`) or just another variant of `activated\deactivated` with a `"focused"` and `"blurred"` reason enum? Kinda feels to me like the latter.

capnmidnight:
Should it be different from the focused/blurred events that already exist
for the window?

borismus:
@toji, having this be part of activated/deactivated SGTM.

A random naming inconsistency is that in general, events on the web are named in the present tense (eg. `click`, `touchmove`, `resize`), but WebVR varies a bit. `vrdisplaypresentchange` is present tense, but `vrdisplayconnected` and `vrdisplayactivated` for example, are past tense.

Should we rename them for consistency?

toji:
That's tricky. I've been looking through a bunch of other events and they're not consistent about tense. Especially the ones this API is modeled after. The connection events mimic the Gamepad API, which uses ["connected/disconnected"](https://w3c.github.io/gamepad/#dfn-gamepadconnected), and the `presentchange` event is modeled after the fullscreen APIs ["fullscreenchange"](https://developer.mozilla.org/en-US/docs/Web/Events/fullscreenchange) event.

If anything I'd be more willing to change the tense of the connection and activation events, because nobody is actively using them yet, but pretty much every WebVR 1.0 page to date will be listening for `vrdisplaypresentchange`

toji:
@capnmidnight: I've thought about it but I'm not totally convinced that it will work. Two cases come to mind.

1. On desktop, I use the system button on either a Rift or Vive. This pulls up the platform's dashboard, so the VR content is blurred. The desktop, however, maintains it's own concept of focus and according to it the browser is still focused. It would be inappropriate to issue a window blur event as a result.

2. In a setting like the Samsung Browser on GearVR where the browser is in VR as well you can imagine a situation where I want to bring up the browser UI (Back button, URL bar, etc) without explicitly exiting VR presentation. This is another case where the VR content is blurred, but the browser itself is not.

So I'm hesitant to tie VR state events to the existing browser events. It seems like in many cases they need to exist in parallel.

Unrelated to that: I've put out a Chrome build that exposes `vrdisplayactivated`, `vrdisplaydeactivated`, and `vrdisplaydisconnected`. (And TECHNICALLY `vrdisplayconnected`, though it can't actually fire in any situation where it would be useful just yet.) I'm still happy to change the tense of these events, but if I'm going to do so I think we should come to consensus fairly quickly so that devs don't become too attached to the current form.

capnmidnight:
Ok, that certainly makes sense.

On tense, I've always kept in mind

   - present-tense for the method that triggers the event (click, blur),
   but could also be used for events that occurred frequently and in
   repetition (mouse-move, wheel, key-down, key-up),
   - future tense for announcing intermittent, state-changing, cancelable
   events that were about to occur (closing, validating), and
   - past tense for intermittent, state-changing events that have occurred
   and are finished (closed, blurred).

I know not all systems are consistent with this. It's also somewhat of a
bikeshedding issue. So long as a not-awful naming scheme is used--you
wouldn't want it to be something like "EVENT_CODE_31"--then it's a very low
priority issue what the event is specifically called. I think there is
enough variability in all the different parts of the naming and what people
expect out of naming that you're never going to be able to come up with a
name that is easily guessable by the vast majority of people. A significant
group is always going to have to look up the name and rote memorize it.

On Mon, Jul 11, 2016 at 12:58 PM, Brandon Jones <notifications@github.com>
wrote:

> @capnmidnight <https://github.com/capnmidnight>: I've thought about it
> but I'm not totally convinced that it will work. Two cases come to mind.
>
>    1.
>
>    On desktop, I use the system button on either a Rift or Vive. This
>    pulls up the platform's dashboard, so the VR content is blurred. The
>    desktop, however, maintains it's own concept of focus and according to it
>    the browser is still focused. It would be inappropriate to issue a window
>    blur event as a result.
>    2.
>
>    In a setting like the Samsung Browser on GearVR where the browser is
>    in VR as well you can imagine a situation where I want to bring up the
>    browser UI (Back button, URL bar, etc) without explicitly exiting VR
>    presentation. This is another case where the VR content is blurred, but the
>    browser itself is not.
>
> So I'm hesitant to tie VR state events to the existing browser events. It
> seems like in many cases they need to exist in parallel.
>
> Unrelated to that: I've put out a Chrome build that exposes
> vrdisplayactivated, vrdisplaydeactivated, and vrdisplaydisconnected. (And
> TECHNICALLY vrdisplayconnected, though it can't actually fire in any
> situation where it would be useful just yet.) I'm still happy to change the
> tense of these events, but if I'm going to do so I think we should come to
> consensus fairly quickly so that devs don't become too attached to the
> current form.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/w3c/webvr/pull/59#issuecomment-231796362>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AASMPh-2m7_rmco4v6qtpu8oV6PrJ1mjks5qUnZAgaJpZM4JD6su>
> .
>



-- 
*Sean T. McBeth*
Lead VR Engineer
www.notiontheory.com

index.bs
kearwood: I would expect that a site would not technically stop presenting when entering a system menu.  If the content is obscured by browser UX or a system menu, perhaps the browser could simply throttle the VRDisplay's requestAnimationFrame callbacks.

kearwood:
This looks good to me, but I would be hesitant to bundle the blur / focus events with this.  I would expect content to continue presenting, but use a separate event to indicate that gameplay or physics should be paused/resumed if applicable.

Perhaps if the content is obscured by the system menu, the browser could optimize for this case by throttling the VRDisplay's requestAnimationFrame callbacks.  In some cases, the system menu wouldn't obscure the content when blurred, and it may be desirable for HMD orientation changes to be reflected while the system menu is displayed.

jbaicoianu:
Correct me if I'm wrong, but wouldn't the vrdeviceactivated / vrdevicedeactivated events make more sense to fire on the VRDisplay object, not window?  What about cases with multiple HMDs?  I know we get the "display" attribute on the event which we can use to differentiate, just seems like it would be a bit more natural to fire on the object which originated the event.

toji:
Okay, for the sake of keeping things moving here, I'm going to merge this. It seems like the functionality has broad support, and the exact details can be tweaked later. A few last notes:

* Events are all present tense now. `vrdisplayactivate`, `vrdisplayconnect`, etc. This does indeed seem more consistent with the majority of the web platform.
* Events will continue to fire on window. @jbaicoianu: If activate/deactivate were to fire on the display it would have to become derived from a DOM node. I don't think that's appropriate for this API.
* Bikeshed fixes were pulled into a separate commit and already merged for clarity and cleanliness.

PR #70, opened 2016-08-10
Specifies the type of matrix returned by sittingToStandingTransform

dmarcos:
@kearwood is this what you meant?


kearwood:
Yes, like that.

I've merged the patch, but not clear if we have bikeshed running as a hook or if that should be run manually to update the html files.

toji:
Bikeshed is a manual process right now (though an easy one: just run "make").

My github-hook-fu is weak, but I love the idea of updating the HTML automatically when changes are made to the .bs file. Ideally that should also be part of something like a Travis CI check so we can tell if pull requests will produce valid HTML. I won't have time to look at such a setup for a while, though. :(

Issue #18, opened 2016-06-01
Add a bikeshed watcher / Webhook


cvan:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [cvan](https://github.com/cvan)**
_Saturday Feb 20, 2016 at 00:41 GMT_
_Originally opened as https://github.com/MozVR/webvr-spec/issues/10_

---

Can borrow [this script](https://github.com/whatwg/loader/blob/master/local-watch.js) from https://github.com/whatwg/loader.

toji:
What's the state of this issue? Trying to clear out older issues in favor or focusing on immediate priorities.

cvan:
see PR #168

toji:
This should actually be fixed now. Key was updating to Travis' built-in deploy function and ensuring it deployed to gh-pages rather than master.

PR #74, opened 2016-08-21
requestPresent throws an Error when VRLayers source is null (fix #72)

### LINKED ISSUE #72: VRDisplay.requestPresent should throw when VRLayer.source is null
I propose that we continue to allow VRLayer.source to be a null value, with the expectation that future extensions to VRLayer will perhaps define new kinds of layers that do not require a source value.

Perhaps VRDisplay.requestPresent should throw when it does not have enough information to create a valid VRLayer.  This would imply that until there are additional attributes to define new layer types, that VRDisplay.requestPresent would throw when VRLayer.source is null

dmarcos:
Trying to capture https://github.com/w3c/webvr/issues/72 opened by @kearwood

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -140,9 +140,9 @@ interface VRDisplay : EventTarget {
   /**
    * Begin presenting to the VRDisplay. Must be called in response to a user gesture.
    * Repeat calls while already presenting will update the VRLayers being displayed.
-   * If the number of values in the leftBounds/rightBounds arrays is not 0 or 4 for any of the passed layers requestPresent throws an Error
+   * If the number of values in the leftBounds/rightBounds arrays is not 0 or 4 for any of the passed layers the promise is rejected
+   * If the source of any of the layers is not present (null), the promise is rejected.
    */
-  [Throws]
   Promise&lt;void&gt; requestPresent(sequence&lt;VRLayer&gt; layers);
 
   /**
--- a/index.html
+++ b/index.html
@@ -457,7 +457,7 @@
 		font-style: normal;
 	}
 	dt dfn code, code.idl {
-		font-size: inherit;
+		font-size: normal;
 	}
 	dfn var {
 		font-style: normal;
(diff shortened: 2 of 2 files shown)


kearwood:
LGTM, Thanks!

cvan:
Thanks. Can you run `make`, rebase, and push again? (we'll add an auotmated GitHub Webhook in #18.)

index.bs
```diff
@@ -140,7 +140,9 @@ interface VRDisplay : EventTarget {
   /**
    * Begin presenting to the VRDisplay. Must be called in response to a user gesture.
    * Repeat calls while already presenting will update the VRLayers being displayed.
+   * If the source of any of the layers is not present (null) requestPresent throws an Error
    */
+  [Throws]
   Promise&lt;void&gt; requestPresent(sequence&lt;VRLayer&gt; layers);
```
cvan: Does `exitPresent` not throw as well?

index.bs
```diff
@@ -140,7 +140,9 @@ interface VRDisplay : EventTarget {
   /**
    * Begin presenting to the VRDisplay. Must be called in response to a user gesture.
    * Repeat calls while already presenting will update the VRLayers being displayed.
+   * If the source of any of the layers is not present (null) requestPresent throws an Error
```
cvan: can you add a comma after ``null)` and period at the end?

toji:
Sorry, I skimmed over this previously, should have looked more closely. Why are we having this throw? Seems like the cleaner response is to reject the promise with an error message.

dmarcos:
@kearwood Should we throw exceptions or reject the promise?

kearwood:
I think we should reject the promise rather than throw exceptions

dmarcos:
@kearwood I addressed the comments and updated the bikeshed output

toji:
LGTM. Merging.

PR #78, opened 2016-09-01
Timestamp value is specific to each VRDisplay (fix #68)

### LINKED ISSUE #68: Need to describe time-base of VRPose.timestamp
In order to prevent incorrect usage of the  timestamp attribute of VRPose, the time-base should be described in more detail:

file:///Users/kearwood/dev/webvr-spec/index.html#vrpose-state-attributes

I suggest adding text to the end of timestamp's description:

The time-base of timestamp is only guaranteed to be consistent for one VRDisplay instance. The timestamp attribute should not be used to determine the chronological order of VRPose's from separate VRDisplay's or compared with timestamp values outside of VRPose.

dmarcos:
It tries to capture conversation in https://github.com/w3c/webvr/issues/68

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -356,7 +356,7 @@ interface VRPose {
 Monotonically increasing value that allows the author to determine if position
 state data been updated from the hardware. Since values are monotonically
 increasing, they can be compared to determine the ordering of updates, as newer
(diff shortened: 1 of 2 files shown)


dmarcos:
@kearwood ?

PR #79, opened 2016-09-01
Describe when submitFrame is not called on every frame (fix #69)

### LINKED ISSUE #69: Describe what happens when submitFrame is not called every frame.
Inspired by #58, there should be some documentation of what the expected results are when you don't call `submitFrame` at the headset's native framerate.

For the short term it's probably enough to note that the last submitted frame will be re-projected. Looking forward, though, to a time when WebGL is not the soul VR source some interesting use cases emerge: A `VRLayer` that described placing a video element at a certain point in space, for example, could update automatically without any Javascript interaction. (This type of layer could also take advantage of DRM content that can't be rendered normally in WebGL.) At that point it feels redundant to repeatedly call submitFrame, and waiting on it may introduce jank where none is necessary. The same could go for things like a static skybox or the "loading" spinner described in #58. Since there's no script interaction necessary taking JS out of the render loop is beneficial. 

I'm not sure what the right language is to describe it, but in that context it seems like submitFrame is primarily a mechanism to inform the browsers VR compositor that programatically generated contents (WebGL, 2D canvas?) have been updated. (And to continue allowing for greater control over content mirroring for those sources.)

Just some random thoughts. None of this is v1 material, aside from the aforementioned note about re-projection.

dmarcos:
This is my attempt at capturing the conversation happened in https://github.com/w3c/webvr/issues/69

More comfortable to read:

http://swimminglessonsformodernlife.com/webvr/#proc

@toji @kearwood I'm happy to iterate on this PR as much as necessary.


index.bs
```diff
@@ -643,6 +643,21 @@ partial interface Gamepad {
...
+
+* The application calls submitFrame consistently below the expected framerate to a degree that could make the user feel umconfortable. The User agent will show a semi transparent overlay on top of the application with a notification icon (an hourglass for instance) until the application improves performance. If the low framerate persists for a long time the User Agent will show an opaque overlay indicating that the application is too slow to be presented.
+
+* An application can stop calling submitFrame all toghether under several circumstances:
+    * A content exception or error has ocurred. In this case the User Agent will notify the user with an opaque overlay.
+    * The User Agent transitions to a different site that will eventually call submitFrame (the User Agent will statically determine if a site will call submitFrame by the presence of meta name="vr"). The User Agent will render an opaque overlay with a loading spinner to inform the user of the situation. Once the new site starts calling submitFrame a semi transparent overlay with the new URL will be shown to notify the user of the site transition.
```
DigiTec: It is not possible to both show an immersive experience and a comfortable overlay reliably. To show readable text it must be presented at depth and at very specific font sizes and contrasts. Most VR text is considered uncomfortable to the majority of VR users and is an area of a lot of investment, but this won't be resolved overnight. Once you've shown this readable text, at depth, you've potentially rendered it into a scene where you lack the depth information of the author content and so you end up with things at 2.5-3.5m displaying over top of things that are "closer" in the 0.5m to 2m range. For this reason, a more likely approach is a browser interstitial rather than a browser overlay. I'd like to avoid browser overlays showing up in specifications until someone gets 1000 users to give the thumbs up on an actual implementation (without first receiving 1000 thumbs downs ;-)

DigiTec: The author has to acquire the VRDisplay, requestPresent, and eventually call submitFrame. There are restrictions on when and how this can be done that would likely need to be encapsulated in other areas of the specification before being consumed here. I'd like to see those updates land before relying on them here. A meta tag is a "stronger" indication, but it shouldn't be the only indication. For instance, hooking up event handlers for onvr* should also count. We need to iterate here a bit on the right set of content analysis checks.

index.bs
```diff
@@ -643,6 +643,21 @@ partial interface Gamepad {
...
+    * The User Agent transitions to a different site that won't invoke requestPresent / submitFrame. The User Agent will render an opaque overlay indicating that there is not content available.
+    * The User Agent transitions to a new URL that returns an HTTP error. The User Agent will show an opaque overlay informing the user about the error.
```
DigiTec: I'd replace overlay with browser interstitial

index.bs
dmarcos: A metatag is a way to statically declare the intent of a website to call `submitFrame` (before running any JS) In the case of the absence of a tag a website might or not might call `submitFrame` or hook up event handlers for `onvr*`.  Would User Agents use a timeout before determining that a website won't present any VR content?

kearwood:
Thanks for capturing the details so well in the processing model text.  If users are uncomfortable with overlays as presented to a UA, perhaps a "browser interstitial" as described by @DigiTec could be sufficient for a feature complete, spec compliant browser; however, I believe the "transparent" or "opaque" effect of the display should be a UA specific UX decision.

There are many methods to be explored for maintaining user comfort while presenting non-opaque browser UX at depth.  Examples:
- Switch to monoscopic rendering of the WebVR scene
- Blur the WebVR content for a bokeh-like effect
- Tween to a 2d magic window view in another space, like in Minecraft VR.

Perhaps rather than "transparent", there is a more generic term that covers UX paradigms that we haven't invented yet that allow you to continue seeing the content, while not strictly requiring "transparency" in the traditional sense.

index.bs
kearwood: The meta tag is uniquely associated with the interstitial, as we need such a mechanism to statically determine that a site will present WebVR content in response to the onvrdisplayactivate event.  In this case either an opaque intersistial or continual rendering of the prior site would be necessary to avoid the flash of 2d content during the transition.  Perhaps the concept of "flash of 2d content" included in the text may help give context?

kearwood: In the context of a browser that includes a VR shell or other mechanism for viewing non-WebVR pages, we can not avoid the flash of 2d content in the HMD without either adding an arbitrary delay that would apply to even 2d pages or the static check with the meta tag that does not require waiting for the linked resources to load or javascript to execute.

DigiTec:
We should be strong in our description, but not precise. Maybe the right answer for now is something like:

The browser SHOULD produce a UI that the user can trust while navigating between content. The UI SHOULD indicate to the user the content that they are about to load and give them some indication of progress. The UI MAY provide a period where the user can cancel or reverse the navigation.

I agree that there are as of yet ways that we haven't invented. However, we aren't 2 years from implementation either and transparent and opaque overlays using our existing technologies would likely be terrible end user experiences. Even if they look good, they may still be spoofable or untrustable. We need something that the author can't SPOOF, that the user can TRUST and that the user finds comfortable.

index.bs
DigiTec: The meta tag sounds good in practice, but fails on modern content. The meta tag MAY be in an iframe, not in the top level page. We still need the page to attach to onvrdisplayactivate since they can't otherwise call requestPresent or submitFrame. 

To me the largest bother in the current text is that it implies a page can call submitFrame. Which it can't. During a navigation, the page MUST hook onvrdisplayactivate, then call getVRDevices, then call requestPresent IN THE CONTEXT of the onvrdisplayactivate (user intent, provided by the browser because it can detect the user navigated while in VR) and finally, after it does a bunch of other stuff, call submitFrame. We should reword the section to clarify what the target of a VR navigation would actually do and not imply that they can call submitFrame directly.

And a browser would still need some timeout, interstitial and user controls. (a combination of them all maybe) since it is always possible for the content not to achieve frame submission for any number of reasons. The annotations of meta tags can simply increase thee confidence that the browser should wait longer before recommending an alternative to a user. Today we say, "Hey this web page isn't responding", and thats annoying on some pages that are just taking a long time, but handy in others. Its a heuristic. So we should provide some minimum set of guarantees on time to wait and then allow browsers to make those longer via other means if they choose (such as a history entry for the page saying that it did indeed make it into VR on a previous navigation).

index.bs
```diff
@@ -643,6 +643,21 @@ partial interface Gamepad {
 <dfn attribute for="Gamepad" id="gamepad-getvrdisplays-attribute">displayId</dfn>
 Return the {{VRDisplay/displayId}} for the associated {{VRDisplay}}.
 
+# Processing Model # {#proc}
+
+## submitFrame ## {#submitframe-proc}
+
+The are are multiple situations where applications won't invoke submitFrame at the expected rate (the VRDisplay native framerate). The User Agent must intervene to protect the user from motion sickness or to consistently inform her about different situations like errors or navigation between pages. This is the breakdown of the different scenarios:
```
cvan: * typo: `are are`
* can you change `VRDisplay` to `{{VDisplay}}` and `submitFrame` to `{{VRDisplay}}.{{submitFrame()}}` so Bikeshed can properly link things up?
* `The User Agent must` to `The User Agent MUST`
* `or to consistently` to `or consistently`
* `different situations like` to `different situations such as`
* personally, I would change `about different situations like errors or navigation between pages. This is the breakdown of the different scenarios:` to `about such scenarious:`

index.bs
```diff
@@ -643,6 +643,21 @@ partial interface Gamepad {
...
+
+* The application calls submitFrame at the expected rate but a few frames are dropped. The User Agent will re-project the last submitted frame.
```
cvan: * `but` to `, but`

dmarcos:
Thanks for the feedback. I tried to incorporate the comments:
- I made the UI descriptions non prescriptive by removing words like ```overlay``` or ```semi transparent```
- I reworded the content navigation scenario based on @DigiTec suggestions.

http://swimminglessonsformodernlife.com/webvr/#submitframe-proc

I'm happy to keep iterating on this as much as necessary.

dmarcos:
@toji any feedback on this?

index.bs
```diff
@@ -643,6 +643,20 @@ partial interface Gamepad {
 <dfn attribute for="Gamepad" id="gamepad-getvrdisplays-attribute">displayId</dfn>
 Return the {{VRDisplay/displayId}} for the associated {{VRDisplay}}.
 
+# Processing Model # {#proc}
+
+## submitFrame ## {#submitframe-proc}
+
+The are multiple situations where applications won't invoke {{VRDisplay}}.{{submitFrame()}} at the expected rate (the {{VRDisplay}} native framerate). The User Agent MUST intervene to protect the user from motion sickness or consistently inform her about different situations such as errors or navigation between pages. This is the breakdown of the different scenarios:
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} at the expected rate, but only a few frames are dropped. The User Agent MUST re-project the last submitted frame.
```
toji: We're relying on the APIs to do the reprojection for us. At the moment they all do, but I don't know if that can be assumed going forward?

index.bs
```diff
@@ -643,6 +643,20 @@ partial interface Gamepad {
...
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} consistently below the expected framerate to a degree that could make the user feel umconfortable. The User agent SHOULD render a trusted interstitial on top of the application until the application improves performance. If the low framerate persists for a long time the User Agent MUST show a message indicating that the application is too slow to be presented.
```
toji: maybe

"The User agent SHOULD render a trusted interstitial **that obscures or minimizes the application** until the application improves performance." 

and 

"If the low framerate persists for a long time the User Agent MUST show a message indicating that the application is too slow to be presented **and terminate presentation**."?

index.bs
```diff
@@ -643,6 +643,20 @@ partial interface Gamepad {
 <dfn attribute for="Gamepad" id="gamepad-getvrdisplays-attribute">displayId</dfn>
 Return the {{VRDisplay/displayId}} for the associated {{VRDisplay}}.
 
+# Processing Model # {#proc}
+
+## submitFrame ## {#submitframe-proc}
+
+The are multiple situations where applications won't invoke {{VRDisplay}}.{{submitFrame()}} at the expected rate (the {{VRDisplay}} native framerate). The User Agent MUST intervene to protect the user from motion sickness or consistently inform her about different situations such as errors or navigation between pages. This is the breakdown of the different scenarios:
```
shaoboyan: 'The are'  A typo ?

index.bs
shaoboyan: 'umconfortable' -> 'unconfortable'

shaoboyan: Agree with **terminate presentation** and I think may be the message should guide User how to deal with it or give User a proper terminate way.

index.bs
dmarcos: @kearwood What do you think? Can we always rely on the VR SDKs to provide reprojection? Should this be excluded from the spec?

index.bs
```diff
@@ -683,6 +683,20 @@ partial interface Gamepad {
 <dfn attribute for="Gamepad" id="gamepad-getvrdisplays-attribute">displayId</dfn>
 Return the {{VRDisplay/displayId}} for the associated {{VRDisplay}}.
 
+# Processing Model # {#proc}
+
+## submitFrame ## {#submitframe-proc}
+
+There are multiple situations where applications won't invoke {{VRDisplay}}.{{submitFrame()}} at the expected rate (the {{VRDisplay}} native framerate). The User Agent MUST intervene to protect the user from motion sickness or consistently inform her about different situations such as errors or navigation between pages. This is the breakdown of the different scenarios:
```
cvan: what do you think of rewording this for conciseness?

> at the expected rate (the {{VRDisplay}} native framerate).

to

> at the expected native frame rate of the {{VRDisplay}}.

cvan: `such as` -> `, such as`

index.bs
```diff
@@ -683,6 +683,20 @@ partial interface Gamepad {
...
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} at the expected rate, but only a few frames are dropped. The User Agent MUST re-project the last submitted frame.
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} consistently below the expected framerate to a degree that could make the user feel ucnonfortable. The User agent SHOULD render a trusted interstitial that obscures or minimize the application until its performance improves. If the low framerate persists for a long time the User Agent MUST show a message indicating that the application is too slow to be presented and terminate presentation.
```
cvan: a few typos, and suggested rephrasing:

> The application calls {{VRDisplay}}.{{submitFrame()}} consistently below the expected framerate to a degree that could make the user feel ucnonfortable.

-> 

> The application calls {{VRDisplay}}.{{submitFrame()}} at a frame rate slower than is expected and likely to cause the user discomfort (e.g., nausea or physical sickness).

cvan: > The User agent SHOULD render a trusted interstitial that obscures or minimize the application until its performance improves. If the low framerate persists for a long time the User Agent MUST show a message indicating that the application is too slow to be presented and terminate presentation.

to

> The User Agent SHOULD render a trusted interstitial that obscures or minimizes the application until a consistent, acceptable frame rate is achieved. If the low frame rate persists for an extended time, the User Agent MAY show a message indicating that the application is too slow to be presented, and the User Agent MAY subsequently terminate its presentation.

index.bs
```diff
@@ -683,6 +683,20 @@ partial interface Gamepad {
...
+
+* An application can stop calling {{VRDisplay}}.{{submitFrame()}} all toghether under several circumstances:
```
cvan: typo: "toghether" should be "toghether"

index.bs
cvan: can we note some ranges or examples on what we expect to be native frame rates? I'm sure the info will be out of date in a matter of months, but it could be useful IMO. thoughts?

index.bs
```diff
@@ -683,6 +683,20 @@ partial interface Gamepad {
 <dfn attribute for="Gamepad" id="gamepad-getvrdisplays-attribute">displayId</dfn>
 Return the {{VRDisplay/displayId}} for the associated {{VRDisplay}}.
 
+# Processing Model # {#proc}
+
+## submitFrame ## {#submitframe-proc}
+
+There are multiple situations where applications won't invoke {{VRDisplay}}.{{submitFrame()}} at the expected rate (the {{VRDisplay}} native framerate). The User Agent MUST intervene to protect the user from motion sickness or consistently inform her about different situations such as errors or navigation between pages. This is the breakdown of the different scenarios:
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} at the expected rate, but only a few frames are dropped. The User Agent MUST re-project the last submitted frame.
```
cvan: what do we consider `a few`? can we quantify this number?

index.bs
```diff
@@ -683,6 +683,20 @@ partial interface Gamepad {
...
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} consistently below the expected framerate to a degree that could make the user feel ucnonfortable. The User agent SHOULD render a trusted interstitial that obscures or minimize the application until its performance improves. If the low framerate persists for a long time the User Agent MUST show a message indicating that the application is too slow to be presented and terminate presentation.
+
+* An application can stop calling {{VRDisplay}}.{{submitFrame()}} all toghether under several circumstances:
+    * A content exception or error has ocurred. In this case the User Agent SHOULD notify the user of the error with a trusted UI.
```
cvan: what do we consider a `content exception`? is that term defined in some other spec? (if so, we should link to it a glossary, for example.)

cvan: `In this case ` -> `In this case, `

cvan: where is the term `trusted UI` defined? the reason I ask because the [Permissions API](https://w3c.github.io/permissions/) is now the preferred mode of prompting users for (in-context) permission changes and notifications.

index.bs
```diff
@@ -683,6 +683,20 @@ partial interface Gamepad {
...
+    * The User Agent transitions to a different site that can eventually present content on the {{VRDisplay}}. The User Agent SHOULD produce a UI that the user can trust while navigating between content. The UI SHOULD indicate to the user they are about to load new content and give her some indication of progress. The UI MAY provide a period where the user can cancel or reverse the navigation. Once the new site starts calling {{VRDisplay}}.{{submitFrame()}} the UI MAY temporarly display the new URL to notify the user of the site transition. If after a period of time {{VRDisplay}}.{{submitFrame()}} is not invoked the User Agent SHOULD display a message indicating that there's no VR content available in the current URL. The time the User Agent will wait for {{VRDisplay}}.{{submitFrame()}} to be invoked will be determined by different factors that express intent of presenting VR like the presence of a metatag that statically declares the presence of VR content or an application attaching an event listener to {{onvrdisplayactivate}}.
```
cvan: do we want to say `The User Agent transitions` or perhaps `The User Agent navigates`?

cvan: I've tried following the spec terminology, but maybe you know better - is this pseduo-non-normative text (though I know it's not, it reads as such) **SHOULD** be using `SHOULD` or `MAY`?

index.bs
```diff
@@ -683,6 +683,20 @@ partial interface Gamepad {
...
+    * The User Agent transitions to a new URL that returns an HTTP error. The User Agent will SHOULD display a trusted UI informing the user about the error.
```
cvan: lots of useful terminology and context here: https://w3c.github.io/html/browsers.html#browsers

dmarcos:
Thanks everybody. I think I have addressed all the comments so far. Please let me know if you have any other feedback

cvan:
@kearwood @toji what do y'all think about this PR: shall we fix it up and merge - or shall we keep these matters as responsibilities/choices of the User Agent?

cvan:
@dmarcos can you rebase? @kearwood @toji, looking to merge this after.

dmarcos:
I rebased this

kearwood:
Perhaps we should incorporate concepts such as Oculus ASW (Asynchronous Space Warp) into the text.  Perhaps there is a way to describe this in a implementation-independent manner.

index.bs
```diff
@@ -716,6 +716,21 @@ partial interface Gamepad {
...
 
+# Processing Model # {#proc}
+
+## submitFrame ## {#submitframe-proc}
+
+There are multiple situations where applications won't invoke {{VRDisplay}}.{{submitFrame()}} at the {{VRDisplay}} native framerate. The User Agent MUST intervene to protect the user from motion sickness or consistently inform her about different situations, such as errors or navigation between pages. This is the breakdown of the different scenarios:
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} at the expected rate, but only a few frames are dropped. The User Agent MUST re-project the last submitted frame.
+
+* The application calls {{VRDisplay}}.{{submitFrame()}} below the expected framerate to a degree that could cause discomfort to the user (simulator sickness). The User agent SHOULD render a trusted interstitial that obscures or minimize the application until a consistent, acceptable frame rate is achieved. If the low framerate persists for an extended period of time the User Agent MUST communicate the user that the application is too slow to be presented and the User Agent MAY terminate presentation.
```
kearwood: Another consideration is that the "expected framerate" may change at any time with ASW.  Essentially, if performance drops too much then VRDisplay.RAF will start to tick at 45hz...

brianchirls: Are there any plans to [report the expected framerate](https://github.com/w3c/webvr/issues/37)?

toji:
Another wrinkle that's worth considering: I've heard from a couple of sources recently (one of which being John Carmack) that the recommended way of implementing a loading splash screen in native VR (especially on mobile) is to submit a single frame with a black background and then let it reproject while you do all the loading in the background. Trying to keep up with a draw loop at that stage is likely to be jittery, and just slows down the load. It would be nice if we could ensure this pattern (or something like it) works well for WebVR, but I'm not clear on it's feasibility cross-platform.

DigiTec:
I think the existence of features like ATW and ASW make our specing of what you should do when the author fails to submitFrame somewhat irrelevant. Right now we don't have enough information about the layers to determine if the Carmack loading screen advice should be applied, or if we should drop out. I think that if anything this comes down to user safety. But user safety is an issue even if the frames are NOT being submitted, since I can still walk someone off a cliff while hitting my 90fps. So the relevant bits are really around:

1. What provides new value to the user?
2. What provides new value to the author?

I think to the user, ATW and ASW can be a great experience regardless of whether or not frames are submitting. Also, for the author, debugging is a big thing. I often am broken into the debugger, with the Rift happily chugging away. Now, the Rift SDK might choose to pop a warning if ATW gets too far out of sync, but that again has nothing to do with frame rate and is instead due to how far out of view the reprojected frame has gotten.

For the author, I'm all for giving them more feedback on the performance of their page. Maybe some of these clauses would again be better written as "dev toolbar messages" or even events/properties that are exposed which allow the author to adjust their content for better performance.

Users will self select to remove themselves from VR if they are uncomfortable, regardless of the performance of the rendering. So I'm kind of wondering why we are trying to give them more reasons to get out of VR. Do we think that there are cases where users will sit inside of VR at low framerates and be subjected to a bad experience from the page that this prescriptive guidance is protecting against?

NellWaliczek:
I'm inclined to agree with @DigiTec on this one.  

As long as a user has a clear understanding of how to get out of an unpleasant experience, then we should let them decide for themselves what meets that bar.  Given the wide range of hardware and stabilization technologies, it seems preemptive to be overly prescriptive about best practices.

And as for the developer, diagnostics sound reasonable.  @DigiTec are you suggesting exposing the ideal framerate of a VRDisplay to allow developers to write their own diagnostics?

dmarcos:
@cvan Why has this PR been closed?

cvan:
I didn't close the PR, but this branch was based off the `gh-pages` branch which I deleted, so it automatically closed this PR. didn't know that would happen - my apologies.

unfortunately, GitHub doesn't let you change the upstream base branch. so there are two options:

1. I can re-add the `gh-pages` branch so we can merge this PR's branch (then merge to `master` and delete the `gh-pages` branch again).
2. you can open a new PR containing this PR's commits (`c85085a` `f356163` `905282d` `4a4f644` `06c2f9`), but base the PR off the `master` branch.

let me know which option you'd prefer. sorry for the inconvenience; I didn't realise this PR would get closed as a result.

PR #80, opened 2016-09-06
Update spec to WebVR "1.1"

toji:


index.bs
```diff
@@ -350,59 +355,51 @@ interface VRPose {
...
+Linear velocity of the sensor given in meters per second. MAY be NULL if the
+sensor is incapable of providing linear velocity. When not NULL MUST be a
 three-element array.
 
 <dfn attribute for="VRPose">linearAcceleration</dfn>
-Linear acceleration of the sensor at {{timestamp}} given in meters per second
-squared. MAY be NULL if the sensor is incapable of providing linear
-acceleration. When not NULL MUST be a three-element array.
+Linear acceleration of the sensor given in meters per second squared. MAY be
+NULL if the sensor is incapable of providing linear acceleration. When not NULL
```
DigiTec: Is capital NULL a thing? I would expect this to be null per JavaScript.

index.bs
```diff
@@ -453,6 +450,69 @@ function poseToMatrix (pose) {
...
+
+<dfn attribute for="VRFrameData">leftProjectionMatrix</dfn>
+A 4x4 matrix describing the projection to be used for the left eye's rendering,
+represented as a 16 element array in column major order. This value may be
+passed directly to WebGL's uniformMatrix4fv function. It is highly recommended
+that applications use this matrix without modification. Failure to use this
+projection matrix when rendering may cause the presented frame to be distorted
+or badly aligned, resulting in varying degrees of user discomfort.
+
+<dfn attribute for="VRFrameData">leftViewMatrix</dfn>
```
DigiTec: Do we want to describe this as the inverse of the model matrix for the eye? Since to ray-cast each eye I need to know this to know that I should take the inverse in order to create a camera to pass to Raycaster. We accurately described sittingToStandingTransform which is great, but I expect this to be less used than Raycasting. I might use the pose for Raycasting, but I suspect there will be some interesting applications for the eye model matrix and getting back to it and knowing how and that I should seems worth calling out.

This is duplicated with rightViewMatrix right now so maybe its worth having a reference from each of these to the "conversions" section so that we don't have to type it twice (this includes the sittingtoStandingTransform.

index.bs
dmarcos: I just picked three random specs:
https://www.w3.org/TR/gamepad/
https://www.w3.org/TR/webrtc/
https://www.w3.org/TR/workers/
I searched for null and found they use lowercase `null`

toji: I'm a C++ developer, damnit! (Oh, wait.... that's `nullptr` now. Doh!)

Fixed, will push soon.

index.bs
toji: Updated these descriptions, hope that makes it clearer?

index.bs
```diff
@@ -480,36 +480,16 @@ starts at 0 the first time {{getCurrentFrameData()}} is invoked for a given
 {{VRDisplay}}.
 
 <dfn attribute for="VRFrameData">leftProjectionMatrix</dfn>
-A 4x4 matrix describing the projection to be used for the left eye's rendering,
-represented as a 16 element array in column major order. This value may be
-passed directly to WebGL's uniformMatrix4fv function. It is highly recommended
-that applications use this matrix without modification. Failure to use this
-projection matrix when rendering may cause the presented frame to be distorted
-or badly aligned, resulting in varying degrees of user discomfort.
+A 4x4 matrix describing the projection to be used for the left eye's rendering, given as a 16 element array in column major order. This value may be passed directly to WebGL's uniformMatrix4fv function. It is highly recommended that applications use this matrix without modification. Failure to use this projection matrix when rendering may cause the presented frame to be distorted or badly aligned, resulting in varying degrees of user discomfort.
```
ssylvan: Saying "column major" identifies the storage, but not the convention of the matrix. OpenGL uses column vectors that get multiplied from the right, whereas D3D convention is (usually) row vectors that get multiplied from the left. The only difference is a transpose, but we should point out that these are OpenGL style matrices where vectors are columns that get multiplied from the right.

toji: I'm not sure I follow. We're not describing a matrix math libraries, and when passing into WebGL it's implied that it expects OpenGL style matrices. As such the storage convention seems like the only relevant thing to specify here. Or am I missing something?

ssylvan: Regardless of how the matrix is stored in memory, there's a difference in convention for how they get multiplied together and with vectors. The historical convention for OpenGL (although  you can do whatever  you want in your shader) is:
newVertex = Matrix * vertex

Whereas for D3D the historical convention is:
newVertex = vertex*Matrix

The former assumes that vectors are columns (and thus can only go on the right in order for the dimensions to make sense), the latter assumes that they're rows and can only go on the left. If we say nothing, people will likely assume the "traditional OpenGL style" (which is what we want), but might be worth calling out to be explicit. E.g. say 

It does have implications for order of multiplications, as  you might expect. E.g.

finalMatrix = projection * view * model; // "OpenGL" convention
finalMatrix = model * view * projection; // "D3D" convention

I call them D3D vs OpenGL convention but you can really go either way depending on how you write your shaders, so there's nothing enforced, just historical convention. The key thing is that the matrices have to be transposed if you switch from one convention to the next. I.e. the x component of the result of the product "vertex*M" is the dot product of vertex and the first *column* of M, whereas for "M*vertex" the x component is the dot product of the first *row* and the vertex, and so on.

index.bs
```diff
@@ -87,20 +87,20 @@ interface VRDisplay : EventTarget {
   readonly attribute DOMString displayName;
 
   /**
+   * Populates the passed VRFrameData with the information required to render
+   * the current frame.
+   */
+  void getCurrentFrameData(VRFrameData frameData);
```
toji: The more code and docs I write about this the more this function name feels overly verbose to me. Anyone object to simply `getFrameData()`?

Also, I think it should return a bool to indicate whether or not the frame data was populated.

DigiTec: Agree on both fronts. And I'm testing if responding to these mails will auto comment for me. :-)

toji: Yes, it does! :D

toji:
I think we need to continue iterating on the spec language, but it feels like the function changes are pretty well agreed on. Merging so that we can more easily facilitate community input. Further refinement of various areas should probably happen in targeted issues.

index.bs
RafaelCintron: @Toji, under what circumstances would getFrameData return false?

toji: With the OpenVR SDK (and I *think* Oculus, but I'm not 100% certain) there
are times where you can't poll a valid pose but do know that the hardware
is connected. This happens mostly during initialization of the associated
runtime or when the headset become completely occluded. It may also be
necessary to artificially prevent the frame data from being populated for
privacy or security concerns, which I would see as another valid reason for
returning false.

RafaelCintron: @toji Thanks, that makes sense.

PR #96, opened 2016-09-21
Make VRLayer.leftBound and VRLayer.rightBounds non-nullable (Fix #95)

### LINKED ISSUE #95: VRLayer.leftBounds and VRLayer.rightBounds should not allow null
There is ambiguity between setting a null value and an empty array to the leftBounds and rightBounds attributes.

We can remove the nullability and default the value to an empty array to remove the ambiguity.

dictionary VRLayer {
  VRSource? source = null;
  sequence<float>? leftBounds = [ ];
  sequence<float>? rightBounds = [ ];
};

Should become:

dictionary VRLayer {
  VRSource? source = null;
  sequence<float> leftBounds = [ ];
  sequence<float> rightBounds = [ ];
};

This change has already been implemented in Firefox and the change was not propagated to the spec here.

kearwood:
Rather than assigning null values, we should use the default, empty
array.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -265,8 +265,8 @@ typedef (HTMLCanvasElement or
 dictionary VRLayer {
   VRSource? source = null;
 
-  sequence&lt;float&gt;? leftBounds = [ ];
-  sequence&lt;float&gt;? rightBounds = [ ];
+  sequence&lt;float&gt; leftBounds = [ ];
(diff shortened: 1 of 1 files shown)


toji:
I thought this change had already been made in out earlier discussions about the bounds. LGTM!

Issue #111, opened 2016-10-13
Should `vrdisplayactivate` and `vrdisplayactivate` be consolidated to single event?


cvan:
Per https://w3c.github.io/webvr/#window-vrdisplayactivate-event:

> **`onvrdisplayactivate`***\* A user agent MAY dispatch this event type to indicate that something has occured which suggests the `VRDisplay` should be presented to. For example, if the `VRDisplay` is capable of detecting when the user has put it on, this event SHOULD fire when they do so with the reason "mounted".

Per https://w3c.github.io/webvr/#window-vrdisplaydeactivate-event:

> **`onvrdisplaydeactivate`** A user agent MAY dispatch this event type to indicate that something has occured which suggests the VRDisplay should exit presentation. For example, if the VRDisplay is capable of detecting when the user has taken it off, this event SHOULD fire when they do so with the reason "unmounted".

À la [`vrdisplaypresentchange`](https://w3c.github.io/webvr/#window-vrdisplaypresentchange-event),  perhaps there should instead be a single event called `vrdisplaymountchange`?

And, in the veins of #109, perhaps `vrdisplaymounterror` as well? Though, I'm not sure when that would be fire, to be honest. Maybe like if there's a VR display connected but it times out on mounting?

toji:
This issue was moved to immersive-web/webvr#7

Issue #86, opened 2016-09-20
Use HTML's "allowed to use" or permission delegation for allowvr


foolip:
https://html.spec.whatwg.org/multipage/embedded-content.html#allowed-to-use

The language in https://w3c.github.io/webvr/#allowvr-attribute doesn't say to traverse all frames to the top-level frame, so it's easy to circumvent by just creating an iframe with the attribute on it.

As with https://github.com/w3c/mediacapture-main/issues/268#issuecomment-231339319 I wonder if permission delegation might work here.

toji:
Thanks! I intended to bring up the idea of using permission delegation ([spec link](https://noncombatant.github.io/permission-delegation-api/) for easy reference) with the other implementors today. Seems like a more forward looking path than the `allowvr` attribute.

foolip:
Yes, I am secretly hoping that `allowfullscreen` will be the only attribute of this sort ever shipped, but of course if you think an `allowvr` still makes the most sense in the end, then that's fine.

toji:
My only concern is that it's not clear to me how committed the other implementing browser are to the permission delegation spec. If nobody strenuously objects, I'm happy to go that route.

raymeskhoury:
Just a heads up that we're looking at getting this added to the initial version of [Feature Policy](https://docs.google.com/document/d/1k0Ua-ZWlM_PsFCFdLMa8kaVTo32PeNZ4G7FFHqpFx4E/edit) (the permission delegation successor) right now.

ddorwin:
Feature Policy seems like the general direction. See also https://groups.google.com/a/chromium.org/d/msg/blink-dev/vM3RM1OQNFU/FcuMLjVfAgAJ for implementation issues with the iframe attribute.

In the meantime, please remove the `allowvr` text or mark it as under reconsideration / do not implement.

ddorwin:
https://github.com/WICG/feature-policy/issues/49 tracks this on the Feature Policy side.

cvan:
Just commenting so we don't forget that the spec needs to be updated to reflect the [Feature Policy](https://docs.google.com/document/d/1k0Ua-ZWlM_PsFCFdLMa8kaVTo32PeNZ4G7FFHqpFx4E/edit) dependency. This in scope for 1.1, 1.2?

Also, do we want to call the feature `vr` or `webvr`? I say `vr`, per `navigator.vr` (#116/#152). 

```html
// vr
<iframe enable="vr" src="example.html"></iframe>
[...]

Issue #30, opened 2016-06-01
Linking between VR web pages


cvan:
<a href="https://github.com/brianchirls"><img src="https://avatars.githubusercontent.com/u/492291?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [brianchirls](https://github.com/brianchirls)**
_Thursday Mar 17, 2016 at 15:09 GMT_
_Originally opened as https://github.com/MozVR/webvr-spec/issues/22_

---

I seem to recall some discussion of enabling linking directly from one VR web site to another without having to stop presenting. The spec as it is doesn't seem to handle that.

This is a potentially tricky case, so I understand wanting to hold off on implementation until more progress is made in the field. But has there been any discussion about what this might look like in an API? Are we confident that the API as it currently stands leaves room for that in the future?

cvan:
<a href="https://github.com/AlbertoElias"><img src="https://avatars.githubusercontent.com/u/913117?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [AlbertoElias](https://github.com/AlbertoElias)**
_Tuesday Mar 22, 2016 at 14:58 GMT_

----

From what I understood, once there has been a user interaction to start presenting in the HMD, you can move to different VR capable sites without having to take the headset off.

cvan:
<a href="https://github.com/brianchirls"><img src="https://avatars.githubusercontent.com/u/492291?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [brianchirls](https://github.com/brianchirls)**
_Tuesday Mar 22, 2016 at 17:34 GMT_

----

So how does the destination web page know that it's already presenting when it starts running?

And what happens if that destination web page doesn't know how to handle VR?

cvan:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [cvan](https://github.com/cvan)**
_Thursday Mar 24, 2016 at 01:48 GMT_

----

> So how does the destination web page know that it's already presenting when it starts running?

> And what happens if that destination web page doesn't know how to handle VR?

[...]

PR #116, opened 2016-10-14
Update navigator interface and event targets

toji:
_THIS IS A BACKWARDS-COMPATIBILITY BREAKING CHANGE_

TL;DR: `navigator.getVRDisplays()` is now `navigator.vr.getDisplays()`, connect and disconnect events now fire from the `navigator.vr`, and all other events fire from the `VRDisplay`. This is to enable use in workers or other non-main-thread contexts.

This updates the VR interfaces to remove dependencies on the `window` object and enable use of the API in contexts other than the main page thread, such as workers. (Note that the WorkerNavigator interface and worker exposure meta tags will be added in a follow-up PR.)

Because workers don't have access to the `window` object the API cannot depend on events that are fired from the `window` object if we want it to be useable in a worker or similar context. To facilitate this most events will be fired from the `VRDisplay` that the event is associated with, with the exception of events that describe display lifetime, such as `vrdisplayconnect` and `vrdisplaydisconnect`, which will now be fired off of the `vr` "namespace" object.

Also added a `vrdisplaynavigate` event that fires from the `vr` object which is functionally similar to `vrdisplayactivate` but is specifically for page-to-page navigation within VR. With the new changes it's not practical to have `vrdisplayactivate` handle navigation events since the event is intended to be fired early in the page lifetime and the page may not have had a chance to call `getDisplays` and install the appropriate listeners.


index.bs
```diff
@@ -643,12 +656,34 @@ The {{VRDisplay}} associated with this event.
...
+<dfn event for="VREventHandlers" id="vreventhandlers-vrdisplaynavigate-event">onvrdisplaynavigate</dfn>
+A user agent MAY dispatch this event type to indicate that the current page has been navigated to from a page that was actively presenting VR content to the {{VRDisplay}}. The current page can call {{requestPresent()}} in response to this event in order to stay in VR presentation mode.
+
+
+## VRDisplayEventHandlers Interface ## {#interface-vrdisplayeventhandlers}
+
+<pre class="idl">
+[NoInterfaceObject]
+interface VRDisplayEventHandlers {
   attribute EventHandler onvrdisplayactivate;
```
DigiTec: If we are making the breaking change, we should remove on* events and have people use addEventListener. on* should be used only for legacy cases that require its semantics.

toji: Good point. I'll update this in a bit.

index.bs
```diff
@@ -559,35 +562,45 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
-  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getVRDisplays();
-  readonly attribute FrozenArray&lt;VRDisplay&gt; activeVRDisplays;
-  readonly attribute boolean vrEnabled;
+interface VR {
+  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getDisplays();
+  readonly attribute FrozenArray&lt;VRDisplay&gt; activeDisplays;
+  readonly attribute boolean isEnabled;
 };
+
+VR implements EventTarget;
```
DigiTec: You can't implements EventTarget, this breaks the ability to do event listener logging by swapping out the EventTarget.prototype values since EventTarget as an implements is not part of your prototype chain.

index.bs
```diff
@@ -559,35 +562,45 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
-## Navigator Interface extension ## {#interface-navigator}
+## VR Interface ## {#interface-vr}
 
 <pre class="idl">
-partial interface Navigator {
-  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getVRDisplays();
-  readonly attribute FrozenArray&lt;VRDisplay&gt; activeVRDisplays;
-  readonly attribute boolean vrEnabled;
+interface VR {
+  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getDisplays();
```
DigiTec: Consider: getDevices, but only if we have a good reason for it. Same for active* if we change the getter here.

index.bs
```diff
@@ -168,6 +168,9 @@ interface VRDisplay : EventTarget {
    */
   void submitFrame();
 };
+
+VRDisplay implements EventTarget;
+VRDisplay implements VRDisplayEventHandlers;
```
DigiTec: Is this the direction everyone is heading? We preferred partials over implements UNLESS multiple things would use the implements. At this time, only VR would use one set of handlers and only VRDisplay would use the other set. So why implements instead of partial or simply adidng it on the interface directly in this case since we are in a single spec? Curious what others think here.

index.bs
toji: Interesting, didn't know that. This syntax was picked up from the WebBluetooth spec, so they've probably got it wrong. Fix incoming.

index.bs
toji: Again, this was picked up from WebBluetooth. Not sure what the reasoning was there. Worth asking them.

index.bs
```diff
@@ -643,12 +656,34 @@ The {{VRDisplay}} associated with this event.
...
+User agents implementing this specification MUST provide the following new DOM events. The corresponding events must be of type {{VRDisplayEvent}} and must fire on the {{vr}} object. Registration for and firing of the events must follow the usual behavior of DOM4 Events.
+
+<dfn event for="VREventHandlers" id="vreventhandlers-vrdisplayconnect-event">onvrdisplayconnect</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been connected.
+
+<dfn event for="VREventHandlers" id="vreventhandlers-vrdisplaydisconnect-event">onvrdisplaydisconnect</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been disconnected.
+
+<dfn event for="VREventHandlers" id="vreventhandlers-vrdisplaynavigate-event">onvrdisplaynavigate</dfn>
+A user agent MAY dispatch this event type to indicate that the current page has been navigated to from a page that was actively presenting VR content to the {{VRDisplay}}. The current page can call {{requestPresent()}} in response to this event in order to stay in VR presentation mode.
```
DigiTec: We need to know when to fire this event since this will be one of the indications that we should stay in VR presentation mode.

toji: Right, we definitely need to flesh out the order of operations on all the events really. I just wanted to make sure this was declared to head off questions about "wait, if `activate` is on the display then how does navigation work?"

index.bs
```diff
@@ -643,12 +656,34 @@ The {{VRDisplay}} associated with this event.
...
   attribute EventHandler onvrdisplayconnect;
   attribute EventHandler onvrdisplaydisconnect;
+  attribute EventHandler onvrdisplaynavigate;
+};
+</pre>
+
+User agents implementing this specification MUST provide the following new DOM events. The corresponding events must be of type {{VRDisplayEvent}} and must fire on the {{vr}} object. Registration for and firing of the events must follow the usual behavior of DOM4 Events.
+
+<dfn event for="VREventHandlers" id="vreventhandlers-vrdisplayconnect-event">onvrdisplayconnect</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been connected.
```
DigiTec: We need to define more what this means. We've been having issues with the expectations around this event. For instance, a Gear VR is technically always connected, but there is also the USB drop in connection and the mounted indicators. We don't want people to wait to hear a `vrdisplayconnect` event before they `requestPresent` or otherwise try to use VR.

Now that we have Rift, Gear and DayDream, let's define what we think this means for all 3. Rift is kind of easy since maybe it means you've either plugged it in via USB or that you've told the browser you want to be in VR available mode (this is the toggle I've been proposing for the UI so someone could say, please don't enable VR experiences right now I don't want a page to automatically use my Rift).

index.bs
DigiTec: Oh yeah, a lot of people get this wrong. In general, the pro of EventTarget is that it goes into your prototype chain and becomes a single source for the functions add/remove/dispatch, but the con is that it goes into your prototype chain so you have to design it in.

If you do implements it, then you get your own "copy" of those functions and people have to crawl the entire type system if they want to do any sort of event listener logging via script. This is how IE 11 was implemented and then I fixed this in Edge.

index.bs
toji: Completely agreed, but maybe as a different PR? (Since we're not changing the definition here, just where it fires from)

cvan:
here's some proposed pseudo-code I've written for the common use cases of the APIs. thoughts?

## Detection

```js
// Exposes whether a VR display is available and accessible (e.g., permissions) on the user's system.

let vrButton = document.querySelector('#vr-button');

vrButton.dataset.available = false;

// This would replace `navigator.vrEnabled`.
navigator.vr.getAvailability().then(isAvailable => {
  vrButton.dataset.available = isAvailable;
});

navigator.vr.addEventListener('availabilitychanged', e => {
  vrButton.dataset.available = e.value;
});
```

<hr>

## Querying for displays

```js
// This would replace `navigator.getVRDisplays()`.
navigator.vr.requestDisplays().then(displays => {
  console.log(displays);
  // >> [VRDisplay, VRDisplay]

  console.log(displays[0]);
  // >> VRDisplay {displayId: 1, displayName: "Oculus Rift", isConnected: true…}

  console.log(displays[1]);
  // >> VRDisplay {displayId: 2, displayName: "HTC Vive", isConnected: true…}
}).catch(err => {
  console.error(err);
});

navigator.vr.requestDisplays({
  filters: [{displayName: 'Oculus Rift'}]
}).then(displays => {
  console.log(displays);
  // >> [VRDisplay]

  console.log(displays[0]);
  // >> VRDisplay {displayId: 1, displayName: "Oculus Rift", isConnected: true…}
});

navigator.vr.requestDisplays({
  filters: [{displayNamePrefix: 'Oculus'}]
}).then(displays => {
  console.log(displays);
  // >> [VRDisplay]

  console.log(displays[0]);
  // >> VRDisplay {displayId: 1, displayName: "Oculus Rift", isConnected: true…}
});

navigator.vr.requestDisplays({
  filters: [
    {displayName: 'Oculus Rift'},
    {displayName: 'HTC Vive'}
  ]
}).then(displays => {
  console.log(displays);
  // >> [VRDisplay]

  console.log(displays[0]);
  // >> VRDisplay {displayId: 1, displayName: "Oculus Rift", isConnected: true…}

  console.log(display[1]);
  // >> VRDisplay {displayId: 2, displayName: "HTC Vive", isConnected: true…}
});
```

<hr>

## Presenting to displays

```js
// Present to a connected HTC Vive.
navigator.vr.requestDisplays({
  filters: [{displayNamePrefix: 'HTC Vive'}]
}).then(displays => {
  if (!displays.length) {
    return;
  }

  return displays[0].requestPresent([
    {source: canvas}
  ]).then(
    render(display)
  );
});


// Very theoretical use case.
// Present to multiple connected headsets
// if we have both an Oculus Rift and HTC Vive.
navigator.vr.requestDisplays({
  filters: [
    {displayNamePrefix: 'Oculus Rift'},
    {displayNamePrefix: 'HTC Vive'}
  ]
}).then(displays => {
  return Promise.all(displays.map(displays => {
    return display.requestPresent([
      {source: canvas}
    ]).then(
      render(display)
    );
  });
}).catch(err => {
  console.error(err);
});


function render (display) {
  var onVRFrame = () => {
    display.requestAnimationFrame(onVRFrame);
    display.getFrameData(frameData);
    // …
    display.submitFrame();
  };
  return () => {
    display.requestAnimationFrame(onVRFrame);
  };
}
```

<hr>

## Link traversal

If display(s) is presenting, after the page is unloaded, the browser sets `navigator.vr.referringDisplay` (or `navigator.vr.referringDisplays`).

The next page can look at `navigator.vr.referringDisplay` and automatically present (without any user-gesture requirement):

```js
if (navigator.vr.referringDisplay) {
  presentOrRender(navigator.vr.referringDisplay);
} else {
  navigator.vr.requestDisplays().then(displays => {
    if (!displays.length) {
      return Promise.reject();
    }

    return presentOrRender(displays[0]);
  });
}

function presentOrRender (display) {
  if (display.isPresenting) {
    return render(display);
  }
  return display.requestPresent([
    {source: canvas}
  ]).then(
    render(display)
  );
}
```

AlbertoElias:
What would the benefit be for the API itself to support filtering in `#getDisplays`, wouldn't that encourage making content only for certain devices?

toji:
I'm not particularly in favor of filtering based on the name. For one, that encourages needlessly exclusionary pages and two it relies on browser all exposing the name the same way AND never changing it. (Even the underlying SDK could change how it's reported and possibly break things.) True, you can already do this if you really want by just looping through the detected headsets and selecting one with name == "...", but that's an antipattern that I'd rather not see baked into the API.

I can see an argument for filtering on capabilities. Querying only 6DoF devices or only those with motion controls seems valuable. But again, it encourages exclusionary uses when I'd rather see developers focusing on progressive enhancements 99% of the time. (If you build a Tilt Brush clone, sure, require motion controllers, but don't block me from your 360 photo viewer because all I have is cardboard.)

Given the shape of the API filtering could be added in later without much trouble, so I'd vote to hold off on that till v2.

I'm curious about your intentions for `getAvailability`. Seems like it's got the potential to allow feature testing without spinning up the entire API, but that sort of depends on the API having an option for that. OpenVR has `vr::VR_IsHmdPresent()` which would work well here, but it can give false positives. On the Oculus side you could call `ovr_Initialize` without `ovr_Create`, but that seems less lightweight and I'm not sure when that fails. Certainly I would only expect the Oculus runtime to be installed on machines that are very likely to have a Rift attached, but again there's a possibility for false positives. As such, we'd have to be very careful about how we specced such a function to indicate that returning true only indicates that the system MAY have an HMD available.

cvan:
thanks, good feedback. I agree the filtering is probably out of scope. It just seems like a common thing to want to get a Vive if it's present. what are your thoughts on the `navigator.vr.referringDisplay`? and I know we just went through a rename, but do you think it's a good idea to keep calling these `displays`? the API description and spec mentions displays _and_ sensors. it seems like `devices` is a better name, IMO - at least that's my current thinking. thoughts?

toji:
Hm... As I recall it was Mozilla that wanted the "display" name in the first place. ;)

I'm not really invested in either name, but I'm not excited to churn verbiage simply for the sake of churn. I also think there's something to be said for making it explicit that this API is intended to deal with devices that display content. That nicely separates it from things like controllers, which are arguable a "VRDevice" but definitely not a "VRDisplay".

index.bs
```diff
@@ -642,45 +681,6 @@ The {{VRDisplay}} associated with this event.
...
 {{VRDisplayEventReason}} describing why this event has has been fired.
 
-
-## Window Interface extension ## {#interface-window}
-
-<pre class="idl">
-partial interface Window {
-  attribute EventHandler onvrdisplayconnect;
-  attribute EventHandler onvrdisplaydisconnect;
-  attribute EventHandler onvrdisplayactivate;
```
NellWaliczek: Perhaps this is a silly question... but where did this portion of the idl get moved to?  I thought the events were going to be added to VRDisplay and while I see the comments moved above, but I don't see the EventHandlers themselves.  Am I misunderstanding how this works?

index.bs
```diff
@@ -559,35 +577,56 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
-
-## Navigator Interface extension ## {#interface-navigator}
+## VR Interface ## {#interface-vr}
 
 <pre class="idl">
-partial interface Navigator {
-  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getVRDisplays();
-  readonly attribute FrozenArray&lt;VRDisplay&gt; activeVRDisplays;
-  readonly attribute boolean vrEnabled;
+interface VR : EventTarget{
```
toji: Let's remove the IDL.

index.bs
```diff
@@ -559,35 +577,56 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
+<dfn attribute for="VR" id="vr-isenabled-attribute">isEnabled</dfn>
+The {{isEnabled}} attribute's getter must return true if the context object is allowed to use the feature indicated by attribute name {{allowvr}} and VR is supported, and false otherwise.
+
+### Events ### {#vr-events}
+
+User agents implementing this specification MUST provide the following new DOM events. The corresponding events must be of type {{VRDisplayEvent}} and must fire on the {{vr}} object. Registration for and firing of the events must follow the usual behavior of DOM4 Events.
 
-<dfn attribute for="Navigator" id="navigator-vrenabled-attribute">vrEnabled</dfn>
-The {{vrEnabled}} attribute's getter must return true if the context object is allowed to use the feature indicated by attribute name {{allowvr}} and VR is supported, and false otherwise.
+<dfn event for="VR" id="vrdisplayconnect-event">vrdisplayconnect</dfn>
```
toji: Good call on dropping the prefix.

As for the tense, I recall that we've discussed that before and found that the web at large is inconsistent. I would LOVE to be pointed at some proper codified guidance that says one or the other. I have no personal preference, as long as the spec is at least internally consistent.

index.bs
cvan: sgtm! opened #140

index.bs
annevk: https://w3ctag.github.io/design-principles/#casing-rules by @domenic covers some but not tense. I think past tense is generally not used for events. We use past sense for promise-state attributes.

domenic:
> If we are making the breaking change, we should remove on* events and have people use addEventListener. on* should be used only for legacy cases that require its semantics.

No!! Event handler IDL attributes are here to stay and should be on all new interfaces as well as old ones. https://github.com/w3ctag/design-principles/issues/29

> We preferred partials over implements UNLESS multiple things would use the implements.

Agreed.

index.bs
domenic: Looking through https://html.spec.whatwg.org/#events-2 it looks like present tense is correct (with some legacy exceptions as noted). I'll add that to the design principles list.

index.bs
```diff
@@ -559,35 +579,56 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
-<dfn attribute for="Navigator" id="navigator-vrenabled-attribute">vrEnabled</dfn>
-The {{vrEnabled}} attribute's getter must return true if the context object is allowed to use the feature indicated by attribute name {{allowvr}} and VR is supported, and false otherwise.
+<dfn event for="VR" id="vr-displayconnect-event">displayconnect</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been connected.
+
+<dfn event for="VR" id="vr-displaydisconnect-event">displaydisconnect</dfn>
+A user agent MAY dispatch this event type to indicate that a {{VRDisplay}} has been disconnected.
+
+<dfn event for="VR" id="vr-navigate-event">navigate</dfn>
+A user agent MAY dispatch this event type to indicate that the current page has been navigated to from a page that was actively presenting VR content to the {{VRDisplay}}. The current page can call {{requestPresent()}} in response to this event in order to stay in VR presentation mode.
```
RafaelCintron: "... has been navigated **to from** a page ...." The words to and from being next to each other confuses me.  I presume what was meant to go here is " ... has been navigated **from** a page ... "

index.bs
RafaelCintron: As at @domenic suggests, the way to do this is to add "attribute EventHandlers".  The HTML 5 spec does this for events like onload, onplaying and many others.

index.bs
cvan: good call on the phrasing.

borrowing some verbiage from the [WHATWG's HTML Standard](https://html.spec.whatwg.org/multipage/browsers.html#windows), perhaps we could change the text to read like this:

```
… the current browsing context is navigated from the initial document to another …
```

domenic: It would be good to get @annevk's opinion on anything that ties into navigation. From a brief skim this seems redundant with the Window's load event, but I'm sure I'm missing something.

annevk: "the current browsing context is navigated from the initial document to another" would be way too unspecific. You basically need to add a hook in HTML's processing model somewhere that defines exactly when this event is dispatched, provided it is indeed needed.

That is, you need to write a patch for the HTML Standard to enable this kind of thing.

MortimerGoro:
I suggest adding a new vrdisplay event type, called something like `change `or `parameterschange` or `configchange `. This event would fire when users reconfigure some HDM params such as: offset between eyes, chaperone size or other configurable params.

What do you think?

toji:
Just rebased this patch. I think at this point it's non-controversial, though we'll probably need to make a few follow up tweaks. I do want to get this merged soon because we will be discussing several other significant changes to be built on top of it. I'll leave it up for the day, but barring any significant concerns being raised I plan on merging this tonight.

cvan:
@toji I did a bunch of testing, and I have some proposed changes to this interface. I'll show you a mock interface I made for `navigator.vr` that essentially polyfills the existing WebVR 1.1 API. Noticed a few issues, inconsistencies with the proposed changes. I think my proposed changes handle everything right now, without making any more significant changes.

toji:
That would be great, @cvan! Please do send them over. I've definitely got some changes to propose as well beyond what this patch does, but my proposals are more disruptive so I wanted to bring them up separately.

RafaelCintron:
@toji , this change LGTM. Separately, we are discussing whether the current set of events should be collapsed as part of #111 but I think we should deal with that as part of a separate PR.

cvan:
In my extension that polyfills `navigator.vr` I got the events nicely consolidated (especially with Chromium, since it fires more).

It's probably a bit hard to follow without all the context. We've talked about this a lot in various issues, including #30, this PR, and at the W3C Workshop and in meetings.

Let me know if this helps convey what I was thinking. Pay attention to `navigator.getAvailability()` (which would respect the `Feature-Policy` - the alternative to `<iframe allowvr>`: `<iframe enable="vr">`) and `navigator.vr.addEventListener('display…', function (evt) { })` and `navigator.vr.referringDisplay`.

I researched Web Bluetooth, WebUSB, Presentation API, Web Audio, Web MIDI, among others, and this felt the most right to me. Open to suggestions. This will probably be easier to go through over our meeting next week, but I'd love to get people thinking about this now.

* https://github.com/WebVRRocks/extension/blob/events/test.html#L326-L360
* https://gist.github.com/cvan/9c0a2e6e1a6839816128aeb461eb3184#file-sample-usage-js
* https://gist.github.com/cvan/9c0a2e6e1a6839816128aeb461eb3184#file-navigator-vr-js-L71

toji:
@cvan: I'm assuming the changes that you talked about are what you showed in the IDL you posted to #179? As I mentioned in that issue, I think the concepts you're presenting have merit but should probably be broken out into individual issues/proposals/pull requests. In that light, it seems like there's value to merging this PR as-is, even if we want to make some adjustments on top of it, simply so that the delta of future updates is smaller and more manageable?

index.bs
```diff
@@ -559,31 +583,56 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
+## VR Interface ## {#interface-vr}
 
 <pre class="idl">
-partial interface Navigator {
-  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getVRDisplays();
-  readonly attribute FrozenArray&lt;VRDisplay&gt; activeVRDisplays;
+interface VR : EventTarget {
+  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getDisplays();
+  readonly attribute FrozenArray&lt;VRDisplay&gt; activeDisplays;
+
```
cvan: I was suggesting adding `navigator.vr.getAvailability` (as a replacement for the `navigator.vrEnabled` property, which should obviously be removed).

This would respect the [Feature Policy](https://docs.google.com/document/d/1k0Ua-ZWlM_PsFCFdLMa8kaVTo32PeNZ4G7FFHqpFx4E/edit#heading=h.4yubgixv5l6b) (which supports `<iframe enable="vr" src="https://example.com/scene.html"></iframe>` – see issue #86).

Besides the aforementioned `<iframe>` case, there are cases (e.g., browser extension) when a UA may not want to (or cannot attempt to) expose VR capabilities. (A VR display could still be connected, but the page may not be privileged to use it.)

Some similar methods in other web APIs:
* [`new PresentationRequest([urls]).getAvailability().then(…)`](https://www.w3.org/TR/presentation-api/#dom-presentationrequest-getavailability)
* [`navigator.bluetooth.getAvailability().then(…)`](https://webbluetoothcg.github.io/web-bluetooth/#availability)
* [`navigator.usb.requestDevice({filters: [], optionalServices: [], acceptAllDevices: true})`.then(…)](https://wicg.github.io/webusb/#dom-usb-requestdevice)

Alternatively, instead of a `Promise<boolean> getAvailability()`, we could have the following:

```idl
    readonly attribute VRAvailability availability;
    attribute EventHandler onavailabilitychange;
    enum VRAvailability {
        'available',
        'unavailable',
        'unknown'
    };
```

index.bs
```diff
@@ -559,31 +583,56 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
-## Navigator Interface extension ## {#interface-navigator}
+## VR Interface ## {#interface-vr}
 
 <pre class="idl">
-partial interface Navigator {
-  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getVRDisplays();
-  readonly attribute FrozenArray&lt;VRDisplay&gt; activeVRDisplays;
+interface VR : EventTarget {
+  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getDisplays();
+  readonly attribute FrozenArray&lt;VRDisplay&gt; activeDisplays;
```
cvan: This doesn't make sense to me. I'm unclear of the use case, and it's ambiguous.

See my proposed `navigator.vr.referringDisplay(s)` idea: https://gist.github.com/cvan/9c0a2e6e1a6839816128aeb461eb3184#file-sample-usage-js-L21-L32

<hr>

À la [`navigator.languages`](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages)), I suggest having introducing a **`navigator.vr.preferredDisplays`, an array of display name strings for user-preferred VR headsets**.

If we think of VR devices/displays à la languages/locales, you can imagine a user flow something like this:

1. When a user first plugs in a headset, the UA remembers that headset name in some internal persisted list of "known VR displays" in that user's UA profile.
2. A user opens the UA's "Preferences" to add/remove/rearrange the preferred headsets to use when display WebVR content.
3. When a requests _any_ page, the browser now starts sending a `Accept-VR-Display` **request header** (e.g., `Accept-VR-Display: HTC Vive, Oculus Rift`)  – à la the [`Accept-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation#The_Accept-Language_header) request header [e.g., `Accept-Language: en-US, fr; q=0.7`]).
4. If a page's **response header** is `Content-VR-Display: Gear VR, HTC Vive` (or a `<meta>` tag: `<meta http-equiv="content-vr-display" content="available HTC Vive, Oculus Rift">` – [or [CSP-Style](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP): `<meta http-equiv="Content-VR-Policy" content="default-display 'HTC Vive'; supported-displays 'HTC Vive', 'Oculus Rift'; prefetch-navigation-to 'https://*.neocities.org'">`]), then the UA knows which displays are supported by this WebVR page (or scenes within the page, if we're going to get complicated).
5. The UA now knows that when the page loads, the user may want to *enter VR* by clicking on some in-content button that calls `<VRDisplay>.requestPresent(…)` (where `<VRDisplay>` would be a VR Display that exists in `navigator.vr.preferredDisplays` and is also supported by the content creator of the WebVR scene).

Here's what I mocked up in Firefox's Preferences, for example:

screen shot 2017-01-26 at 5 50 52 pm
screen shot 2017-01-26 at 6 32 47 pm


<hr>

If while a WebVR source is presenting to a headset _and_ a window navigation occurs, the UA can ought to *fire the `navigate` event on `navigator.vr` _in addition to_ pushing the display to `navigator.vr.referringDisplays` (i.e., `attribute FrozenArray<VRDisplay> referringDisplays`).

A developer could then check that and immediately request presentation back to the display the user was using before/during the page navigation.

AlbertoElias: I think allowing sites to say which headsets they support would be as dangerous as letting them say which browsers or OS they support and increase fragmentation

index.bs
```diff
@@ -169,6 +169,12 @@ interface VRDisplay : EventTarget {
    * created without preserveDrawingBuffer set to true will be cleared.
    */
   void submitFrame();
+
+  attribute EventHandler onactivate;
+  attribute EventHandler ondeactivate;
+  attribute EventHandler onblur;
+  attribute EventHandler onfocus;
+  attribute EventHandler onpresentchange;
```
toji: This was modeled after [`fullscreenchange`](https://developer.mozilla.org/en-US/docs/Web/Events/fullscreenchange) and [`pointerlockchange'](https://developer.mozilla.org/en-US/docs/Web/Events/pointerlockchange), which seemed to me like similar state transitions. I also see that the Page Visibility API include `visibilitychange`, so I'd say this verbiage is consistent with the rest of the web platform.

index.bs
toji: I was giving some thought to this as well, since it's going to be very difficult to get pages to always properly handle multiple headsets. (Even my WebVR.info samples don't, though there's comments in the source alluding to the fact that you should.) I think it's going to be very very very very common to have sited just use `display[0]`.

It seems like the dialog that you mocked up above could be pretty effective if the only thing it did was control the order that headsets are exposed in through `getDisplays()`. Want to use your Vive instead of your Rift? Make it always show up as `display[0]`. Sites that do the dumb thing will almost always pick it up, sites that are better behaved will allow you to pick anyway.

(A page can still break it by actively filtering out displays unless `displayName.includes('Rift')`, but at that point they're just being actively user hostile and I don't think we can help them much.)

index.bs
toji: SGTM, thanks for the research and suggestion! I'll make the update soon.

index.bs
```diff
@@ -215,6 +221,24 @@ Returns an array with the {{VRLayer}} currently being presented. MUST return an
...
+
+<dfn event for="VRDisplay" id="vrdisplay-ondeactivate-event">ondeactivate</dfn>
+A user agent MAY dispatch this event type to indicate that something has occured which suggests the {{VRDisplay}} should exit presentation. For example, if the {{VRDisplay}} is capable of detecting when the user has taken it off, this event SHOULD fire when they do so with the reason "unmounted".
+
+<dfn event for="VRDisplay" id="vrdisplay-onblur-event">onblur</dfn>
+A user agent MAY dispatch this event type to indicate that presentation to the {{VRDisplay}} by the page is paused by the user agent, OS, or VR hardware. While a {{VRDisplay}} is blurred it does not lose it's presenting status ({{isPresenting}} continues to report true) but {{getFrameData()}} returns false without updating the provided {{VRFrameData}} and {{getPose()}} returns null. This is to prevent tracking while the user interacts with potentially sensitive UI. For example: A user agent SHOULD blur the presenting application when the user is typing a URL into the browser with a virtual keyboard, otherwise the presenting page may be able to guess the URL the user is entering by tracking their head motions.
```
toji: Added a "blur" reason and text indicating it must be used here.

index.bs
```diff
@@ -215,6 +221,24 @@ Returns an array with the {{VRLayer}} currently being presented. MUST return an
...
+
+<dfn event for="VRDisplay" id="vrdisplay-onfocus-event">onfocus</dfn>
+A user agent MAY dispatch this event type to indicate that presentation to the {{VRDisplay}} by the page has resumed after being blurred.
```
toji: Added a "focus" reason and text indicating it must be used here.

index.bs
```diff
@@ -215,6 +221,24 @@ Returns an array with the {{VRLayer}} currently being presented. MUST return an
...
+
+<dfn event for="VRDisplay" id="vrdisplay-onpresentchange-event">onpresentchange</dfn>
+A user agent MUST dispatch this event type to indicate that the {{VRDisplay}} has begun or ended VR presentation. This event should not fire on subsequent calls to {{requestPresent()}} after the {{VRDisplay}} has already begun VR presentation.
```
toji: Added "presentstart" and "presentend" reasons and text indicating they must be used here.

index.bs
annevk: It would be a little easier to discuss design in dedicated threads, rather than subthreads of a PR.

What matters here are the use cases and what the typical programming experience would be like. Having said that, exposing state transitions through promises is the latest trend and if that makes sense for your API that'd the best to align on.

index.bs
```diff
@@ -559,31 +583,60 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
-
-## Navigator Interface extension ## {#interface-navigator}
+## VR Interface ## {#interface-vr}
 
 <pre class="idl">
-partial interface Navigator {
-  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getVRDisplays();
-  readonly attribute FrozenArray&lt;VRDisplay&gt; activeVRDisplays;
+interface VR : EventTarget {
+  Promise&lt;boolean&gt; getAvailability();
```
RafaelCintron: @cvan Out of curiosity, what is the rational for marking `getAvailability` with `[SecureContext]`?

index.bs
```diff
@@ -559,31 +583,60 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
+  Promise&lt;sequence&lt;VRDisplay&gt;&gt; getDisplays();
+
+  attribute EventHandler ondisplayconnect;
```
RafaelCintron: @cvan , Are you suggesting we add `displayadd` in addition to `displayconnect` or replace `displayconnect` with `displayadd`?  In your mind, what is the difference between 'adding' a display and 'connecting' a display?

index.bs
cvan: ah, I thought `SecureContext` is required for Feature Policy stuff. but, it appears – as I should have known – that `SecureContext` literally means for `https:` window contexts.

index.bs
```diff
@@ -559,31 +583,60 @@ Width of the play-area bounds in meters. The bounds are defined as an axis-align
...
+  attribute EventHandler ondisplaydisconnect;
+  attribute EventHandler onnavigate;
```
annevk: How does this event integrate with HTML? Is it properly hooked?

index.bs
cvan: sorry, replace. I find the `displayadd` verbiage make a bit more sense.

cvan:
@toji: this looks quite good to me. if you can address the last few comments, I'm comfortable with this getting merged. these in particular could use comments, and, if necessary, spec tweaks (or new issues filed):

* https://github.com/w3c/webvr/pull/116/files#r98351045
* https://github.com/w3c/webvr/pull/116/files#r98622176

awesome work. and, thanks for the patience. apologies for derailing the conversation; moving some properties and changing the interfaces involved dropping some and considerable rethinking. doing this in piecemeal is smart, and this is starting to shape up nicely. thanks!

toji:
Addressed several more points of feedback from @cvan, especially around how the events are documented (now has an attribute doc for the `on*` attributes which points to a separate event) but I'm leaving out any further changes to the events themselves because I'd like that conversation to be in a place that's easier to follow. :)

Going to merge this now, and follow up with creating some separate issues for items that weren't addressed in the initial pull.

PR #124, opened 2016-10-17
add auto Bikeshed deploy script via Travis CI (for issue #18)

cvan:


deploy.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+set -e # Exit with nonzero exit code if anything fails.
+
+SOURCE_BRANCH="master"
+TARGET_BRANCH="gh-pages"
```
toji: I thought we were moving away from gh-pages? Or is that the only way this can work?

PR #127, opened 2016-10-17
Describe what happens when submitFrame is not called every frame

dmarcos:
Reopening. The original https://github.com/w3c/webvr/pull/79 was closed after deleting the gh-pages branch


toji:
This is pretty stale. Closing for housekeeping purposes.

PR #131, opened 2016-10-20
Fixes missing comma in enum for VRDisplayEventReason

AVGP:
I encountered an error when running `make` due to this missing comma (and a stray trailing comma).


toji:
Thanks, I had this as a fix in a different PR but that one is taking a while to push through so this is a good thing to merge standalone.

Issue #100, opened 2016-09-27
Why a Float32Array for @offset in VREyeParameters


paulhiggs:
The @offset attibite in VREyeParameters is defined as a Float32Array but the description does not indicate why this would contain multiple values.

toji:
It's a 3 element vector, which matches many of the underlying APIs
implementations. We typically only expect the X values (0-th index) to be
anything meaningful, but some HMDs may report slight offset to Y or Z as
well for more exact calibration of the view.

paulhiggs:
Thanks for the definition. Could this be somehow incorporated into the description? As written now, it could be considered as a single value (as I did) as the last sentence states "Values for the left eye MUST be  negative; values for the right eye MUST be positive" but this is only for the X-component.

Perhaps use "3D Vector" as per the description of @position in VRPose

toji:
Definitely. I'll update this soon to reduce future confusion.

NellWaliczek:
@paulhiggs Out of curiosity, what are you using this data for?

toji:
Offset is deprecated in favor of providing view matrices for each eye directly. (This enabled more complex relationships between the eyes and the head pose than a simple offset.) Closing this issue.

PR #132, opened 2016-10-21
Explaining that VREyeParameters.offset is a vector rather than a scal…

AVGP:
…ar value


AVGP:
This is my shot at addressing #100

toji:
LGTM

Issue #125, opened 2016-10-17
Clarify renderWidth/Height: recommended size or 1:1?


toji:
On some devices, especially mobile, there are two potentially width and height values for the render targets we could report. One is the size required for the pixels at the center of the users vision to be 1:1 with the physical screen pixels post-distortion. The other is the size of the buffer that the platform feels provides the best balance of quality/performance. For example: on GearVR the reported value is 1024x1024, even though the screen could benefit from higher.  It's explained in their documentation that this value will hit a sweet spot for most apps, though something like a photo viewer may want to use a larger target.

Which of these values should WebVR report? Or should we report both?

brianchirls:
I've been pretty happy so far with modifying the resolution on the fly to keep the frame rate up, so I'd recommend sticking with the greater of the two values. You can always step it down. Also, I correct me if I'm wrong here but the fill rate could be widely variable depending on the complexity of the fragment shader(s) and the amount of overdraw. So I'm not convinced the recommended value is useful. Even if you do use a fixed resolution, you're still gonna have to benchmark on a few devices.

toji:
Glad to hear that dynamic resolution is working well for you! My concern is that a lot of naive applications will just allocate a buffer at the size that we report and be done with it, even if it's not what they really want from a performance perspective. Then again maybe we should just build dynamic resolution adaptation into Three.js and that'll solve it for 90% of devs. ;)

Regardless of what we chose, the spec text should be more explicit about what the value represents.

mkeblx:
Ideally every app would likely do some level (basic or advanced) of dynamic scaling and that is true also of non-VR WebGL apps where currently not usually employed. I don't see a way around besides app/framework handling in some fashion.

I don't know if we need two different values sets though.. "good balance of quality and performance for most applications" is going to be pretty arbitrary and likely on the conservative side well below 1:1 i.e. Oculus Mobile SDK recommendation of 1024x1024 even for native (with specific apps like photo app using larger (or using OVR Layers)) for all supported phones (all within a certain ~small factor of perf, same screen resolution [2560x1440], which won't be true of what WebVR content running on) but I don't know how they will change with increasing resolutions, more GPU power, .. 

[...]

PR #139, opened 2016-10-30
Removed VREyeParameters and related refactors

toji:
This is born from discussions I've had with Nell recently about a related to handling multiple coordinate systems that should be seeing it's own pull request soon-ish. Basically the thinking is that if we're going to break backwards compatibility in a significant way anyway (as is suggested for Worker support in #116) might as well take the opportunity to tidy up the API all around.

Refactored the information it contained into VRFrameData (offset is now *PoseToEyeMatrix for more flexibility) and the renderWidth/Height moved onto the VRDisplay as recommendedRenderWidth and recommendedRenderHeight. They now describe the ideal canvas size, containing both eyes, rather than the ideal render target size for a single eye. Field of view was already deprecated and replaced with the projection matrix on VRFrameData.

The core desire behind this change is to localize everything that may change frame to frame into `VRFrameData`. It was awkward having `getEyeParameters` be something that you may have to repeatedly query in addition to `getFrameData`, and the vector `offset` alone couldn't describe if the eye transform had a rotational component. Now that's explicit in the `*PoseToEyeMatrix` attributes. (That value COULD be derived from the view matrices and `pose`, but the math was pretty cumbersome and tricky to get right. No need to make users do it if we already know the answer.) Three.js in particular should be able to make direct use of those matrices. As for adding the matrix to the pose, that's just because it's awkward to have everything else be a matrix but still expect the user to compute that one. The implementations can (and probably will) choose to lazily compute either the matrix or the individual components so that we're not sending "duplicate" data internally. 

At the end of the day all of it is to simply say to users "We know you will need these values frequently. We'll give you them up front rather than making everyone do the same math over and over again."

Also Removed getPose, because if we’re gonna break backwards compatibility
let’s go all in and get rid of “deprecated” functions that never officially launched.

Feedback is very much wanted and appreciated!


brianchirls:
I see you've included helpful descriptions of `recommendedRenderWidth` and `recommendedRenderHeight` which are more detailed than their previous counterparts. This would appear to partially address #125.

It might also be useful to have a second pair of dimensions that indicate the resolution at which the center pixels render at 1:1. The values may be the same, but it looks like so far at least for Daydream and GearVR (and presumably Cardboard), they'll be different.

I wonder if the recommended values might even change dynamically when the browser or OS determines the device needs to cool down or preserve battery. [related](https://github.com/w3c/webvr/issues/37#issuecomment-256232767)

index.bs
```diff
@@ -402,54 +379,10 @@ Angular acceleration of the sensor given in radians per second squared. MAY be
 null if the sensor is incapable of providing angular acceleration. When not null
 MUST be a three-element array.
 
-<div class="example">
-The following code snippet creates a WebGL-compatible matrix from a
-{{VRPose}}:
+<dfn attribute for="VRFrameData">poseMatrix</dfn>
+A 4x4 matrix describing the tranform defined by the position and orientation of
+the pose, given as a 16 element array in column major order.
```
RafaelCintron: If poseMatrix describes the position and orientation of the pose, seems like we want to change the position and orientation Float32Arrays into booleans that indicate whether position and/or orientation are available.

index.bs
```diff
@@ -215,6 +202,17 @@ Returns an array with the {{VRLayer}} currently being presented. MUST return an
...
+
+<dfn attribute for="VRDisplay">depthFar</dfn>
+Z-depth defining the far plane of the eye view frustum. Enables mapping of values in the render target depth attachment to scene coordinates. Initially set to 10000.0.
+
+<dfn attribute for="VRDisplay">recommendedRenderWidth</dfn>
+Recommended width of the {{VRLayer}} source in pixels. The width reported SHOULD ideally be the resolution at which the final image presented to the user has a 1:1 pixel ratio at the center of the user's view after any distortion the {{VRDisplay}} will apply. The width reported MAY be lower than that, however, if the underlying platform recommends it for performance reasons. The width reported MUST be large enough to contain both the left and right eye viewports.
+
+<dfn attribute for="VRDisplay">recommendedRenderHeight</dfn>
+Recommended height of the {{VRLayer}} source in pixels. The height reported SHOULD ideally be the resolution at which the final image presented to the user has a 1:1 pixel ratio at the center of the user's view after any distortion the {{VRDisplay}} will apply. The height reported MAY be lower than that, however, if the underlying platform recommends it for performance reasons.
```
RafaelCintron: Now that recommendedRenderWidth and recommendedRenderHeight are for the entire VRLayer, it becomes a bit confusing to the web developer what the layout of the left and right eye are expected to be.  

For instance, is the developer expected to detect if the user agent returns e.g. 3840x1080 then that means it’s “supposed” to render side by side, but if the user agent returns e.g. 1920x2160 it’s top/bottom? I think we should either go back to using per-eye sizes, or just specify that we always want the data in a side by side fashion.

toji: In the default use case today (which this pull request doesn't change) content is expected to be rendered side-by-side because that's what the default `VRLayer` bounds define. You can always provide your own bounds if you want something more exotic, like top/bottom rendering.

There are several updates in this PR that are aimed at reducing the amount of computation JS apps are expected to do to derive the desired values. IE: the poseMatrix isn't hard to compute, but if we're providing matrices for everything else it's good for API consistency and general usability if the JS apps don't have to copy/past that logic everywhere. Likewise, while this was by no means a complicated calculation if 99% of WebVR apps are going to need to run through some equation to get the canvas size we should just give them the canvas size up front and make the less commonly needed value (the size per eye) the one that requires some math.

That being said, I don't feel so strongly about this that I'm opposed to changing it back to a "size per eye" value if there's a compelling argument for doing so.

ssylvan: I don't feel super strongly either, but I do think it's a bit weird to provide the size of a combined rect when we don't mandate any particular way of making that combination. If we're letting people choose the layout, it's a bit strange to make assumptions about that layout when providing properties about it.

After your clarification it makes sense that it is to be interpreted using the default layout of the eye rects, but before the clarification it was confusing to me.

toji: Yeah, no question that said clarifications should be present in the spec text.

toji:
Updated the comments for `recommendedRenderWidth/Height` for clarity and added `maximumRenderWidth/Height` to further reduce confusion and ensure that apps like photo viewers can targets the highest quality rendering effectively. I'd love the other browsers/HMD vendors take on that! For example, is that a property that GearVR can support?

mkeblx:
Should 1:1 pixel ratio post-distort be the defining characteristic for `maximum`? Possibly, but it's not really a maximum as even higher would be better, not fitting the language at least of "Applications should use a {{VRLayer}} source with this width to achive maximum quality." (Also not sure 1:1 pixel ratio applicable to next-gen displays.)

That does provide some useful information though that otherwise developers wouldn't have access to otherwise easily.

brianchirls:
@mkeblx How would you define a maximum? At what point is it just not worth going any higher?

cvan:
can you rebase?

mkeblx:
Never! :) There would be a maximum in terms of canvas size but can get already.. (Are you ready to render at 16,384 x 16,384 yet? :) ). Maybe max => maximum along pipeline. But I would just drop & maybe go with `recommended` I think.

As doing some things to look more future oriented, say a 16k screen mobile HMD comes out next year near same power envelope (to primarily eliminate screen door or something) with true 1:1 ratio post-distort being 20k so `recommended` by that would be 20k... maybe not really _recommended_ resolution you'd want to render at.

screenRes kinda has a weak relationship with renderRes. I'm not sure a good solution for figuring out the actual resolution they should render at. Again, app/dev responsibility.

index.bs
NellWaliczek: At the very least we need some way to indicate that tracking was lost for 6dof supporting devices whose fallback "3dof" mode still contains neck modeling.

toji:
This PR has been surpassed by the conversations around the 2.0 explainer, so I'm closing it.

PR #143, opened 2016-11-04
Revert "Convert VRLayer to an interface"

toji:
Reverts w3c/webvr#122

There's apparently a desire to discuss this further.


toji:
Closing. I need to apply this specifically against the 1.1. archived spec, which I'll do as a separate commit.

PR #147, opened 2016-11-09
Clarification about preserveDrawingBuffer behavior

toji:
It seems to me that many WebVR implementations will require the default framebuffer for a context to be replaced during VR presentation. Preserving the existing contents of the previous framebuffer is probably feasible but will require copying the data over when, in most cases, the data will be rendered invalid by resizing the canvas for VR anyway. This is especially important as the "easy path" for desktop mirroring is to set `preserveDrawingBuffer`.

This spec clarification simply allows the browser to drop the data when presentation begins for performance reasons.


toji:
Given that I haven't received any feedback about this yet and that the new text doesn't require the behavior, only allows it, I'm going to just merge it. Still, speak up if you feel strongly about it!

PR #151, opened 2016-11-15
Change requestAnimationFrame handle to unsigned long

MortimerGoro:
Hello,

While working on the Servo WebVR implementation I found that the requestAnimationFrame handle types are incosistent between Window and VRDisplay:

- Window uses `unsigned long`: https://www.w3.org/TR/html51/browsers.html#window-window
- VRDisplay uses `long`

We should unify the type. It isn't a problem in the JavaScript side but it avoids some type casts in native implementation, specially when doing the fallback from VRDisplay.requestAnimationFrame  to Window.requestAnimationFrame;

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -137,13 +137,13 @@ interface VRDisplay : EventTarget {
    * make no assumptions of frame rate or vsync behavior as the HMD runs
    * asynchronously from other displays and at differing refresh rates.
    */
-  long requestAnimationFrame(FrameRequestCallback callback);
+  unsigned long requestAnimationFrame(FrameRequestCallback callback);
 
   /**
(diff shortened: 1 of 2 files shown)


cvan:
Thanks! This looks good to me. We talked about this but I forgot to file an issue/fix the spec. Implementers, this good to merge?

DigiTec:
lgtm - lets get this merged in.

toji:
Thanks for pointing this out. Committed the change [here](https://github.com/w3c/webvr/commit/0789f4c6d7e8490aef1e0586e14a0f58d2952e64) to avoid the merge conflict. Sorry, probably wouldn't have been a problem if I had reviewed this earlier.

PR #155, opened 2016-11-28
Archive for previous versions of the spec

toji:
Sets up a folder to retain older versions of the spec in for reference.


toji:
Given no complaints, merging.

PR #157, opened 2016-12-02
Fix #interface-interface-vrfieldofview ID typo

foolip:


foolip:
Editing online and didn't update the HTML file, just thought this was the quickest way to point out the problem.

cvan:
thanks - yep, I was close but have to finish my work on automating the Bikeshed generation (#18); don't worry about updating the HTML 😄

Issue #166, opened 2016-12-20
VRLayer's leftBounds/rightBounds attributes: Sequences must not be used as the type of an attribute


ddorwin:
The spec currently has:
```
interface VRLayer {
  ...
  readonly **attribute sequence**<float> leftBounds;
  readonly **attribute sequence**<float> rightBounds;
};
```

Per https://heycam.github.io/webidl/#idl-sequence:
>Sequences must not be used as the type of an attribute or constant.
>>Note: This restriction exists so that it is clear to specification writers and API users that sequences are copied rather than having references to them passed around. Instead of a writable attribute of a sequence type, it is suggested that a pair of operations to get and set the sequence is used.

toji:
Closing since this is planned to be addressed in 2.0. If anyone has issues with the 2.0 proposal, please file a new issue. For details see the "VRCanvasLayer" interface described here: https://github.com/w3c/webvr/blob/master/explainer.md#appendix-b-proposed-idl

Issue #156, opened 2016-12-01
Redefine layer bounds to be standard OpenGL texture coords.


toji:
In 1.1 the layer `leftBounds` and `rightBounds` are defined as a kind of strange combination of a UV and a width and height, defined in UV space. It also uses the order [left, top, width, height], which implies a flipped UV space from OpenGL. I've brought this up before and the consensus was generally "Don't break it if we don't have to", but since then I've been bitten by this system a few time and I think that Chrome's implementation has it "upside down" as a result. :P 

I suggest that for clarity in 1.2 we redefine the layer bounds to explicitly be OpenGL UVs with the order [left, bottom, right, top]. (In OpenGL's UV space positive Y is up, so the bottom left corner is the origin.) Additionally we should define that bounds where left > right or bottom > top are invalid.

This means that instead of the default bounds being:
```
{
leftBounds: [0.0, 0.0, 0.5, 1.0],
rightBounds: [0.5, 0.0, 0.5, 1.0]
}
```

They become:
```
{
leftBounds: [0.0, 0.0, 0.5, 1.0],
rightBounds: [0.5, 0.0, 1.0, 1.0]
}
```

This feels more consistent with the rest of the platform to me (sticks to WebGL's conventions) and more predictable for developers. Any complaints?

mkeblx:
This seems quite unobjectionable (and none expressed), and minor. I don't see a reason for not redefining & closing, as we don't care about backwards compat too much and easily polyfillable.

toji:
Pushed this change as d6219a8c3e3dd892be60901937c483b1747bcf43. We're likely redefining how layers work soon, so it's somewhat moot, but I wanted to close the ticket down and ensure than change carried forward.

RafaelCintron:
LGTM after the fact.

Issue #142, opened 2016-11-03
Constructable VRLayer instead of dictionary?


bfgeek:
How likely is it that there will be different types of VRLayers in the future?

It might make more sense for VRLayer to be constructible directly? e.g.

```js
const fovLayer = new VRFOVLayer({canvas: canvas});
fov.eyeSomethingSomething = something;

const rectLayer = new VRRectLayer({...});
rectLayer.position.m11 = something;

display.requestPresent([fovLayer, rectLayer]);
```

@domenic might have thoughts here.

Ian

domenic:
As far as I can tell it's already constructible.

bfgeek:
Oh, I must have been looking at an old version of the spec in my browser, `requestPresent` seems broken then? @toji

toji:
Going to close this as a duplicate of #107, and note that a solution has been proposed in #186

Issue #107, opened 2016-10-12
VRLayer should be a real interface and not a Dictionary


esprehn:
The API exposes getLayers() which returns a sequence<VRLayer>, and the API would also benefit from allowing real methods on VRLayers. In general returning dictionaries has a bunch of drawbacks, lets use real interfaces instead. :)

@tabatkins @ojan @bfgeek

DigiTec:
What APIs would be useful to add on VRLayer? Can you give some examples. This would help inform the overall design. Specifically for backwards compat we have to continue to take the dictionary in requestPresent, but could also take an array of "some other object" that the user could construct (but then we'd also have to add constructors).

cvan:
I'm thinking something like this:

```js
window.addEventListener('vrdisplaypresentactivate', e => {
  if (!e.display) { return; }
  if (e.reason === 'navigation') {
    e.display.requestPresent([
      {source: webglCanvasVRContent}
    ]);
  } else if (e.reason === 'mounted') {
    e.display.requestPresent([
      {source: webglCanvasVRContent}
    ]);
  }
});

window.addEventListener('vrdisplaypresentblur', e => {
  if (!e.display) { return; }
    e.display.requestPresent([
      {source: webglCanvasVRPauseScreen},
      {source: webglCanvasVRContent}
    ]);
  }
});
```

tabatkins:
@DigiTec Elliot is talking about the methods that *return* a VRLayer dictionary.  This results in you returning a plain JS object with values set on it as data properties.  Instead using an interface type for return values gives you a more standard sort of object, which has getters/setters on the prototype, and which can actually have methods added to it in the future.  (We should make returning dictionary types be an error in WebIDL instead; they're never a good idea.)

But also:

> Specifically for backwards compat we have to

Uh, what backwards compat? There's no shipping API to be back-compatible with.

toji:
> Uh, what backwards compat? There's no shipping API to be back-compatible with.

True, but there's a non-trivial amount of content built against experimental builds and flagged features. We can break it when necessary, (and have a few times) but in general we'd like to be as kind as possible to developers who have invested in the feature early.

So to clarify, even if we continue accepting dictionaries in `requestPresent` it would be preferable to  return an actual interface from `getLayers` that represents the same values, correct? I don't have a problem with that. I'd like to continue consuming dictionaries, though. This particular mechanism is intended to be used as an API extension point in the future, so flexibility in what it accepts will be beneficial.

[...]

PR #169, opened 2017-01-07
Proposal for converting VRLayers into constructible objects

toji:
Attempts to address #142, #107, #166, #156, and #125. Feedback welcome!

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -262,8 +262,13 @@ vrDisplay.requestPresent([ new CanvasVRLayer(canvas) ]).then(function() {
 The {{VRLayer}} is the base interface for the various layer types that can be
 provided to a {{requestPresent()}} call to be presented in the HMD.
 
+Each type of {{VRLayer}} should have a corresponding {{VRLayerReadOnly}} that
+will be returned by {{getLayers()}}.
+
 <pre class="idl" id="vrlayer-interface">
 interface VRLayer {};
+
+interface VRLayerReadOnly {};
 </pre>
 
 ## CanvasVRLayer ## {#interface-canvasvrlayer}
@@ -282,13 +287,23 @@ interface CanvasVRLayer : VRLayer {
   attribute float renderScale;
   attribute boolean resizeSourceOnPresent;
 
-  readonly attribute unsigned long renderWidth;
-  readonly attribute unsigned long renderHeight;
-
   void setLeftBounds(float left, float bottom, float right, float top);
   sequence&lt;float&gt; getLeftBounds();
 
   void setRightBounds(float left, float bottom, float right, float top);
+  sequence&lt;float&gt; getRightBounds();
+};
+
+interface CanvasVRLayerReadOnly : VRLayerReadOnly {
+  readonly attribute CanvasVRSource source;
+
+  readonly attribute float renderScale;
+  readonly attribute boolean resizeSourceOnPresent;
+
+  readonly attribute unsigned long renderWidth;
+  readonly attribute unsigned long renderHeight;
+
+  sequence&lt;float&gt; getLeftBounds();
   sequence&lt;float&gt; getRightBounds();
 };
 </pre>
@@ -301,7 +316,7 @@ The {{CanvasVRLayer/source}} attribute defines the canvas whose contents will be
 <dfn attribute for="CanvasVRLayer">renderScale</dfn>
 Specifies the ratio of the number of canvas pixels to the {{VRDisplay}}'s physical pixels at the center of the output image. Note that a value of <code>1.0</code> will generally result in a larger canvas resolution than the {{VRDisplay}}'s physical screen in order to allow for corrective distiorion. Specifying a lower value can improve performance. A value of <code>0.0</code> instructs the user agent to select a ratio that provides a good balance between quality and performance for most applications on the current {{VRDisplay}}. Calls to {{requestPresent()}} MUST be rejected if the renderScale of any {{CanvasVRLayer}} provided is less than <code>0.0</code>.
 
-If this value was set to <code>0.0</code> when the {{CanvasVRLayer}} is passed to {{requestPresent()}} the {{renderScale}} of the corresponding {{CanvasVRLayer}} returned from {{getLayers()}} should be set to the value selected by the user agent.
+If this value was set to <code>0.0</code> when the {{CanvasVRLayer}} is passed to {{requestPresent()}} the {{renderScale}} of the corresponding {{CanvasVRLayerReadOnly}} returned from {{getLayers()}} should be set to the value selected by the user agent.
 
 The default value for {{renderScale}} MUST be <code>0.0</code>.
 
@@ -309,12 +324,6 @@ The default value for {{renderScale}} MUST be <code>0.0</code>.
 If <code>true</code> the {{CanvasVRLayer/source}} of any {{CanvasVRLayer}} provided to {{requestPresent()}} will be resized to the width and height computed from the {{renderScale}} before the promise resolves or the vrdisplaypresentchange events are fired. If {{requestPresent()}} is rejected the {{CanvasVRLayer/source}} dimensions will not be changed.
 
 The default value for {{resizeSourceOnPresent}} MUST be <code>true</code>.
-
-<dfn attribute for="CanvasVRLayer">renderWidth</dfn>
-When a {{CanvasVRLayer}} is returned from {{getLayers()}} this attribute MUST be set to the width computed from the {{renderScale}} when the {{CanvasVRLayer}} was passed to {{requestPresent()}}.
-
-<dfn attribute for="CanvasVRLayer">renderHeight</dfn>
-When a {{CanvasVRLayer}} is returned from {{getLayers()}} this attribute MUST be set to the height computed from the {{renderScale}} when the {{CanvasVRLayer}} was passed to {{requestPresent()}}.
 
 <dfn method for="CanvasVRLayer">setLeftBounds()</dfn>
 Sets the portion of the canvas to present to the left eye. Bounds are specified as texture coordinates in UV space (0.0 - 1.0), where <code>(0, 0)</code> represents the bottom left corner of the canvas and <code>(1, 1)</code> represents the top right corner. Calls to {{requestPresent()}} MUST be rejected if the left bounds of any {{CanvasVRLayer}} provided specifies bounds where <code>right</code> is less than <code>left</code> or <code>top</code> is less than <code>bottom</code>.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -261,39 +259,78 @@ vrDisplay.requestPresent([{ source: canvas }]).then(function() {
...
+
+<dfn attribute for="CanvasVRLayer">renderScale</dfn>
+Specifies the ratio of the number of canvas pixels to the {{VRDisplay}}'s physical pixels at the center of the output image. Note that a value of <code>1.0</code> will generally result in a larger canvas resolution than the {{VRDisplay}}'s physical screen in order to allow for corrective distiorion. Specifying a lower value can improve performance. A value of <code>0.0</code> instructs the user agent to select a ratio that provides a good balance between quality and performance for most applications on the current {{VRDisplay}}. Calls to {{requestPresent()}} MUST be rejected if the renderScale of any {{CanvasVRLayer}} provided is less than <code>0.0</code>.
+
+If this value was set to <code>0.0</code> when the {{CanvasVRLayer}} is passed to {{requestPresent()}} the {{renderScale}} of the corresponding {{CanvasVRLayer}} returned from {{getLayers()}} should be set to the value selected by the user agent.
+
+The default value for {{renderScale}} MUST be <code>0.0</code>.
+
+<dfn attribute for="CanvasVRLayer">resizeSourceOnPresent</dfn>
```
toji: This is the part of the change I'm most unsure about. We generally want to make it easy to do the right thing when using the API, and one of the areas I frequently see people get wrong is properly sizing the canvas. This would make it so that the default API behavior "magically" sets the canvas to the correct size unless you choose to do it manually. But it feels a little weird to me to have WebVR changing other element's attributes. Thoughts?

index.bs
```diff
@@ -261,39 +259,78 @@ vrDisplay.requestPresent([{ source: canvas }]).then(function() {
...
+<pre class="idl" id="canvasvrlayer-interface">
 typedef (HTMLCanvasElement or
-         OffscreenCanvas) VRSource;
+         OffscreenCanvas) CanvasVRSource;
 
-[Constructor(optional VRLayerInit layer)]
-interface VRLayer {
-  readonly attribute VRSource? source;
+[Constructor(optional CanvasVRSource source)]
+interface CanvasVRLayer : VRLayer {
```
toji: One area that I'm still waffling on is if we want to have two interfaces for each layer type: One for passing into `requestPresent` and one to be returned by `getLayers` (Which would likely be `CanvasVRLayerReadOnly` or something like that.) That way attributes that can't be set by the user (`renderWidth/Height`) are only in the ReadOnly variant and there's no confusion about what happens if you change the layer values after getting them out of `getLayers` like so:

```
var layers = vrDisplay.getLayers();
layers[0].setLeftBounds(l, b, r, t);
// Hey, WTF! Why didn't my layer bounds change like I asked?
```

index.bs
```diff
@@ -261,39 +259,78 @@ vrDisplay.requestPresent([{ source: canvas }]).then(function() {
 
 ## VRLayer ## {#interface-vrlayer}
 
-The {{VRLayer}} interface is provided to a {{VRDisplay}} and presented in the HMD.
+The {{VRLayer}} is the base interface for the various layer types that can be
+provided to a {{requestPresent()}} call to be presented in the HMD.
 
-<pre class="idl" id="vrlayer-dictionary">
+<pre class="idl" id="vrlayer-interface">
+interface VRLayer {};
```
toji: One thing I want to call out that I like about this proposal is that it makes feature detection of supported layers really easy once we start defining new ones. It becomes simply:

```
if (EMEVideoVRLayer) {
  // Yay! I can use this layer type!
}
```

index.bs
```diff
@@ -261,39 +259,78 @@ vrDisplay.requestPresent([{ source: canvas }]).then(function() {
...
+  sequence&lt;float&gt; getRightBounds();
 };
 </pre>
 
-### Attributes ### {#vrlayer-attributes}
+### Attributes ### {#canvasvrlayer-attributes}
+
+<dfn attribute for="CanvasVRLayer">source</dfn>
+The {{CanvasVRLayer/source}} attribute defines the canvas whose contents will be presented for this layer by the {{VRDisplay}} when {{VRDisplay}}.{{submitFrame()}} is called. Upon being passed into {{requestPresent()}} the current backbuffer of the source's context MAY be lost, even if the context was created with the <code>preserveDrawingBuffer</code> context creation attribute set to <code>true</code>. Calls to {{requestPresent()}} MUST be rejected if the {{CanvasVRLayer/source}} of any {{CanvasVRLayer}} provided does not represent a canvas with a <code>webgl</code> or <code>webgl2</code> context.
+
```
RafaelCintron: I presume the case where the canvas becomes lost upon calling requestPresent is for multi-adapter scenarios?  If so, we should indicate that in the text and point out that the requestPromise should be rejected if the canvas becomes lost upon calling requestPresent or if it is already lost before calling requestPresent.

index.bs
RafaelCintron: I'm inclined to have the object you pass to requestPresent be a different type than the object returned from getLayer. The current approach can lead to confusion as to which attributes are initialization parameters and which are informational ones owned by the user agent.

index.bs
```diff
@@ -261,39 +259,78 @@ vrDisplay.requestPresent([{ source: canvas }]).then(function() {
...
+<dfn attribute for="CanvasVRLayer">renderScale</dfn>
+Specifies the ratio of the number of canvas pixels to the {{VRDisplay}}'s physical pixels at the center of the output image. Note that a value of <code>1.0</code> will generally result in a larger canvas resolution than the {{VRDisplay}}'s physical screen in order to allow for corrective distiorion. Specifying a lower value can improve performance. A value of <code>0.0</code> instructs the user agent to select a ratio that provides a good balance between quality and performance for most applications on the current {{VRDisplay}}. Calls to {{requestPresent()}} MUST be rejected if the renderScale of any {{CanvasVRLayer}} provided is less than <code>0.0</code>.
+
+If this value was set to <code>0.0</code> when the {{CanvasVRLayer}} is passed to {{requestPresent()}} the {{renderScale}} of the corresponding {{CanvasVRLayer}} returned from {{getLayers()}} should be set to the value selected by the user agent.
+
```
RafaelCintron: If the developer sets this value to 0.0, how will the user agent set the value?  Is the user agent expected to dynamically resize the canvas in response to general performance frame over frame?  Or is this meant to be a general recommended value based on the characteristics of the device?  If it is the former, I have misgivings about the user agent attempting to do this for all content on the Internet. I would prefer that we give the developer the tools to determine the sizing on their own and adjust it if they wish.

toji: Good question. Demonstrates that the text needs clarification.

The intent is not to have the system doing dynamic scaling. The developer should be able to do that via viewport alteration, but the user agent won't attempt to guess at how the scaling should work. Instead the 0.0 case here indicates that the size reported should be a static value that gives the developer some wiggle room with the fill rate while still looking decent. Up to the UA what this means, and it's intended to be a value that we can tune fairly opaquely when needed. I would expect that in the real world most desktop devices will report a 1:1 value anyway, since they have the hardware to back it up, and mobile devices will report something lower. (GearVR docs recommend 1024x1024 per eye, for example, and Daydream sample code suggests using 7/10 the full 1:1 size)

index.bs
toji: This actually wasn't aimed at a multiadapter scenario, but instead accounts for cases where the render target for the canvas needs to be swapped out for something VR specific. If we were attempting to strictly adhere to preserveDrawingBuffer semantics that would require a copy of the existing content over to the new render target, something that will be lost 99% of the time anyway because the canvas will likely be resized to match the VRDisplay size. So this text simply allows the UA to not guarantee that the render target content will be preserved.

The multi-adapter scenario does need to be accounted for, though. And unfortunately I'm still not sure where or how. :(

index.bs
toji: Okay, thanks for the vote of confidence in that route. I've updated this now, let's see if there's any strenuous objections.

index.bs
RafaelCintron: OK, in this case, we should probably say that the contents of the canvas are deleted upon calling requestPresent.

index.bs
```diff
@@ -261,39 +259,93 @@ vrDisplay.requestPresent([{ source: canvas }]).then(function() {
...
+If this value was set to <code>0.0</code> when the {{CanvasVRLayer}} is passed to {{requestPresent()}} the {{renderScale}} of the corresponding {{CanvasVRLayerReadOnly}} returned from {{getLayers()}} should be set to the value selected by the user agent.
+
+The default value for {{renderScale}} MUST be <code>0.0</code>.
+
+<dfn attribute for="CanvasVRLayer">resizeSourceOnPresent</dfn>
+If <code>true</code> the {{CanvasVRLayer/source}} of any {{CanvasVRLayer}} provided to {{requestPresent()}} will be resized to the width and height computed from the {{renderScale}} before the promise resolves or the vrdisplaypresentchange events are fired. If {{requestPresent()}} is rejected the {{CanvasVRLayer/source}} dimensions will not be changed.
```
RafaelCintron: WindowsHolographic has a similar property to renderScale called [ViewportScaleFactor](https://msdn.microsoft.com/en-us/library/windows/apps/windows.graphics.holographic.holographiccamera.viewportscalefactor.aspx). Here is what the docs have to say:   
`Each frame, the system decides the size of the viewport within the render target that the app must render that frame. Sometimes the system may lower this to free up system resources, for example when video capture is occurring. Regardless of the viewport size, the rendered image is always scaled up to the full size of the displays. An app looking to lower its rendering cost for other reasons can request a smaller viewport.`

`You can specify a scale factor between 0.0 and 1.0, which is multiplied by the default viewport size. In future frames, the system will specify that you render to the nearest valid viewport size to this requested size.`

This means that the viewport you're given depends on:

1. What the system decides is best for you based upon some the capabilities of the system and some other operation happening in the system that the developer can't control or have knowledge about.
2. A rounding of the requested viewport scale with valid viewport sizes. 

I still feel that the browser should not be resizing the framebuffer based on what is happening on the webpage.

Given this, I would prefer that we make the following changes:
1. Change `renderScale` to `requestedRenderScale`
2. Add the notion of a viewport size that the web developer needs to render to within the Canvas.  

What is the rationale for setting resizeSource to false?  Do we expect that web developers will want to super size their content?  If so, why can't they do this themselves via an intermediate framebuffer?  Seems like having resizeSource=true as the only option would be a simplification we can make to the API.  Same questions apply to renderScale > 1.0.

What do you think?

index.bs
```diff
@@ -261,39 +259,93 @@ vrDisplay.requestPresent([{ source: canvas }]).then(function() {
...
+  sequence&lt;float&gt; getRightBounds();
 };
 </pre>
 
-### Attributes ### {#vrlayer-attributes}
+### Attributes ### {#canvasvrlayer-attributes}
+
+<dfn attribute for="CanvasVRLayer">source</dfn>
+The {{CanvasVRLayer/source}} attribute defines the canvas whose contents will be presented for this layer by the {{VRDisplay}} when {{VRDisplay}}.{{submitFrame()}} is called. Upon being passed into {{requestPresent()}} the current backbuffer of the source's context MAY be lost, even if the context was created with the <code>preserveDrawingBuffer</code> context creation attribute set to <code>true</code>. Calls to {{requestPresent()}} MUST be rejected if the {{CanvasVRLayer/source}} of any {{CanvasVRLayer}} provided does not represent a canvas with a <code>webgl</code> or <code>webgl2</code> context.
+
```
RafaelCintron: Please also accept experimental-webgl.

toji:
Replaced by #186. Closing.

PR #175, opened 2017-01-12
Strip timestamp out of VRFrameData

toji:
As discussed at Kirkland F2F. This can be added back in later after
further discussion to more precisely define what the timestamp is
relative to.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -23,9 +23,6 @@ Abstract: This specification describes support for accessing virtual reality (VR
 </pre>
 
 <pre class="anchors">
-urlPrefix: http://www.w3.org/TR/hr-time/
-    type: typedef; text: DOMHighResTimeStamp
-    type: dfn; text: timestamp origin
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -504,7 +491,7 @@ A 4x4 matrix describing the projection to be used for the right eye's rendering,
 A 4x4 matrix describing the view transform to be used for the right eye's rendering, given as a 16 element array in column major order. Represents the inverse of the model matrix of the right eye in sitting space. This value may be passed directly to WebGL's uniformMatrix4fv function. It is highly recommended that applications use this matrix when rendering.
 
 <dfn attribute for="VRFrameData">pose</dfn>
-The {{VRPose}} of the {{VRDisplay}} at {{timestamp}}.
+The {{VRPose}} of the {{VRDisplay}} at the point the {{VRFrameData}} was produced.
```
mthiesse: "the point the {{VRFrameData}} was produced" is confusingly worded. Is "the time the {{VRFrameData}} was produced" better?

PR #190, opened 2017-02-18
Introducing onvrdisplaypointerrestricted and onvrdisplaypointerunrestricted events

NellWaliczek:
On a few occasions, I've brought up the importance of protecting users from unintended
consequences while trying to use their mouse while in VR.  On some systems the OS, platform,
or user agent may implement functionality to help protect the user, but it leaves web
developers unclear about when they should attempt to take pointerlock to behave properly.
This proposal attempts to address the issue by providing optional events for user agents to
implement that communicate this state to web developers.


cvan:
FWIW, there's been some discussion on whether `requestPointerLock` should be gated on the [Permissions API](https://w3c.github.io/permissions/) ([`navigator.permissions`](https://w3c.github.io/permissions/#navigator-and-workernavigator-extension)): https://github.com/w3c/permissions/issues/67

Since `pointerlock` is a considered a gesture-based API, there's [some hesitation](https://github.com/w3c/permissions/issues/67#issuecomment-235818464) to add `pointerlock` to the [Permissions API's enum](https://w3c.github.io/permissions/#permission-registry). So, if other folks are fine with handling this events in the WebVR API, I suppose this change is fine.

I'll let others weigh in. Thanks for the PR!

NellWaliczek:
Hey @cvan,
Thanks for the reference!  The language changes I made regarding "engagement gesture" were to better align with the language used in the pointerlock spec, but the need for the events themselves is about system state rather permissions.  

Specifically, if a developer wants to use mouse input for a VR experience, @toji has previous mentioned that the best practice would be to use pointerlock (which I wholeheartedly agree with!).  The reason is that the user cannot see their mouse while in VR and without pointerlock the user could really be clicking on anything (and even accidentally delete things!)  The trick is, on systems with affordances to protect users from that sort of unhappy accident, how can a web developer know when the right time to take pointerlock is (especially given the need to support demo driver like scenarios)?  The events I'm proposing would formalize the appropriate time to rely only on pointer delta information vs. complete pointer information.

If there's anything I can add that would help make the explanation or usage clearer, please let me know!

archive/prerelease/1.1/index.bs
```diff
@@ -675,11 +675,17 @@ A user agent MAY dispatch this event type to indicate that something has occured
...
 
 <dfn event for="Window" id="window-onvrdisplayfocus-event">onvrdisplayfocus</dfn>
 A user agent MAY dispatch this event type to indicate that presentation to the display by the page has resumed after being blurred.
 
+<dfn event for="Window" id="window-onvrdisplaypointerrestricted-event">onvrdisplaypointerrestricted</dfn>
+A user agent MAY dispatch this event type to indicate that pointer input is restricted to consumption via a <a href="http://w3c.github.io/pointerlock/">pointerlocked</a> element. This event MUST only be dispatched while the {{VRDisplay}} is presenting. Pages that wish to consume pointer input for HMD experiences are encouraged to call requestPointerLock in response to this event.
```
toji: Just clarifying: My reading is that this indicates that the page is entering a mode where normal mouse events/pointer events are no longer fired at all unless it's against a pointerlocked element, correct?

toji:
Minor clarification requested, but otherwise I think this looks good. I'll give @cvan (or anyone else) a day or so in case they have any further comments before merging.

NellWaliczek:
@toji Yep, that's correct.

PR #198, opened 2017-03-02
Name Nell Wallczek as a spec editor

toji:
@NellWaliczek has asked if she could be listed as a spec editor, and I personally think that's a great idea given Microsoft's contributions thus far. Putting this up as a pull request to allow for comments, but I doubt there's going to be any opposition to the idea. :)


kearwood:
I agree!  This is long overdue.

kearwood:
I took the liberty of merging

PR #206, opened 2017-03-13
Moved the spec around a bit to make it easier for people to find 1.1

toji:
Getting some issues filed against a 1.1 implementation from looking at
the ToT spec. Re-organizing the spec archive to make it easier to find
the spec that people want to see.

This makes the top level page a directory which links to the 1.1 and "latest" versions, with a brief explanation of both.


kearwood:
Thanks @toji !  This is greatly appreciated and should make it much clearer.

toji:
No complaints, so I'm going to merge this. Apologies if I break somebody's pending changes!

Issue #22, opened 2016-06-01
Add Acknowledgements


cvan:
<a href="https://github.com/cvan"><img src="https://avatars.githubusercontent.com/u/203725?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [cvan](https://github.com/cvan)**
_Monday Feb 29, 2016 at 00:32 GMT_
_Originally opened as https://github.com/MozVR/webvr-spec/issues/14_

---

toji:
nothing against having an Acknowledgements section, but nobody seems quite sure what to put in it. :) Removed for now in 84d8d8bcbd6ec661815f30e86acb89b60d4df8ba, feel free to add it back in at any time if you think of something/someone to acknowledge.

PR #209, opened 2017-03-14
spec: replace hardcoded list of editors with link to Contributors on GitHub

cvan:


toji:
I don't agree with this change. Spec Editor is a pretty well defined position that comes with various responsibilities. I have zero problems acknowledging the work of the larger community, but that would be more appropriate for an Acknowledgements section. (Which I guess means we've figured out what to use for #22!)

cvan:
that's totally fine. I've seen a number of W3C specs make this change, so I thought I'd suggest it. but that's definitely fine by me to use the Acknowledgements section for mentioning contributors from non-core contributors!

Issue #213, opened 2017-04-04
Move depthNear/Far to VRCoordinateSystem or VRFrameOfReference


toji:
For a while now the presence of depthNear and depthFar on the VRSession has been gnawing at me because they felt out of place, and there's been questions in the past about their exact behavior. (#120) Specifically, what's the expected result of this:

```js
session.depthFar = 10;
let poseA = session.getPose(coordSystem);
session.depthFar = 100;
let poseB = session.getPose(coordSystem);
```

In that scenario it seems like `poseA` and `poseB` should return different projection matrices, but that conflicts with the idea that the pose returned should stay the same throughout the duration of the frame callback.

To resolve that (and clean up the session interface a bit) I propose that we make the depth parameters const members of either the `VRCoordinateSystem` or the `VRFrameOfReference` (and I'd love @NellWaliczek or @RafaelCintron's advice on which of those seems more appropriate.) The code would look like this:

```js
let frameOfRef = await vrSession.createFrameOfReference({
    type: "EyeLevel",
    depthNear: 0.1,
    depthFar: 10 });

let poseA = session.getPose(frameOfRef);
frameOfRef.depthFar = 100; // ERROR! Read only member!
```

This enforces that the same Frame of Reference always returns the same pose for the duration of the frame callback. Making `createFrameOfReference` take a dictionary also allows us to provide sensible defaults for everything (including `type`?) and makes it more extensible in the future if needed. It also seems like it may play better with APIs like Windows Holographic, but I might just be mis-reading the docs on that one.

The one downside is that it becomes more difficult to adjust depth parameters on the fly, since you'd need to create a new frameOfReference for that. I'm not sure how frequently that need actually arises in the real world, though.

Proposed IDL:
```webidl
dictionary VRFrameOfReferenceInit {
  VRFrameOfReferenceType type = "EyeLevel";
  double depthNear = 0.01;
  double depthFar = 10000.0; // Defaults from existing spec. Worth re-evaluating?
};

interface VRFrameOfReference : VRCoordinateSystem {
  readonly VRFrameOfReferenceType type;
  readonly double depthNear; // Keeping it all on VRFrameOfReference here for simplicity
  readonly double depthFar;
  readonly attribute VRStageBounds? bounds;
};

partial interface VRSession {
  Promise<VRFrameOfReference> createFrameOfReference(VRFrameOfReferenceInit init);
};
```

NellWaliczek:
O_o
[...]

PR #214, opened 2017-04-11
Multiple Frame of Reference changes

toji:
- move depthNear/Far to FrameOfReference
- Create FrameOfReference with constructor instead of async function

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -124,11 +124,6 @@ function BeginVRSession() {
     glCanvas.width = sourceProperties.width;
     glCanvas.height = sourceProperties.height;
 
-    // The depth range of the scene should be set so that the projection
-    // matrices returned by the session are correct.
-    vrSession.depthNear = 0.1;
-    vrSession.depthFar = 100.0;
-
     // The content that will be shown on the device is
     // defined by the current layer.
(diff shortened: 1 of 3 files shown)


toji:
Landing would fix #213

NellWaliczek:
Hey @toji! Sorry I didn't a chance to respond to the issue that prompted this sooner :-/

I totally agree that it's unclear when depth plane changes will be applied with the current shape of the API.  I'm not sure though that depthNear and depthFar are really properties of the Frame of Reference as much as of the VR Device itself...  For example, from frame-to-frame developers can choose to use whichever frame of reference they like, and as such we risk developers accidentally having the wrong projection matrices during tracking loss because they only updated one of the frames of reference.

What about if we have these values be readonly attributes of the VRSession and have a Promise<void> setDepthNear() to change them that fulfills the promise on the rendering frame boundary.

toji:
Updated the pull request. There were several non-controversial documentation updates rolled up into this that I pulled out and landed separately on master. Now this PR focuses more strictly on the proposed changes. Doesn't address Nell's feedback yet.

toji:
Closing this since it's obvious they need more discussion and it's got conflicts now anyway.

PR #219, opened 2017-04-15
Adding a reference to the Windows Mixed Reality headsets

NellWaliczek:
As of the Windows 10 Creators Update, Microsoft Edge now supports WebVR and developers can even get started without a headset using the [Windows Mixed Reality Simulator](https://developer.microsoft.com/en-us/windows/mixed-reality/using_the_windows_mixed_reality_simulator). Note that while WebVR is enabled by default in Microsoft Edge, using the Windows Mixed Reality Portal or Mixed Reality dev kits currently requires Developer Mode to be turned on in Windows settings.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -1,7 +1,7 @@
-# WebVR Explained
+﻿# WebVR Explained
 
 ## What is WebVR?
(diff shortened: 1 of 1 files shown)


toji:
👍

Issue #223, opened 2017-04-20
Managing multi-GPU systems


toji:
On systems with multiple graphics adapters (SLI/Crossfire or an integrated/discreet pair like many laptops) the underlying graphics context driving the VR hardware must be the one that the headset video output is actually physically connected to. Currently the web has no mechanisms for handling that, which is a problem for WebVR.

This issue and the high level solutions are courtousy of Microsoft. I'm fleshing out the proposal and filing here so they have a bit more bandwidth to work on other high priority items (like VR Input).

## Overview: ##
There's two different categories of app that developers may fall into:

 - **VR Centric:** The app's primary use case is VR, and as such it doesn't mind initializing resources in a VR-centric fashion, which may include asking users to select a headset as soon as the app starts. An example would be a game which is dependent on VR presentation and input.
 - **VR Enhanced:** The app can take advantage of VR, but it's used as a progressive enhancement rather than a core part of the experience. Most users will probably _not_ interact with the app's VR features, and as such asking them to make VR-centric decisions early in the app lifetime would be confusing and innapropriate. An example would be a news site with an embedded 360 photo gallery or video. (We expect the large majority of early WebVR content to fall into this category.)

In the first case, it makes sense to allow developers to indicate at WebGL context creation time what `VRDevice` the context will be used with so that it can be initialized on the correct adapter. The suggestion is that this be passed as an optional `WebGLContextAttributes`, which acts as a hint to the system for which VR device will be using the context in the future. The system should be able to ignore this in cases where it's simply not applicable (Mobile device, only one GPU, etc.)

In the second case, though, initializing the canvas this way may force developers to have their users pick very early on which VR display they intend to use the device with. Going to a news site and having it immediately ask for which headset you intend to use even when there's no obvious VR content on screen is not a desirable user experience. In this case, the canvas can only know which device it's going to be used with at the point that the VR session begins using it.

In this case, it seems reasonable to use the existing WebGL [context loss and restoration](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.13) mechanism. If an exclusive session makes use of a canvas that isn't already on the right adapter the context is lost and (if properly handled) will be recreated with the correct GPU. (This makes an argument for something I proposed in #218, which is to have sessions explicitly bound to a canvas element at creation time. That makes session creation a perfect point for the context check, and the session creation can be rejected if the context isn't compatible or able to be made compatible.)

[...]

PR #226, opened 2017-05-02
Update explainer to account for multi-GPU handling

toji:
Closely follows Microsoft’s suggestion from #223, though I’ve moved the
ensureCompatibility call to the VRDevice because it feels a bit better
localized that way.


NellWaliczek:
Thanks!  This generally looks great :)
A few small things:
- Your framing about "VR-Centric" vs "VR-Enhanced" was really helpful and it might be useful to include something similar here to assist developers in determining which approache is right for them.
- Can we call out something about the fact that calling `ensureContextCompatibility` may (will always?) have impact on more than just the glcontext supplied?  I'd love to know how folks are approaching the implementation of the power preference attribute.  It may make sense for us to align with that pattern...
- Can we say that `ensureContextCompatibility` may reject for other reasons (such as already having been paired with an incompatible VRDevice already?)

RafaelCintron:
> I'd love to know how folks are approaching the implementation of the power preference attribute. It may make sense for us to align with that pattern...

@grorg detailed how Webkit is approaching the powerPreference attribute in a [post to the WebGL mailing list](https://www.khronos.org/webgl/public-mailing-list/public_webgl/1703/msg00021.php).  The latest I've heard from @kenrussell is Chrome will be following a similar approach.  The [language in the spec](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2.1) is written such that contextLost/restored can happen at any time to regulate power and memory consumption.

explainer.md
```diff
@@ -499,6 +512,8 @@ interface VRDevice : EventTarget {
 
   Promise<boolean> supportsSession(VRSessionCreateParametersInit parameters);
   Promise<VRSession> requestSession(VRSessionCreateParametersInit parameters);
+
+  Promise<void> ensureContextCompatibility(WebGLRenderingContext context);
```
RafaelCintron: We should use `WebGLRenderingContextBase` instead of `WebGLRenderingContext` so that the API can be used with WebGL 2.

kenrussell: Agree with this, or perhaps use a union type if you want to refer to a data type that's exposed to JavaScript. (WebGLRenderingContextBase isn't.)

toji: Good catch. I swear I had already done that.

I don't _think_ there's a problem with using `WebGLRenderingContextBase` here, even if it isn't Javascript exposed. The IDLs should still parse correctly. (But if not, then yes. A union would work just fine.)

toji:
Updated the pull request based on feedback. Felt silly adding a couple of lines around some unrelated topics that I know we'll be replacing soon, but it was necessary to maintain a decent flow in the document. 🤷‍♂️ 

PTAL!

explainer.md
```diff
@@ -147,17 +145,51 @@ function OnSessionStarted(session) {
...
+### Setting up a VRLayer
+
+The content to present to the device is defined by a [`VRLayer`](https://w3c.github.io/webvr/#interface-vrlayer). In the initial version of the spec only one layer type, `VRCanvasLayer`, is defined and only one layer can be used at a time. This is set via the `VRSession.baseLayer` attribute. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
+
+In order for a WebGL canvas to be used with a `VRCanvasLayer`, it's context must be _compatible_ with the `VRDevice`. This can mean different things for different environments. For example, on a desktop computer this means that the context should be created against the graphics adapter that the `VRDevice` is physically plugged into. On most mobile devices though, that's not a concern and so the context will always be compatible. In either case, the WebVR application will need to take steps to ensure WebGL context compatibility before using it with a `VRLayer`.
+
+When it comes to ensuring canvas compatibility there's a two broad categories that apps will fall under.
```
NellWaliczek: there's a two -> there's two

explainer.md
```diff
@@ -147,17 +145,51 @@ function OnSessionStarted(session) {
...
+}
+```
 
-  // Start the render loop
-  vrSession.commit().then(OnFirstVRFrame);
+### Setting up a VRLayer
+
+The content to present to the device is defined by a [`VRLayer`](https://w3c.github.io/webvr/#interface-vrlayer). In the initial version of the spec only one layer type, `VRCanvasLayer`, is defined and only one layer can be used at a time. This is set via the `VRSession.baseLayer` attribute. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
+
+In order for a WebGL canvas to be used with a `VRCanvasLayer`, it's context must be _compatible_ with the `VRDevice`. This can mean different things for different environments. For example, on a desktop computer this means that the context should be created against the graphics adapter that the `VRDevice` is physically plugged into. On most mobile devices though, that's not a concern and so the context will always be compatible. In either case, the WebVR application will need to take steps to ensure WebGL context compatibility before using it with a `VRLayer`.
```
NellWaliczek: on a desktop computer this means that the context should be created -> on a desktop computer this may mean the context must be created

NellWaliczek: the WebVR application will need to take  -> the WebVR application must take

NellWaliczek: Perhaps...
compatibility before using it with a `VRLayer`. ->compatibility before using it with a `VRCanvasLayer`.

explainer.md
```diff
@@ -147,17 +145,51 @@ function OnSessionStarted(session) {
...
 }
 ```
 
+**VR Centric:** The app's primary use case is VR, and as such it doesn't mind initializing resources in a VR-centric fashion, which may include asking users to select a headset as soon as the app starts. An example would be a game which is dependent on VR presentation and input. These types of applications can to avoid the need to call `ensureContextCompatibility` and the possible context loss that it may trigger by passing the `VRDevice` that the context will be used with as a context creation argument.
+
+```js
+let gl = glCanvas.getContext("webgl", { compatibleVrDevice: vrDevice });
+```
+
+Ensuring context compatibility with a `VRDisplay` through either method may have side effects on other graphics resources in the page, such as causing the entire user agent to switch from rendering using an integrated GPU to a discreet GPU.
```
NellWaliczek: Is there a way to make this last sentence more visible? Since this could have a big impact, it's pretty important it not get lost in the explainer ^_^

RafaelCintron: One disadvantage of having the developer pass the VRDevice as a context creation attribute and/or calling ensureContextCompatibility is that multiple documents on the page can fight about the GPU if each one passes a different VRDevice to these two mechanisms.  Perhaps we only allow one selection at a time per document.  For iFrames, we can use the permissions API to mitigate this.

explainer.md
```diff
@@ -147,17 +145,51 @@ function OnSessionStarted(session) {
...
+
+This style of application should call `VRDisplay.ensureContextCompatibility` with the WebGL context in question. This will set a compatibility bit on the context that allows it to be used. Contexts without the compatibility bit will fail when attempting to create a `VRLayer` with them. In the event that a context is not already compatible with the `VRDisplay` the [context will be lost and attempt to recreate itself](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.13) using the compatible graphics adapter. It is the page's responsibility to handle WebGL context creation loss properly, recreating any necessary WebGL resources in response. If the context loss is not by the page handled the promise returned by `ensureContextCompatibility` will fail. The promise may also fail for a variety of other reasons, such as the context being actively used by a different, incompatible `VRDevice`.
```
toji: Oof. 😫 Yoda my fingers think I am.

toji:
Thanks for calling out my atrocious grammar. 😅 Updated with the above suggestions.

toji:
Given the lack of negative feedback I'm going to merge this. As always, nothing is final yet so please file issues or follow up pull requests if there's some aspect of this change that needs clarification.

PR #229, opened 2017-05-11
Handling non-WebGL contexts for requestPresent and submitFrame

dlibby-:
Update spec per discussion at https://github.com/w3c/webvr/issues/153
 - requestPresent should reject for non-WebGL contexts, as they are not supported.
 - If the VRSource has no context, submitFrame will throw at a later point if the context was later created (i.e. after requestPresent) as a non-WebGL context


spec/1.1/index.bs
```diff
@@ -210,7 +210,7 @@ Functionally equivalent to <a href="https://www.w3.org/TR/html51/webappapis.html
...
+Begins presenting the contents of the specified {{VRLayerInit}} array on the {{VRDisplay}} and fulfills the returned promise when presentation has begun. If {{canPresent}} is false the promise MUST be rejected. If the {{VRLayerInit}} array contains more than {{maxLayers}} elements the promise MUST be rejected. If any of the {{VRSource}} properties specified in the {{VRLayerInit}} array contain a context that is not WebGL (i.e. WebGLRenderingContextBase) the promise MUST be rejected. If {{requestPresent()}} is called outside of an engagement gesture the promise MUST be rejected unless the {{VRDisplay}} was already presenting. This engagement gesture is also sufficient to allow <a href="http://w3c.github.io/pointerlock/">requestPointerLock</a> calls until presentation has ended. The user agent MAY reject the promise for any other reason. If the {{VRDisplay}} is already presenting when {{requestPresent()}} is called the {{VRDisplay}} SHOULD update the {{VRLayerInit}} list being presented. If a call to {{requestPresent()}} is rejected while the {{VRDisplay}} is already presenting the {{VRDisplay}} MUST end presentation.
```
cvan: to improve ambiguity, I'd recommend adding commas for these conditional sentences:

- `is false the` -> `is false, the`
- `elements the promise` -> `elements, the promise`
- `not WebGL (i.e. WebGLRenderingContextBase) the promise` -> `not WebGL (i.e., WebGLRenderingContextBase), the promise`
- `gesture the promise` -> `gesture, the promise`
- `user agent` -> `UA` (?)
- `is called the` -> `is called, the`
- `presenting the` -> `presenting, the`

spec/1.1/index.bs
```diff
@@ -219,7 +219,7 @@ Ends presentation to the {{VRDisplay}} and fulfills the returned promise when fu
...
-Captures the current state of the {{VRLayerInit}} currently being presented and displays it on the {{VRDisplay}}. It is assumed that the frame was rendered using the {{VRPose}} and matrices provided by the last call to {{getFrameData()}}. If {{getFrameData()}} was not called prior to calling {{submitFrame()}} the user agent MAY warn the user of potentially malformed visuals or prevent the frame from being shown at all.
+Captures the current state of the {{VRLayerInit}} currently being presented and displays it on the {{VRDisplay}}. It is assumed that the frame was rendered using the {{VRPose}} and matrices provided by the last call to {{getFrameData()}}. If {{getFrameData()}} was not called prior to calling {{submitFrame()}} the user agent MAY warn the user of potentially malformed visuals or prevent the frame from being shown at all. If any {{VRSource}} that was specified in the {{VRLayerInit}} array at {{requestPresent()}} time had no context, but a non-WebGL context was created subsequently after the fact, a NotSupportedError DOMException MUST be thrown, and the {{VRDisplay}} MUST NOT be updated with the contents of the layers specified by {{VRLayerInit}}.
```
cvan: `displays it on` -> `displayed to`
`to calling {{submitFrame()}} the user agent` -> `to calling {{submitFrame()}}, the UA`

spec/1.1/index.bs
toji: FWIW in this version of the spec we've been using "User Agent" (capitalized) instead of UA, so let's stick with that for consistency. Might want to re-evaluate for the 2.0 spec.

The rest of the comments seem fine.

spec/1.1/index.bs
toji: Sorry, I don't agree with the suggestions. For one, neither is a change that was introduced by this PR.

With the first suggestion the full sentence would read:

> Captures the current state of the VRLayerInit currently being presented and displayed to the VRDisplay.

 "presented and displayed to" sounds like we're repeating ourselves. Granted it could probably be clarified since there's obviously some confusion here, but I don't think that change would help.

With the second suggestion the added comma is fine but, as established above, we should use "User Agent" in the 1.1 spec for consistency.

toji:
LGTM, thanks!

PR #230, opened 2017-05-11
Update WebVR 1.1 spec to account for multi-GPU handling

dlibby-:
Following the language from the PR for the 2.0 explainer (https://github.com/w3c/webvr/pull/226), this adds a paragraph that describes the expected sequence of events during requestPresent, in the event that a WebGL context is not compatible with the VRDisplay's graphics adapter. The webglcontextlost will be fired, and if Event.preventDefault() is called, the user agent will at that point switch to the VRDisplay's graphics adapter and continue the requestPresent process.


spec/1.1/index.bs
```diff
@@ -212,6 +212,8 @@ Passing the value returned by {{requestAnimationFrame()}} to will unregister the
...
...the requestPresent promise is outstanding, the <a href="https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.13">context will be lost and attempt to recreate itself</a> using the compatible graphics adapter. If the user agent requests user consent in response to {{requestPresent()}}, the context lost operation MUST NOT happen prior to the user granting consent. It is the webpage's responsibility to handle WebGL context loss properly, calling preventDefault() during webglcontextlost and recreating any necessary WebGL resources during the webglcontextrestored event. If the context lost event is not handled (i.e. preventDefault was not called during the firing of webglcontextlost) the {{requestPresent()}} promise MUST be rejected. If the context lost event is handled, all other WebGL contexts on the page MAY also have their context lost as byproduct of using the correct graphics adapter instead of the default one. The promise may also fail for a variety of other reasons, such as the context being actively used by a different, incompatible {{VRDisplay}}. The fulfillment of the {{requestPresent()}} promise implies that the WebGL context is now compatible with the {{VRDisplay}}.
```
cvan: few grammatical nits:

- `context lost operation` -> `context-lost operation`
- `user granting consent` -> `user's granting consent`
- `webpage's` -> `web page's` (or `document's`)
- `context lost event` -> `context-lost event`
- `(i.e. preventDefault` -> `(i.e., preventDefault`
- `the {{requestPresent()}}` -> `, the {{requestPresent()}}`
- `context lost event is handled` -> `context-lost event is handled`

toji: Just a couple notes:

> If the context lost event is not handled (i.e. preventDefault was not called during the firing of webglcontextlost) the {{requestPresent()}} promise MUST be rejected.

Maybe change this to something like "If the `webglcontextlost` event is fired but preventDefault is not called the {{requestPresent()}} promise MUST be rejected." to reinforce the idea that the context won't be lost every single time.

> If the user agent requests user consent in response to {{requestPresent()}}, the context lost operation MUST NOT happen prior to the user granting consent.

This feels too early in the paragraph. Might be more natural before "If the context lost event is handled, all other WebGL..."

But other than that and @cvan's feedback, looks good.

toji:
LGTM, thanks!

PR #233, opened 2017-05-23
Change VRWebGLLayer to take a context rather than a canvas

toji:
Given the recent changes in the render loop there no longer seems to be
a reason to track the canvas itself, so I propose here switching the
VRWebGLLayer to accept the rendering context instead.

Also did some research and found that, as @kenrussell warned, we can’t use
WebGLRenderingContextBase in our IDL directly. So instead we get to
play with more typedefs! *sigh*

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -172,7 +172,7 @@ function setupWebGLLayer() {
     // The content that will be shown on the device is defined by the session's
     // baseLayer. In non-exclusive The baseLayer is not used for presentation, but
     // the canvas dimensions are used to construct the projection matrices.
-    vrSession.baseLayer = new VRWebGLLayer(vrSession, glCanvas);
+    vrSession.baseLayer = new VRWebGLLayer(vrSession, gl);
   });
 }
 ```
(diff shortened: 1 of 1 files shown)


toji:
Rebased to resolve conflicts with #232

kearwood:
This LGTM.

Only feedback may be naming of 'VRWebGLRenderingContext'.  Today this name makes sense and is easily understandable; however, if future web rendering API standards are not called "WebGL", would we want to use the same interface?

Would it make sense to call it something more agnostic, such as 'VRRenderingContext'?

NellWaliczek:
@kearwood That thought crossed my mind too... Though it also occurred to me that we'd probably need a new layer type at that point, so I figured it was ok.  What do you think?

toji:
I wondered the same thing, but I feel like we want to differentiate WebGL contexts from WhateverComesNext contexts. Specifically, it wouldn't make sense to have a `VRWebGLLayer` accept a generic `VRRenderingContext` if said context eventually expanded to include `WebMetalVulkan12`. I anticipate we'd add a new layer type at that point to accommodate the new context type.

`ensureContextCompatibility` is less straightforward, but could be handled via overloading or by introducing another typedef that _does_ include all the context types. (I'm of the opinion that overloading is cleaner in that case.)

kearwood:
This sounds reasonable to me.  A next-generation web graphics API is also likely to need different attributes than what `VRWebGLLayer` provides.  The corresponding next-generation layer type could use the same language as the new graphics api to resolve ambiguity.

RafaelCintron:
LGTM also.  

Not only will WebGLNext have a new context type, it will likely have a different "framebuffer" type as well.

toji:
Great, sounds like everyone commenting is on the same page. Merging!

PR #234, opened 2017-05-24
Make VRView.eye non-nullable

toji:
This struck me as a bit odd when I saw it, so I asked around and it
seems like only one other web spec has a nullable enum (WebPayments)
and it’s not clear if it’s intentional or necessary that that was the
case. As such I think we should stick with the majority and make this
non-nullable.

That of course leaves a question of “What do we return when there is no
intrinsically associated eye?” My vote would be: Just return “left” and
be done with it. This attribute will probably only ever be used when
rendering content with pre-rendered stereo, and such content will
always need to pick an eye to present regardless of display mechanism.
(Realtime content generally won’t care about which eye is being
rendered, and will just use the prescribed view matrices.) By making
the API choose a reasonable default we make the cases where it’s needed
that much easier for developers.

Happy to hear if there’s scenarios I haven’t considered that require
more nuance. If there is I’m in favor of adding a “none”, “unknown”, or
<empty string> entry to the enum rather than leave it nullable.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -618,7 +618,7 @@ enum VREye {
 };
 
 interface VRView {
-  readonly attribute VREye? eye;
+  readonly attribute VREye eye;
   readonly attribute Float32Array projectionMatrix;
 
   VRViewport? getViewport(VRLayer layer);


NellWaliczek:
If I'm remembering right, the original reasoning was to support both magic window and CAVE systems.  My inclination is to make it non-nullable as you've suggested but add a "" like the gamepad mapping enum.  Thoughts?

toji:
I could live with "". But with both the magic window and CAVE systems (presumably mono) if I loaded up a WebVR app that rendered a stereo video, I'd still have to make a judgement call about which eye's stream I show. It could be nice to trust that the API will give you one by default. Truth be told in either case we'll probably end up with most developers writing code like this:

```js
if (view.eye == "left") {
  drawVideoSubrect(0.0, 0.0, 1.0, 0.5);
} else {
  drawVideoSubrect(0.0, 0.5, 1.0, 1.0);
}
```

As such, the interesting question is simply "Is there something useful the developer can do if they know the view is not explicitly associated with any eye?" I can't think of any off the top of my head.

toji:
Given that this is a *one character change* that doesn't have any opposition on the basic principle I'm going to merge it now. Adding a "" or "none" to the enum is a pretty simple follow up change if we feel it's needed.

PR #235, opened 2017-05-30
Proposed Name change: ensureContextCompatibility -> setCompatibleVrDevice

toji:
While reviewing the current state of the API a few days ago I had a request to try and make the naming of the `compatibleVrDevice` context attribute and `ensureContextCompatibility` function more similar to enforce the fact that they both set the same internal state. I don't personally see it as being problematic, but I understand the request.

This is my proposed change, though in practice `vrDevice.setCompatibleVrDevice(canvas)` sounds a bit repetitive and slightly backwards. Very happy to hear other suggestions!


NellWaliczek:
It definitely seems weird to have the naming be inconsistent, but even moreso perhaps what's weird is that the functionality is on inconsistent types.  What if we renamed it to setCompatibleVRDevice and moved it off of VRDevice and put it on the WebGLRenderingContextBase?  That way both functions will be on the context

toji:
I can get behind that. I'm generally not a fan of duct taping new methods into one API from the spec of another, but in this case I think it's probably justified. I'll update the pull request.

toji:
Given there's been no further comments on this, I'm going to merge.

PR #236, opened 2017-05-31
Allow multiple active sessions at once

toji:
Step one in working towards a more robust magic window solution now that
we've decided it should be part of 2.0

This change removes any concepts that assume a device may only have a
single active session at a time. This allows multiple non-exclusive
VRSessions to be active at once, which will be important for enabling
various interesting magic window scenarios in the future. With this
change starting an exclusive session suspends the non-exclusive ones,
but doesn’t end them.

IDL Changes:
 - Remove VRDevice.activeSession
 - VRDevice.onsessionchange -> VRSession.onended

Comments welcome! Working on some follow up PRs to this one that should
provide more context if this seems confusing. I wanted to avoid having a
single huge "Magic Window" explainer overhaul, though.


NellWaliczek:
Hey @toji,
I'm curious about what scenario is motivating this change?  We'd talked about single vs. multiple sessions before, but I thought we'd come to the conclusion only one was needed?  I'm a bit nervous about how this might relate to input routing and security considerations, but if there's something concrete motivating the change let's talk through it!

toji:
@NellWaliczek and I talked directly about this, but to recap for the benefit of the rest of the group:

Motivation for allowing multiple non-exclusive sessions is (mostly) galleries of magic window content in a world where each non-exclusive session is bound to an output canvas as suggested by #237. This is especially important when dealing with a "punch through" magic window scenario, since each canvas will need it's own subtly different projection and view matrices.

More generally, it would be fairly limiting if you could only manage one magic window piece of content per page. Imagine if a third party library creates one without you being aware, at which point your request for a non-exclusive session just starts failing for unknown reasons.

Input routing IS a problematic case, but mostly for inputs that have no implicit ray/direction associated with them: standard gamepads, bluetooth clickers, etc. Touch and mouse inputs have pretty obvious mappings to a specific session/canvas pair, as would using a VR controller to interact with a 2D page viewed in VR. (The pick ray will cast into a specific canvas.)

If Nell wants to fill in any relevant details of our discussion that I've forgotten I'd appreciate it!

toji:
Rebased this against the recent master changes, would still like to land it. Again, the primary motivator is the idea that pages may want multiple magic window elements active at once, and if we only allow a single active session at a time that would both discourage potentially interesting use cases AND force motivated developers to fall back to potentially ugly hacks (like full-screen canvases with carefully scissored content rects. 🤢)

Happy to hear any and all feedback!

toji:
I'm a little concerned about the lack of feedback here (outside of a conversation with Nell), especially given that this PR has been open for 3 weeks. I feel like this is a pretty important change that will ultimately make the API much more useful and would like to land it soon.

Would love to hear from @kearwood, @DigiTec, or @mkeblx for example. Otherwise I guess I'll assume that lack of protest is a positive sign?

kearwood:
Hi @toji !  I've had a chance to review and find this to be a good model, compatible with concepts we have been exploring on our end.

Would it be worth expanding this to explain the behavior of link traversal when changing urls in an active exclusive or non-exclusive session?

Perhaps we could say something to the effect of..  "If the content presenting an exclusive session changes the URL, the destination page is also granted an exclusive session without a user gesture in order to create a seamless link traversal without breaking immersion.  If content does not have an active WebVR session or has a non-exclusive WebVR session and changes the URL, a user gesture is required to grant permission to the new page when it requests an exclusive session."

toji:
I don't mind the intent behind the link traversal text you gave. Looking at the entire navigation section we've got right now though reminds me that it's been woefully neglected and needs a much larger overhaul. That should probably be handled in a separate pull request. I can start working on that, keeping your suggestion in mind.

kearwood:
@toji Sounds good to me.  I'd be glad to help with the overhaul on the link traversal section.

explainer.md
```diff
@@ -76,9 +76,15 @@ navigator.vr.getDevices().then(devices => {
...
 
-In this case, we only ask if the ability to have `exclusive` access (Which includes the ability to display imagery on the headset). Note that `exclusive: true` is actually the dictionary default, and so does not need to be specified here. It’s made explicit in this example for clarity.
+Sessions can be created with two levels of access:
+
+**Exclusive Access**: The default mode, but can be explicitly requested with the `exclusive: true` dictionary argument. Exclusive sessions present content directly to the `VRDevice`, enabling immersive VR presentation. Only one exclusive session is allowed at a time across the entire UA. Exclusive sessions must be created within a user gesture event.
```
ddorwin: The "Only one" statement needs to be more specific. A single UA could have multiple if there are multiple HMDs.

It can't be per `VRDevice` because there may be many multiple `VRDevice` objects in the UA backed by a single HMD. Is how this works described in this doc? If not, does it need to be? On a related note, what happens if there is already an exclusive session and a different document requests exclusive access?

explainer.md
```diff
@@ -76,9 +76,15 @@ navigator.vr.getDevices().then(devices => {
...
+
+**Non-Exclusive Access**: Requested with the `exclusive: false` dictionary argument. Non-exclusive sessions do not have the ability to display anything on the `VRDevice`, but are able to access device tracking information and use it to render content on the page. This technique, where a scene rendered to the page is responsive to device movement is sometimes referred to as "Magic Window" mode. It's especially useful for mobile devices, where moving the device can be used to look around a scene. Devices like Tango phones and tablets with 6DoF tracking capabilities may expose them via non-exclusive sessions even if the hardware is not capable of immersive, stereo presentation. Multiple non-exclusive sessions can be active at once, but all non-exclusive sessions are suspended when an exclusive session is active. Non-exclusive sessions are not required to be created within a user gesture event.
```
ddorwin: Similarly, how are requests from multiple documents across the UA handled?

ddorwin: What restrictions are there on which frames can request non-exclusive access at the same time? Are multiple iframes allowed to request access at once? Must they be same-origin to each other or the top-level document?

Requiring that the frame receiving poses currently has focus is an important privacy mitigation, especially because there is no gesture requirement. Allowing multiple frames to receive poses could be problematic.

ddorwin: nit: Remove the comma before "but."

explainer.md
ddorwin: It is almost certain that a user gesture event will not be required in some cases, such as the `activate` handler. The current explainer text says, "The activate event acts as a user gesture..." I don't think we want to treat that as a generic gesture as has previously been discussed. Should we file a separate bug to clean up gesture-related language throughout?

explainer.md
```diff
@@ -76,9 +76,15 @@ navigator.vr.getDevices().then(devices => {
...
+
+In this case, we ask if the `VRDevice` supports sessions with `exclusive` access, since we want the ability to display imagery on the headset.
```
ddorwin: This text location and example no longer fit the flow of this section.

explainer.md
```diff
@@ -255,29 +257,26 @@ To stop presenting to the `VRDevice`, the page calls [`VRSession.endSession`](ht
...
+  vrSession = null;
+
   // Ending the session stops executing callbacks passed to requestFrame().
   // To continue rendering, use the window's AnimationFrame callback
   window.requestAnimationFrame(onDrawFrame);
 }
 ```
 
-In addition to the application ending the session manually, the UA may force the session to end at any time for a variety of reasons. Well behaved applications should monitor the `sessionchange` event on the `VRDevice` to detect when that happens.
+In addition to the application ending a session manually, the UA may force sessions to end at any time for a variety of reasons. Well behaved applications should monitor the `ended` event on the `VRSession` to detect when that happens.
```
ddorwin: Maybe this is where some of the scenarios I mentioned above should be covered.

explainer.md
```diff
@@ -445,17 +444,19 @@ Similarly, the `deactivate` event can be used to detect when the user removes th
...
+  if (vrSession) {
+    vrSession.endSession().then(OnSessionEnded);
   }
 });
 ```
 
 ### Responding to a suspended session
 
-The UA may temporarily suspend a session if allowing the page to continue reading the headset position represents a security risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. While suspended the page may either refresh the vr device at a slower rate or not at all, and poses queried from the headset may be less accurate. The UA is expected to present a tracked environment to the user when the page is being throttled to prevent user discomfort.
+The UA may temporarily suspend a session if allowing the page to continue reading the headset position represents a security risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. Additionally, non-exclusive sessions are suspended while an exclusive session is active.
```
ddorwin: security *or privacy* risk

explainer.md
```diff
@@ -445,17 +444,19 @@ Similarly, the `deactivate` event can be used to detect when the user removes th
...
+
+While suspended the page may either refresh the vr device at a slower rate or not at all, and poses queried from the device may be less accurate. If the user is wearing a headset the UA is expected to present a tracked environment when the page is being throttled to prevent user discomfort.
```
ddorwin: Should this be normatively specified? It could be problematic if some UAs provide throttled poses but others provide none. Also, it may be useful to provide clear privacy/security guidance in the spec. This will at least ensure the implementers have the benefit of the collective community.

ddorwin: Explain what a tracked environment is.

explainer.md
```diff
@@ -445,17 +444,19 @@ Similarly, the `deactivate` event can be used to detect when the user removes th
...
 vrDevice.addEventListener('deactivate', vrDeviceEvent => {
-  if (vrDevice.activeSession) {
-    vrDevice.activeSession.endSession().then(OnSessionEnded);
+  if (vrSession) {
+    vrSession.endSession().then(OnSessionEnded);
   }
 });
 ```
 
 ### Responding to a suspended session
```
ddorwin: It may help address comments above and below to add a separate section that discusses when a UA should suspend a session. That could be referenced both above and from here.

explainer.md
toji: I know we've talked about it off and on in the past, but I don't think I've ever gotten a strong sense of a beneficial scenario where a single piece of hardware is exposed as multiple `VRDevice`s. Did you have something in mind?

Regardless of that, though, you're right that this needs clarification.

toji: Good point, and yes. That's probably worth it's own bug.

explainer.md
toji: This definitely needs to be expanded on. I'll add a section dedicated to session security.

explainer.md
ddorwin: I simply meant that multiple frames or documents could call `getDevices()` and have a `VRDevice` referencing the same HMD. For my first paragraph, I think we need to say something like there can only be one exclusive session for each underlying device. For the second paragraph, we probably need some text about conflicts. For example, an exclusive request when there is already an active exclusive session for the underlying device somewhere else in the UA (or client) will fail. Maybe there is language in WebUSB or Web Bluetooth that we can use as a model.

ddorwin: I filed #248.

explainer.md
toji: I think it's going to depend on the hardware/API/OS capabilities and the privacy needs. I'd like to address this more directly, but it's a big enough topic that it's probably worth a whole new issue and/or PR.

explainer.md
toji: Agreed, but probably as a new PR.

explainer.md
toji: Ah, got it. Thanks for the clarification.

toji:
Thanks for all the feedback @ddorwin! Addressed a bunch of it, but I've deferred some of the security-centric additions to a different PR for the sake of easier review.

explainer.md
```diff
@@ -52,9 +52,10 @@ The basic steps any WebVR application will go through are:
...
+[`navigator.vr.getDevices`](https://w3c.github.io/webvr/#navigator-getvrdevices-attribute) returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a list of available devices. Each [`VRDevice`](https://w3c.github.io/webvr/#interface-vrdevice) represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktops this will usually be a headset peripheral; on mobile devices it may represent the device itself in conjunction with a viewer harness (e.g., Google Cardboard or Samsung Gear VR). It may also represent devices without stereo presentation capabilities but more advanced tracking, such as Tango devices.
```
ddorwin: Separate issue: These links don't work anymore. Why are they absolute? I guess `../../spec/latest/` might work.

ddorwin: nits:
"On desktop clients,"
"on mobile devices,"

ddorwin: Does the spec currently require that all VRDevices "can present imagery to the user somehow"? For example, in a pure magic window case where there is no headset, the VRDevice is not actually capable of presenting imagery.

explainer.md
```diff
@@ -74,15 +75,20 @@ navigator.vr.getDevices().then(devices => {
...
 
-If a `VRDevice` is available and has the appropriate capabilities the page will usually want to add some UI to trigger activation of "VR Presentation Mode", where the page can begin sending imagery to the device. Testing to see if the device supports the capabilities the page needs is done via the `supportsSession` call, which takes a dictionary of the desired functionality and returns whether or not the device can create a session supporting them. Querying for support this way is necessary because it allows the application to detect what VR features are available without actually engaging the sensors or beginning presentation, which can incur significant power or performance overhead on some systems and may have side effects such as launching a VR status tray or storefront.
+### Sessions
+
+A `VRDevice` indicates the presence of a VR hardware device but provides very little information about it outside of a name that could be used to select it from a list. In order to do anything that involves the hardware's presentation or tracking capabilities the page will need to request a [`VRSession`](https://w3c.github.io/webvr/#interface-vrsession) from the `VRDevice`.
```
ddorwin: It's nice to somehow indicate that you are referencing a definition and link to that definition. This may not be easy to do in MD, but in the spec, you'd probably want to make "VR hardware device" a link.

ddorwin: nits: "...capabilities, the application must request..."

I believe "application" should be used instead of "page" in nearly all cases. This can be addressed separately.

explainer.md
```diff
@@ -74,15 +75,20 @@ navigator.vr.getDevices().then(devices => {
...
 
 Sessions can be created with two levels of access:
```
ddorwin: nit: "one of two"

explainer.md
```diff
@@ -74,15 +75,20 @@ navigator.vr.getDevices().then(devices => {
...
+
+### Detecting and advertising VR mode
+
+If a `VRDevice` is available and able to create an exclusive session the page will usually want to add some UI to trigger activation of "VR Presentation Mode", where the page can begin sending imagery to the device. Testing to see if the device supports the capabilities the page needs is done via the `supportsSession` call, which takes a dictionary of the desired functionality and returns whether or not the device can create a session supporting them. Querying for support this way is necessary because it allows the application to detect what VR features are available without actually engaging the sensors or beginning presentation, which can incur significant power or performance overhead on some systems and may have side effects such as launching a VR status tray or storefront.
```
ddorwin: nit: "...session, the application..."

explainer.md
ddorwin: This comment was resolved by moving the text above.

Does it make sense to say "In the following example" instead? "In this case" usually refer to something in the preceding text.

explainer.md
```diff
@@ -105,7 +111,7 @@ async function OnVRAvailable() {
...
+Clicking that button will attempt to acquire a `VRSession` by callling `VRDisplay.requestSession`. This returns a promise that resolves to a `VRSession` upon success. When requesting a session the capabilities that the returned session must have are passed in via a dictionary, exactly like the `supportsSession` call. If `supportsSession` returned true for a given dictionary then calling `requestSession` with the same dictionary values should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user gesture for an exclusive session or another page currently having an exclusive session for the same device.)
```
ddorwin: You might want to use "VR hardware device" at the end.

nits: Commas after "When requesting a session" and before "then."

explainer.md
ddorwin: Okay, please file the new issue since you have a better idea of what it should be.

explainer.md
toji: Hmm... I think that's probably a matter of semantics. (I tend to think of Magic Window as "Presenting imagery to the user") but the point stands that this is confusing.

toji: Ah, I do recall that we wanted to call this out somehow to differentiate a VR headset from a VR controller. Not sure how to mitigate that...

toji:
Made some more updates to take into account @ddorwin's continued feedback. (Thanks!) Didn't strip out URLs yet or do full-doc page->application replace. Would prefer to do those separately.

explainer.md
ddorwin: Speaking of semantics, is a phone "a physical unit of VR hardware?" ;)

explainer.md
```diff
@@ -91,19 +103,15 @@ async function OnVRAvailable() {
...
-
-Requesting a new type of session will end any previously active ones.
+Clicking that button will attempt to acquire a `VRSession` by callling `VRDisplay.requestSession`. This returns a promise that resolves to a `VRSession` upon success. When requesting a session, the capabilities that the returned session must have are passed in via a dictionary, exactly like the `supportsSession` call. If `supportsSession` returned true for a given dictionary, then calling `requestSession` with the same dictionary values should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user gesture for an exclusive session or another application currently having an exclusive session for the same VR hardware device.)
```
NellWaliczek: "another application currently having an exclusive session for the same VR hardware device" isn't true for all VR platforms.  :)  Can we reword this to be a bit clearer by perhaps removing this reason and just saying that it's up to the underlying platform to decide if it can honor the request?

explainer.md
```diff
@@ -112,9 +120,9 @@ function BeginVRSession(isExclusive) {
   vrDevice.requestSession({ exclusive: isExclusive })
       .then(OnSessionStarted)
       .catch(err => {
-        // May fail for a variety of reasons, including another page already
-        // having exclusive access to the device. Probably just want to render
-        // the scene normally without any tracking at this point.
+        // May fail for a variety of reasons, including another application
```
NellWaliczek: Same comment as above

explainer.md
```diff
@@ -445,17 +451,19 @@ Similarly, the `deactivate` event can be used to detect when the user removes th
...
+  if (vrSession) {
+    vrSession.endSession().then(OnSessionEnded);
   }
 });
 ```
 
 ### Responding to a suspended session
 
-The UA may temporarily suspend a session if allowing the page to continue reading the headset position represents a security risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. While suspended the page may either refresh the vr device at a slower rate or not at all, and poses queried from the headset may be less accurate. The UA is expected to present a tracked environment to the user when the page is being throttled to prevent user discomfort.
+The UA may temporarily suspend a session if allowing the page to continue reading the headset position represents a security or privacy risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output. Additionally, non-exclusive sessions are suspended while an exclusive session is active.
```
NellWaliczek: Why suspend non-exclusive sessions rather than end them?

explainer.md
```diff
@@ -445,17 +451,19 @@ Similarly, the `deactivate` event can be used to detect when the user removes th
...
 
-In general the page should continue drawing frames in response to `VRSession.requestFrame()` callbacks. The UA may use these frames as part of it's tracked environment, though they may be partially occluded, blurred, or otherwise manipulated. Still, some applications may wish to respond to this suspension by halting game logic, purposefully obscuring content, or pausing media. To do so, the application should listen for the `blur` and `focus` events from the `VRSession`. For example, a 360 media player would do this to pause the video/audio whenever the UA has obscured it.
+While suspended the page may either refresh the vr device at a slower rate or not at all, and poses queried from the device may be less accurate. If the user is wearing a headset the UA is expected to present a tracked environment (a scene which remains responsive to user's head motion) when the page is being throttled to prevent user discomfort.
```
NellWaliczek: What does this mean in the context of a suspended session? "If the user is wearing a headset the UA is expected to present a tracked environment (a scene which remains responsive to user's head motion) when the page is being throttled to prevent user discomfort."

explainer.md
toji: Thinking about the SketchFab gallery scenario: if you have a grid of magic windows, then click on one of them to jump into immersive mode, then exit again, you ideally want to be back to the same grid of magic windows. That's still do-able if the sessions end upon creating an exclusive session, but it's much more of a pain for page authors. Suspensions keeps the basic wiring around without requiring that the frame loop be pumped.

toji:
In the WebVR implementors call today we agreed that it would be good to get the majority of the content from this PR merged in, but defer the bit that explicitly states that multiple non-exclusive sessions are allowed at once since there's still some questions around that. We _should_ however keep the removal of elements that would explicitly prevent that in the future (such as the `activeSession` attribute.) I've made that change, and am merging now.

Next, a followup PR should focus on clearly defining the terms "suspended" and "ended" and the conditions under which a session could be either suspended or ended.

Issue #224, opened 2017-04-27
Proposal: Mirroring/Magic Window


toji:
We need to have a clear idea of how mirrored and magic window content get processed and displayed on the page before we wrap up the spec, even though most of the complexity won't come into play until we get to a multi-layered world further down the road. (The terms "Mirroring" and "Magic Window" as used in this issue are defined in #222)

(This issue assumes that the managed framebuffer feature described in #218 will be part of the spec in some form)

## Challenges: ##
While a simple, single layer magic window mode could be managed by simply allowing developers to draw to a canvas backbuffer that method falls apart when support multiple layers are introduced. At that point, if the developer wants to accurately display the same contents in magic window mode that the user would see in the headset they either have to:

 - Reproduce the rendering logic for all layer types in use and render them manually or
 - Rely on the UA to do the compositing and render the output to the screen.

From a developer's point of view the second option is obviously preferable, and this issue will make the assumption that the first option is untenable. This puts magic window rendering in the same category as mirroring in that they both now rely on the UA to composite the various inputs and provide that composited content back to the page. The key differences between the two modes are:

 - When mirroring the on-screen content does not need to be as high quality or low latency as the in-headset content. Mirrored views should be allowed to downscale content, crop output to a more natural FOV, drop frames, and display stale frames because the intent is only to provide external viewers an idea of what the user is seeing in the headset. Whether or not that quality should be controllable by the developer is an open question. Not needed at all on devices without an external monitor
 - When in Magic Window mode the content quality should be under developer control and treated more strictly, since it's now the primary method of content consumption.

Native API support is also worth considering. In many cases it would be desirable for UAs to map certain layer types to a native API compositor feature for performance or quality purposes, but in order to do that fully the native API would also have to have a way to provide it's own mirrored view of the final composited content. Of the desktop APIs that I've looked at both OpenVR and the Oculus API have a method for retrieving a mirror texture for this exact use. (It's worth noting that the OpenVR mirror texture, at least, also includes chaperone visualization and system menus, so using it directly COULD represent a security concern if not handled carefully.) OpenXR is still under development but has not committed to making a mirror texture available, and WinMR doesn't seem to have any mirroring APIs but that's likely because it also doesn't appear to have any form of native layering. Mobile APIs don't have any applicable mirroring concessions for obvious reasons.

[...]

PR #237, opened 2017-06-01
Detail magic window support

toji:
Described how magic window mode works with the `outputCanvas` proposal.
Also, because they’re natural variants of the same mechanism, described
mirroring and tracking only behavior. Follows the proposal in #224.

Also changed VRPresentationFrame -> VRFrame because if we moved forward
with the tracking only portion of this there would be scenarios in
which the VRPresentationFrames had no Presentation component.


toji:
Updated this pull request after some conversations with @NellWaliczek to drop the concept of a "Tracking only" session for the moment. While I'd like to discuss that further in the future it's the least important part of this proposal and I don't care for it to complicate things.

Very keen to get feedback, since I consider this to be one of the last big issues to work out prior to TAG review and starting on the spec proper.

explainer.md
```diff
@@ -108,10 +108,10 @@ async function OnVRAvailable() {
 Clicking that button will attempt to initiate a [`VRSession`](https://w3c.github.io/webvr/#interface-vrsession), which manages input and output for the display. When creating a session with `VRDevice.requestSession` the capabilities that the returned session must have are passed in via a dictionary, exactly like the `supportsSession` call. If `supportsSession` returned true for a given dictionary then calling `requestSession` with the same dictionary values should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user gesture for an exclusive session or another page currently having an exclusive session for the same device.)
 
 ```js
-function BeginVRSession(isExclusive) {
+function BeginVRSession() {
   // VRDevice.requestSession must be called within a user gesture event
   // like click or touch when requesting exclusive access.
-  vrDevice.requestSession({ exclusive: isExclusive })
```
NellWaliczek: It seems a bit odd to pass the exclusive entry explicitly in the supportsSession() sample code and then skip it here.  Was there a specific reason for that?

explainer.md
```diff
@@ -123,7 +123,7 @@ function BeginVRSession(isExclusive) {
 ```
 Once the session has started, some setup must be done to prepare for rendering.
 - A `VRFrameOfReference` must be created to define the coordinate system in which the `VRDevicePose` objects will be defined. See the Advanced Functionality section for more details about frames of reference.
-- The depth range of the session should be set to something appropriate for the application. This range will be used in the construction of the projection matricies provided by `VRPresentationFrame`.
+- The depth range of the session should be set to something appropriate for the application. This range will be used in the construction of the projection matricies provided by `VRFrame`.
```
NellWaliczek: This naming change perhaps isn't as relevant with the removal of the non-visual sessions?

explainer.md
```diff
@@ -336,6 +325,62 @@ if(frameOfRef.bounds) {
 }
 ```
 
+### Mirroring
```
NellWaliczek: Should this go in the advanced functionality section?  On mobile devices, this won't be particularly useful and on Windows Mixed Reality headsets, there is automatic mirroring in the Mixed Reality Portal.

explainer.md
```diff
@@ -336,6 +325,62 @@ if(frameOfRef.bounds) {
...
+
+In order to mirror WebVR content to the page, developers specify an `outputCanvas` in the `VRSessionCreateParameters` of an exclusive session. The canvas supplied must have a [`ImageBitmapRenderingContext`](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmapRenderingContext) or the promise returned by `requestSession` will be rejected. Once the session has started any content displayed on the headset will then be mirrored into that canvas.
+
+When mirroring only one eye's content will be shown, and it should be shown undistorted. The UA may choose to crop the image shown, display it at a lower resolution than originally rendered, and the mirror may be multiple frames behind the image shown in the headset. The mirror may include or exclude elements added by the underlying VR system (such as visualizations of room boundaries) at the UA's discretion. Pages should not rely on a particular timing or presentation of mirrored content, it's really just for the benefit of bystanders or demo operators.
+
+On devices without an external display to mirror to, such as mobile or all-in-one systems, the `outputCanvas` will be ignored.
```
NellWaliczek: Are you operating under the assumption that external display == need to mirror?  What if a user agent/OS is already doing its own mirroring (such as the Windows Mixed Reality Portal)?  It seems like a better end user experience might come from not mirroring a second time.  It might be a bit heavy handed to say that a UA can only ignore the `outputCanvas` if it doesn't have an external display.

Of course with the current design the page has no way of distinguishing this situation.  We'd be likely to see pages always supplying an outputCanvas and then there might be a big gaping hole in the middle of the page in some UA.  Perhaps it's worth indicating the need for mirroring (as a attribute off an object returned by calls to supportsSession?)  There is still the risk that developers will ignore it though.  Hrmmmm....

explainer.md
```diff
@@ -336,6 +325,62 @@ if(frameOfRef.bounds) {
...
+
+Exclusive and non-exclusive sessions use the same render loop code, but there are some differences in behavior to be aware of. The sessions may run their render loops at at different rates. During exclusive sessions the UA runs the rendering loop at the `VRDevice`'s native refresh rate. During non-exclusive sessions the UA runs the rendering loop at the refresh rate of page (aligned with `window.requestAnimationFrame`.) Also, when rendering with an exclusive session the projection matrices provided each frame are based on the device optics, while non-exclusive sessions provide projection matrices based on the output canvas dimensions.
```
NellWaliczek: That last sentence might be accidentally a bit too restrictive.  The "punch-through" scenarios we've discussed also take into account the headpose the same as I imagine zSpace would.

explainer.md
```diff
@@ -336,6 +325,62 @@ if(frameOfRef.bounds) {
...
+
+Most instances of non-exclusive sessions will only provide a single `VRView` to be rendered, but UA may request multiple views be rendered if, for example, it's detected that that output medium of the page supports stereo rendering. As a result pages should always draw every `VRView` provided by the `VRFrame` regardless of what type of session has been requested.
```
NellWaliczek: An example here might help to clarify a lot :)

explainer.md
```diff
@@ -336,6 +325,62 @@ if(frameOfRef.bounds) {
...
+
+UAs may have different restrictions on non-exclusive contexts that don't apply to exclusive contexts. For instance, a different set of `VRFrameOfReference` types may be available with a non-exclusive session versus an exclusive session.
+
+The UA may reject requests for a non-exclusive sessions for a variety of reasons, such as the inability of the underlying hardware to provide tracking data without actively rendering to the device. Pages should be designed to robustly handle the inability to aquire non-exclusive sessions.
```
NellWaliczek: Did I miss it, or are we calling out somewhere that no user initiated action is required for non-exclusive sessions?

explainer.md
```diff
@@ -578,10 +623,12 @@ interface VRDevice : EventTarget {
 
 dictionary VRSessionCreateParametersInit {
   boolean exclusive = true;
+  HTMLCanvasElement outputCanvas = null;
```
NellWaliczek: Out of curiosity, why HTMLCanvasElement rather than the ImageBitmapRenderingContext given the hard requirement listed above?

RafaelCintron:
Here's what [`ImageBitmapRenderingContext`](https://html.spec.whatwg.org/multipage/canvas.html#the-imagebitmap-rendering-context) looks like today:

> 4.12.5.2.2 The ImageBitmapRenderingContext interface
interface ImageBitmapRenderingContext {
  readonly attribute HTMLCanvasElement canvas;
  void transferFromImageBitmap(ImageBitmap? bitmap);
};
 
> **context . transferFromImageBitmap(imageBitmap)**
Transfers the underlying bitmap data from imageBitmap to context, and the bitmap becomes the contents of the canvas element to which context is bound.

Calling `transferFromImageBitmap` replaces contents of the canvas with the passed in `ImageBitmap`. It’s as if the internal canvas texture is switched out with the backing store of the `ImageBitmap`.  The corresponding [`transferToImageBitmap`](https://html.spec.whatwg.org/multipage/canvas.html#dom-offscreencanvas-transfertoimagebitmap) has been explained to me as something you “tear off” the corresponding offscreen canvas.  
 
The problem with using `ImageBitmapRenderingContext` for mirroring is that its contents belong to the developer.  The rendering context we want is one that belongs to the user agent.  What happens when transferFromImageBitmap is called while the UA is busy mirroring?  Will mirroring be disabled?  Is the developer telling the UA “I want the ImageBitmap to now be this other thing, please stop messing with it”?  Or will the user see the contents of their image bitmap for a fleeting moment until the user agent stomps on it again?   

If we introduce spec language that says “transferFromImageBitmap is disabled while we’re mirroring” then I think we’re better off just making our own “mirroring” rendering context type that doesn’t have this function.  At that point, we can introduce `getImageData` functions to this rendering context type so that developers can get the contents of what the UA has rendered for testing purposes.

explainer.md
toji: This is in the advanced functionality section, actually. (Not always easy to tell with the sparse diffs)

explainer.md
toji: Fair point. I'll revert these for now. (I think Input might push them away from the `Presentation` verbiage again, though.)

explainer.md
toji: Agreed that it's a bit confusing, reverting for now. I also think that we need to push a change to `supportsSession` soon, as you've been suggesting, and perhaps at that point it'll be easier to demonstrate effectively.

explainer.md
toji: Expanded the language in upcoming commit.

explainer.md
toji: Primarily because the intent is to create a link to a specific canvas on the page, but I'm flexible on this point. Probably want to address @RafaelCintron's comments about the `ImageBitmapRenderingContext` before we make a decision here, though.

explainer.md
toji: It's called out in the "Detecting and advertising VR mode" section, but didn't show up in this diff because it was made as part of #236, which I (accidentally) made the upstream for this patch. Hm... We'll have to sort that out.

explainer.md
toji: I guess I could, but it would literally be a copy/paste of a previous render loop example with a comment saying: "This is literally a copy/paste of the previous render loop example." ;)

explainer.md
toji: I could probably find a way to highlight this better, but the intent is not so much "external display == need to mirror" but "no external display == pointless to mirror". UA should still have the option to ignore a requested mirror when it's appropriate, even if the system indicates that it has an external display.

toji:
Made several updates based on @NellWaliczek's feedback, haven't yet taken into account @RafaelCintron's comments because I need some more time to consider them.

PR #238, opened 2017-06-06
Update 1.1 spec html by running make on index.bs (auto update tool up…

dlibby-:
…on PR merges appears to be broken)


toji:
Thanks for the manual update. Looking into the auto update problems now.

PR #240, opened 2017-06-08
Change animation loop from promise-based to callback-based

toji:
While talking with @NellWaliczek today she pointed out that Microsoft
was concerned about how to deal with VRPresentationFrame promises that
the app holds a reference to and calls `.then()` on well after the
frame has been processed. Upon reflection this is definitely a sticky
problem: Ideally we want any frame callbacks registered to simply pick
up the next frame’s data, but with a promise you could effectively hold
onto a stale frame for a long time.

It could be made to work, but the spec gymnastics required to ensure
sensible behavior would be significant and hard to parse, while the
existing `requestAnimationFrame` syntax is already well understood by
devs and gives us the desired behavior of never being able to fire a
frame callback for a stale frame. It will require a bit of re-hashing
the `requestAnimationFrame` syntax in the WebVR spec, but that’s
probably preferable to trying to push a Promise into a rAF-shaped hole.

Thanks for pointing out this issue, @NellWaliczek and @RafaelCintron!

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -145,7 +145,7 @@ function OnSessionStarted(session) {
   // Ensure the canvas context is compatible and create the VRLayer.
   setupWebGLLayer().then(() => {
     // Start the render loop
-    vrSession.requestVRPresentationFrame().then(onDrawFrame);
+    vrSession.requestFrame(onDrawFrame);
   });
 }
 ```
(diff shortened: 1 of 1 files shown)


DigiTec:
lgtm

The callback enables you to more concretely specify within the HTML 5 event loop as to when the frame callback fires. Having a Promise, resolved, just places it into the micro-task checkpoint queue which is unfortunately described to run under a number of conditions, many of which you probably wouldn't have expected. The micro-task checkpoint, as spec'ed, requires a macro-task to commit as well. For this reason you still need a major event to tickle the queue.

toji:
Thanks for the reviews!

Issue #228, opened 2017-05-10
Support polygonal stage bounds


huningxin:
**Problem Statement**
Current [VRStageBounds](https://w3c.github.io/webvr/spec/latest/#vrstagebounds-interface) interface only supports rectangular stage bounds.

```
interface VRStageBounds {
  readonly attribute double minX;
  readonly attribute double maxX;
  readonly attribute double minZ;
  readonly attribute double maxZ;
};
```

It can't support VR/MR devices/systems which can provide polygonal bounds defined by end users. 

With the polygonal bounds information, app is able to align the scene more precisely and create visual warning when user moves close to the bounds. 

Examples of native VR/MR systems provide polygonal bounds:

* [Oculus Guardian System](https://developer3.oculus.com/documentation/pcsdk/latest/concepts/dg-guardian-system/)

When boundary type is `ovrBoundary_Outer`, the `ovr_GetBoundaryGeometry` returns all of the points (in `ovrVector3f *`) that define the outer boundary.

* [Windows Mixed Reality](https://docs.microsoft.com/en-us/uwp/api/windows.perception.spatial.spatialstageframeofreference)

`TryGetMovementBounds` of `Spatial​Stage​Frame​Of​Reference` class returns the boundary vertices (in `Vector3[]`) of the open area defined by the user where they intend to move while using their headset.

* [OpenVR Chaperone](https://github.com/ValveSoftware/openvr/wiki/IVRChaperone_Overview)

`GetPlayAreaRect` returns the 4 corner positions of the PlayArea. It is rectangle which can be treated as a kind of polygon. 

**Proposal**

Remove `VRStageBounds` interface. Change `bounds` of `VRFrameOfReference` to a `Float32Array`. It supports the polygonal stage bounds, including the rectangular stage bounds.

```
interface VRFrameOfReference : VRCoordinateSystem {
  readonly attribute Float32Array? bounds;
};
```
The `bounds` is the array of all 3D points that define the bounds (in clockwise). Each 3D point is defined by 3 `float32`. 

When the bounds is rectangle, it contains 4 3D points.

kearwood:
This sounds reasonable to me.

Would a helper-function be appropriate to add to the spec, which calculates the largest rectangular play area within the polygonal bounds?  Or should this be left to content to calculate?

Also, while we are at it, would it be beneficial expose an additional value for the height of the boundary area?

I propose that a simple float value representing the height be added to VRFrameOfReference if that is useful.

[...]

PR #241, opened 2017-06-09
Add polygonal stage bounds.

toji:
Addresses #228. I’m proposing keeping the bounds rect in the spec,
though slightly modified to make use of the new point type, because I
believe that it’s going to be the more generally useful value for
developers. It’s true that the same value could be computed from the
bounds geometry, but the methods for doing that is not completely
trivial and possibly something better handled by the UA for consistency
(also, the native APIs may provide it anyway).

Even if we drop the play area rect, though I’d like to keep the
distinction between the bounds object and the geometry array to allow
for future extension.


explainer.md
```diff
@@ -687,13 +687,18 @@ enum VRFrameOfReferenceType {
 
 interface VRFrameOfReference : VRCoordinateSystem {
   readonly attribute VRStageBounds? bounds;
+  attribute EventHandler onboundschange;
```
huningxin: Should `VRFrameOfReference` or `VRCoordinateSystem`  inherit from `EventTarget`?

explainer.md
```diff
@@ -316,18 +316,18 @@ vrSession.createFrameOfReference("stage").then(frame => {
...
   // floor whose left back corner is at (0, 0, 0) and is 1 meter square.
-  scene.translateX(frameOfRef.bounds.minX);
-  scene.translateZ(frameOfRef.bounds.minZ);
-  scene.scaleX(frameOfRef.bounds.maxX - frameOfRef.bounds.minX);
-  scene.scaleZ(frameOfRef.bounds.maxZ - frameOfRef.bounds.minZ);
+  scene.translateX(frameOfRef.bounds.playAreaMin.x);
+  scene.translateZ(frameOfRef.bounds.playAreaMin.z);
+  scene.scaleX(frameOfRef.bounds.playAreaMax.x - frameOfRef.bounds.playAreaMin.x);
+  scene.scaleZ(frameOfRef.bounds.playAreaMax.z - frameOfRef.bounds.playAreaMin.z);
 } else {
```
huningxin: Do we need to add samples of `geometry` , e.g. visualization of bounds, and `onboundschange` event?

toji:
Okay, update the PR as per @huningxin's suggestions. Thanks!

explainer.md
```diff
@@ -316,27 +316,50 @@ vrSession.createFrameOfReference("stage").then(frame => {
...
+    // Reposition the scene to match the reported bounds. Assumes a scene with a
+    // floor whose left back corner is at (0, 0, 0) and is 1 meter square.
+    scene.translateX(frameOfRef.bounds.playAreaMin.x);
+    scene.translateZ(frameOfRef.bounds.playAreaMin.z);
+    scene.scaleX(frameOfRef.bounds.playAreaMax.x - frameOfRef.bounds.playAreaMin.x);
+    scene.scaleZ(frameOfRef.bounds.playAreaMax.z - frameOfRef.bounds.playAreaMin.z);
+
+    // Visualize the bounds geometry as 2 meter high quads
+    boundsMesh.clear();
+    for (let i = 0; i < frameOfRef.bounds.geometry.length - 1; ++i) {
```
DigiTec: You never connected the last point to the first point. Is that by design? Is the implication that the last point would equal the first point and so you don't have to complete the loop?

explainer.md
```diff
@@ -316,27 +316,50 @@ vrSession.createFrameOfReference("stage").then(frame => {
...
-  scene.scaleZ(frameOfRef.bounds.maxZ - frameOfRef.bounds.minZ);
-} else {
-  // Use default bounds. (2 meters square, centered at the origin)
-  scene.translateX(-1.0);
-  scene.translateZ(-1.0);
-  scene.scaleX(2.0);
-  scene.scaleZ(2.0);
+// Demonstrated here using a fictional 3D library to simplify the example code.
+function OnBoundsUpdate() {
+  if(frameOfRef.bounds) {
```
DigiTec: space after if, there seems to be a consistent coding style so I'll try to note deviations.

explainer.md
```diff
@@ -687,13 +710,18 @@ enum VRFrameOfReferenceType {
...
-  readonly attribute double minZ;
-  readonly attribute double maxZ;
+  readonly attribute FrozenArray<VRStageBoundsPoint> geometry;
+  readonly attribute VRStageBoundsPoint playAreaMin;
+  readonly attribute VRStageBoundsPoint playAreaMax;
+};
+
+interface VRStageBoundsPoint {
+  readonly attribute double x;
+  readonly attribute double z;
```
DigiTec: In general, I think describing a 2D point as x,y is safe, even though we know it represents location on the xz plane. No matter what we choose we will confuse some users, but I think x,z will confuse more users than x,y in this case.

explainer.md
DigiTec: Oddly placed colon after area: leading into First,... Should probably be a period.

DigiTec: The second comment here, ended up showing on the wrong line. It is in reference to the preceding paragraph from the code block.

explainer.md
```diff
@@ -675,7 +698,7 @@ interface VRWebGLLayer : VRLayer {
 // Coordinate Systems
 //
 
-interface VRCoordinateSystem {
+interface VRCoordinateSystem : EventTarget {
```
huningxin: And it needs to add `VRCoordinateSystemEvent` and `VRCoordinateSystemEventInit` accordingly?

toji:
Updated the PR again in regards to the feedback give. Thanks! A couple of notes:

> And it needs to add `VRCoordinateSystemEvent` and `VRCoordinateSystemEventInit` accordingly? 

This is a valid point, but it also got me thinking about where the event actually was being fired from and if it was the most appropriate place. Here's the thing: The bounds are reported in terms of a `VRFrameOfReference`, but it seems to me that they're part of the overall VR session. So I've speculatively moved the `boundschange` event to `VRSession`, with the implication being that any time bounds change it probably changes for all of the `VRFrameOfReference`'s that have them at once. (And as a side benefit it simplifies the IDL a bit, so that's nice.) I'd very much like feedback on this if anyone feels it's misguided or inaccurate!

> In general, I think describing a 2D point as x,y is safe...

I don't necessarily disagree, but I felt like erring on the side of correctness was better here, since confusion is essentially a given. Plus if we ever DO find a need to extend it with a Y component I'd prefer to have it actually be able to be `y`. Would LOVE some other opinions here, though! If consensus is that `x, y` is clearer I'm happy to run with that.

huningxin:
> So I've speculatively moved the boundschange event to VRSession, with the implication being that any time bounds change it probably changes for all of the VRFrameOfReference's that have them at once.

It sounds reasonable. And, does it imply all stage `VRFrameOfReference` instances actually represent one underlying stage (with same bounds, and origin etc.,) in use? I can't find related info in current explainer.

toji:
Actually, after bringing up the matter of the event location on an implementors call there was some pushback to move it back to the FrameOfReference. I've done so in the latest commit, along with adding the new event and init types.

toji:
I've updated the pull request to remove the play area rect for the moment. To be clear, I still feel strongly that a play area rect is valuable, but I want to give the structure of how we expose it some additional consideration given @DigiTec’s comments about using it to imply a preferred forward vector for the user.  While that gets figured out, though, I don't want to block the rest of this change from landing.

explainer.md
```diff
@@ -316,21 +316,14 @@ vrSession.createFrameOfReference("stage").then(frame => {
...
+When using a stage `VRFrameOfReference` the device will frequently have a configured "safe area" that the user can move around in without fear of bumping into real world objects. WebVR can communicate the rough boundaries of this space via the `VRFrameOfReference.bounds` attribute. It provides a polygonal boundary given in the 'geometry' point array, which represents a clockwise loop of points at the edges of the safe space. The shape it describes is not guaranteed to be convex. The values reported are relative to the stage origin, but do not necessarily contain it. The `bounds` attribute is null if the bounds are unavailable for the current frame of reference.
```
kearwood: Should we make any guarantees related to the winding order of the bounds points?  (ie. The points will be in a clockwise winding order as viewed from above the play area.)

toji: This text does guarantee the winding order ("which represents a clockwise loop of points") but point taken that the position of the viewer matters in that regard. I think that most sensible people would understand it to mean "as seen from above", but it certainly doesn't hurt to make that explicit.

kearwood:
Regarding the "x,y" vs "x,z" coordinates, it seems a bit unnatural to have a 2d point with any members other than "x,y", so I feel good about the call to use "x,y" even though "x,z" is technically more correct.  If we were to later revise the bounds with ceiling and floor height information, it wouldn't be properly expressed by extending the coordinates with a 3rd value -- otherwise it would lead to an awkward decision of what the 3rd scalar would represent.  It would help to accompany the docs with an image illustrating the bounds as a 2d plane on the floor of the play area to eliminate any ambiguity.

tl;dr - As long as we don't intend to add a 3rd value to the bounds coordinate later on, I feel best with "x,y".

explainer.md
kearwood: Adding "from above" certainly removed the ambiguity, thanks!

toji:
> tl;dr - As long as we don't intend to add a 3rd value to the bounds coordinate later on, I feel best with "x,y".

Thanks for the feedback. I think @NellWaliczek was going to consult with the WinMR side of things and let us know how important they felt that 3rd coordinate was to them. I'll hold off on changes till we hear back on that, but assuming they don't feel strongly about preserving the need for a 3rd coord it seems like the general consensus is that `x, y` is more natural.

explainer.md
NellWaliczek: This may be one of those things that's a matter of personal preference... My instinct is that x,y will confuse users more than x,z ^_^

explainer.md
```diff
@@ -690,13 +706,16 @@ enum VRFrameOfReferenceType {
 
 interface VRFrameOfReference : VRCoordinateSystem {
   readonly attribute VRStageBounds? bounds;
+  attribute EventHandler onboundschange;
```
NellWaliczek: Do we want to pass in the update bounds as part of the event?  If so, doesn't the event need to be a VRFrameOfReferenceEvent rather than a VRCoordinateSystemEvent?

explainer.md
```diff
@@ -316,27 +316,43 @@ vrSession.createFrameOfReference("stage").then(frame => {
...
-  scene.translateZ(frameOfRef.bounds.minZ);
-  scene.scaleX(frameOfRef.bounds.maxX - frameOfRef.bounds.minX);
-  scene.scaleZ(frameOfRef.bounds.maxZ - frameOfRef.bounds.minZ);
-} else {
-  // Use default bounds. (2 meters square, centered at the origin)
-  scene.translateX(-1.0);
-  scene.translateZ(-1.0);
-  scene.scaleX(2.0);
-  scene.scaleZ(2.0);
+// Demonstrated here using a fictional 3D library to simplify the example code.
```
NellWaliczek: Given that many platforms will draw their own version of these bounds as a safety feature already, do we want to encourage developers to draw their own bounds by providing this sample code?  Perhaps a better sample would be to validate that a raycast from the head is within bounds?

toji: I don't expect that long term developers will be referring to this doc for code samples, and this is the most concise demonstration of the data being provided that I can think of. I don't really think it's necessary to change at the moment?

explainer.md
toji: I wondered about the same thing myself, but I figured that a `VRCoordinateSystemEvent` would be a bit more futureproof. Especially given that Javascript will duck-type the hell out of these things for us. ;)

I don't see the need to provide the bounds in the event, since it's just as easy to access them off the frame of reference. There's not a lot of difference between `event.coordinateSystem.bounds` and `event.bounds`.

toji:
Merged after discussion on the implementor call. If you really care about the coordinates being X and Y instead of X and Z open a separate issue and we can discuss that aspect more narrowly.

Issue #145, opened 2016-11-09
Should we have the ability to distinguish between opaque and see-through devices?


ssylvan:
Right now WebVR apps have to assume that the HMD is opaque, but in some cases the app might want to do something different if they know the devices is transparent (e.g. they may skip rendering "background" visuals).

Should we have some way of checking that? It would be tempting to put in the device capabilities, but it seems like some HMDs will use pass-through video feeds to simulate the HMD being see-through, which means that the value can change from frame to frame, and might not even be a binary choice.

Perhaps the VRFrameData could have an HMDOpacity value that goes from 0 to 1? On typical VR devices it would be 1, on something like a Hololens it would be 0, and if any future devices has the ability to change it continuously it would be somewhere in-between.

Thoughts?

toji:
What would a HMDOpacity of 0.5 represent? Having a hard time picturing such a device. Definitely think that this is a valuable thing to communicate, though.

A few thoughts:
- If a device supports pass-through video, that seems like something that the app will want to/need to explicitly account for, even if it's just a flag that says "yes I want passthrough/no I don't". Event if it's just for performance purposes. (Using the passthrough camera feed will always have an associated perf cost.)
- That same flag is nonsensical for see-through devices like HoloLens. You can't "turn off" the real world, but you can paint over the whole thing.
- Seems like Windows Holographic already needs to deal with both classes of devices. Any documentation that we can be pointed at for how it's handled there?
- It does seem like whether we communicate opacity as a binary state or a float we'll want an event that communicates if it changes. I'd like to think that should be in the apps control completely, but I'm guessing that won't prove true long term.

@judax has been looking at dealing with passthrough cameras on Tango, he'd probably have some valuable insight on this subject.

ssylvan:
I'm not fully convinced having opacity be continuous is desirable - so would love to hear from others. 

Imagine something like an Oculus and a Vive but with two cameras on the outside and a "pass through" mode that just shows you the real world instead of VR. Since the "real world" is a matter of blending textures, they could choose to make the real world 10% opaque or any arbitrary fraction (e.g. for safety reasons? or in response to user controls?).

brianchirls:
For that matter, I wonder if there's relevant information about how the
graphical content is overlayed on the real world. Is the FOV different? Is
the overlay fully opaque? Does black work? Is there blending? What kind?

Maybe it's too early to fix this in place.

blairmacintyre:
[...]

PR #242, opened 2017-06-13
Added `VRDisplay.isTransparent` attribute and related example.

toji:
Addresses #145. Not sure if I like the idea that this resides on the `VRDisplay` (more fingerprinting avenues) but could be useful in cases where a transparent framebuffer needs to be allocated.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -483,6 +483,31 @@ vrSession.addEventListener('resetpose', vrSessionEvent => {
 });
 ```
 
+### Handling Transparent Displays
+
+The majority of current VR devices are opaque and do not enable the user to see their surrounding environment while using it. However some devices, like Microsoft's HoloLens, utilize transparent displays in the service of AR features. While the initial version of the WebVR API does not attempt to provide access to AR-centric features it does communicate whether or no the device display is see-through with the `isTransparent` attribute.
+
+WebVR content may want to refrain from drawing a background over the full frame in cases where the display is transparent, to allow the virtual content to appear as if it's floating in the real environment. (Note that this version of the WebVR API provides no mechanism for aligning or occluding content againts the environment.) To do so rendering of virtual backgrounds should be skipped when the display indicates it is transparent.
+
+```js
+function onDrawFrame(vrFrame) {
+  let pose = vrFrame.getDevicePose(vrFrameOfRef);
+  gl.bindFramebuffer(vrSession.baseLayer.framebuffer);
+
+  for (let view in vrFrame.views) {
+    let viewport = view.getViewport(vrSession.baseLayer);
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -483,6 +483,31 @@ vrSession.addEventListener('resetpose', vrSessionEvent => {
 });
 ```
 
+### Handling Transparent Displays
+
+The majority of current VR devices are opaque and do not enable the user to see their surrounding environment while using it. However some devices, like Microsoft's HoloLens, utilize transparent displays in the service of AR features. While the initial version of the WebVR API does not attempt to provide access to AR-centric features it does communicate whether or no the device display is see-through with the `isTransparent` attribute.
```
ddorwin: s/no/not/

explainer.md
```diff
@@ -483,6 +483,31 @@ vrSession.addEventListener('resetpose', vrSessionEvent => {
...
+
+WebVR content may want to refrain from drawing a background over the full frame in cases where the display is transparent, to allow the virtual content to appear as if it's floating in the real environment. (Note that this version of the WebVR API provides no mechanism for aligning or occluding content againts the environment.) To do so rendering of virtual backgrounds should be skipped when the display indicates it is transparent.
```
ddorwin: In the last sentence: Comma after "so." Also, maybe change the voice from passive.

ddorwin: Since this is an explainer, it would be helpful to describe use cases that this very limited mechanism is intended to address.

explainer.md
```diff
@@ -483,6 +483,31 @@ vrSession.addEventListener('resetpose', vrSessionEvent => {
 });
 ```
 
+### Handling Transparent Displays
```
ddorwin: We discussed needing to know _how_ to specify a pixel should be transparent. Unless we want to commit to a specific color or alpha value - something that seems bound to be wrong on some hardware - this attribute should tell the app how to render transparent. (That may also allow non-transparent displays to pretend to be transparent by, for example, passing through video to such pixels. Such functionality would probably need to be requested somehow, though.)

Also, if some color or alpha value is going to render as transparent (and this is not clearly obvious), we need a way to indicate to authors that they should not use such values. This seems problematic for apps that are just trying to write a generic WebVR app. (This may also warrant some type of feature detection and/or request to "prevent me from drawing transparent pixels.)

domenic:
Drive-by naming concern: we _mostly_ don't use "is" prefixes for boolean options. https://github.com/w3ctag/design-principles/issues/53

toji:
Thanks for the pointer, @domenic, but in this case I think there's enough concern about this particular feature/patch that I'm going to close it till we can do a deeper dive on AR-related things.

PR #251, opened 2017-06-28
Link all definitions in the explainer to the respective definitions in the latest WebVR API spec

anssiko:
Did a pass on the explainer, and while doing so, linked the definitions in the explainer to the respective definitions in the latest WebVR API spec, and other related specs. This should make it easier for a less familiar reader to navigate between the explainer and the spec. This also highlights the API surface that still needs to be added to the latest spec (i.e. is not linked).

No other changes apart from obvious typo fixes, so you can consider this an editorial pass.

[I notice there are some PRs open to the explainer that may be easier to land before this, so feel free to rebase this after those PRs have landed. I'll be OoO for a while, so feel free to rebase this on my behalf as needed if this PR is deemed useful.]

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -1,7 +1,7 @@
-﻿# WebVR Explained
+# WebVR Explained
 
 ## What is WebVR?
-[WebVR](https://w3c.github.io/webvr/) is an API that provides access to input and output capabilities commonly associated with Virtual Reality hardware like [Google’s Daydream](https://vr.google.com/daydream/), the [Oculus Rift](https://www3.oculus.com/rift/), the [Samsung Gear VR](http://www.samsung.com/global/galaxy/gear-vr/), the [HTC Vive](https://www.htcvive.com/), and [Windows Mixed Reality headsets](https://developer.microsoft.com/en-us/windows/mixed-reality). More simply put, it lets you create Virtual Reality web sites that you can view in a VR headset.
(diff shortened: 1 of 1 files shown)


toji:
Thanks for doing this, I meant to go through and do a similar update myself in the next few days so you've saved me some work. :)

Once concern I have: Do you feel it's necessary to link every single instance of a defined term? Certainly they should all be linked at least once, probably on first use, but there's some terms that are used _very_ frequently (as I'm sure you discovered) and the repeated links make it harder to parse the non-rendered markdown as well as iterate on the explainer text.

anssiko:
I figured the same thing out, but went ahead and linked 'em all similarly to as we do in specs, where bikeshed et al. do this grunt work for us :-) I'm a believer in the [priority of constituencies](https://www.w3.org/TR/html-design-principles/#priority-of-constituencies) i.e. "the editor(s) should take the cost of editing if that'd benefit a much larger group of readers". A WYSIWYG markdown editor will help editing. That said, feel free to amend this PR as you see fit.

toji:
I've actually gone the other way on this and stripped out the linking of various IDL terms to the spec. The spec itself is just a shell right now and there's enough fluidity to the explainer that the links are frequently either broken or not useful. I'll take it as an action item to go back and re-link the entire explainer more thoroughly once the spec has actually stabilized into something useful.

anssiko:
Sounds good. Let me do another pass when the explainer churn has settled.

PR #252, opened 2017-06-29
Clarify the definitions and uses of "suspended" vs. "ended" sessions

toji:
The need to make this more explicit was identified by discussion around #236.


toji:
Given the lack of negative feedback I'm going to merge this now.

PR #253, opened 2017-06-30
Specify some 1.1 methods can only be called in a VRDisplay.rAF callback

### LINKED ISSUE #246: getFrameData usage while not presenting inconsistent in 1.1 spec
The current 1.1 spec is inconsistent about how poses can be used while not presenting ("magic window" mode), it says:

> [...] Subsequent calls to getFrameData() MUST provide the same values until the next call to submitFrame(). Returns true if the the provided VRFrameData object was successfully updated, false otherwise.

However, it's only legal to call submitFrame while presenting (it requires VRLayerInit and related presentation-only setup), so as written the frame data could never be updated in magic window mode.

Suggestion: change this to explicitly state _The value returned will not change until JavaScript has returned control to the browser_, without mentioning submitFrame, this is what it already says in the getPose() API comment which is different from the text description following that.

Also, it's unclear if the intention is to permit use of getFrameData() from both within window.rAF and vrDisplay.rAF callbacks. The vrDisplay.rAF spec says it's _Functionally equivalent to window.requestAnimationFrame when the VRDisplay is not presenting_, but implementations that tie pose handling to vrDisplay.rAF specifically may not be able to provide poses easily for window.rAF callbacks. Can this be clarified? Either require getFrameData() to be used from a vrDisplay.rAF context (similar to submitFrame), or make it explicit that getFrameData should return reasonably-fresh poses for either window.rAF or vrDisplay.rAF while not presenting?

toji:
Specifically `getFrameData` and `submitFrame`.

Fixes #246.


spec/1.1/index.bs
```diff
@@ -193,7 +193,7 @@ The {{capabilities}} attribute MUST return the {{VRDisplay}}'s {{VRDisplayCapabi
...
+Populates the provided {{VRFrameData}} object with the {{VRPose}} and view and projection matricies for the current frame. These values describe the position, orientation, acceleration, and velocity of the {{VRDisplay}} that should be used when rendering the next frame of a scene. The User Agent MAY optionally use predictive techniques to estimate what these values will be at the time that the next frame will be displayed to the user. The value returned will not change until JavaScript has returned control to the browser. Returns true if the the provided {{VRFrameData}} object was successfully updated, false otherwise. Will return false and not populate the {{VRFrameData}} object if called outside a {{VRDisplay}}.{{requestAnimationFrame()}} callback.
```
klausw: FYI, there's also a section above with the method description block which duplicates some of the info here, i.e. for `[NewObject] VRPose getPose();`. Please check to make sure these stay in sync.

toji: Updated the method comments to raise the same points. (Did not update getPose because it's deprecated)

PR #255, opened 2017-07-07
Detail magic window support

toji:
This is a continuation of #237, which Github helpfully closed without notifying my after some branch rebasing shenanigans and now won't let me re-open. Ugh. 😠

Anyway, to follow up on @RafaelCintron's comments from the closed PR: after giving it a lot of thought and having some conversations on the topic with various people I agree that we probably want to create our own `CanvasRenderingContext` type. The primary driver for this in my mind is that on some devices (like zSpace displays) we'll want to allow for stereo imagery to be shown in the canvas, and trying to hack images in an `ImageBitmap` is not something I care to do. Instead, we'll want a purpose built set of test-centric APIs and disallow imagery to come from anywhere but the `VRSession` so we don't have to deal with issues like a canvas that's been showing stereo imagery suddenly switching to mono when an `ImageBitmap` is pushed to it and then back again on the next VR frame.

I've updated the pull request to reflect that, and am really eager to have somebody suggest better names for the new context type. 😝 Also, this PR is avoiding attempting to add the testing-centric APIs mentioned by Rafael to the context as I figure that could get to be a lengthy discussion for a follow up patch.

(Also, sorry about the delayed reply! Between the US holiday and this simply being a tough issue I've been really slow to gather my thoughts on it.)


toji:
Ping for comments. This is one of the critical pieces we identified for TAG readiness, so I'd definitely like some additional feedback on it. In particular:

- Is this general approach (1:1 association between VRSessions and an output canvas) agreeable?
- There is surely a better name for the canvas context in question than `VRRenderingContext`.
- I feel like mirroring is a pretty natural, uncontroversial extension of this behavior, which is why I've included it. If anyone feels it's problematic as presented here (or simply wants to give it more attention as a separate issue) I'd like to hear it. I definitely don't want the PR to get hung up on the mirroring aspect.

explainer.md
```diff
@@ -772,4 +824,15 @@ partial dictionary WebGLContextAttributes {
...
 };
+
+//
+// RenderingContext
+//
+interface VRRenderingContext {
+  readonly attribute HTMLCanvasElement canvas;
+};
+
+dictionary VRRenderingContextSettings {
```
NellWaliczek: What's this for? (did I miss something?)

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
...
+  let mirrorCanvas = document.createElement('canvas');
+  let mirrorCtx = mirrorCanvas.getContext('vr');
+  document.body.appendChild(mirrorCanvas);
+
+  vrDevice.requestSession({ outputContext: mirrorCtx })
+      .then(OnSessionStarted);
+}
+```
+
+### Non-exclusive sessions ("Magic Windows")
```
NellWaliczek: Should we be considering requiring an outputContext for non-exclusive sessions?

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
 
 If the UA needs to halt use of a session temporarily the session should be suspended instead of ended. (See previous section.)
 
+## Rendering to the Page
+
+There are a couple of scenarios in which developers may want to present content rendered with the WebVR API on the page instead of (or in addition to) a headset.
+
+### VRRenderingContext
```
toji: Reading this back I definitely see what you mean. I'll take another stab at it.

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
...
+
+### Mirroring
+
+On desktop devices, or any device which has an external display connected to it, it's frequently desirable to show what the user in the headset is seeing on the exernal display. This is usually referred to as mirroring.
+
+In order to mirror WebVR content to the page, developers provide a `VRRenderingContext` as the `outputContext` in the `VRSessionCreateParameters` of an exclusive session. Once the session has started any content displayed on the headset will then be mirrored into the canvas associated with the `outputContext`.
+
+When mirroring only one eye's content will be shown, and it should be shown undistorted. The UA may choose to crop the image shown, display it at a lower resolution than originally rendered, and the mirror may be multiple frames behind the image shown in the headset. The mirror may include or exclude elements added by the underlying VR system (such as visualizations of room boundaries) at the UA's discretion. Pages should not rely on a particular timing or presentation of mirrored content, it's really just for the benefit of bystanders or demo operators.
```
toji: I meant without any post processing to correct for optical distortion. I'll try to make this clearer.

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
 
 If the UA needs to halt use of a session temporarily the session should be suspended instead of ended. (See previous section.)
 
+## Rendering to the Page
+
+There are a couple of scenarios in which developers may want to present content rendered with the WebVR API on the page instead of (or in addition to) a headset.
+
+### VRRenderingContext
+
+Any time WebVR content is presented on the page it's done through a `VRRenderingContext`. Like a `WebGLRenderingContext`, developers acquire a `VRRenderingContext` by calling `HTMLCanvasElement.getContext()` or `OffscreenCanvas.getContext()` with the context id of "vr". (Editor's note: Naming is hard. Please _please_ *please* suggest something better!) The returned `VRRenderingContext` is permenantly bound to the canvas.
```
toji: The thought had crossed my mind too. I worry that's unintuitive for magic window case, though. Also toyed with "vroutput" but it just didn't sit with me well.

Hmm... what about "vrpresent" or "vrpresentation" paired with a "VRPresentationContext"? I worry that a "Rendering" context sounds like something that you should be producing imagery with rather than a mechanism for displaying it. "Presentation" carries a bit better implications in that regard.

explainer.md
toji: Given that the concept of a tracking-only context has proven controversial, yes. :)

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
...
+  document.body.appendChild(mirrorCanvas);
+
+  vrDevice.requestSession({ outputContext: mirrorCtx })
+      .then(OnSessionStarted);
+}
+```
+
+### Non-exclusive sessions ("Magic Windows")
+
+There are several scenarios where it's beneficial to render a scene on the page who's view is controlled by device tracking. For example:
```
toji: Derp. Thanks!

explainer.md
toji: Ah, I never mention that anywhere, do I? `getContext` allows a dictionary of context creation args for any context type and I was just trying to think of what properties might be applicable here. Honestly I'm not even sure this is necessary, though. I can drop if for now.

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
...
+
+ - Using phone rotation to view panoramic content.
```
toji: Good point. I'll try to improve that.

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
...
+ - Taking advantage of 6DoF tracking on devices (like [Tango](https://get.google.com/tango/) phones) with no associated headset.
+ - Making use of head-tracking features for devices like [zSpace](http://zspace.com/) systems.
+
+These scenarios can make use of non-exclusive sessions to render tracked content to the page. Similar to mirroring, to make use of this mode a `VRRenderingContext` is provided as the `outputContext` at session creation time, as well as the `exclusive: false` flag. At that point content rendered to the `VRSession.baseLayer` will be rendered to the canvas associated with the `outputContext`. In the future, if multiple `VRLayers` are used their composited result will be what is displayed in the `outputContext`. Given that the session is tied to a specific `outputContext` it's permissible to have multiple non-exclusive sessions active at once to handle multiple output surfaces.
```
toji: Good question. I guess I was thinking that outputContext would be tied to the session for the session's lifetime, regardless of exclusivity. I guess it could be a convenience to re-use a context to switch between magic window and mirroring, but I don't see that as a huge concern.

In the meantime, I'll take out the reference to multiple sessions.

explainer.md
```diff
@@ -312,6 +300,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
...
+
```
toji: I do have language like that for mirroring, but you're asking for the option to do boundary rendering for magic window? I'm not sure if I'm clear on the benefit of that.

toji:
Okay, updated the PR to take into account @NellWaliczek's feedback. (Thanks!)

explainer.md
```diff
@@ -313,6 +301,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
...
+Exclusive and non-exclusive sessions can use the same render loop, but there are some differences in behavior to be aware of. The sessions may run their render loops at at different rates. During exclusive sessions the UA runs the rendering loop at the `VRDevice`'s native refresh rate. During non-exclusive sessions the UA runs the rendering loop at the refresh rate of page (aligned with `window.requestAnimationFrame`.) Also, when rendering with an exclusive session the projection matrices provided each frame are based on the device optics, while non-exclusive sessions provide projection matrices based on the output canvas dimensions and possibly on the position of the users head in relation to the canvas if that can be determined.
```
NellWaliczek: > Also, when rendering with an exclusive session the projection matrices provided each frame are based on the device optics, while non-exclusive sessions provide projection matrices based on the output canvas dimensions and possibly on the position of the users head in relation to the canvas if that can be determined.

I'm a bit nervous that this may give folks the impression that there are circumstances under which they don't need to check the projection matrices each frame. 

Also, the projection matrices are also based on depthNear and depthFar, so maybe there's a way to word this a little less specifically?

explainer.md
```diff
@@ -313,6 +301,67 @@ vrSession.addEventListener('ended', OnSessionEnded);
 
 If the UA needs to halt use of a session temporarily the session should be suspended instead of ended. (See previous section.)
 
+## Rendering to the Page
+
+There are a couple of scenarios in which developers may want to present content rendered with the WebVR API on the page instead of (or in addition to) a headset: Mirroring and "Magic Window". Both methods display WebVR content on the page via a Canvas element with a `VRPresentationContext`. Like a `WebGLRenderingContext`, developers acquire a `VRPresentationContext` by calling `HTMLCanvasElement.getContext()` or `OffscreenCanvas.getContext()` with the context id of "vrpresent". The returned `VRPresentationContext` is permenantly bound to the canvas.
```
NellWaliczek: > The returned `VRPresentationContext` is permenantly bound to the canvas.

Should we call out that supplying an invalid (already used) VRPresentationContext should not cause session creation to fail? I assume we'd just want the mirroring part to fail?

explainer.md
```diff
@@ -557,10 +606,12 @@ interface VRDevice : EventTarget {
 
 dictionary VRSessionCreateParametersInit {
   boolean exclusive = true;
+  VRPresentationContext outputContext = null;
```
NellWaliczek: Do we want to allow this value to be reset?  Or should it only allowed to be set at session creation time?  My inclination is to start as you have it (not resettable) and see if there's use cases that would drive adding the feature.

explainer.md
```diff
@@ -90,7 +90,7 @@ Sessions can be created with one of two levels of access:
...
 
-In the following example we ask if the `VRDevice` supports sessions with `exclusive` access, since we want the ability to display imagery on the headset.
+In the following examples we will focus on using exclusive sessions, and cover non-exclusive session use in the [`Advanced Functionality`](#non-exclusive-sessions-magic-windows) secion. With that in mind, we ask here if the `VRDevice` supports sessions with `exclusive` access, since we want the ability to display imagery on the headset.
```
toji: Yeah, I am too but I think consistency is important. I'll remove it, but I'm also going to add an example in the magic window section of using `supportsSession` to test for magic window support.

explainer.md
toji: That's my feeling too. It's easier to add the ability to reset it in the future than it is to revoke it if it proves problematic.

toji:
Thanks again @NellWaliczek! I've made some verbiage tweaks to account for your feedback, though I've left the names alone for now. I agree that the context ID name especially is not my favorite, but that's easy to change as soon as someone suggests a better one. Describing the behavior is more important for now.

I think I'm going to merge this soon so that we've got the basics in place for the TAG review. Obviously we can continue to iterate on it after that, but I'm feeling pretty good about the general shape of this part of the API.

explainer.md
NellWaliczek: It was more that I wanted to leave us the option of having the VR system merge in its own pixels (for whatever reason) rather than just act as layer composition.

toji: I would love to have a more concrete example of what you have in mind for this, but I suspect that if you could have shared one you would have already. ;) In any case, I don't object to the idea of the UA being able to push additional pixels so I've added a line to that effect.

toji:
Merged!

PR #259, opened 2017-07-19
Change supportsSession promise to return void

toji:
Suggestion to make this change came up during an API ergonomics review I did today. This lets us focus on the simple behavior of resolve/reject being the signal for support and allows us to extend the promise to return more information in the future if needed.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -88,7 +88,7 @@ Sessions can be created with one of two levels of access:
 
 ### Detecting and advertising VR mode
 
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -99,19 +99,20 @@ async function OnVRAvailable() {
...
-  let exclusiveMode = await vrDevice.supportsSession({ exclusive: true });
-  if (exclusiveMode) {
+  vrDevice.supportsSession({ exclusive: true }).then(() => {
     var enterVrBtn = document.createElement("button");
     enterVrBtn.innerHTML = "Enter VR";
     enterVrBtn.addEventListener("click", BeginVRSession);
     document.body.appendChild(enterVrBtn);
-  }
+  }).catch(() => {
+    console.log("VRDevice does not support exclusive sessions.");
```
NellWaliczek: Nit: Perhaps we want to word this error message a bit differently in case we add reasons for rejection that aren't simply because of hardware capability.  For example, in the past we've talked about the idea of the user being able to "disable VR" via some setting in the UA?

toji: Good call. I've updated the example code to log out a "reason" argument provided by the promise rejection.

Issue #248, opened 2017-06-23
Do not treat activate events as a user gesture


ddorwin:
The explainer currently says, "The activate event acts as a user gesture..." The intent is that WebVR presentation can start from the event handler without requiring a user gesture, such as a click.

WebVR should not generate a generic user gesture state/bypass because this would affect all other specs and implementations that use a user gesture for privacy and/or security mitigation.

Instead, WebVR should differentiate user gestures and other cases that allow the user gesture requirement to be bypassed. In the spec, it may be helpful to have internal state/variables and/or algorithms related to "Is presentation allowed?"

ddorwin:
This was about the 2.0 explainer, but we may want to clean up the 1.1 text as well.

cvan:
this is touched on in PR #256. @toji: do you want to assign this to yourself and/or close?

toji:
Yeah, I'm going to consider this closed by #256.

PR #260, opened 2017-07-21
Remove the `navigation` event from the explainer

toji:
Requested by @NellWaliczek. This is for the sake of clarifying things
for TAG review if we don’t get a better solution in place before they
look at the explainer, since we know this ISN’T the right way to handle
this. Ideally we’ll put it back in with #256 or a similar PR.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -479,28 +479,6 @@ function onDrawFrame() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


NellWaliczek:
I imagine @ddorwin would agree since he said something similar in #248 ;)

toji:
Done!

PR #271, opened 2017-08-04
Fixes pointer events missing from idl in 1.1 spec

NellWaliczek:
The descriptions for the vrdisplaypointerresticted and vrdisplaypointerunrestricted events were added to the webvr 1.1 spec in February, but the actual events themselves were missing from the idl.  This commit corrects that mistake.


toji:
LGTM!

PR #279, opened 2017-08-15
Input Explainer

toji:
Sharing the current input explainer for feedback.


toji:
Squashed down some a bunch of the commits for easier review.

AlbertoElias:
Really like how this will simplify Web Inputs, some very good ideas that I think could be brought to the rest of the Web. This seems like a good way to test them out but just wanted to mention that it might be worth keeping in mind that in the future this could be merged with the Gamepad API as many of these ideas would be useful there and having too many APIs has its difficulties.

AlbertoElias:
Another thing, how do Vive Trackers fit in all of this?

toji:
That's an excellent question, and one that I only have a partial answer to. At it's most basic a Vive tracker would likely show up in the `VRSession.getControllers()` results. They wouldn't have any elements like triggers or touchpads, and would never generate select gestures, but would deliver a valid `gripPoseMatrix` when tracking. `supportsPointing` would likely be false, since there's no natural pointing vector associated with a bare tracker and we don't want to instruct WebVR apps to draw pointer lasers coming off of people's feet.

When you start considering that the trackers can have attachments, though, things change significantly. Many of the attachments I've seen (guns, tools) DO have a natural pointing direction and frequently have a sensible primary action to generate selection gestures with. I'm not familiar enough with the APIs that back the trackers to know how much about the device we can query in terms of attachment behavior, and whether they give us enough info to automatically map those types of things. Need to look into it, but if they do then hopefully we can "just map it" and the average WebVR app will never have to worry about what the tracked object looks like.

Event if we can do that, though, one thing that's not really accounted for in the spec right now is how to differentiate a tracker with an attachment from a standard tracked controller. It's easy to imagine a company wanting to build a custom web app to show off their attachment, but if all that we surface is 3 mostly-identical seeming controllers developers are going to start falling back to some nasty tricks to try and differentiate them. (Oh, I noticed that the delta between the grip and pointer rays is 15 degrees on this controller while it's 30 degrees on a standard Vive controller. This must be my AR-15 device!) Not sure how to do it in a ecosystem-conscious way, though. We're trying to stay away from advertising controller names due to both fingerprinting and cross-platform concerns. (We already see people scanning the gamepad array for "Vive" or "Daydream" and only supporting those devices, which is an anti-pattern that we need to squash.)

I'm going to do some more research on the tracker API (and see if I can get ahold of one myself) but thoughts on how to make this all more developer friendly are appreciated!

toji:
So for what it's worth HTC's recommendation on resolving tracker vs. accessory coordinate systems is "Measure the physical angles and hard code it." ([Page 27](https://dl.vive.com/Tracker/Guideline/HTC_Vive_Tracker_Developer_Guidelines_v1.3.pdf)) So that's cool and helpful and definitely won't cause us any problems ever. *sigh*

For the other inputs everything coming from the tracker gets mapped as one of the existing Vive wand buttons with no other semantic information. So it looks like the expectation is that you'll just look at the device/manufacturer name and look up the mapping and metrics in a table somewhere. :sob:

AlbertoElias:
> "Measure the physical angles and hard code it." 

Woah, harsh.

> everything coming from the tracker gets mapped as one of the existing Vive wand buttons

What is everything? Isn't it just a pose?

> supportsPointing would likely be false,

I do think it would be a good idea to add `supportPointing` as these could be inputs for people how are unable to hold a controller. It's something that hasn't been explored yet, so we could hold off for now, but then that would prevent the experimentation being done with WebVR.

> When you start considering that the trackers can have attachments

I have no idea how they work as attachments. Maybe they aren't read as trackers but as that other input device which could help, as these may have buttons, like the guns you mentioned having triggers and maybe even a joystick.

> We already see people scanning the gamepad array for "Vive" or "Daydream"

Currently doing that and removing the need for it is one of my favourite things about this explainer

toji:
> What is everything? Isn't it just a pose?

"Everything" is the pose and any buttons or axes exposed on the Vive wand, which can be simulated through the use of the tracker's Pogo pins. Specifically: Trigger, Grip button, touchpad, app button.

kearwood:
In the "Controller Rendering" section:

> The glTF asset should represent the various movable elements (buttons, triggers, joysticks, etc) of the controller with different nodes in the scene hierarchy so that they can be transformed independently.

We should perhaps be explicit in saying that the "moveable elements" may include bone nodes for a rigged model, enabling the representation of things like articulated hands.

kearwood:
Also, regarding the "Controller Rendering"..  There might not be a 1:1 relationship between the input controls (buttons, triggers, etc) and the set of nodes that are moved around to update the visual representation of the controller.

IMHO, there should be an opaque array of matrices returned per-frame associated to the nodes that need to be transformed for the visual representation.  A separate array of transformed points and/or sub-meshes to highlight could be returned to enable contextual highlighting of the associated buttons for use in rendering "in-vr controller legends / mapping UI".

kearwood:
As mentioned in the F2F meeting, we may need to define the relationship of the events to the gestures.  This could be something as simple as "The event will fire when the gesture has been determined by the UA, and not necessarily when the gesture has started or finished."

toji:
> There might not be a 1:1 relationship between the input controls (buttons, triggers, etc) and the set of nodes that are moved around to update the visual representation of the controller.

This is a good point, and something I'm not totally clear on a solution for. Example: Many systems like to show a visual representation of your finger sliding around on a trackpad, usually with a dot or sphere that becomes visible on touch. Changing colors of the buttons on press is another common pattern for inputs with low travel. So do we need the ability to change mesh component visibility too? Or swap out textures/colors?

kearwood:
It would be annoying to see a ray cast from your controller or head through the entire duration of a VR session for experiences that have a "select"-able object always visible, so content should try to hide the ray when possible.

I would like to suggest that either a pair of events or a polled boolean value be added that informs content when the user may wish to see the ray.  For example, some hardware includes capacitive sensors that know when the user is touching the "select" button on the controller and can cause the ray to become visible in anticipation of the button being pressed.

toji:
> content should try to hide the ray when possible.

Agreed, but my concern is that with the exception of the scenario that you mentioned (capacitive select button) the context for when a ray is appropriate it hide usually lies within the app and not the UA. I think that until we get to a declarative world there's limited recommendations we can make for things like that. A-Frame is probably in a far better position to determine visibility.

kearwood:
> > There might not be a 1:1 relationship between the input controls (buttons, triggers, etc) and the set of nodes that are moved around to update the visual representation of the controller.
> 
> This is a good point, and something I'm not totally clear on a solution for. Example: Many systems like to show a visual representation of your finger sliding around on a trackpad, usually with a dot or sphere that becomes visible on touch. Changing colors of the buttons on press is another common pattern for inputs with low travel. So do we need the ability to change mesh component visibility too? Or swap out textures/colors?

Agree, we may need more than the node transforms returned in the array.  Simply including a boolean "visible" flag may allow for more freedom for now, with possible future expansion to include other material attributes.

If we were to include material attributes in the per-frame animation, this may also enable an api such as "highlight the button-n control and give me a transformed point for a label", enabling the button-mapping UI's for more advanced content to be implemented agnostically.

kearwood:
If supportsGrabbing or supportsPointing are false, should the corresponding arrays in VRInputPose have no value?

toji:
> If supportsGrabbing or supportsPointing are false, should the corresponding arrays in VRInputPose have no value?

This was addressed on the call, but to recap for the benefit of GitHub: These are terrible, awful, no good attribute names that desperately need to be replaced with something better, because they cause confusion such as this.

The intent was not to communicate whether or not the input source has a pointer or grip matrix but whether it can be used for at least 3DoF movement ("pointing") and 6DoF movement ("grabbing". Yes, I know. I'm sorry too.) We want a term that will communicate the controllers capability level's without explicitly falling back to the terms 3DoF and 6DoF.

Suggestions welcome and needed!

toji:
A couple of question for the group, but especially for @NellWaliczek since I don't recall if we covered this already and decided against it for some reason:

**First**, is it necessary to have a `supportsPointing` attribute to indicate pointing rather than gaze devices, or can we infer that from the presence of the grip pose? Especially for the purposes of rendering. Seems like we've got a few possible configurations inside the frame loop. (ignoring the select gesture for a moment):

1. No `VRController`s are available. This probably indicates screen-space input. Obviously nothing should be rendered.
2. No `inputPose` is available for a controller. This likely indicates a temporary lack of tracking. Again, nothing should be rendered because there's no basis for where to render it.
3. A pointer pose is given but not grip pose. This can probably be safely assumed to indicate gaze-targeted input, because if we have _any_ basis for the pointer pose _and_ it's coming from a tracked controller we don't really have an excuse for not providing a grip pose as well, right? So only render the cursor in this case.
4. A pointer and a grip pose are given. This is obviously not a gaze cursor, because we shouldn't be providing grip poses for the head (IMO).  Render both the cursor and ray (and probably controller too) in this case.

**Second**, In the case of a clicker-style input (Cardboard, bluetooth clicker) should we have a canonical element ID? If so, what? None of the ones we've identified so far (trigger, grip, touchpad, joystick) feel entirely correct. The exception is GearVR, who's gaze input clearly falls under "touchpad." "trigger" is maybe the closest, but many users will understandably assume that refers to an analog input instead of a digital button. Maybe we want to add "clicker" or something similar to the canonical list?

toji:
Another issue that's come up for me: many systems will want a way to detect that a controller in the `getControllers()` array is the same device from frame to frame. It's especially useful if the app wants to track any state along with the controller (such as GPU resources, player state, etc.) At the moment the explainer offers no guarantees about this.

The two simplest routes I can see would be to:
- Provide an integer ID on the controller that can be used as an easy lookup mechanism. Straightforward and easy to understand, if not particularly elegant. I imagine that the web platform folks would prefer we shy away from this route.
- Guarantee array index stability. The Gamepad API goes this route, I'll leave it to the reader to determine if it's a successful strategy. This does mean that we'll end up with an array with "holes" in it as earlier items disconnect, but as long as we clearly specify how the holes are presented (null, undefined, etc.) it shouldn't be a problem. There is some confusion that can arise from a controller disconnecting and a new one connecting in the same frame and filling in the first controller's slot.

Curious to hear preferences or alternatives!

input_explainer.md
```diff
@@ -0,0 +1,406 @@
...
+  "",
+  "left",
+  "right"
+};
+
+interface VRController : VRInputSource {
+  readonly attribute VRHandedness handedness;
+
+  // TODO: Better names for these two properties.
+  readonly attribute boolean supportsPointing;
```
dmarcos: This comes from my perspective of higher level APIs so not sure if this is relevant in this context. If we expect to consider more gestures in the future it might be cleaner to return an object to be able to do:
````javascript
controller.gestures.pointing
````

File: input_explainer.md [hunk 2]
```diff
@@ -0,0 +1,406 @@
...
+};
+
+//
+// Session
+//
+
+partial interface VRSession {
+  attribute EventHandler onselect;
+
+  attribute EventHandler oncontrolleradded;
```
dmarcos: maybe `oncontrollerconnect` \ `oncontrollerdisconnect` for consistency with `ondeviceconnect` \ `ondevicedisconeect` of the WebVR 2.0 API (https://w3c.github.io/webvr/spec/latest/#dom-vr-ondeviceconnect)

input_explainer.md
```diff
@@ -0,0 +1,406 @@
...
+};
+
+interface VRController : VRInputSource {
+  readonly attribute VRHandedness handedness;
+
+  // TODO: Better names for these two properties.
+  readonly attribute boolean supportsPointing; 
+  readonly attribute boolean supportsGrabbing; 
+
+  readonly attribute VRControllerElementMap elements;
```
dmarcos: `elements` seems a bit of a generic word, not very descriptive. The SteamController API calls this  `inputSources` but it collides with `VRInputSource` in our case.  Idea: Rename `VRInputSource` to `VRInputDevice` and `elements` to `inputSources`.

slomchinski3:
What is the best place to study code?


Jennifer slomchinski

On Aug 22, 2017 1:23 PM, "Brandon Jones" <notifications@github.com> wrote:

So for what it's worth HTC's recommendation on resolving tracker vs.
accessory coordinate systems is "Measure the physical angles and hard code
it." (Page 27
<https://dl.vive.com/Tracker/Guideline/HTC_Vive_Tracker_Developer_Guidelines_v1.3.pdf>)
So that's cool and helpful and definitely won't cause us any problems ever.
*sigh*

For the other inputs everything coming from the tracker gets mapped as one
of the existing Vive wand buttons with no other semantic information. So it
looks like the expectation is that you'll just look at the
device/manufacturer name and look up the mapping and metrics in a table
somewhere. 😭

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/w3c/webvr/pull/279#issuecomment-324111022>, or mute the
thread

input_explainer.md
toji: This seems face-palm-worthy obvious now that you've pointed it out. 😅 I'm all for the change.

I do seem to recall that @NellWaliczek and friends may have preferred the "added" verbiage simply because it may apply better to inputs like hand tracking where hands may fall in and out of tracking but not be "connecting" and "disconnecting" in the traditional sense. I feel like the terminology isn't a big deal and I'd prefer consistency, but I am curious to hear the Microsoft team's opinion.

input_explainer.md
toji: Frankly I'm super happy to take _any_ suggestions for a better name than `elements`. I think the input/device distinction again comes from Microsoft feeling that "device" is an awkward term for hand-based gestures, but similar to "added/connected" I'm not terribly concerned about the verbiage here outside of making sure names are intuitive. (In which case `elements` is really pushing it.)

cvan:
@toji know this is a bit old, but do you need this reviewed, or should I wait on updates?

NellWaliczek:
Probably worth waiting on for right now.  We're trying to figure out how to respond to some architectural issues that were raised a few weeks ago.

cvan:
@NellWaliczek thanks for the update

toji:
I'm closing this down, as the input proposal as given here is effectively a non-started due to native API compatibility issues, and a revised proposal is better off in it's own PR.

Issue #282, opened 2017-08-21
Feature Request: Emulated standing space


toji:
Throw this one into the "probably a 2.1 feature" bucket.

It would be a huge developer convenience to enable emulated "stage" frame of references where where, if an actual calibrated room layout is not available the developer could specify a fallback height to be used, at which point the data returned would effectively be an "eyeLevel" frame of ref with the origin offset on the Y axis by the requested height.

This would allow developers to use the API-provided matrices directly in more cases, simplifying code significantly. It also encourage apps to gracefully fall back and support a wider variety of devices rather than just throw up an error message if standing space isn't available.

kearwood:
This may provide a lot of benefit through enabling users to configure their own height and having it consistent across sites.

toji:
This was brought up on the most recent implementer call and was well received as a potential quality of life improvement for developers in 2.0. I'll draw up a pull request with a more formal proposal ASAP.

toji:
This is now in the explainer.

PR #283, opened 2017-08-22
Added section about `stage` emulation

toji:
Following proposal in #282. This would definitely simplify a lot of the code in some of the samples I've been working on and I believe would make it easier for developers to create experiences that work on the widest possible range of devices.

Comments welcome!

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -396,17 +396,9 @@ Some VR devices have been configured with details about the area they are being 
 ```js
 // Try to get a frame of reference where the floor is at Y = 0
 vrSession.requestFrameOfReference("stage").then((frameOfRef) => {
+  // Will always succeed due to stage emulation. See the section titled
+  // "Emulated stage frame of reference" for details.
   vrFrameOfRef = frameOfRef;
-}).catch((err) => {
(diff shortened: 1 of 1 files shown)


kearwood:
This is looking very good -- I only have one suggestion:

Instead of "stageEmulationAllowed: true" could we have this the default, and allow override with "disallowStageEmulation: true"?

I imagine that web developers may ignore this part of the spec entirely once they get a site working, resulting in content that may break when viewed on browsers/devices that only provide an emulated stage.  By explicitly disabling emulation, the developer has confirmed that they will be expecting the stage to be potentially not set.

toji:
We discussed Kip's suggestion on the WebVR call yesterday and it was well received, definitely seems like the right path for ensuring developers hit a "pit of success" as frequently as possible. Updated the pull request to reflect that. (I used `disableStageEmulation` instead of `disallowStageEmulation` for the simple reason that it's one character shorter. 😅)

toji:
Haven't heard any disagreement with this feature, so merging now and specific concerns can be handled via individual bugs.

PR #284, opened 2017-08-23
Fix typo

shaoboyan:
I think this is a typo.
discreet GPU -> discrete GPU
informaton -> information


shaoboyan:
@toji find a possible typo. PTAL :)

toji:
LOL. Well, maybe it's just a really careful GPU, right? 😁 Good catch, thanks for the fix!

PR #289, opened 2017-09-28
Refactor getDevices into requestDevice

toji:
Got an initial pull request up, but this is going to take some discussion.

The basic idea is to drop `getDevices` in favor of a `requestDevice` call that can take an optional set of filters. Credit where credit is due, this is basically what @cvan suggested we do [way back in January](https://github.com/w3c/webvr/issues/179#issuecomment-275284599). I felt that we should try and keep the API simpler. Jokes on me, I guess. ;)

Some of the important points:

This follows very closely the pattern already happily established by [WebUSB](https://wicg.github.io/webusb/#enumeration) and [WebBluetooth](https://webbluetoothcg.github.io/web-bluetooth/#device-discovery), right down the the exact function name. That immediately puts my mind at ease a bit, in that it means we feel more at home next to those APIs and means we're able to borrow from the expertise that went into establishing those standards.

Our filters are much more minimal than the other two APIs, however, since there is no perceived benefit and several real downsides to being overly specific in this case. We generally want to encourage developers to take whatever device matches their most coarse requirements (ie: needs to support exclusive presentation, in the future needs to support AR passthrough?). We definitely don't want to allow or encourage "I want a device with at least a screen resolution of X, FoV no smaller than Y, and that supports at least Z simultaneous layers."

There was a question on the most recent call about if we should drop the `VRDevice` object, since we could feasibly just request sessions directly from navigator.vr with this pattern. After stewing on it for a bit I think the answer is "Yes... but there's an issue." Specifically, we still need a way to ensure that WebGL contexts are created with a compatible adapter, and I like the pattern we've established of using the `VRDevice` to indicate what adapter the context needs to be compatible with. (I personally feel it's a hard requirement to allow developers to create compatible contexts prior to the creation of a session.) So in my opinion `VRDevice` stays and simply loses the name string.

That creates a minor complication when it comes to multiple devices connected to a single PC, though. I had this fanciful idea that the `VRDevice` didn't need to represent a physical piece of hardware, and could instead represent all the potential devices that could satisfy the filters, and only when the user attempted to create a `VRSession` would the system throw up a prompt (if necessary) and say "Hey, which of these two headsets do you want to present to?" That breaks the ability to use `VRDevice` as the handle for context compatibility, though, since it's entirely possible those two headsets are connected to different adapters. (It's also just kinda weird when set next to WebUSB and WebBluetooth, both of whom treat their respective devices as a physical piece of hardware.)

So I think a `VRDevice` needs to continue representing just one piece of hardware, which then begs the question of how the poor power user with 4 GPUs chooses between their Rift and Vive when viewing WebVR content.

However, I don't think it's a big enough issue for us to try and bend over backwards for. A few things to consider:

- Most users will only have one possible device, either because it's their phone, it's a standalone HMD, or because headsets are expensive and power hungry so it doesn't make much sense to have more than one per machine.
- Even in the case that you do have N devices available, you'll often have a clear indication that the user wants to use a specific device. (If the user is viewing a VR browser in one HMD, they definitely don't want WebVR content to show up on the other.)
- Hardware trends and basic financial factors seem to be favoring standalone and mobile systems long term anyway.
- In the case that it truly is ambiguous the browser could have some internal settings page where the default VR device is set.

Something I don't want to see happen, though, is for pages to pop up a "Which device do you mean?" dialog upon calls to `requestDevice`, because I would expect that to happen early on in most pages lifetime as they try to determine if they should advertise VR features. A UA-driven selection dialog at that step would mean that many pages would immediately ask you to pick a headset upon navigation even though you've made no indication you want to use VR. That's severely user-hostile behavior.

One thing I'd really appreciate is hearing from @grorg if these changes would satisfy Apple's fingerprinting concerns sufficiently. I think we can work out the above issues, but if this change doesn't get us closer to enabling eventual implementation in WebKit then it's something of a moot point.

I need to add a section about how to use the filters, so this PR isn't quite complete even if everyone agrees on the functionality, but I wanted to get the conversation going. As a heads up, I'll be unavailable for the entirety of the week of Oct 2nd, but I hope that the rest of the community group can work towards a resolution in the meantime. This is an important change to finalize soon so that we don't get stuck in spec limbo.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -42,36 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
 
 The basic steps any WebVR application will go through are:
 
- 1. Request a VR device that supports the presentation modes the application needs.
+ 1. Request a VR device.
  1. If a device is available, application advertises VR functionality to the user.
- 1. User performs an action that indicates they want to enter VR mode.
- 1. Request a VR session from the device to present VR content with.
- 1. Begin a render loop that produces graphical frames to be displayed on the VR device.
+ 1. Request a VR session from the device in response to a user activation event.
+ 1. Use the session to run a render loop that produces graphical frames to be displayed on the VR device.
  1. Continue producing frames until the user indicates that they wish to exit VR mode.
  1. End the VR session.
 
 ### Acquiring a Device
 
-The first thing that any VR-enabled page will want to do is request a `VRDevice` and, if present, advertise VR functionality to the user.
+The first thing that any VR-enabled page will want to do is request a `VRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the DOM that the user can click to start VR content.)
 
-`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard or Samsung Gear VR). It may also represent devices without stereo presentation capabilities but more advanced tracking, such as ARCore/ARKit enabled devices.
+`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. In no `VRDevice` is available it will resolve to null. The promise will be rejected if an error occurs, such as the the page not having the appropriate permissions to access VR capabilities.
+
+A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung GearVR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices.
 
 ```js
-navigator.vr.requestDevice().then(device => {
-  onVRAvailable(device);
-}, (err) => {
-  // Could not find any available VR hardware or an error occurred querying VR
-  // hardware.
-});
+function checkForVR() {
+  navigator.vr.requestDevice().then(device => {
+    if (device) {
+      onVRAvailable(device);
+    }
+  }, err => {
+    // An error occurred while requesting a VRDevice.
+    console.error('Unable to request a VR device.');
+  });
+}
 ```
 
-If there are multiple VR hardware devices available the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the page lifetime when the user has not made any indication that they wish to use VR features.
+Future revisions of the API may add filter criteria to `navigator.vr.requestDevice`.
 
-Future iterations of the API may add filter criteria to `requestDevice`.
+> **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without and user activation.
+
+It's possible that even if no VR device is available initially one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected at disconnected at any time. To respond to these changes pages can listen to the `devicechange` event on the `navigator.vr` object.
+
+```js
+navigator.vr.addEventListener('devicechange', checkForVR);
+```
 
 ### Sessions
 
-A `VRDevice` indicates the presence of a VR hardware device but provides very little information about it. In order to do anything that involves the hardware's presentation or tracking capabilities the application will need to request a `VRSession` from the `VRDevice`.
+A `VRDevice` only indicates the availabilty of a VR device. In order to do anything that involves the device's presentation or tracking capabilities, the application will need to request a `VRSession` from the `VRDevice`.
 
 Sessions can be created with one of two levels of access:
 
@@ -620,7 +631,7 @@ partial interface Navigator {
 };
 
 [SecureContext, Exposed=Window] interface VR : EventTarget {
-  attribute EventHandler ondevicechanged;
+  attribute EventHandler ondevicechange;
   Promise<VRDevice> requestDevice();
 };
 
--- a/explainer.md
+++ b/explainer.md
@@ -44,7 +44,7 @@ The basic steps any WebVR application will go through are:
 
  1. Request a VR device.
  1. If a device is available, application advertises VR functionality to the user.
- 1. Request a VR session from the device in response to a [user-activation event](https://html.spec.whatwg.org/multipage/interaction.html#activation).
+ 1. Request an exclusive VR session from the device in response to a [user-activation event](https://html.spec.whatwg.org/multipage/interaction.html#activation).
  1. Use the session to run a render loop that produces graphical frames to be displayed on the VR device.
  1. Continue producing frames until the user indicates that they wish to exit VR mode.
  1. End the VR session.
@@ -55,7 +55,7 @@ The first thing that any VR-enabled page will want to do is request a `VRDevice`
 
 `navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. If no `VRDevice` is available, it will resolve to null. The promise will be rejected if an error occurs, such as the the page not having the appropriate permissions to access VR capabilities.
 
-A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung GearVR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices.
+A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung Gear VR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices.
 
 ```js
 function checkForVR() {
@@ -72,7 +72,7 @@ function checkForVR() {
 
 Future revisions of the API may add filter criteria to `navigator.vr.requestDevice`.
 
-> **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without and user activation.
+> **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device-selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without and user activation.
 
 It's possible that even if no VR device is available initially, one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected or disconnected at any time. Pages can listen to the `devicechange` event emitted on `navigator.vr` to respond to changes in device availability after the page loads. (VR devices already available when the page loads will not cause a `devicechange` event to be fired.)
 
--- a/explainer.md
+++ b/explainer.md
@@ -54,19 +54,20 @@ The basic steps any WebVR application will go through are:
 
 The first thing that any VR-enabled page will want to do is request a `VRDevice` and, if present, advertise VR functionality to the user.
 
-`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an available device which matches the given filter criteria. A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard or Samsung Gear VR). It may also represent devices without stereo presentation capabilities but more advanced tracking, such as Tango devices.
+`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard or Samsung Gear VR). It may also represent devices without stereo presentation capabilities but more advanced tracking, such as ARCore/ARKit enabled devices.
 
 ```js
-let vrDevice = null;
-
-navigator.vr.requestDevice().then((device) => {
-  vrDevice = device;
-  onVRAvailable();
+navigator.vr.requestDevice().then(device => {
+  onVRAvailable(device);
 }, (err) => {
   // Could not find any available VR hardware or an error occurred querying VR
   // hardware.
 });
 ```
+
+If there are multiple VR hardware devices available the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the page lifetime when the user has not made any indication that they wish to use VR features.
+
+Future iterations of the API may add filter criteria to `requestDevice`.
 
 ### Sessions
 
@@ -85,7 +86,11 @@ If a `VRDevice` is available and able to create an exclusive session, the applic
 In the following examples we will focus on using exclusive sessions, and cover non-exclusive session use in the [`Advanced Functionality`](#non-exclusive-sessions-magic-windows) section. With that in mind, we ask here if the `VRDevice` supports sessions with `exclusive` access (the default), since we want the ability to display imagery on the headset.
 
 ```js
-async function onVRAvailable() {
+let vrDevice = null;
+
+async function onVRAvailable(device) {
+  vrDevice = device;
+
   // Most (but not all) VRDevices are capable of granting exclusive access to
   // the device, which is necessary to show imagery in a headset. If the device
   // has that capability the page will want to add an "Enter VR" button (similar
@@ -614,17 +619,9 @@ partial interface Navigator {
   readonly attribute VR vr;
 };
 
-dictionary VRDeviceFilter {
-  boolean exclusive = false;
-};
-
-dictionary VRDeviceRequestOptions {
-  required sequence<VRDeviceFilter> filters;
-};
-
 [SecureContext, Exposed=Window] interface VR : EventTarget {
-  attribute EventHandler ondeviceschanged;
-  Promise<VRDevice> requestDevice(optional VRDeviceRequestOptions options);
+  attribute EventHandler ondevicechanged;
+  Promise<VRDevice> requestDevice();
 };
 
 //
--- a/explainer.md
+++ b/explainer.md
@@ -40,7 +40,7 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
 
 ## Lifetime of a VR web app
 
-The basic steps any WebVR application will go through are:
+The basic steps most WebVR applications will go through are:
 
  1. Request a VR device.
  1. If a device is available, application advertises VR functionality to the user.
@@ -72,7 +72,7 @@ function checkForVR() {
 
 Future revisions of the API may add filter criteria to `navigator.vr.requestDevice`.
 
-> **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device-selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without and user activation.
+> **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device-selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without user activation.
 
 It's possible that even if no VR device is available initially, one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected or disconnected at any time. Pages can listen to the `devicechange` event emitted on `navigator.vr` to respond to changes in device availability after the page loads. (VR devices already available when the page loads will not cause a `devicechange` event to be fired.)
 
--- a/explainer.md
+++ b/explainer.md
@@ -44,16 +44,16 @@ The basic steps any WebVR application will go through are:
 
  1. Request a VR device.
  1. If a device is available, application advertises VR functionality to the user.
- 1. Request a VR session from the device in response to a user activation event.
+ 1. Request a VR session from the device in response to a [user-activation event](https://html.spec.whatwg.org/multipage/interaction.html#activation).
  1. Use the session to run a render loop that produces graphical frames to be displayed on the VR device.
  1. Continue producing frames until the user indicates that they wish to exit VR mode.
  1. End the VR session.
 
 ### Acquiring a Device
 
-The first thing that any VR-enabled page will want to do is request a `VRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the DOM that the user can click to start VR content.)
+The first thing that any VR-enabled page will want to do is request a `VRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the page that the user can click to start VR content.)
 
-`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. In no `VRDevice` is available it will resolve to null. The promise will be rejected if an error occurs, such as the the page not having the appropriate permissions to access VR capabilities.
+`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. If no `VRDevice` is available, it will resolve to null. The promise will be rejected if an error occurs, such as the the page not having the appropriate permissions to access VR capabilities.
 
 A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung GearVR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices.
 
@@ -65,7 +65,7 @@ function checkForVR() {
     }
   }, err => {
     // An error occurred while requesting a VRDevice.
-    console.error('Unable to request a VR device.');
+    console.error('Unable to request a VR device:', err);
   });
 }
 ```
@@ -74,7 +74,7 @@ Future revisions of the API may add filter criteria to `navigator.vr.requestDevi
 
 > **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without and user activation.
 
-It's possible that even if no VR device is available initially one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected at disconnected at any time. To respond to these changes pages can listen to the `devicechange` event on the `navigator.vr` object.
+It's possible that even if no VR device is available initially, one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected or disconnected at any time. Pages can listen to the `devicechange` event emitted on `navigator.vr` to respond to changes in device availability after the page loads. (VR devices already available when the page loads will not cause a `devicechange` event to be fired.)
 
 ```js
 navigator.vr.addEventListener('devicechange', checkForVR);
@@ -82,7 +82,7 @@ navigator.vr.addEventListener('devicechange', checkForVR);
 
 ### Sessions
 
-A `VRDevice` only indicates the availabilty of a VR device. In order to do anything that involves the device's presentation or tracking capabilities, the application will need to request a `VRSession` from the `VRDevice`.
+A `VRDevice` indicates only the availability of a VR device. In order to do anything that involves the device's presentation or tracking capabilities, the application will need to request a `VRSession` from the `VRDevice`.
 
 Sessions can be created with one of two levels of access:
 
--- a/explainer.md
+++ b/explainer.md
@@ -42,43 +42,35 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
 
 The basic steps any WebVR application will go through are:
 
- 1. Request a list of the available VR devices.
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -42,43 +42,35 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an available device which matches the given filter criteria. A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard or Samsung Gear VR). It may also represent devices without stereo presentation capabilities but more advanced tracking, such as Tango devices.
```
cvan: `Tango` -> `ARCore/ARKit`?

explainer.md
```diff
@@ -42,43 +42,35 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
-    // which device to use.
-    vrDevice = devices[0];
-    onVRAvailable();
-  } else {
-    // Could not find any VR hardware connected.
-  }
-}, err => {
-  // An error occurred querying VR hardware. May be the result of blocked
-  // permissions by a parent frame.
+navigator.vr.requestDevice().then((device) => {
```
cvan: nit: don't need the parentheses since this is a single, non-expanded argument

explainer.md
```diff
@@ -42,43 +42,35 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+  vrDevice = device;
+  onVRAvailable();
```
cvan: do you think we might as well have this example pass `vrDevice` as a parameter to the [`onVRAvailable` function](https://github.com/w3c/webvr/blob/8cd20faf/explainer.md#detecting-and-advertising-vr-mode)? as "best practices" sample code, having the `onVRAvailable` example function accept a single argument could be nicer than exposing a global variable.

explainer.md
```diff
@@ -622,19 +614,24 @@ partial interface Navigator {
...
 
-[SecureContext, Exposed=Window] interface VR : EventTarget {
-  attribute EventHandler ondeviceconnect;
-  attribute EventHandler ondevicedisconnect;
+dictionary VRDeviceFilter {
+  boolean exclusive = false;
+};
+
+dictionary VRDeviceRequestOptions {
+  required sequence<VRDeviceFilter> filters;
```
cvan: they're simple, but can we add a sample code block showing usage of filters?

explainer.md
```diff
@@ -622,19 +614,24 @@ partial interface Navigator {
...
+};
 
-  Promise<sequence<VRDevice>> getDevices();
+[SecureContext, Exposed=Window] interface VR : EventTarget {
```
cvan: can we have this (and others) exposed on `(Window,Worker)`? or do you feel that warrants a separate issue and discussion?

explainer.md
```diff
@@ -622,19 +614,24 @@ partial interface Navigator {
...
+  attribute EventHandler ondeviceschanged;
```
cvan: so we don't have the plural vs. singular and active vs. past tenses, do you think `onchange` might read better? (the event names in the event interfaces of the [Web Bluetooth](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetooth-onavailabilitychanged) and [Sensor](https://www.w3.org/TR/generic-sensor/#the-sensor-interface) APIs are inconsistent, but the [Permissions API uses `onchange`](https://w3c.github.io/permissions/#status-of-a-permission), which I feel is pretty unambiguous. still doesn't disambiguate from `change` vs. `changed`, but it's still better I think.)

explainer.md
```diff
@@ -622,19 +614,24 @@ partial interface Navigator {
...
+  Promise<VRDevice> requestDevice(optional VRDeviceRequestOptions options);
```
cvan: to avoid adding the new dictionary interface, `VRDeviceRequestOptions`, can this instead be a plain `object`? (inspiration: [`navigator.permissions.query`](https://w3c.github.io/permissions/#permissions-interface)) I seem to recall @annevk's making a similar suggestion in prior discussions.

explainer.md
cvan: on second thought, perhaps it's better to be explicit. but, if only one event is emitted per `VRDevice` that is changed, IMO this should be `ondevicechanged` (singular) instead of `ondeviceschanged` (plural).

explainer.md
cvan: why's this `required` if it's [`optional` as an argument](https://github.com/w3c/webvr/pull/289/files#diff-817a23c792520ce222813d3eff70d031R627)?

explainer.md
cvan: shouldn't there be some `state` associated with the event emitted? from a web developer's POV this is a bit hard to work with:

```js
navigator.vr.addEventListener('devicechange', evt => {
  // Which VR device changed? And to which state did it change?
});
```

see the [Permissions API](https://w3c.github.io/permissions/#status-of-a-permission) for similar cases.

explainer.md
annevk: I'd try to avoid `object` as it means you have to write your own binding logic in prose.

cvan: gotcha, makes sense. I assume that was the rationale behind this issue you filed: https://github.com/w3c/permissions/issues/158

cvan:
Apologies in advance for talking about this yet again! I know we've talked about this in the past; both other web specs and my thinking have changed (for the better). (Spoiler: I still think **No UI** is the best path forward.)

But to address your concerns about the Permissions aspect of `navigator.vr.requestDevice` …

I found this comment in the Permissions API spec: https://github.com/w3c/permissions/issues/158#issuecomment-331551550

> In my view, it's less a request for permission elevation, more a request to turn on the accelerometer (now or forever).

From https://github.com/w3c/permissions/issues/158#issuecomment-333530815:

> We could also [encourage UAs to merge prompts](https://w3c.github.io/mediacapture-main/getusermedia.html#handling-devices) within a narrow window of time.

> If a script calls getUserMedia() multiple times before reaching a stable state, this document advises the UI designer that the permission dialogs should be merged, so that the user can give permission for the use of multiple cameras and/or media sources in one dialog interaction. The constraints on each getUserMedia call can be used to decide which stream gets which media sources.

(Note: these comments aren't directly related to the parent issue in w3c/permissions#158 [https://github.com/w3c/permissions/issues/158#issue-259866485], but they're still useful for the context of this discussion.)

<hr>

This reminds me of how the [Fullscreen API](https://fullscreen.spec.whatwg.org/) works with [Pointer Lock](https://www.w3.org/TR/pointerlock/). Neither API is spec'd to use the [Permissions API](https://w3c.github.io/permissions/#permission-registry) (I'll assume for historical, legacy reasons), and I do also share the concern about compounding the problem with introducing mandatory UI prompts.

Today, browsers don't require two clicks for web sites to engage Pointer Lock ([`requestPointerLock`](https://fullscreen.spec.whatwg.org/#dom-element-requestfullscreen
)) after engaging Full Screen ([`requestPointerLock`](https://w3c.github.io/pointerlock/#dom-element-requestpointerlock)). This behaviour isn't yet defined in the Permissions API (but [could be](https://github.com/w3c/sensors/issues/183)).

From the [Pointer Lock spec](https://w3c.github.io/pointerlock/#dom-element-requestpointerlock):

> Pointer lock is commonly combined with fullscreen [[FULLSCREEN](https://w3c.github.io/pointerlock/#bib-FULLSCREEN)], and if an [engagement gesture](https://w3c.github.io/pointerlock/#dfn-engagement-gesture) is used to enter fullscreen it is sufficient for a subsequent requestPointerLock.

<hr>

**I would suggest that we build `navigator.vr.requestDevice` with the Permissions API in mind** (but at this time a `navigator.permissions.query({name: 'vr'})` request ought not to be required).

Similar to the [ongoing spec changes](https://github.com/w3c/webvr/commit/5d3050c9#commitcomment-24722876) to the [Sensor API](https://www.w3.org/TR/generic-sensor/#permissions) (i.e., `deviceorientation`, `devicemotion`), I do not think we need to introduce UI prompts but access could be gated by the Permissions API.

> Access to sensor readings are controlled by the Permissions API [PERMISSIONS]. User agents use a number of criteria to grant access to the readings. Note that access can be granted without prompting the user.

If we can add similar text in this PR, I'd be a happy camper!

Thanks for taking this one on. It's a tricky issue that hasn't yet been fully addressed in other APIs either, so thanks. :+1:

explainer.md
annevk: No, that had a different reason (and doesn't seem to be going anywhere).

annevk:
How could `navigator.permissions.query({name: 'vr'})` be required? It's just a permission check. If the browser grants that permission by default, it would end up returning "granted".

Also, to be clear, you cannot require any kind of specific UX. User agents have to be free to implement a flow they think makes sense. (It helps enormously to sketch these out in advance and verify them with security UX folks, as it can influence the API a bit.)

cvan:
> How could `navigator.permissions.query({name: 'vr'})` be required? It's just a permission check. If the browser grants that permission by default, it would end up returning "granted".

1. User loads a WebVR site
2. `navigator.permissions.query({name: 'vr'})` gets called
3. UA shows Permissions dialog box
4. User clicks `Deny`
5. Promise from `navigator.permissions.query({name: 'vr'})` gets resolved with `{state: 'rejected'}`
6. Content should subsequently fail whenever `navigator.vr.getDevices()` is called

Am I correct in thinking that this behaviour should be spec'd?

How it's spec'd right now and implemented, as long it the action was [user initiated](https://github.com/w3c/webvr/blob/8cd20faf/explainer.md#beginning-a-vr-session), a call to `navigator.vr.getDevices()` is always granted, regardless of browser permissions.

FWIW, this is what the [Sensor API](https://w3c.github.io/sensors/#permissions) spec says:

> Access to <a data-link-type="dfn" href="https://w3c.github.io/sensors/#sensor-readings" id="ref-for-sensor-readings①⓪">sensor readings</a> are controlled by the Permissions API <a data-link-type="biblio" href="https://w3c.github.io/sensors/#biblio-permissions">[PERMISSIONS]</a>.
User agents use a <a data-link-type="dfn" href="https://w3c.github.io/permissions/#new-information-about-the-users-intent" id="ref-for-new-information-about-the-users-intent">number of criteria</a> to grant access to the <a data-link-type="dfn" href="https://w3c.github.io/sensors/#sensor-readings" id="ref-for-sensor-readings①①">readings</a>.
Note that access can be granted without prompting the user.

<hr>

> Also, to be clear, you cannot require any kind of specific UX. User agents have to be free to implement a flow they think makes sense. (It helps enormously to sketch these out in advance and verify them with security UX folks, as it can influence the API a bit.)

Understood, thanks. It's certainly something we've been careful to not describe, even in non-normative text. I see [this passage](https://w3c.github.io/permissions/#requesting-more-permission) repeated in the Permissions API:

> This is intentionally vague about the details of the permission UI and how the UA infers user intent. UAs should be able to explore lots of UI within this framework.

<hr>

As an aside, but related:

Firefox 55+ always requires a user-initiated event (can be manually disabled with the `dom.vr.require-gesture` flag) for the first time content presents to VR.

In Chrome for Android, `navigator.getVRDisplays()` was gated on a user-initiated gesture (usually a `touch*` event) . That's [changed slightly in Chrome 62](https://developers.google.com/web/updates/2017/09/webvr-origin-trial-chrome-62):

> The synthetic click event at viewport (0, 0) has been removed (for both Cardboard and the Daydream controller touchpad) (bug). The `vrdisplayactivate` event is now considered a user gesture, and may be used to request presentation and begin media playback, without relying on the click event.

annevk:
I think you misunderstand `query()`. `query()` never triggers UI.

Note that requiring "user activation" (that's the HTML Standard term for "gesture") should be written down: https://html.spec.whatwg.org/#triggered-by-user-activation. E.g., Fullscreen does require that.

cvan:
> I think you misunderstand `query()`. `query()` never triggers UI.

From your original comment, I now see what you mean.

The Permissions API just returns a `state` of `'prompt'` if the API is blocked on user activation. Chrome, for example, will only return `'granted'` or `'denied`' after a user has clicked the `Allow`/`Block` buttons in the permissions UI prompt. And the only way Chrome's permissions UI prompt will show up is when a "powerful API" is invoked upon a user-initiated event.

Thanks for the clarification. This helps a ton.

toji:
FYI: I just got back from vacation so I haven't had a chance to parse all this yet but thank you for the detailed discussion! I'll respond to things individually as I get a chance.

explainer.md
toji: Agreed that we need example code that uses filters, I'll try to add that. As for the required on an optional, it's a pattern I've seen elsewhere before. Simply saying that the dictionary itself is optional, but if you include one then you must specify filters. (Otherwise why are you passing the dictionary at all?)

explainer.md
toji: Definitely want to get there, but I do feel it's worth having it's own issue. I don't _expect_ it to be controversial, but any conversation that does crop up around it deserves its own space rather than being intermixed with this change.

explainer.md
toji: I'm not too opinionated on the exact verbiage (though I think "devices" is more accurate than "device", since there may in fact be more than one.)

I don't know what we could appropriately surface in the event without brushing up against the fingerprinting concerns that prompted this change, though. If we deliver the devices in question via this event then developers could presumably just monitor the event to get a more complete view of the system than the query delivers.

The purpose of the event is simply to indicate to developers that if they had previously asked for a device and didn't get an appropriate one back they may now want to try again. Given that, is there an appropriate bit of contextual information that you can think of that would be useful here without describing the available hardware explicitly? (It's a bit of an annoying song and dance, I know, but: Web.)

toji:
Okay, catching up now. The permissions discussion is very interesting to me, because I also had some misconceptions about what it was doing. To clarify, this is how I think it works now based on the above discussion:

- Calling `navigator.permissions.query({name: 'vr'})` would return "granted" if the page is allowed to use VR features, "denied" if the page has been prevented from using VR features (likely from a previously shown prompt), or "prompt" if the next attempt to use VR features would result in UI being shown. The call to `query` itself will never cause UI to be shown.
- Any UI that does show up will be the result of designated WebVR API calls (such as potentially `vrDisplay.requestSession()`), and although we can specify that it requires user activation we must only suggest places where UI _may_ be appropriate, not require it.
- The WebVR spec can independently outline what actions have permissions associated with them without blocking on inclusion or browser implementation of the permissions API, which merely provides a way to surface the current state of any permissions we define.

Correct?

annevk:
Yeah, that looks correct to me. @jyasskin you might want to double check the previous comment in this issue.

explainer.md
toji: Done

explainer.md
toji: Done

explainer.md
toji: Done, though I still set the global variable in onVRAvailable simply because it's easier to track over the various example snippets than constantly passing the device around.

explainer.md
toji: After discussion with the rest of the implementers we've decided to remove the filter criteria in the initial version so that we can have more detailed discussions about what's appropriate to filter on before we add it back in down the road. As such the need for example code is now moot. 😉

explainer.md
toji: Changed to `ondevicechanged`

toji:
Should also mention that we got a tentative "OK" from Apple on this route addressing the core of their fingerprinting concerns. Dean is going to get additional confirmation, but for the moment it looks like this gets us to a better place in terms of cross vendor implementability.

toji:
One final note: My impression at this point is that the latest version of this proposal has broad vendor support so unless concerns are raised I'm inclined to merge it within a day or so. As always we can iterate on individual details in separate issues after that.

explainer.md
```diff
@@ -623,18 +620,15 @@ partial interface Navigator {
 };
 
 [SecureContext, Exposed=Window] interface VR : EventTarget {
-  attribute EventHandler ondeviceconnect;
-  attribute EventHandler ondevicedisconnect;
-
-  Promise<sequence<VRDevice>> getDevices();
+  attribute EventHandler ondevicechanged;
```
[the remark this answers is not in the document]
cvan: I know I've asked this like a dozen different times, sorry lol…

should this be past or activetense?

explainer.md
```diff
@@ -623,18 +620,15 @@ partial interface Navigator {
...
+  Promise<VRDevice> requestDevice();
```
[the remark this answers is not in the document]
cvan: would you be opposed to `getAvailableDevice`? (borrowing a bit from the [Remote Playback API](https://w3c.github.io/remote-playback/) which is making its way to [CR](https://github.com/w3c/remote-playback/issues/73) - which obviously doesn't mean it has absolutely the best spec API but it's probably a good signal.)

explainer.md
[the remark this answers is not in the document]
cvan: IMO, this change is probably deserving of its own PR. but I think perhaps this needs to be a bit more nuanced. a `state` property ([example](https://w3c.github.io/remote-playback/#dfn-remote-playback-devices)) seems like it'd be valuable.

jyasskin:
https://github.com/w3c/webvr/pull/289#issuecomment-336536775 looks correct to me. I haven't fully digested the rest of this thread, or the rest of WebVR.

I'll point out that the Permissions Spec defines a couple operations around requesting permission, which are intended for use from other specs, and which you can use even in browsers that don't implement the Permissions API (`permissions.query()`). For a `requestDevice()` kind of operation, you probably want https://w3c.github.io/permissions/#prompt-the-user-to-choose, which gives the UA a list of options for the user to choose between, and leaves it up to the UA how it wants to let the user choose.

You may instead want https://w3c.github.io/permissions/#request-permission-to-use if you intend a yes/no kind of request.

Using these makes sure that you don't have to change your spec if folks later want to give UAs more flexibility in how they show prompts.

explainer.md
```diff
@@ -42,43 +42,36 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
- 5. Request a VR session to present VR content with.
- 6. Begin a render loop that produces graphical frames to be displayed on the VR device.
- 7. Continue producing frames until the user indicates that they wish to exit VR mode.
- 8. End the VR session.
+ 1. Request a VR device that supports the presentation modes the application needs.
+ 1. If a device is available, application advertises VR functionality to the user.
+ 1. User performs an action that indicates they want to enter VR mode.
+ 1. Request a VR session from the device to present VR content with.
+ 1. Begin a render loop that produces graphical frames to be displayed on the VR device.
+ 1. Continue producing frames until the user indicates that they wish to exit VR mode.
```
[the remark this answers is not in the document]
toji: I'm working on rewording this a bit, but I think this is the core of it: This section is to give a quick overview of the expected API use pattern, so it should favor brevity. I'll try to apply these suggestions throughout the rest of the doc, though. (Possibly as a follow up PR for clarity?)

explainer.md
```diff
@@ -42,43 +42,36 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+ 1. End the VR session.
 
-### Device enumeration
+### Acquiring a Device
 
-The first thing that any VR-enabled page will want to do is enumerate the available VR hardware and, if present, advertise VR functionality to the user.
+The first thing that any VR-enabled page will want to do is request a `VRDevice` and, if present, advertise VR functionality to the user.
```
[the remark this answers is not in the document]
toji: "Advertise to the user" in this case meaning "Add a VR button to the DOM" or similar. I've added some text to that effect to the explainer.

explainer.md
```diff
@@ -42,43 +42,36 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard or Samsung Gear VR). It may also represent devices without stereo presentation capabilities but more advanced tracking, such as ARCore/ARKit enabled devices.
```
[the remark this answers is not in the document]
toji: Made the suggested changes, and added a section about the `devicechange` event, which is the appropriate way to monitor connects/disconnects.

explainer.md
```diff
@@ -42,43 +42,36 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
-  // permissions by a parent frame.
+navigator.vr.requestDevice().then(device => {
+  onVRAvailable(device);
+}, (err) => {
+  // Could not find any available VR hardware or an error occurred querying VR
+  // hardware.
 });
 ```
 
+If there are multiple VR hardware devices available the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the page lifetime when the user has not made any indication that they wish to use VR features.
```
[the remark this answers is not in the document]
toji: Given that this isn't a spec proper the concept of "non-normative" is a little weird, but I've tried to make it a bit more apparent anyway.

explainer.md
```diff
@@ -623,18 +620,15 @@ partial interface Navigator {
 };
 
 [SecureContext, Exposed=Window] interface VR : EventTarget {
-  attribute EventHandler ondeviceconnect;
```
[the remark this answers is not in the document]
toji: Which documentation are you referring to in this case? There's not much in the way of docs for the 2.0 API yet.

explainer.md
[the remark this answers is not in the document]
toji: FWIW: Changed this to `devicechange` after reading through it a few more times. I think that is indeed the preferred tense. Not sure if I agree this should be pulled out into it's own PR, though. I feel like the individual connect/disconnect events don't make sense once the function has changed over.

explainer.md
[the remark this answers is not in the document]
toji: Both WebUSB and WebBluetooth use `getDevice`. (`navigator.usb.getDevice` and `navigator.bluetooth.getDevice` respectively.) As a result I'd really like to carry that verbiage through to this API to reinforce the sense that this is part of the same Web API family.

toji:
Thanks for the additional feedback, @cvan! Made some updates and left some responses. And thanks for the clarification @jyasskin. I'll definitely make use of those definitions in the spec text.

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
- 2. Checks to see if the desired device supports the presentation modes the application needs.
- 3. If so, application advertises VR functionality to the user.
- 4. User performs an action that indicates they want to enter VR mode.
- 5. Request a VR session to present VR content with.
- 6. Begin a render loop that produces graphical frames to be displayed on the VR device.
- 7. Continue producing frames until the user indicates that they wish to exit VR mode.
- 8. End the VR session.
+ 1. Request a VR device.
+ 1. If a device is available, application advertises VR functionality to the user.
+ 1. Request a VR session from the device in response to a user activation event.
```
cvan: `user activation event` -> `user-activation event`

and can you link to https://html.spec.whatwg.org/multipage/interaction.html#activation?

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+ 1. Use the session to run a render loop that produces graphical frames to be displayed on the VR device.
+ 1. Continue producing frames until the user indicates that they wish to exit VR mode.
+ 1. End the VR session.
 
-### Device enumeration
+### Acquiring a Device
 
-The first thing that any VR-enabled page will want to do is enumerate the available VR hardware and, if present, advertise VR functionality to the user.
+The first thing that any VR-enabled page will want to do is request a `VRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the DOM that the user can click to start VR content.)
```
cvan: since you're already saying `page` above, I'd say `page` instead of `DOM`

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
 
-`navigator.vr.getDevices` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a list of available devices. Each `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard or Samsung Gear VR). It may also represent devices without stereo presentation capabilities but more advanced tracking, such as Tango devices.
+`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. In no `VRDevice` is available it will resolve to null. The promise will be rejected if an error occurs, such as the the page not having the appropriate permissions to access VR capabilities.
```
cvan: can you add a comma after `available`?

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+
+A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung GearVR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices.
```
cvan: 👍 thanks

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
 
 ```js
-let vrDevice = null;
+function checkForVR() {
+  navigator.vr.requestDevice().then(device => {
+    if (device) {
+      onVRAvailable(device);
```
cvan: looks good 👍

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+    }
+  }, err => {
+    // An error occurred while requesting a VRDevice.
+    console.error('Unable to request a VR device.');
```
cvan: since `err` is used as an argument, should probably add print that too:

```js
    console.error('Unable to request a VR device:', err);
```

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
-    vrDevice = devices[0];
-    onVRAvailable();
-  } else {
-    // Could not find any VR hardware connected.
-  }
-}, err => {
-  // An error occurred querying VR hardware. May be the result of blocked
-  // permissions by a parent frame.
-});
+Future revisions of the API may add filter criteria to `navigator.vr.requestDevice`.
```
cvan: it'd be nice to add links/references (I assume that can come in the spec)

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+
+> **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without and user activation.
```
cvan: 👍

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+
+It's possible that even if no VR device is available initially one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected at disconnected at any time. To respond to these changes pages can listen to the `devicechange` event on the `navigator.vr` object.
```
cvan: can you add a comma after `initially`?

cvan: `connected at disconnected` -> `connected or disconnected`?

cvan: perhaps alternate phrasing:

> To respond to these changes pages can listen to the `devicechange` event on the `navigator.vr` object.

to

> Pages can listen to the `devicechange` event emitted on `navigator.vr` to respond to new information from the device.

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+
+```js
+navigator.vr.addEventListener('devicechange', checkForVR);
```
cvan: will `devicechange` always fire when a document loads and a VR device is detected? or in the presence of an event listener on `navigator.vr`? what's the trigger?

if so, when? in relation to `DOMContentLoaded`, `load`, etc.?

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
 ```
 
 ### Sessions
 
-A `VRDevice` indicates the presence of a VR hardware device but provides very little information about it outside of a name that could be used to select it from a list. In order to do anything that involves the hardware's presentation or tracking capabilities the application will need to request a `VRSession` from the `VRDevice`.
+A `VRDevice` only indicates the availabilty of a VR device. In order to do anything that involves the device's presentation or tracking capabilities, the application will need to request a `VRSession` from the `VRDevice`.
```
cvan: `only indicates the availabilty` -> `indicates only the availability`

(typo: `availability`; and the `only` should be placed next to the noun, which is `the availability`)

explainer.md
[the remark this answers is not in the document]
cvan: I thought I had read a passage on these events - but I was wrong. I was thinking of the spec.

this is fine 👍

explainer.md
cvan: I think the link would be helpful, because this topic seems to come up a lot of `How is user intent defined in WebVR?`

explainer.md
cvan: or `document`?

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
- 2. Checks to see if the desired device supports the presentation modes the application needs.
- 3. If so, application advertises VR functionality to the user.
- 4. User performs an action that indicates they want to enter VR mode.
- 5. Request a VR session to present VR content with.
- 6. Begin a render loop that produces graphical frames to be displayed on the VR device.
- 7. Continue producing frames until the user indicates that they wish to exit VR mode.
- 8. End the VR session.
+ 1. Request a VR device.
+ 1. If a device is available, application advertises VR functionality to the user.
+ 1. Request a VR session from the device in response to a [user-activation event](https://html.spec.whatwg.org/multipage/interaction.html#activation).
```
cvan: thanks :+1:

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
+`navigator.vr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to a `VRDevice` if one is available. If no `VRDevice` is available, it will resolve to null. The promise will be rejected if an error occurs, such as the the page not having the appropriate permissions to access VR capabilities.
+
+A `VRDevice` represents a physical unit of VR hardware that can present imagery to the user somehow, referred to here as a "VR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung GearVR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices.
```
cvan: nittiest nit: `GearVR` -> `Gear VR`

explainer.md
cvan: nit: `device selection UI` -> `device-selection UI`

cvan: typo: `without and`

explainer.md
```diff
@@ -42,43 +42,47 @@ VR provides an interesting canvas for artists looking to explore the possibiliti
...
-  // permissions by a parent frame.
-});
+Future revisions of the API may add filter criteria to `navigator.vr.requestDevice`.
+
+> **Non-normative Note:** If there are multiple VR devices available, the UA will need to pick which one to return. The UA is allowed to use any criteria it wishes to select which device is returned, including settings UI that allows users to manage device priority. Calling `navigator.vr.requestDevice` should not trigger device selection UI, however, as this would cause many sites to display VR-specific dialogs early in the document lifecycle without and user activation.
+
+It's possible that even if no VR device is available initially, one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected or disconnected at any time. Pages can listen to the `devicechange` event emitted on `navigator.vr` to respond to changes in device availability after the page loads. (VR devices already available when the page loads will not cause a `devicechange` event to be fired.)
```
cvan: thanks for adding the last sentence - it’s helpful

explainer.md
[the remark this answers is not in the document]
toji: No, the verbiage just got confusing after applying some suggested edits. I was kind of viewing this list as the steps of a "basic" VR experience, which is mostly about getting the exclusive session running. I've made some tweaks to try and clarify.

toji:
@NellWaliczek: The fingerprinting concern was two fold as I understood it. The device name string was certainly part of it, but there was also the fact that if I did happen to have multiple VR devices this identifies me as one of a very small group, making it far more effective as a fingerprinting bit. Just returning "Yup, here's a device" on the other hand probably just means that you have a newish phone, and that will become less of a unique bit as time goes on.

Then again, maybe Apple was thinking of something completely different. 😅 Either way this change appears to address the concern.

explainer.md
```diff
@@ -623,18 +620,15 @@ partial interface Navigator {
...
+  attribute EventHandler ondevicechanged;
+  Promise<VRDevice> requestDevice();
 };
 
 //
 // Device
 //
 
 [SecureContext, Exposed=Window] interface VRDevice : EventTarget {
-  readonly attribute DOMString deviceName;
```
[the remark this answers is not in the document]
huningxin: I would understand the impact to developer on making "responsive" webvr apps cross HMDs. I mean webvr apps are able to check the `devcieName` and load corresponding assets, let's say different complexity of models, different size of textures etc., to deliver best experience for that HMD. Are there any other ways to do this without `deviceName`?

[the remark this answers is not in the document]
toji: I don't have a great answer for you, unfortunately. (This is something that WebGL in general struggles with.) But I can say with confidence that scaling scene complexity based on the device name is a bad idea. At the very least it would mean that you'd have to update your code every time a new device hit the market, and it's very easy to get wrong (Not only would browser possibly expose different names but developers might check for, say, the string "Vive" and load high-end assets not realizing that HTC has plans to [use the Vive branding for standalone devices](https://www.vive.com/us/product/standalone/) as well.)

There's a variety of methods out there to determine if you're on a mobile device or not, so that could be used as a baseline performance differentiator. Beyond that you could ask users to pick a detail level (high, medium, low) or try to scale it dynamically as your experience runs.

I wish I had a better answer, but this is a problem that extends well beyond WebVR and we wouldn't be doing the platform any favors by encouraging another form of user agent sniffing.

PR #290, opened 2017-09-29
Remove 'Deprecated API' label from WebVR 1.1 spec text. Fixes #288

### LINKED ISSUE #288: Possible to soften DEPRECATED language in the 1.1 spec?
Until the next evolution of the WebVR specification is named and ratified, would it be possible to remove or soften the "Deprecated API" language in the 1.1 spec?
 https://github.com/w3c/webvr/blob/master/spec/1.1/index.bs#L39

The text following that notice pointing out that there are substantial changes and additions underway is fine, but it's not clear to me at this point what the timeline for that is, much less what the browser engine implementation and rollout plans will be (though certainly there will be a polyfill in place earlier, which https://github.com/mozilla/webxr-polyfill shows we should be able to do).

The current version of the specification was already seeing fairly widespread usage, especially on mobile, before Mozilla and Microsoft turned it on by default (previously working via the polyfill), and the current text is leading to tool vendor and developer confusion. I cannot speak for other engines, but I know that Mozilla is committed to supporting this specification either directly or via polyfill until such time as that usage drops below an acceptable threshold.

Thanks!

cc @kearwood @cvan

juj:
Ran `make` to rebuild the docs, which seems to pull in some other updates as well. (earlier commits did not `make` during commit time?)

### DIFF
--- a/spec/1.1/index.bs
+++ b/spec/1.1/index.bs
@@ -36,9 +36,7 @@ urlPrefix: https://www.w3.org/TR/html5/
 </pre>
 
 
-<b style="color: red; font-size: 1.3em">Deprecated API</b>
-
-<b>The version of the WebVR API represented in this document is does not represent the final shape of the API. While the API is being finalized support for this version may temporarily be available in some browsers but is expected to be replaced eventually.</b>
+<b>Development of this version of the WebVR API has halted in favor of being replaced by a newer variant of the API. See latest specification draft <a href="https://w3c.github.io/webvr/spec/latest/">here</a>. Several browsers will continue to support this version of the API in the meantime.</b>
 
 # Introduction # {#intro}
 
--- a/spec/1.1/index.html
+++ b/spec/1.1/index.html
@@ -1176,7 +1176,7 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version aa82ed4ababc6275b796a6e8bae93f32d8a46d2d" name="generator">
+  <meta content="Bikeshed version 5834a15f311a639c0b59ff2edbf3a060391d15ff" name="generator">
   <link href="https://w3c.github.io/webvr/" rel="canonical">
 <style>/* style-md-lists */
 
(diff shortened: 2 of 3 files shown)


spec/1.1/index.bs
```diff
@@ -36,9 +36,7 @@ urlPrefix: https://www.w3.org/TR/html5/
 </pre>
 
 
-<b style="color: red; font-size: 1.3em">Deprecated API</b>
-
-<b>The version of the WebVR API represented in this document is does not represent the final shape of the API. While the API is being finalized support for this version may temporarily be available in some browsers but is expected to be replaced eventually.</b>
+<b>Development of WebVR 1.1 specification is finished. The next version of the WebVR specification is being developed under a WebVR 2.0 version label.</b>
```
toji: Thanks for the update, but I don't agree with the verbiage here. "specification is finished" gives a sense of long-term stability that we explicitly do not want to imply. Additionally, we don't want to lean to heavily on the 2.0 verbiage in public docs, as it again implies that this version is stable and will stick around in tandem to the new API.

I'd suggest something like:
"Development of this version of the WebVR API has halted, and is expected to be replaced by a newer variant (possibly link to latest spec here?) of the API. Several browser will continue to support this version of the API in the meantime."

spec/1.1/index.html
```diff
@@ -1437,10 +1437,10 @@ <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="cont
      <dd class="editor p-author h-card vcard"><a class="p-name fn u-email email" href="mailto:kgilbert@mozilla.com">Kearwood Gilbert</a> (<a class="p-org org" href="https://mozilla.org/">Mozilla</a>)
      <dd class="editor p-author h-card vcard"><a class="p-name fn u-email email" href="mailto:cvan@mozilla.com">Chris Van Wiemeersch</a> (<a class="p-org org" href="https://mozilla.org/">Mozilla</a>)
      <dt>Participate:
-     <dd><span><a href="https://github.com/w3c/webvr/issues/new">File an issue</a> (<a href="https://github.com/w3c/webvr/issues">open issues</a>)</span>
-     <dd><span><a href="https://lists.w3.org/Archives/Public/public-webvr/">Mailing list archive</a></span>
-     <dd><span><a href="irc://irc.w3.org:6665/">W3C’s #webvr IRC</a></span>
-     <dd><span></span>
+     <dd><a href="https://github.com/w3c/webvr/issues/new">File an issue</a> (<a href="https://github.com/w3c/webvr/issues">open issues</a>)
```
toji: These sort of "unnecessary" changes are the result of Bikeshed updates and can't really be avoided, so don't worry about them getting rolled into you patch.

spec/1.1/index.bs
NellWaliczek: How about "Design of the 1.1 version of the WebVR API has halted."

juj: How about

`"Development of this version of the WebVR API has halted in favor of being replaced by a newer variant of the API. See latest specification draft [here](https://w3c.github.io/webvr/spec/latest/). Several browsers will continue to support this version of the API in the meantime."`

to directly link that WebVR 2.0 is the reason for WebVR 1.1 not being worked on anymore, and that there is no pause implied to occur in development.

toji: That sounds good to me!

juj:
Updated the pull request, how does it look now?

toji:
Looks good, thanks! Sorry for the review delay, I was out for a week on vacation and am just now catching up on some of these issues. :)

PR #293, opened 2017-10-17
Remove [EnforceRange] annotations

peria:
Now we have `[EnforceRange]` annotations on **VRWebGLLayerInit**'s framebufferScaleFactor and **VRFrameOfReferenceOptions**'s stageEmulationHeight, and their type is `double`.
According to [Web IDL spec](https://heycam.github.io/webidl/#EnforceRange), `[EnforceRange]` annotation can be associated with only integer types.

It seems to difficult to change their types to integer types, so I think it is good to remove these annotations to make IDLs valid.

***
[Preview](https://s3.amazonaws.com/pr-preview/peria/webvr/RemoveEnforceRange.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webvr/26d1812...peria:e8b054f.html)


toji:
Good catch, I didn't realize this was an integer-only attribute. Thanks!

PR #295, opened 2017-10-24
add `Tests` section to README (for `w3c/web-platform-tests`)

cvan:
per https://github.com/w3c/gamepad/pull/64/files#r146330440

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,17 @@ To generate the spec document (`index.html`) from the `index.bs` [Bikeshed](http
 make
 ```
 
+
+## Tests
+
+For normative changes, a corresponding
+[web-platform-tests](https://github.com/w3c/web-platform-tests) PR is highly appreciated. Typically,
+both PRs will be merged at the same time. Note that a test change that contradicts the spec should
+not be merged before the corresponding spec change. If testing is not practical, please explain why
+and if appropriate [file a web-platform-tests issue](https://github.com/w3c/web-platform-tests/issues/new)
+to follow up later. Add the `type:untestable` or `type:missing-coverage` label as appropriate.
+
+
 ## License
 
 Per the [`LICENSE.md`](LICENSE.md) file:


NellWaliczek:
Love this on principle!  In practice, there are several open discussions around how we can actually *do* such testing.
https://github.com/w3c/webvr/issues/187
https://github.com/w3c/web-platform-tests/pull/5535 

I'm comfortable with merging this change under the understanding that we're still working out the specifics of how to do it and also that we may need some flexibility at the beginning of the process of moving from explainer to spec.

anssiko:
There's a wider movement to adopt [test as you commit](https://github.com/w3c/testing-how-to/#test-as-you-commit) policy for web specs, championed by @foolip et al.

For WebVR API, or for any spec still undergoing rapid iteration, it may not make sense to enforce this rule, but apply it selectively when it makes sense. But as soon as there is an implementation that conforms to the spec, this policy becomes very helpful. E.g. it ensures spec assertions are actually testable, enables TDD.

In the Device and Sensors WG, we have put the following language to the soon-to-be-renewed charter to give us a bit leeway:

https://w3c.github.io/dap-charter/DeviceAPICharter.html#deliverables

That is, we recommend the policy to specs at Candidate Recommendation and beyond, and allow spec editors and test facilitators to optionally adopt the policy at earlier maturity stages.

To signal whether a particular spec has opted in to the policy, the best practice that is shaping up is to use the standard boilerplate in CONTRIBUTING.md in the spec's repo, e.g.:

https://github.com/w3c/sensors/blob/master/CONTRIBUTING.md#tests

Shameless plug: if you are attending TPAC, there are testing-related session on Tue and Wed:

https://www.w3.org/2017/11/TPAC/schedule.html#tue
https://www.w3.org/wiki/TPAC/2017/SessionIdeas#web-platform-tests

foolip:
It's still up for debate how to best phase in testing, but https://github.com/foolip/testing-in-standards/blob/master/lifecycle.md is the best ideas I've come up with. Basically, as soon as there is an implementation, any changes to things already implemented should be backed by an understanding of that implementation and tests that will begin to fail for it.

At some point I hope it'll be redundant to sprinkle advice about testing everywhere, but for I think I'd have nothing initially, and then add the standard CONTRIBUTING.md wording as soon as there's an implementation.

NellWaliczek:
@anssiko I will indeed be at TPAC! It's my first time going and I'd love to chat about this more.  I know @ddorwin is also interested on this topic too.

toji:
Thanks @cvan! My feelings mirror @NellWaliczek's in this case: Happy to merge this with the understanding that we're still in a highly flexible period of the spec lifetime and it's not necessary or appropriate to require a test for every change until we're at a more stable place. Encouraging tests is never a bad thing, though!

anssiko:
@NellWaliczek, great to hear you'll be there! It seems you're already planning to attend the Testing Project meeting on Tuesday, so you're all set. (Personally I'm chairing another group Mon/Tue and cannot attend the Testing Project meeting, sadly.)

WebVR API is certainly an interesting API to test, and may require some novel approaches, and sharing the early experiences will be helpful.

@toji @cvan, I applaud the group for taking testing seriously this early. It'll mean spec churn will be more expensive, but it will pay off in the end.

PR #300, opened 2017-11-09
requestFrame -> requestAnimationFrame

toji:
This applies the API shape proposed in
https://github.com/w3ctag/design-reviews/issues/141#issuecomment-3395591
67 without defining the underlying common interface.

The proposal has been well received, especially after reviewing it at
TPAC, and we can probably comfortably switch over to using that
verbiage while we make the full proposal to the other appropriate
standards bodies.


explainer.md
```diff
@@ -297,7 +297,7 @@ function endVRSession() {
 function onSessionEnd() {
   vrSession = null;
 
-  // Ending the session stops executing callbacks passed to requestFrame().
+  // Ending the session stops executing callbacks passed to requestAnimationFrame().
```
NellWaliczek: Now that the function names are the same, this is a bit confusing.  Perhaps prefix with VRSession?

explainer.md
```diff
@@ -672,13 +672,13 @@ dictionary VRSessionCreationOptions {
...
-  long requestFrame(VRFrameRequestCallback callback);
-  void cancelFrame(long handle);
+  long requestAnimationFrame(VRFrameRequestCallback callback);
+  void cancelAnimationFrame(long handle);
 
   Promise<void> end();
 };
 
-callback VRFrameRequestCallback = void (VRPresentationFrame frame);
+callback VRFrameRequestCallback = void (DOMHighResTimeStamp time, VRPresentationFrame frame);
```
NellWaliczek: Did you want to add any language about the purpose of this timestamp?

toji:
Thanks for the feedback! Made those changes and a couple of other cleanup's I spotted along the way.

explainer.md
```diff
@@ -206,7 +206,7 @@ Ensuring context compatibility with a `VRDevice` through either method may have
...
+`VRWebGLLayer` objects are not updated automatically. To present new frames, developers must use `VRSession`'s `requestAnimationFrame()` method. When the callback function is run, it is passed both a timestamp (identical to the one given by `window.requestAnimationFrame()`'s callback) and a `VRPresentationFrame` containing fresh rendering data that must be used to draw into the `VRWebGLLayer`'s `framebuffer` during the callback. This framebuffer is created by the UA and behaves similarly to a canvas's default framebuffer. Using `framebufferTexture2D`, `framebufferRenderbuffer`, `getFramebufferAttachmentParameter`, and `getRenderbufferParameter` will all generate an INVALID_OPERATION error. Additionally, attempting to render to this framebuffer outside of the `requestAnimationFrame()` callback will generate an INVALID_OPERATION error.
```
NellWaliczek: > identical to the one given by `window.requestAnimationFrame()`'s callback

Not to be pedantic, but this could mean a lot of things.  I'm a bit worried that people will interpret it as being the same number as the last window.RAF or something.  IIRC we'd chatted about returning 0 for right now so folks wouldn't be enticed to use the data?

toji: Better?

PR #313, opened 2018-01-03
Corrects Multiview Extension specification URL

NellWaliczek:
I just noticed that the WebGL multiview extension spec moved to a slightly different location and the link in the WebXR explainer needs to be updated


toji:
Thanks!

PR #314, opened 2018-01-04
 Policy change: reject `requestDevice` with `NotFoundError` if no devices found

toji:
Simple change, but one that I want other vendors to LGTM before merging: Had a recent conversation where how we handle the case of not finding a device came up, and it made me question how other similar APIs do it. Previously I had advocated for resolving the promise with `null` in that scenario, but I realized there's prior art here. Turns out that WebUSB and WebBluetooth reject their `requestDevice` calls with a `NotFoundError` when there's no device that matches the given criteria. Therefore, for the sake of consistency on the platform, I feel pretty strongly that WebXR should do the same.

Any complaints?

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -63,7 +63,7 @@ The basic steps most WebXR applications will go through are:
 
 The first thing that any XR-enabled page will want to do is request an `XRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the page that the user can click to start XR content.)
 
-`navigator.xr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an `XRDevice` if one is available. If no `XRDevice` is available, it will resolve to null. The promise will be rejected if an error occurs, such as the page not having the appropriate permissions to access VR capabilities.
+`navigator.xr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an `XRDevice` if one is available. If no `XRDevice` is available, it will reject with a `NotFoundError`. The promise will also be rejected if an error occurs, such as the page not having the appropriate permissions to access VR capabilities.
 
 A `XRDevice` represents a physical unit of XR hardware that can present imagery to the user somehow, referred to here as an "XR hardware device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung Gear VR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices.
 
--- a/explainer.md
+++ b/explainer.md
@@ -63,7 +63,7 @@ The basic steps most WebXR applications will go through are:
 
 The first thing that any XR-enabled page will want to do is request an `XRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the page that the user can click to start XR content.)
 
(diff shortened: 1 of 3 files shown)


explainer.md
```diff
@@ -63,19 +63,22 @@ The basic steps most WebXR applications will go through are:
 
 The first thing that any XR-enabled page will want to do is request an `XRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the page that the user can click to start XR content.)
 
-`navigator.xr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an `XRDevice` if one is available. If no `XRDevice` is available, it will resolve to null. The promise will be rejected if an error occurs, such as the page not having the appropriate permissions to access VR capabilities.
+`navigator.xr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an `XRDevice` if one is available. If no `XRDevice` is available, it will reject with a `NotFoundError`. The promise will also be rejected if an error occurs, such as the page not having the appropriate permissions to access VR capabilities.
```
NellWaliczek: In the last sentence, do we need to call out that this error will not necessarily be the same?

cvan: are there any other notable [error names](https://heycam.github.io/webidl/#idl-DOMException-error-names) to call out here?

btw, for reference, the [Generic Sensor API](https://w3c.github.io/sensors/) throws a `SecurityException` if [permissions](https://w3c.github.io/sensors/#feature-detection) are not granted (and you didn't check using `navigator.permissions.query`).

explainer.md
```diff
@@ -63,19 +63,22 @@ The basic steps most WebXR applications will go through are:
...
 function checkForXR() {
   navigator.xr.requestDevice().then(device => {
-    if (device) {
-      onXRAvailable(device);
-    }
+    onXRAvailable(device);
   }, err => {
-    // An error occurred while requesting an XRDevice.
-    console.error('Unable to request an XR device:', err);
+    if (err instanceof NotFoundError) {
```
cvan: `NotFoundError` isn't an exposed `window` global. I'd change to this:

```js
    if (err.name === 'NotFoundError') {
```

explainer.md
```diff
@@ -63,7 +63,7 @@ The basic steps most WebXR applications will go through are:
 
 The first thing that any XR-enabled page will want to do is request an `XRDevice` and, if one is available, advertise VR functionality to the user. (For example, by adding a button to the page that the user can click to start XR content.)
 
-`navigator.xr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an `XRDevice` if one is available. If no `XRDevice` is available, it will reject with a `NotFoundError`. The promise will also be rejected if an error occurs, such as the page not having the appropriate permissions to access VR capabilities.
+`navigator.xr.requestDevice` returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to an `XRDevice` if one is available. If no `XRDevice` is available, it will reject with a "NotFoundError". The promise will also be rejected with an appropriate error if an error occurs during the device query. For example, if the page does not have the appropriate permissions to access XR capabilities it would reject with a "NotAllowedError".
```
cvan: per my earlier comment, actually, it would be a `SecurityError` (see [Web Bluetooth's `requestDevice`](https://webbluetoothcg.github.io/web-bluetooth/#requestDevice-user-gesture) for measure)

toji: So I looked at that, and the case where they throw a `SecurityError` is when the call isn't triggered by user activation. (`requestDevice` in our API does not have that requirement.) Given the [description of the two errors](https://heycam.github.io/webidl/#securityerror):

**"SecurityError":** The operation is insecure.
**"NotAllowedError":** The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.

It seems pretty explicit that a denied permission should use a `NotAllowedError`, while calling a function that requires user activation outside of user activation would produce a `SecurityError`.

cvan: awesome, thanks for digging into this. I learned a few things :)

PR #318, opened 2018-01-11
Add mixin syntax before WebGLRenderingContextBase interface

haoyunfeix:
According to
https://www.khronos.org/registry/webgl/specs/latest/1.0/webgl.idl
WebGLRenderingContextBase is a mixin interface, add mixin in spec.


toji:
Thanks! I had seen that change land in the WebGL spec and meant to pull it over here but wasn't confident in the proper syntax for the partial interface.

PR #323, opened 2018-02-01
advertise VR functionality -> advertise XR functionality

Utopiah:
In the "Acquiring a Device" paragraph of "Lifetime of a VR web app" there is a mention of VR capability whereas it's not an example, thus XR is more generic and should be used here.


toji:
Thanks!

Issue #319, opened 2018-01-18
Simplified V1 Input Proposal


toji:
# Background

A good chunk of this is, primarily, a re-focusing of the previously produced [Input Explainer](https://github.com/immersive-web/webxr/blob/unstable_input/input_explainer.md). As a result nothing here should really be surprising. 

For those who don't already know, we are unable to continue with the design put forward in that document because of concerns about compatibility with upcoming (and as of this writing unreleased) VR/AR standards. I will not dive into a comprehensive evaluation of those incompatibilities, due the said unreleasedness of the standard in question, but will broadly note that it's not known at this time if it will allow the full input device state to be queried in the manner the previous explainer would require.

Given that, and given that we would prefer to have users begin using the WebXR Device API as soon as is reasonable without being blocked on third parties, I propose that we re-focus on exposing a minimal but broadly compatible subset of the previously discussed functionality and have some clear ideas of how it could evolve to fit a variety of underlying input systems in the future.

# Requirements

The "simple" proposal from the previous explainer was one that just allowed developers to listen for basic point-and-click events from an source of VR input, which is enough to enable basic button-based UIs. This is "good enough" for video players, galleries, some simple games, etc. It is insufficent for more complex uses like A-Painter style art apps, complex games, or really anything that involves direct manipulation of objects.

That's regrettable, but a limitation that I feel is worth accepting for the moment in order to enable the significant percentage of more simplistic content that we see on the web today.

So, what we need to enable that level of input is:

  - An event that fires when the input's primary interaction method occurs. Important that this be treated as a user activation event.
  - Frame-to-frame tracking of the input ray

I would also propose that, since this would be all we offer initially, we make this just a teensy bit more useful and future-proof by adding:

  - Notifications of starting and ending the primary interaction method. (ie: Button down/button up)

This would allow a bit more nuance in the interactions allowed by the system, giving the option to drag items around, for example.

# Proposal

I find it easier to talk about these things when looking at an interface, so I'll start with a proposed IDL:

```webidl
enum XRHandedness {
  "",
  "left",
  "right"
};

interface XRInputSource {
  readonly attribute VRHandedness handedness;
};

interface XRInputPose {
  readonly attribute Float32Array? gripMatrix;
  readonly attribute Float32Array? pointerMatrix;
};

//
// Extensions to existing interfaces
//

// Aside: I really think we should consider renaming this to just XRFrame
partial interface XRPresentationFrame {
[...]

PR #325, opened 2018-02-21
Add Basic Input description to the explainer

toji:
Pull request based on the latest conversation from #319 and the WebXR CG call.

I said I would have this out by end of day, which is now technically correct. ;) I expect that I've missed all sorts of dumb typos in my push to get this out before I fall asleep tonight, though. My apologies in advance.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -423,7 +423,7 @@ If an input source can be tracked the `XRInputPose`'s `gripMatrix` will indicate
 
 An input source will also provide its preferred pointing ray, which is defined as a ray originating at `[0, 0, 0]` and extending down the negative Z axis, transformed bt the `pointerMatrix` attribute of an `XRInputPose` object. `pointerMatrix` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `pointerOrigin` attribute:
 
-  * `'head'` indicates the pointer ray will originate at the user's head and follow the direction they are looking. (This is commonly referred to as a "gaze input" device.)
+  * `'head'` indicates the pointer ray will originate at the user's head and follow the direction they are looking. (This is commonly referred to as a "gaze input" device.) There should be at most one `'head'` input source per session.
   * `'hand'` indicates that the pointer ray originates from a handheld device, and represents the device's preferred targeting ray. The exact orientation of the ray relative to the device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance, the pointer ray should most likely point in the same direction as the user's index finger if it was outstretched while holding the controller.
   * `'screen'` indicates that the input source was an interaction with the 2D canvas of a non-exclusive session, such as a mouse click or touch event. See [Magic Window Input](#magic_window_input) for more details.
 
@@ -496,7 +496,7 @@ Three events are fired on the `XRSession` related to these primary actions: `sel
 
 A `selectstart` event indicates that the primary action has been initiated. It will most commonly be associated with pressing a button or trigger.
 
-A `selectend` event indicates that the primary action has ended. It will most commonly be associated with releasing a button or trigger.
+A `selectend` event indicates that the primary action has ended. It will most commonly be associated with releasing a button or trigger. A `selectend` event must also be fired if the input source is disconnected after a primary action has been initiated, or the primary action has otherwise been cancelled. In that case an associated `select` event will not be fired.
 
 A `select` event indicates that a primary action has been completed. `select` events are considered to be [triggered by user activation](https://html.spec.whatwg.org/multipage/interaction.html#triggered-by-user-activation) and as such can be used to begin playing media or other trusted interactions.
 
--- a/explainer.md
+++ b/explainer.md
@@ -397,6 +397,134 @@ function checkMagicWindowSupport() {
 }
 ```
 
+## Input
+
+XR hardware provides a wide variety of input mechanisms, ranging from single state buttons to fully tracked controllers with multiple buttons, joysticks, triggers, or touchpads. While the intent is to eventually support the full range of available hardware, for the initial version of the WebXR Device API the focus is on enabling a more universal "point and click" style system that can be supported in some capacity by any known XR device and in magic window mode.
+
+In this model every input source has a ray that indicates what is being pointed at, called the "Pointer Ray", and reports when the primary action for that device has been triggered, surfaced as a "select" event. When the select event is fired the XR application can use the pointer ray of the input source that generated the event to determine what the user was pointing at and respond accordingly.
+
+### Enumerating input sources
+
+Calling the `getInputSources()` function on an `XRSession` will return a list of all currently available `XRInputSource`s. An `XRInputSource` may represent a tracked controller, inputs built into the headset itself, or more ephemeral input mechanims like tracking of hand gestures. When input sources are added to or removed from the list of available input sources the `inputsourceschange` event will be fired on the `XRSession` object to indicate that any cached copies of the list should be refreshed.
+
+```js
+// Get the current list of input sources.
+let xrInputSources = xrSession.getInputSources();
+
+// Update the list of input sources if it ever changes.
+xrSession.addEventListener('inputsourceschange', (ev) => {
+  xrInputSources = xrSession.getInputSources();
+});
+```
+
+### Input poses
+
+Each input source can queried a `XRInputPose` using the `getInputPose()` function of any `XRPresentationFrame`. Getting the pose requires passing in the `XRInputSource` you want the pose for, as well as the `XRFrameOfReference` the pose values should be given in, just like `getDevicePose()`. Similar to `getDevicePose()` the requested pose may return `null` in cases where tracking has been lost.
+
+If an input source can be tracked the `XRInputPose`'s `gripMatrix` will indicate the device's position and orientation. This will be `null` if the input source isn't trackable. The `gripMatrix` is a transform into a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
+
+An input source will also provide its preferred pointing ray, which is defined as a ray originating at `[0, 0, 0]` and extending down the negative Z axis, transformed bt the `pointerMatrix` attribute of an `XRInputPose` object. `pointerMatrix` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `pointerOrigin` attribute:
+
+  * `'head'` indicates the pointer ray will originate at the user's head and follow the direction they are looking. (This is commonly referred to as a "gaze input" device.)
+  * `'hand'` indicates that the pointer ray originates from a handheld device, and represents the device's preferred targeting ray. The exact orientation of the ray relative to the device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance, the pointer ray should most likely point in the same direction as the user's index finger if it was outstretched while holding the controller.
+  * `'screen'` indicates that the input source was an interaction with the 2D canvas of a non-exclusive session, such as a mouse click or touch event. In this case the pointing ray will originate at the point that was clicked on the canvas, projected onto the near clipping plane (defined by the `depthNear` attribute of the `XRSession`) and extending out into the scene along that projected vector. `'screen'` input sources can only be returned by a `select`, `selectstart`, or `selectend` event.
+
+```js
+// Loop over every input source and get their pose for the current frame.
+for (let inputSource of xrInputSources) {
+  let inputPose = xrFrame.getInputPose(inputSource, xrFrameOfRef);
+
+  // Check to see if the pose is valid
+  if (inputPose) {
+    // Render a visualization of the input source (see next section).
+    RenderInputSource(inputSource, inputPose);
+
+    // Highlight any objects that the pointing ray intersects with.
+    // Presumes the use of a fictionalized rendering library.
+    let hoveredObject = scene.getObjectIntersectingRay(inputPose.pointerMatrix);
+    if (hoveredObject) {
+      renderer.drawHighlight(hoveredObject);
+    }
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


toji:
10 days with no comments? 😢 

I would like to remind everyone that even if this is merged into the explainer there's still room to tweak the described behavior, verbiage, etc. It would be much appreciated if some of the other vendors could at least do a quick pass for anything that seems egregiously bad. Ideally I'd like to be able to have any topics for discussion re: this pull request up prior to next week's call.

toji:
No worries, that's fantastic to hear! I'm not sure how much of the spec you're implementing to test that portion, but if it helps these are the pages I've been checking my implementation against:

https://immersive-web.github.io/webxr-samples/input-tracking.html
https://immersive-web.github.io/webxr-samples/input-selection.html

toji:
Two thoughts that came about while working on a proof-of-concept implementation for this:

1) I had previously said in the pull request that cancelling the `selectstart` event should suppress associated `selectend` events, but I don't feel that's correct any more. Reason being that if you have multiple parts of the code listening to events at once one of them may start, say, a drag operation on `selectstart` and wait for `selectend` to stop dragging, while another listener may cancel the event because they want to suppress a button push due to where the pick ray started. If that suppressed the `selectend` event as well the dragged object would then stay stuck to the controller until the user pressed and released the button again. As a result, consistently firing the `selectend` events seems appropriate even when the `select` itself has been cancelled.

2) I also said in the pull request that input sources with a `pointerOrigin` of `screen` could only come from events, but I think that needs reassessment. If the user places their finger on the magic window canvas and begins dragging it, I think we want to capture that the same way we would clicking and dragging a controller. In order to do that, I think we should push the `screen` input source onto the array returned by `getInputSources()` on touch start/mouse down and remove it on touch end/mouse up. That way the same dragging logic that would work with gaze cursors or controllers also works for screen input.

explainer.md
```diff
@@ -395,13 +395,143 @@ function checkMagicWindowSupport() {
...
+
+// Update the list of input sources if it ever changes.
+xrSession.addEventListener('inputsourceschange', (ev) => {
+  xrInputSources = xrSession.getInputSources();
+});
+```
+
+### Input poses
+
+Each input source can queried a `XRInputPose` using the `getInputPose()` function of any `XRPresentationFrame`. Getting the pose requires passing in the `XRInputSource` you want the pose for, as well as the `XRFrameOfReference` the pose values should be given in, just like `getDevicePose()`. Similar to `getDevicePose()` the requested pose may return `null` in cases where tracking has been lost.
```
thetuvix: getInputPose (and any other APIs that give back poses) should just take an XRCoordinateSystem, to give apps the most flexibility around which coordinate system they reason within.

explainer.md
```diff
@@ -395,13 +395,143 @@ function checkMagicWindowSupport() {
 }
 ```
 
+## Input
+
+XR hardware provides a wide variety of input mechanisms, ranging from single state buttons to fully tracked controllers with multiple buttons, joysticks, triggers, or touchpads. While the intent is to eventually support the full range of available hardware, for the initial version of the WebXR Device API the focus is on enabling a more universal "point and click" style system that can be supported in some capacity by any known XR device and in magic window mode.
+
+In this model every input source has a ray that indicates what is being pointed at, called the "Pointer Ray", and reports when the primary action for that device has been triggered, surfaced as a "select" event. When the select event is fired the XR application can use the pointer ray of the input source that generated the event to determine what the user was pointing at and respond accordingly.
```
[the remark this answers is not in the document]
toji: Good call. Done.

explainer.md
toji: That actually is what it takes in, according to the IDL at the bottom of the doc. I've fixed the reference here.

explainer.md
```diff
@@ -395,13 +395,143 @@ function checkMagicWindowSupport() {
...
+`getInputSources()` will never return input sources with a `pointerOrigin` of `screen`, so they do not need to be accounted for when rendering.
+
+```js
+// Render a visualization of the input source.
+// Presumes the use of a fictionalized rendering library.
+function RenderInputSource(inputSource, inputPose) {
+  // If the input source is a handheld device with a valid gripMatrix, render
+  // a controller mesh using the gripMatrix as a 
+  if (inputSource.pointerOrigin == "hand" && inputPose.gripMatrix) {
+    let controllerMesh = getControllerMesh();
```
thetuvix: Per today's call, XRInputSource could expose some kind of modelId or renderModelId property that could be passed to getControllerMesh here, to enable a request to a CDN.  This could be some kind of common string form of VID/PID for motion controllers, or just the string "hand" for hand sources.  If we are concerned about use of this as a general controller identifier, we could perhaps leave the id blank for controllers with no gripMatrix (and thus no 6DoF controller rendering).

leweaver:
Hi @toji - sorry for the delayed take on this, bad timing with illness on my side!

I've built a prototype to exercise the two samples you linked above with both WMR 6DOF controllers and the HoloLens, and thought I would brain dump my initial comments/feedback. 

One thing that strikes me about the design is that there are two potential ways that a developer will use it - the intended 'point at things' using `pointerMatrix` and `onselect`, but also 'grab and move things around' using `gripMatrix`. While the latter isn't the intended use, it was easy for me to modify your sample to enable dragging the boxes around; so I am sure the web will do this too! (I used the `pointerMatrix` to identify which box would be dragged, then updated its position based on changes to the `gripMatrix` - while crude, this worked pretty well.)

My General Comments: 
- The simple select use case works really well on HoloLens and VR headsets, feels clean.
- From the text, it wasn't clear to me whether `head` input sources should also fire select events when `hand` sources do. Should the UA suppress `head` select events if it emits `hand`, or fire both? Should we specify that `hand` inputsource events are always fired prior to `head` and `screen`, allowing a page to call `preventDefault` if they want to suppress the `head` event?
- Should we specify whether it there is at MOST 1 `head`/`screen` input source; or can there be multiple of either?
- There is no `onselectcancel` event (emitted if the source is disconnected etc.). While I agree with the comments in the other issue that gaze & commit style actions would benefit from this, in my experimentation with dragging on HoloLens it actually felt quite natural for the `end` method to be called when tracking was lost.
- We should absolutely discourage rendering of models based on the presence of the `gripMatrix` alone - there should be a separate `renderable` flag or model string. This was basically the one thing that stopped the HoloLens hands working nicely in my draggable-box scenario.
  - I experimented with a hack to allow the UA to hide controller models when using tracked hands by setting the gripMatrix scale to ~zero... nasty and non-performant, but an effective fallback!

toji:
Thanks Lewis for taking the time to do a prototype implementation! This kind of feedback is invaluable. (Also, no worries about the delay. Glad you're feeling better!)

> One thing that strikes me about the design is that there are two potential ways that a developer will use it...

Just to make sure I'm on the same page, it doesn't sound like you're concerned about the grip use case, right? I think it's a nice side effect that it's an option, even though developers who use that design pattern without a fallback will exclude their app from devices without 6DoF input.

> From the text, it wasn't clear to me whether head input sources should also fire select events when hand sources do. ...

I guess I wasn't expecting to expose a `head` input source when `hand` input sources were available. Or we'd switch back and forth as needed. For example, in Chrome's Oculus implementation currently I only expose the "active" input devices. If you have both the untracked remote and a pair of Touch controllers, the active set of device is whichever was used last. In practice this switches really cleanly from a gaze cursor to 6DoF pointer and back depending on which device you pick up and start using.

I can see why some platforms would want to expose both, though. The pattern you mentioned sounds reasonable in that case. I've got to say, though, that after a lot of careful reading of other web specs I'm not sure `preventDefault()` is supposed to work that way, despite my thinking it did only a week or two back. It prevents secondary native actions (like a `click` event checking a checkbox) but I don't see cases where it suppresses follow-up events? (That's a bit more like `stopPropagation()`, but that's intended to prevent further propagation of the current event, not related later events.)

In any case, my primary concern is simply that we don't end up in a scenario where apps are built in one environment that doesn't expose `hand` and `head` inputs simultaneously, and begin misbehaving when they start getting two events. It is pretty trivial to synthesize a `head` input ray from a `hand` input, so maybe it's sufficient to suggest that developers do that if they want to force a gaze cursor universally?

> Should we specify whether it there is at MOST 1 head/screen input source; or can there be multiple of either?

I can't think of any reason we might have more than one `head` input source. `screen` may be a different story? Text needs to be clarified, but it could make sense to expose one `screen` input per touch source? Also, if you're using a browser in VR and treating a 6DoF controller as a type of stylus technically you could have multiple of them at once. We may want to make an executive decision to keep it simpler for now and only expose one primary `screen` input at a time?

> in my experimentation with dragging on HoloLens it actually felt quite natural for the end method to be called when tracking was lost.

Agreed that this feels like a good way to address the "canceled" case without adding much complexity.   If the associated "select" event isn't fired along with it that should be a sufficient indication that the input was interrupted somehow.

> We should absolutely discourage rendering of models based on the presence of the gripMatrix alone 

This is an excellent point, thanks for bringing it up. I think based on comments from others (especially @Artyom17) the ideal way to handle this would be an identifier for a render model, with null meaning "don't draw me". That's has the potential to be a tricky thing to spec, so I'm not opposed to deferring it in favor of a `renderable` boolean if needed, but we should probably at least investigate the other route first.

The zero scale trick is clever, I'll give you credit for that, but that seems like a pretty ugly spec hack when we can simply add the necessary properties here and now. I'm filing a separate issue to talk through this specific feature, because it's likely to require a bit of work to get to a comfortable spot.

I've rebased the PR and added a couple of clarifications to account for your observations around having at most one `head` source and utilizing `selectend` as a cancel signal. As of the last CG call the other vendors felt broadly comfortable merging this PR as a baseline and iterating on individual points of concern separately from there. We were waiting for your feedback before proceeding. Do you feel comfortable with that now?

toji:
Separate issue for render models now up here: https://github.com/immersive-web/webxr/issues/336

leweaver:
TL;DR - yes I'm comfortable with merging this in now and iterating, makes sense to move the model/renderable discussion to its own issue. Thanks for waiting for me!

> Just to make sure I'm on the same page, it doesn't sound like you're concerned about the grip use case, right?

Right - I'm with you and think this is a nice option for developers to have.

> I guess I wasn't expecting to expose a head input source when hand input sources were available...

That works for me - I'm in favor of this as it reduces the amount of work logic a page needs! You raise a good point that all UA's will need to be consistent here.

> We may want to make an executive decision to keep it simpler for now and only expose one primary screen input at a time?

Having multiple `screen` sources to me sounds like the most correct solution for multi-touch/6DOF stylus. My opinion on whether to include this now/later boils down to: do we make it harder for ourselves to add multi-touch in the future if we set the expectation now that there will only be a single `screen` source? I'm worried that we might end up with pages that then make this assumption and break if we change our mind. Perhaps best to explicitly allow multiple, giving UA's the option to return multiple, or just a primary. Perhaps this is something we need to noodle over in a separate issue.

explainer.md
thetuvix: Per the discussion below, UAs could return a renderModelId of null on headsets where you can see your hands/controller and a renderModelId of "hand" or a VID/PID when you can't see your hands/controller.  That would be a strong hint to the app as to whether it should render the controller.

toji:
Excellent! Thanks! Merging now.

> do we make it harder for ourselves to add multi-touch in the future if we set the expectation now that there will only be a single screen source?

We may, and that's a bit of a concern. (Specifically, I can imagine users testing drag and drop features and thinking it works just fine, but suddenly a second finger is introduced in `<browser_version_current+5>` and now your dragged object starts jittering between two fingers.

With pointer events there's a concept of the "active" pointer, which sounds like it implies that the one-touch scenario is distinct from the multi-touch scenario, but that's largely because pointer events also produced emulated mouse events, of which there can be only one (cursor.) That's not really a problem in our environment, as XR apps are likely to be developed with the assumption that at least two persistent pointers is a common case.

thetuvix:
> I guess I wasn't expecting to expose a head input source when hand input sources were available. Or we'd switch back and forth as needed. For example, in Chrome's Oculus implementation currently I only expose the "active" input devices. If you have both the untracked remote and a pair of Touch controllers, the active set of device is whichever was used last. In practice this switches really cleanly from a gaze cursor to 6DoF pointer and back depending on which device you pick up and start using.

I recall a design principle from our discussions around the original input explainer that we believed that UAs should just expose all known input sources (known sources of committing an action), letting middleware be in the business of deciding which input form is primary at any given moment.  This was what led us to move away from `primaryInputSource` as an API concept in the previous explainer.

For example, looking at the [Choosing a cursor mode](https://github.com/immersive-web/webxr/blob/unstable_input/input_explainer.md#choosing-a-cursor-mode) section from the original explainer, the app itself tracks its own current `cursorMode`.  The sample code happens to choose a policy of tracking the preferred mode of the most recent inputSource observed, but the app could customize that policy as needed.

Some examples of why an app would choose a notion of "cursor mode" that differs from the UA's default:
* A given app may be fundamentally designed to be "gaze-and-commit" rather than "point-and-commit".  Even if the app finds itself with pointing-capable controllers, the app may prefer to always target with the head ray, which it could do by ignoring a select's `inputSource `and just polling for the head's "center eye" pose whenever a select happens.
* An app may support both "gaze-and-commit" and "point-and-commit" based on the input source's capabilities, and typically just rely on the input source's `pointerMatrix`.  However, the app (or its middleware) may implement, for example, some kind of modal grab/release interaction where the first select picks up an object and the second select puts it down.  The app/middleware may consider the input source that started the grab interaction to be "captured" through the end of the grab, continuing to target with that input source until the user selects again, even if the user happens to bump a controller that causes the UA to swap its notion of "active" source.

Looking back at the previous explainer's [cursor mode sample code](https://github.com/immersive-web/webxr/blob/unstable_input/input_explainer.md#choosing-a-cursor-mode), our suggested pattern for an app that decides it's in "gaze" mode was to just use `getDevicePose` to fetch the "head" pose.  As long as apps know they can rely on `getDevicePose` to deliver a "center eye" targeting ray, UAs shouldn't have any reason to expose a dummy "head" input source.

Therefore, even in both of the cases above, I'd always expect a given select to only ever fire one `select` event, with its `inputSource` accurately representing the source on which the user committed the select, be it a hand, left mouse click, gamepad button, headset button, 0DoF controller, 3DoF controller, 6DoF controller or magic window tap.  An app that wants to target in some way other than the commit source's preferred pointing ray is free to look up any other ray itself.

The only blocker there could be if the app had "captured" an input source that a UA like Chrome no longer considers "active" and thus stopped enumerating, even though it's still tracked by the underlying XR system.  We could say that if an app holds onto an `XRInputSource`, it can continue polling for its input poses if available, even if the UA decides through some policy to stop listing it in `getInputSources()`.  That could give apps the best of both worlds.

It might be good for the new input explainer to explore sample code for the two cases above, just to ensure that apps wanting to use a pointing ray other than the preferred ray for their select's commit source are not blocked.

PR #334, opened 2018-03-21
Add informative references to Generic Sensor-based Orientation Sensor

### LINKED ISSUE #332: clean up mentions and discussions of devicemotion
There are a number of discussions of `devicemotion` in the documents.  The API was never standardized, and multiple organizations are pushing to change or remove it.  So, for example, we should not assume a developer can use device motion before or instead of creating a session, and may want to be explicit in ensuring that WebXR handles all the use cases it needs to without assuming `devicemotion` exists.

anssiko:
Fix #332 

PTAL @toji @blairmacintyre @cvan

(Also fixed the bikeshed build with https://github.com/immersive-web/webxr/commit/966eaec48b49ee210240291288e81ca515e78537.)


toji:
This is great, thank you!

Issue #337, opened 2018-03-28
anchors as a core concept


blairmacintyre:
While discussing the https://github.com/immersive-web/anchors proposal, the question of how fundamental we want anchors to be came up.

We've touched on this a number of times, but we should decide if we want to introduce anchors into VR as well as AR, to present an API that encourages a style of programming where all content is attached to anchors (i.e., make it so no content is just specified in a global coordinate system).

As we move from 3DOF and fixed-space VR gear, to inside-out devices that track their space, this will become necessary, so we should consider doing it now.

judax:
With the current proposal for anchors, VR could also use them although in most systems any created anchors won't ever be updated, most likely.

The main benefit is consistency between building VR and AR experiences. In AR experiences the use of anchors is highly recommended. If developers would understand anchors and start using them on any WebXR app they build, a more consistent ecosystem could be achieved. Also, as Blair indicates, a basic anchor could be used to specify the world origin, also justifying that anchors could have meaning in VR too, even in non-inside out systems.

toji:
Following up from the call today, it appears that some of the biggest hangups around using anchors everywhere is simply that the term "Anchor" has gained connotations that don't neatly fit into other uses that we have within WebXR. (Mapping the concept of an Anchor to the "headModel" frame of reference, for example, is awkward.) The proposed solution coming out of the call was to find a different, more neutral term to describe a single XR object type used to track spatial relationships and have the designation of "Anchor" vs "Frame of Reference", etc come from the way they are created. e.g. `requestFrameOfReference`, `requestAnchorAtPoint`, etc.)

At the moment we already have a base `XRCoordinateSystem` type that may serve that purpose well (it was intended to act as the base for both frames of reference and anchors). We could examine other possible names (`XRSpace`?) if needed. If we start using `XRCoordinateSystem` directly we would need to move some of the stage-specific stuff off of `XRFrameOfReference`, but that's likely a good idea anyway. (It would be nice to have the session describe stage bounds and be able to query them in relation to any coordinate system.)

toji:
Started a speculative pull request (#340) based on the above to facilitate more concrete conversation about that proposal.

speigg:
@toji I think not using anchors as the base class for generic spatial relationships is the right choice (doesn’t seem generic enough), although it seems that this is a different question than whether or not anchors (and/or trackables) should be a core concept in WebXR. The biggest challenge I see here is in arriving at a consensus about what anchors/trackables actually mean. 

[...]

PR #340, opened 2018-04-03
Convert XRFrameOfReference to XRCoordinateSystem

toji:
Pull request is part of the conversation in #337. I'm not currently proposing this be merged, but I felt like having the necessary changes available in an form that was easy to comment on would be helpful in driving forward the conversation.

This is a speculative change based on today’s WebXR call. It drops the
concept of a separate XRFrameOfReference object in favor of using
XRCoordinateSystem directly. This necessitated moving where bounds were
queried from, but I kind of like this better?

Naming feels a little awkward at this point, but it’s a start.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -147,13 +147,13 @@ function beginXRSession() {
 }
 ```
 Once the session has started, some setup must be done to prepare for rendering.
-- A `XRFrameOfReference` must be created to define the coordinate system in which the `XRDevicePose` objects will be defined. See the Advanced Functionality section for more details about frames of reference.
+- An `XRCoordinateSystem` must be created to define the coordinate system in which the `XRDevicePose` objects will be defined. See the Advanced Functionality section for more details about frames of reference.
 - The depth range of the session should be set to something appropriate for the application. This range will be used in the construction of the projection matrices provided by `XRPresentationFrame`.
-- A `XRLayer` must be created and assigned to the `XRSession`'s `baseLayer` attribute. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -543,54 +543,55 @@ Beyond the core APIs described above, the WebXR Device API also exposes several
...
+When using a `stage` frame of reference the device will frequently have a configured "safe area" that the user can move around in without fear of bumping into real world objects. WebXR can communicate the rough boundaries of this space via the `XRSession`'s `getStageBoundsGeometry()` function. It provides a polygonal boundary given as an array of `XRStageBoundsPoint`s which represents a loop of points at the edges of the safe space. The points are given in a clockwise order as viewed from above, looking towards the negative end of the Y axis. The shape it describes is not guaranteed to be convex. The values reported are relative to the `XRCoordinateSystem` passed into the call, and do not necessarily contain the origin of that coordinate system. The bounds are unavailable for the device or given `XRCoordinateSystem`, an empty array will be returned.
```
thetuvix: It seems fine to squash XRFrameOfReference itself into XRCoordinateSystem, as XRFrameOfReference doesn't do much over XRCoordinateSystem as currently defined, except to add some stage-specific attributes. (and stages are only one kind of frame)

However, it feels like a design smell that we already find ourselves stashing the stage-specific bounds data off on XRSession.  Especially as we start digging into anchors and other AR world understanding APIs, we are likely to have an increasing number of coordinate system flavors beyond stages that have extra metadata associated with them.

For example:
* Stages will have bounds
* Plane anchors will have extents, which will have change events
* Face anchors will have facial feature points that change each frame
* etc.

While stages will usually have just 1 instance, many of those other coordinate system flavors will have N instances, which will make it more awkward for us to stash such data elsewhere away from the coordinate system.  Rather than finding a one-off solution for stage attributes, we should audition an API pattern here that works for stages, anchors and beyond.

I dig in on ideas there a bit in my comment on [anchors issue 4](https://github.com/immersive-web/anchors/issues/4).

toji:
After discussion on the WebXR call a few weeks ago it was determined that we primarily needed better docs and samples, not this specific change. Closing in light of that.

Issue #339, opened 2018-03-30
Representation of a rays in WebXR


toji:
This came up on the AR-centric call this week: Conversations around the [Hit Test explainer](https://github.com/immersive-web/hit-test/blob/master/explainer.md), which deals with rays quite a bit, have proposed representing them as two vectors: an origin and a direction. This is an easy to understand and use format, and is especially helpful if developers have to specify a ray manually.

The [WebXR explainer](https://github.com/immersive-web/webxr/blob/master/explainer.md#input), however, currently represents rays as a matrix, which indicates the transform to be applied to a ray with an origin at [0, 0, 0] and pointing down the -Z axis. The desire was to make it easy to render with, while still not making it too hard to get the origin/direction form through a couple of vector transforms.

Both approaches have pros and cons, but we should definitely pick one and use it consistently throughout the spec. In my opinion the following should be considered:

### Origin/Direction form
**Pros:**
 - Human readable
 - Easy to build by hand
 - Less values to communicate

**Cons:**
 - When matrix format is needed it's[ harder to generate](http://glmatrix.net/docs/mat4.js.html#line1442).
 - Need to deal with non-normalized directional vectors. (Pretty minor)

### Matrix form
**Pros:**
 - Easy to use directly for rendering
 - Producing Origin/Direction from a matrix is relatively easy
 - Using it would mean WebXR only has to deal with one math primitive type.

**Cons:**
 - Not human readable
 - Difficult to construct by hand
 - Overly expressive (rotation around z axis doesn't matter, scaling and shearing are unnecessary and detrimental)

My gut feeling, taking all that into account, is that using the origin/direction form is likely better for API usability long-term. It's unfortunate that it's more difficult to use for rendering, but if we ever expect users to provide rays back into the system (as the hit-test proposal does) then it seems overly complicated and error prone to ask them to compose and pass full matrices. (It's way easier to normalized the directional input implicitly than it is to try and handle a badly formed matrix.)

If we do go with origin/direction, I would suggest that we use an interface to represent it within the API. Maybe something like this?

```webidl
interface XRRay {
  Float32Array origin;
  Float32Array direction;
}
```
or
```webidl
interface XRRay {
  double originX;
  double originY;
  double originZ;
  double directionX;
  double directionY;
  double directionZ;
}
```

Both of which have their own issues in my mind.

toji:
[...]

Issue #336, opened 2018-03-25
Indicating how to render an input source model


toji:
The topic of how/if we indicate to developers what to render for handheld input sources has come up repeatedly. (For example, see comments from @Artyom17 [here](https://github.com/immersive-web/webxr/issues/319#issuecomment-365113337) and @leweaver [here](https://github.com/immersive-web/webxr/pull/325#issuecomment-375836527)

It was previously my feeling that we should omit any information about controller models for V1 of the API and try to address it more robustly in a followup. But Lewis' prototype revealed and interesting quirk in that AR headsets like HoloLens may need to expose as `'hand'` input source that has a valid `gripMatrix`, and yet not want a generic controller model rendered over the top of the user's actual hand. This makes a lot of sense, and it's probably best to address that scenario early.

We could easily handle that specific case with a `renderable` boolean on the `XRInputSource`. However since we know that the request to give some sort of indication of the controller type is a popular one it may make more sense to address both cases with a single mechanism.

It would be nice to expose a renderable mesh directly to the page, but there's significant logistical issues to doing so. Where do the meshes come from? What format is it delivered in? How should it be animated? Etc. It seems more realistic to instead expose some form of well-formatted ID to indicate what type of device, if any, should be rendered.

In WebVR the Gamepad `id` served this purpose, but there were several issues that made it difficult to use:
- The name needed to serve double duty as a human readable string and mesh ID.
- Different browsers exposed different strings for the same controllers.
- Hands would have had to have been given a name as well, making them more error prone to filter from *not* rendering.

If we're going to expose something similar we can and should do better.

For the sake of discussion I'm going to assume that we'd expose a 'renderModelName' string attribute on `XRInputSource`. (Name subject to bikeshedding) I'd propose that it have the following properties:
- If the input source should not be rendered (hands in AR) it should be empty string.
- If the input source should be rendered, but for whatever reason the device type isn't known a predetermined string ('unknown device'?) should be used.
- Ideally we want to be able to use strings surfaced by the underlying APIs when available, but we'd want to do so in a way that at least attempts to normalize it. Things like specifying that the names should always be lower case, and that they should prefer a format of 'vendor, model' would be a start.
- The string shouldn't contain information that would be duplicated in the `XRInputSource` or `XRInputPose`. Specifically, something like handedness should be omitted. (So 'oculus, touch' rather than 'oculus, left touch' even though the left and right controllers are different devices with different meshes.)
[...]

Issue #330, opened 2018-03-13
Revisit session creation options and flow  


blairmacintyre:
XR session creation is structured as it was in WebVR, which had a much more limited set of use cases and possible display structures.

I would suggest we need to do a few things to make session creation more flexible and usable.  Here are a list of possibilities, to spur discussion.

First, the combination of `supportsSession` and `requestSession` seem reasonable, but we need a way to handle the case where an external action may activate a session.  Is there an implication that the UA can fire off an XRSession event when there is not currently a session?  For example, a UA may have it's own UI for activating VR or AR, which might include allowing the user to specify the sort of session.  Or, if a user follows a link while in AR of VR, the next page should auto-create a session corresponding to the previous page's session.

Second, we should update the XRSessionCreationOptions to reflect the diversity of sessions we might get, and have a corresponding set of attributes on the XRSession.

1. Get rid of “exclusive” … it always seemed like something of a hack to deal with the fact that we may want different sorts of connections to devices.  
2. There 3 sorts of sessions type's we've been talking about displays supporting: "immersive" (i.e., 3D probably head-worn displays), "portal" (i.e., the magic window type, where a non-headworn display is tracked and presents either AR or VR on a display), and "flat"  (e.g., is just 3D graphics in a DIV in page on the display, with no WebXR specific exclusive rendering, but might use some form of sensing).   Additional ones may be added in the future, depending on platform.  E.g., “projective” may support projection AR (like Microsoft's RoomAlive) if that was something someone wanted to support.  Not sure what to call this:  "xrType"?
3. To support some of the current use of "exclusive", add a capability to distinguish between VR and AR. Not sure what to call it, perhaps “reality”?  With possible values “virtual, augmented”?
4. Add “worldAligned” as a capability to request that the local tracking frame of reference is aligned with geospatial components.  
5. A capability to differentiate between 3DOF and 6DOF.  Again, not sure what to call it, perhaps "spatialTracking"?

One of the assumptions I'm making here is that a given implementation would support one or more of these combinations, as they see fit and are able.  A 6DOF device could support 3DOF _if_ it wanted;  it's not required to.  A device that doesn't have a magnetometer may not support "worldAligned".  

A phone like a DayDream and ARCore capable Android phone could end up supporting a range of realities that provide some combinations of
- reality = virtual AND augmented
- spatialTracking = 3DOF AND 6DOF
-  xrType = immersive AND portal AND flat
[...]

PR #342, opened 2018-04-16
Renamed pointerOrigin and enum values to reflect usage pattern, not device type.

leweaver:
The main purpose of this PR is to clarify what input sources are returned from a call to XRSession.getInputSources(). These changes make it easier to expose tracked hands on AR devices are exposed (eg, the HoloLens). 

I've renamed the `pointerOrigin` enum and members to `targetRayMode: { 'gazing', 'pointing', 'tapping' }` to better reflect the input intent, rather than physical representation. This was primarily to reduce confusion: since a tracked hand is actually uses the head gaze as a ray, so should be a `head` not a `hand` in the old nomenclature. 

The implication of this is that while it is up to the user-agent to determine if it is appropriate to merge multiple 0DOF input sources (clicker, gamepad etc.) into a single `gazing` input source, or expose them as multiple (eg, tracked hands that only have a position but not orientation - thus are still considered `gazing`), 
I've removed the responsibility of the user agent for actually choosing which XRInputSource's are 'active'. All available input sources are returned.

~I've also adopted the as-yet unfinalized raycast origin/target, which is under discussion in #339 and immersive-web/hit-test#14. Will need to likely update this when those discussions are resolved.~

Lastly - the controller rendering samples here still depend on the outcome of open issues: #336 and #330

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -399,7 +399,7 @@ function checkMagicWindowSupport() {
 
 XR hardware provides a wide variety of input mechanisms, ranging from single state buttons to fully tracked controllers with multiple buttons, joysticks, triggers, or touchpads. While the intent is to eventually support the full range of available hardware, for the initial version of the WebXR Device API the focus is on enabling a more universal "point and click" style system that can be supported in some capacity by any known XR device and in magic window mode.
 
-In this model every input source has a ray that indicates what is being pointed at, called the "Target Ray", and reports when the primary action for that device has been triggered, surfaced as a "select" event. When the select event is fired the XR application can use the target ray of the input source that generated the event to determine what the user was attempting to interact with and respond accordingly. Additionally, if the input source represents a tracked device a "Grip" matrix will also be provided to indicate where a mesh should be rendered to align with the physical device.~~~
+In this model every input source has a ray that indicates what is being pointed at, called the "Target Ray", and reports when the primary action for that device has been triggered, surfaced as a "select" event. When the select event is fired the XR application can use the target ray of the input source that generated the event to determine what the user was attempting to interact with and respond accordingly. Additionally, if the input source represents a tracked device a "Grip" matrix will also be provided to indicate where a mesh should be rendered to align with the physical device.
 
 ### Enumerating input sources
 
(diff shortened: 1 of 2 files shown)


leweaver:
Since the raycast discussion is as yet unresolved, I'm changing `targetRay` back to a matrix since the primary point of this change is not semantics of raycasting, but the naming of `pointerOrigin`.

explainer.md
```diff
@@ -415,17 +415,21 @@ xrSession.addEventListener('inputsourceschange', (ev) => {
 });
 ```
 
+The properties of an XRInputSource object are immutable. If a device can be manipulated in such a way that these properties can change, the `XRInputSource` will be removed and recreated.
```
toji: I had to think about it for a while to decide if I'm OK with this, but I think I am. The most problematic scenario that I can think of is if you have a system that is not explicitly handed (like the Vive wands) and the system decides that you've switched hands in the middle of a grab-and-drag gesture it would be common for the page implementation to cause the object to be dropped as the "hand" holding it disappears.

The reality is, though, that I rarely see this happen outside of early system initialization and if it's a large concern then the system can silently re-map the physical devices to the previously created left/right hand input sources to hide the transition. Plus there's a significant number of devices for which the handedness is an explicit property of the input device or only changed manually, and therefore will never be surprising to the user.

So TL;DR: I'm good with it!

explainer.md
```diff
@@ -415,17 +415,21 @@ xrSession.addEventListener('inputsourceschange', (ev) => {
...
+Each input source can query a `XRInputPose` using the `getInputPose()` function of any `XRPresentationFrame`. Getting the pose requires passing in the `XRInputSource` you want the pose for, as well as the `XRCoordinateSystem` the pose values should be given in, just like `getDevicePose()`. `getInputPose()` may return `null` in cases where tracking has been lost (similar to `getDevicePose()`), or the given `XRInputSource` instance is no longer connected or available.
+
+If an input source can be tracked the `XRInputPose`'s `gripMatrix` will indicate the device's position and orientation. If only position or orientation is trackable (not both), the `gripMatrix` will return a transform matrix applying only the trackable pose. An example of this case is for physical hands on some AR devices, that only have a tracked position. The `gripMatrix` will be `null` if the input source isn't trackable.
```
toji: "If only position or orientation is trackable (not both), the `gripMatrix` will return a transform matrix applying only the trackable pose"

The wording on this is a little weird to me. I assume you mean that, for example, if the input source can only track position the transform matrix will not contain any rotation component? Not sure how to state that more clearly though.

Also, does this imply that the transform matrix should not apply an arm model for 3DoF controllers? Because if so I'm pretty strongly opposed to that.

explainer.md
```diff
@@ -415,17 +415,21 @@ xrSession.addEventListener('inputsourceschange', (ev) => {
...
 
-An input source will also provide its preferred pointing ray, which is defined as a ray originating at `[0, 0, 0]` and extending down the negative Z axis, transformed bt the `pointerMatrix` attribute of an `XRInputPose` object. `pointerMatrix` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `pointerOrigin` attribute:
+An input source will also provide its preferred target ray on its pose, which is defined as a ray originating at `[0, 0, 0]` and extending down the negative Z axis, transformed by the `targetRayMatrix` attribute of an `XRInputPose` object. `pointerMatrix` will never be `null`. Their values will differ based on the type of input source that produces it, which is represented by the `targetRayMode` attribute:
 
-  * `'head'` indicates the pointer ray will originate at the user's head and follow the direction they are looking. (This is commonly referred to as a "gaze input" device.) There should be at most one `'head'` input source per session.
```
toji: We lost the text indicating that there should be at most one `head` input source, which I think was originally requested by Microsoft? Is that intentional?

explainer.md
```diff
@@ -415,17 +415,21 @@ xrSession.addEventListener('inputsourceschange', (ev) => {
...
+  * `'gazing'` indicates the target ray will originate at the user's head and follow the direction they are looking (this is commonly referred to as a "gaze input" device). While it may be possible for these devices to be tracked (and have a grip matrix), the head gaze is used for targeting. Example devices: 0DOF clicker, regular gamepad, voice command, tracked hands.
+  * `'pointing'` indicates that the target ray originates from a handheld device and represents that the user is using that device for pointing. The exact orientation of the ray relative to the device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance, the target ray should most likely point in the same direction as the user's index finger if it was outstretched while holding the controller.
+  * `'tapping'` indicates that the input source was an interaction with the 2D canvas of a non-exclusive session, such as a mouse click or touch event. See [Magic Window Input](#magic_window_input) for more details.
```
toji: Not sure if the term `tapping` feels accurate here. People typically associate it with touchscreens, and we want it to cover mouse and stylus input as well. Admittedly I don't have a better term in mind.

explainer.md
ransico: Your interpretation is correct, but I'll have a think about wording to make this a bit clearer. However, it was not the intention to disallow arm modeling for 3DoF controllers; would changing "will return" -> "may return" be sufficient?

explainer.md
```diff
@@ -516,22 +527,147 @@ function onSessionStarted(session) {
...
+    // Draw a cursor for gazing and pointing devices only.
+    let cursorPosition = scene.getIntersectionPoint(inputPose.targetRayMatrix);
+    if (cursorPosition) {
+      renderer.drawCursor(cursorPosition);
+    }
+  }
+}
+```
+
+### Complex input
```
toji: "Complex input" is maybe not the phrase I would use here. Maybe "Direct manipulation" or "Grabbing and dragging"?

toji:
I've left a few comments, but in general I feel like this is a good clarification of the input system. Thanks for putting the time into it!

explainer.md
toji: Will return -> may return is good enough for me, especially since this change does preserve the explicit mention of arm models later on.

explainer.md
leweaver: This was intentional since we are now treating AR hands as `gazing` sources that have a `gripMatrix` (whereas in the initial prototypes, we used `hand` pointerOrigin). Since there are multiple hands,  there needs to be one distinct inputSource in the array for each.

explainer.md
leweaver: This is a tough one to name; some other ideas:

- `projecting` - as in, the pointing ray is projected into the scene from the camera as defined by a point on the near plane
- `touching` - mouse, touch, stylus, etc could maybe considered forms of touch?
- `pressing` - same reasoning as touching

toji: `projecting` is probably the most accurate, but it doesn't seem intuitive. Maybe 'screen-projection' or something like that?

While it's not exactly more accurate than `tapping`, but the term `touching` feels more comfortable for whatever reason. At the very least tapping sounds like an instantaneous event while touching implicitly includes drag or long press operations.

leweaver: `screen-projection` seems to be the clearest, but `touching` feels more in line with the other enum values and is my preference by a narrow margin. Thoughts from other vendors welcome!

leweaver:
The code in the samples directory will also need to be modified to handle the new values. Considering that the Ray updates are potentially on the way, does it make sense to update the samples in one go as a separate PR?

toji:
Yes, I think the samples can be updated in a more bundled fashion. We'll also want to add some simple, temporary shims to ensure that the new names map correctly to existing implementations (like Chrome) which will take a couple of browser releases to cycle to the new verbiage.

explainer.md
thetuvix: It is precisely since this is rare that we believe it is safer to remove and re-add the sources - since most apps will experience effectively unchanging sources, they'll likely assume they don't change any would probably miss such changes.  Removing and re-adding the source gets us back onto the mainline path such apps will have tested.

Issue #349, opened 2018-04-25
framebufferScaleFactor seems difficult to use correctly/consistently


klausw:
In addition to being inconsistently documented (see issue #348), I think this setting also seems difficult for application developers to use correctly.

Currently, the application can choose to supply the value 0.0 to use the default scale factor, or it can supply a specific value to replace the default scale factor. However, as far as I can tell there's no way to tell what the default scale factor would have been.

Let's say the goal is for the browser to tune the default so that a moderately-complex application can hit the target framerate on a given device. The specific framebuffer scale needed for this will depend on the device characteristics and potentially also OS and browser version.

As an example, current Daydream headsets use a default framebufferScaleFactor of 0.5 on Android N, and 0.7 on Android O due to a more efficient render path there. Going forward, this may well be subdivided further, for example a Pixel 2 could use a default of 0.8 due to its fast GPU and comparatively low screen resolution, while a first-generation Pixel XL would use 0.6 due to a slower GPU and higher screen resolution.

By contrast, on a Windows headset, there's likely to be a system-provided recommended scale, where using a framebufferScaleFactor of 1.0 would mean to use this unchanged. An application with simple graphics may want to use a higher scale factor, but this doesn't seem to be supported by the current spec which assumes it's a value from 0 to 1.

How is a developer supposed to choose an appropriate setting for their application? In the current system, if the developer sets framebufferScaleFactor=0.6 on an Android device, this would be higher-than-default on some devices and lower-than-default on others, and that's unhelpful if the defaults were tuned to get approximately equivalent performance for typical content.

To work around this, the developer would either need to infer the default from experimentally created sessions, or maintain a database of device/OS/browser version to use for tuning. Both of these seem unpleasant.

I think it would be far more helpful and intuitive to supply a relative scale factor that defaults to 1.0 and that applies on top of the system-selected default framebuffer scale. That way, applications with simple graphics compared to a baseline "moderately-complex" one could use values greater than 1.0, or a smaller value for complex rendering.

A relative scale would also work well with lower-level tuning such as SteamVR's [automatic resolution adjustment](https://steamcommunity.com/games/250820/announcements/detail/1661138371528106606), where the default 1.0 value would be pre-tuned to match the current system's GPU performance.

The low-level framebufferScaleFactor could potentially still be exposed to applications if it's considered helpful for specific use cases, but I think a relative scale would work better.

sjpt:
[...]

PR #353, opened 2018-05-15
Change how framebufferScaleFactor is interpreted.

toji:
Addresses #349.

Changes `framebufferScaleFactor` so that a value of 1.0 corresponds to the "default" framebuffer size and exposes a new static `XRWebGLLayer.getNativeFramebufferScaleFactor(xrSession)` function that reports the scale that should be used to get a 1:1 pixel ratio.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -722,7 +722,7 @@ function setupWebGLLayer() {
   });
 ```
 
-In some cases the developer may want to ensure that their application is rendering at the 'native' size for the device. To do this the developer can query the scale factor that should be passed during layer creation with the `XRWebGLLayer.getNativeFramebufferScaleFactor()` function.
+In some cases the developer may want to ensure that their application is rendering at the 'native' size for the device. To do this the developer can query the scale factor that should be passed during layer creation with the `XRWebGLLayer.getNativeFramebufferScaleFactor()` function. (Note that in some cases the native scale may actually be less than the recommended scale of `1.0` if the system is configured to render "superscaled" by default.)
 
 ```js
 function setupNativeScaleWebGLLayer() {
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -706,19 +706,33 @@ function drawMultiviewScene(views, pose) {
...
 
 ```js
 function setupWebGLLayer() {
   return gl.setCompatibleXRDevice(xrDevice).then(() => {
+    // Create a WebGL layer with a slightly lower than default resolution.
     xrSession.baseLayer = new XRWebGLLayer(xrSession, gl, { framebufferScaleFactor: 0.8 });
   });
 ```
 
+In some cases the developer may want to ensure that their application is rendering at the 'native' size for the device. To do this the developer can query the scale factor that should be passed during layer creation with the `XRWebGLLayer.getNativeFramebufferScaleFactor()` function.
```
thetuvix: After the code, it may be worth mentioning the downside of using the "native" size as shown, if your app ends up on a larger display than expected.

toji: Good feedback, thanks! Addressed it in the latest commit. Also noted that the native resolution may end up being a scale less than 1.0 (For example, when using Valve's automatic supersampling.)

thetuvix: Alternatively, we should consider what we expect apps to do with the native scale factor here.  Some compositors constrain the maximum buffer size an app can request on weaker GPUs, for example when the card does not have enough video memory to support a larger buffer.  That maximum buffer size is primarily a function of the GPU moreso than the headset display - depending on the GPU and headset, the maximum buffer size could end up bigger or smaller than the native buffer size.  As displays scale up, this will ensnare more powerful GPUs as well.

Therefore, even if an app wants to render at native size regardless of the hit to render time, it still may not be able to on a given compositor and a given GPU.  We should confirm whether we have a strong enough motivation for getNativeFramebufferScaleFactor in WebXR 1.0 - with the rest of your change here, we can always add it in later if we find apps that both need it and can live with the limitations above.

thetuvix: I see that your new text addresses this point explicitly.  It may still be worth confirming that we see a strong use for the native resolution in WebXR, to help apps avoid perf pitfalls.

Artyom17:
Oculus approves!

kearwood:
This looks good to me!

explainer.md
toji: I've definitely had feedback in the past that the lack of a method for determining the native resolution was problematic in WebVR. (Jason Marsh, for example, was one individual who called this out explicitly in the past.) So I feel that the fact that querying it will both be off the default path and subject to clamping at the UA's discretion gives us enough leeway to avoid the worst of the performance problems that may arise. Given support voiced by Oculus and Mozilla, I'm going to merge this as-is.

PR #357, opened 2018-05-24
Created an explicit XRRay type

toji:
There's a lot more to unpack here than the size of the diff would suggest, so let me explain some of the intentions behind this.

First off, this definitely will need to be rebased on top of #342, but I'm happy to do that after it's landed. In the meantime I've adopted the name of the ray that is present in that change. (`targetRay` instead of `pointerRay`).

Second, I should point out that the `transformMatrix` given on the ray is intended as a developer convenience, since that's what you'll frequently need to render with. It is not intended to be supplied or modified by the developer, and would be lazily computed when first queried. Since it's effectively syntactic sugar we could easily drop it, but I rather like it.

This change is designed not only to make the target ray case nicer but to support the [AR hit-test proposal](https://github.com/immersive-web/hit-test), which needs a way to communicate rays from the user to perform hit tests against the native understanding of the real environment. It's not appropriate to have users provide rays as a matrix because it's much harder to handle intuitively and because the matrix may include transforms that are illogical for a ray representation (scaling, shears, etc.) By having users pass a purpose-built ray object we can avoid a lot of those issues. It also gives us an opportunity to clearly scope things like normalization of the direction, which would now happen at XRRay creation time and therefore reduce the amount of validation functions which consume the rays must do.

Furthermore, it's possible that we will need to restrict hit-tests to rays that originate from the screen initially, since that's what both the mobile AR APIs currently require. If we leave off a ray constructor for the moment it limits the availability of the rays to only what's provided from elsewhere in the API. That means that you could only perform hit tests with rays retrieved from input events, which in magic window mode can only come from the screen. (It may be necessary to provide a function that also creates a projected ray for an arbitrary output canvas coordinate.) That's not the only way to solve the problem, but it's a possibility.

On the subject of the new `XRVector3` type: I've been extremely reluctant to introduce any sort of vector types to the spec, but after a lot of back an forth I feel like this is the cleanest way to present the API. I'll explain why I didn't feel some of the alternatives don't work:

**Float32Arrays**
```
interface XRRay {
  readonly attribute Float32Array origin;
  readonly attribute Float32Array direction;
  readonly attribute Float32Array transformMatrix;
};
```
This is problematic because even though the attributes are marked `readonly` that only means you can't assign the attribute to point at a new `Float32Array` and that the size can't be changed, not that the contents of the array can't be manipulated. In fact, there is no way that I know of to prevent the modification of the values in a `Float32Array`. This doesn't matter for the matricies the API tosses around because they're never fed back into the system, but if developers could get a ray and then modify the values of, say, the `direction` then we can't offer any guarantees about the value being normalized. As such, I don't feel `Float32Array`s are appropriate for anywhere that we intend to communicate read-only vectors.

**DOMPoint**
```
interface XRRay {
  readonly attribute DOMPointReadOnly origin;
  readonly attribute DOMPointReadOnly direction;
  readonly attribute Float32Array transformMatrix;
};
```
This is tempting, but the [type is still experimental](https://developer.mozilla.org/en-US/docs/Web/API/DOMPointReadOnly) and not widely supported. I don't think it's beneficial for us to tie our spec to another one with an indeterminate timeline. Plus the `DOMPoint` is a four-element vector and I think it would be an ongoing point of confusion that the `w` value in this case has no meaning.

**Separate Values**
```
interface XRRay {
  readonly attribute double originX;
  readonly attribute double originZ;
  readonly attribute double originY;
  readonly attribute double directionX;
  readonly attribute double directionY;
  readonly attribute double directionZ;
  readonly attribute Float32Array transformMatrix;
};
```
_sigh_ Just... no?

Very verbose, and it makes it much harder to pass individual vectors to a function.

So at the end of the say I feel like having a very basic vector type of our own is probably for the best. What I very much don't want to do is turn it into an interface for general vector mathematics. That's why the type itself has no constructor, readonly values, and no member functions. It is purely data storage. If we make the `XRRay` constructable in the future I'd advise against having it take in these vector types as the constructor args and instead consume raw values like so: `new XRRay(ox, oy, oz, dx, dy, dz);`

So that's my proposal for `XRRays`. Very interested to hear the thoughts the other vendors have on it, and especially anyone who's been working on AR hit testing!

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -426,7 +426,9 @@ The `gripMatrix` is a transform into a space where if the user was holding a str
 
 If the input source has only 3DOF, the grip matrix may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripMatrix` will be `null` if the input source isn't trackable. 
 
-An input source will also provide its preferred target ray on its pose, which is defined as a ray originating at `[0, 0, 0]` and extending down the negative Z axis, transformed by the `targetRayMatrix` attribute of an `XRInputPose` object. `targetRayMatrix` will never be `null`. The value will differ based on the type of input source that produces it, which is represented by the `targetRayMode` attribute:
+An input source will also provide its preferred pointing ray, given by the `XRInputPose`'s `targetRay`. The ray, which is an `XRRay` object, includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. The `XRRay` also defines a `transformMatrix` which represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
+
+The `targetRay` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `targetRayMode` attribute:
 
   * `'gaze'` indicates the target ray will originate at the user's head and follow the direction they are looking (this is commonly referred to as a "gaze input" device). While it may be possible for these devices to be tracked (and have a grip matrix), the head gaze is used for targeting. Example devices: 0DOF clicker, regular gamepad, voice command, tracked hands.
   * `'tracked-pointer'` indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing. The exact orientation of the ray relative to a given device should follow platform-specific guidelines if there are any. In the absence of platform-specific guidance or a physical device, the target ray should most likely point in the same direction as the user's index finger if it was outstretched.
@@ -442,11 +444,12 @@ for (let inputSource of xrInputSources) {
     // Render a visualization of the input source (see next section).
     renderInputSource(session, inputSource, inputPose);
 
-    // Highlight any objects that the target ray intersects with
-    let hoveredObject = scene.getObjectIntersectingRay(inputPose.targetRayMatrix);
-    if (hoveredObject) { 
+    // Highlight any objects that the target ray intersects with.
+    let ray = inputPose.targetRay;
+    let hoveredObject = scene.getObjectIntersectingRay(ray.origin, ray.direction);
+    if (hoveredObject) {
       // Render a visualization of the object that is highlighted (see below).
-      drawHighlightFrom(hovered, inputSource);
+      drawHighlightFrom(hoveredObject, inputSource);
     }
   }
 }
@@ -529,7 +532,8 @@ function onSelect(event) {
   let inputPose = event.frame.getInputPose(event.inputSource, xrFrameOfRef);
   if (inputPose) {
     // Ray cast into scene to determine if anything was hit.
-    let selectedObject = scene.getObjectIntersectingRay(inputPose.targetRayMatrix);
+    let ray = inputPose.targetRay;
+    let selectedObject = scene.getObjectIntersectingRay(ray.origin, ray.direction);
     if (selectedObject) {
       selectedObject.onSelect();
     }
@@ -589,12 +593,14 @@ function renderCursor(inputSource, inputPose) {
   // Only render a target ray if this was the most recently used input source.
   if (inputSource.targetRayMode == "tracked-pointer") {
     // Draw targeting rays for tracked-pointer devices only.
-    renderer.drawRay(inputPose.targetRayMatrix);
+    let ray = inputPose.targetRay;
+    renderer.drawRay(ray.origin, ray.direction);
   }
 
   if (inputSource.targetRayMode != 'screen') {
     // Draw a cursor for gazing and tracked-pointer devices only.
-    let cursorPosition = scene.getIntersectionPoint(inputPose.targetRayMatrix);
+    let ray = inputPose.targetRay;
+    let cursorPosition = scene.getIntersectionPoint(ray.origin, ray.direction);
     if (cursorPosition) {
       renderer.drawCursor(cursorPosition);
     }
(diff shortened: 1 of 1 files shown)


rcabanier:
Every browser but Edge is shipping DOMPoint and DOMMatrix. 
Since this API is primarily for authors that handcode JS (as opposed to WASM), I think we should go with the high level types since they are easier to develop in.

toji:
Given that we want Edge to ship this API as well rather than getting stuck on WebVR due a missing type, that omission is significant. Any Microsoft folks want to comment on implementation timeline (or lack thereof?)

Ping @leweaver, @RafaelCintron, @thetuvix? (Sorry, don't recall Yashar's GH handle at the moment)

RafaelCintron:
Note that browser support is not the only factor to consider. The matrices provided by the WebXR API will be funneled into WebGL at some point.  WebGL only accepts Float32Array matrices today.  The fewer conversions (cost) between datatypes we impose on web developers, the better.  

When all browsers and WebGL support DOMMatrix natively, the case in favor of DOMMatrix is clear.

toji:
To be clear, @RafaelCintron, you mentioned the Edge team has no immediate plans for `DOMMatrix` support, but I assume that also applies to `DOMPoint`? I don't feel particularly compelled to switch how we're representing matricies at this point, but if for whatever reason `DOMPoint` was on your TODO list that would make me much more willing to use it.

@rcabanier: Certainly we can, and I outlined some of the possibilities in my first comment along with reasons I felt they didn't work. That doesn't mean we couldn't make them work regardless, or avoid an `XRRay` type all together. However I don't feel that we should avoid declaring new interfaces when they serve a clear ergonomic purpose. If we wanted to avoid new type declarations entirely we could just pass around flat structs of the necessary values, but that gets cumbersome for developers very quickly:

``` js
function onSelect(event) {
  let inputPose = event.frame.getInputPose(event.inputSource, xrFrameOfReference);
  if (inputPose) {
    xrSession.requestHitTest(inputPose.targetRayOriginX,
                             inputPose.targetRayOriginY,
                             inputPose.targetRayOriginZ,
                             inputPose.targetRayDirectionX,
                             inputPose.targetRayDirectionY,
                             inputPose.targetRayDirectionZ).then(/*...*/);     
  }
}
```

vs.

``` js
function onSelect(event) {
  let inputPose = event.frame.getInputPose(event.inputSource, xrFrameOfReference);
  if (inputPose) {
    xrSession.requestHitTest(inputPose.targetRay).then(/*...*/);     
  }
}
```

Seems like a worthwhile tradeoff to me.

RafaelCintron:
@toji and @rcabanier , apologies for not fully answering your question.  `DOMPoint` is in the same "no immediate plans" boat as `DOMMatrix` I'm afraid.  

@rcabanier , I agree `DOMMatrix's` `Float32Array` conversion function is not a big ergonomics hit.  The runtime cost was what I was after when I brought up WebGL.  

In general, I'm not overly concerned about the developer modifying the contents of Float32Arrays we give them.  If they change it, they're the ones that have to deal with the ramifications.  Is there a peril I am not thinking of?

toji:
@RafaelCintron: The issue with allowing developers to change the value is that if the ray is passed back in to the API (such as with the hit testing proposal) then we can no longer offer strong guarantees about the validity of it's contents. For example, it would be preferable (in my opinion) to spec that constructing an `XRRay` with a zero-length direction throws an exception, rather than specifying that every function that accepts an XRRay must check to ensure it's direction is of non-zero length. Same goes for expecting every consuming function to normalize the direction when instead we could ensure it happens on construction.

@rcabanier: Regarding the `DOMPoint` concern, given that we don't want to throw up undue barriers to implementation for any vendors, what if we reframe the matter slightly and ensure that the IDL for the XR point is a strict subset of a `DOMPointReadOnly` without a constructor. Name it such that it's obvious they're meant to be analogs. Then, at a future point when the browsers have all implemented the DOM geometry APIs we can cleanly swap the types out with no developer-facing impact. (Unless they happen to be checking `if (point instanceof XRPointReadOnly)` for some insane reason.)

```webidl
[SecureContext, Exposed=Window]
interface XRPointReadOnly {
    readonly attribute unrestricted double x;
    readonly attribute unrestricted double y;
    readonly attribute unrestricted double z;
};
```

Some browser may even be able to simply point this interface at the `DOMPointReadOnly` implementation and be done with it.

My point being that if the vector issue is the only thing blocking this from going forward I'd rather come to a reasonable compromise that allows us to not be dependent on vendor support timelines for otherwise unrelated APIs.

toji:
Made the above update to the PR. I'd really love some feedback on if this is a workable direction for everyone!

explainer.md
```diff
@@ -1122,16 +1122,16 @@ dictionary XRFrameOfReferenceOptions {
 //
 
 [SecureContext, Exposed=Window]
-interface XRVector3 {
```
rcabanier: Maybe it's easier to define these as dictionaries.
A dictionary in WebIDL is just a JS object that is required to have certain members. 
It also doesn't introduce new classes in the js namespace so you don't need to get premission.

If you just make the change to dictionary, you will also allow the passing of XRVector and XRPoint to function that take a DOMPointInit

toji: Talking to our web API ergonomics contact here, I have been informed that returning dictionaries like this would definitely be "weird" from a web platform perspective. In _most_ cases, dictionaries are used to pass information into the web platform and interfaces are used to pass information back out.

toji:
At the Seattle F2F Microsoft indicated that they are OK with the WebXr spec relying on DOMPointReadOnly, and will implement it and the related interfaces as needed when they implement WebXR.

I'll put up a rebased and revised pull request based on that soon.

toji:
PR updated. Please take a look and provide any feedback you may have! I especially want to call attention to the fact that since `DOMPoint(ReadOnly)` is a 4-element vector this change specifies what the W component must be for each usage. (1 for `origin`, 0 for `direction`.)

explainer.md
```diff
@@ -1160,6 +1167,26 @@ dictionary XRFrameOfReferenceOptions {
 // Input
 //
 
+[SecureContext, Exposed=Window]
+<<<<<<< HEAD
```
ddorwin: Rebase artifacts.

toji: Derp. Thanks, fixed it.

toji:
I'm under the impression that there's broad agreement on this now, but I would really like an indication on this PR that other implementors have looked at the change and approve of it prior to merging.

toji:
Merged, give that there have been no objections over the past week.

Issue #355, opened 2018-05-22
onresetpose should be clear what happens when poses are reset


billorr-google:
For example, it might mean that sites should treat the next pose direction as forwards and update themselves.  It could instead mean that poses and transforms are relative to a new origin.  We should be clear what is meant.

If we mean the second, should we give out the transform between new and old origin?  This could be useful for smoothly animating content from one place to another.

toji:
New pull request is up to clarify this (though the answer is "it's platform specific.")

PR #358, opened 2018-05-24
Specify timing/behavior of the resetpose event

### LINKED ISSUE #354: resetpose event timing requirements
Should the spec require that the resetpose event listeners be triggered before any frame with the new pose data is available for use?

https://immersive-web.github.io/webxr/#eventdef-xrsession-resetpose

toji:
This should resolve #354 and #355. Anyone have concerns about the described behavior?

**TL:DR:** the `resetpose` event must be fired _before_ any poses are delivered with the reset origin/direction.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -753,9 +753,9 @@ function onDrawFrame() {
 
 ### Responding to a reset pose
 
-Many XR systems have a mechanism for allowing the user to reset which direction is "forward." For security and comfort reasons the WebXR Device API has no mechanism to trigger a pose reset programmatically, but it can still be useful to know when it happens. Pages may want to take advantage of the visual discontinuity to reposition the user or other elements in the scene into a more natural position for the new orientation. Pages may also want to use the opportunity to clear or reset any additional transforms that have been applied if no longer needed.
+Many XR systems have a mechanism for allowing the user to reset which direction is "forward" or recenter the scene's origin at their current location. For security and comfort reasons the WebXR Device API has no mechanism to trigger a pose reset programmatically, but it can still be useful to know when it happens. Pages may want to take advantage of the visual discontinuity to reposition the user or other elements in the scene into a more natural position for the new orientation. Pages may also want to use the opportunity to clear or reset any additional transforms that have been applied if no longer needed.
 
-A page can be notified when a pose reset happens by listening for the 'resetpose' event from the 'XRSession'.
+A page can be notified when a pose reset happens by listening for the 'resetpose' event from the 'XRSession'. The exact behavior of a reset is platform specific, but in most cases it's expected that the origin and/or forward direction of all active coordinate systems will shift to be aligned to the users current physical location and orientation. The event must fire prior to any poses being delivered with the newly origin/direction, and all poses queried following the event must be relative to the reset origin/direction. 
 
 ```js
 xrSession.addEventListener('resetpose', xrSessionEvent => {


toji:
No worries! The event should not require new `XRFrameOfReference`s, but may alter the meaning of existing ones. I tried to cover that in the text:

> in most cases it's expected that the origin and/or forward direction of all active coordinate systems will shift to be aligned to the users current physical location and orientation

This is fairly platform specific, and may vary by frame of reference type, so I don't want to get over-prescriptive. For example, on a Mirage Solo (standalone Daydream device) holding the home button triggers a recentering that reset both the forward orientation and origin. The device can detect floor heights, so for a `stage` you'd get and origin X and Z and forward direction reset, but probably not Y, while an `eyeLevel` FoR would reset all of them. Contrast with a Vive where pose resets have zero effect on a `stage` (since the room is calibrated to have a consistent forward direction and center) but will similarly reset the full origin and forward direction of an `eyeLevel` FoR.

NellWaliczek:
Thanks, Brandon!  I think I missed that because it wasn't the type name XRCoordinateSystem.  

Ok, so then how are developers expected to handle existing teleportation transforms?  I'm not sure zeroing back to "origin" is always going to be the right mechanic for the end user experience.  In which case, I wonder if it would be helpful to provide a transform representing the delta for a given XRFrameOfReference object since the last XRPresentationFrame (err... is that what it's still called?)  If so, it probably needs to be an event on the XRFrameOfReference objects, yes?

Again, sorry if I'm rehashing conversations you've already had!

thetuvix:
We should decide if some of the Mirage Solo vs. Vive behavior differences you discussed are OK or will lead to subtle app issues if you test only on one or the other.

My starting point is that I'd love to try defining pose-reset behavior for WebXR abstractions like "stage" that are consistent across underlying devices.  If that's too difficult to achieve given the differences in underlying platform reset behaviors, we could back away and make this platform-specific, although I'd expect OpenXR to lead to alignment here sooner than later anyway.

toji:
This was discussed on the call today, and in addition to Alex's suggestion above we decided to try an enumerate both the modes of reset we're aware of and the actions that we would expect the developer to take in response to this event and see if it's something that merits attempting to normalize the meaning of "reset".

So here's some reset behaviors that I'm aware of. Feel free to add to the list:
  1. Reset has no effect or is not available. (Vive/Rift `stage` behavior, where play area is well defined by static external markers.
  1. Origin is reset to headset position and forward direction is reset to current head yaw. (Mirage Solo, Vive/Rift `eyeLevel` behavior)
      1. Controller orientation may be aligned with head orientation in this scenario if 3DoF.
  1. Forward direction is reset to current head yaw, origin is unchanged. (Oculus mobile, Daydream 3DoF)
      1. Controller orientation may be aligned with head orientation in this scenario if 3DoF.
  1. Reset moves virtual origin X/Z to user's location and reset forward direction, but does not affect Y position because distance to floor is known. (Mirage Solo 'stage' behavior)

And these are actions that I can see the developer wanting to take in response:
  1. Developers will want to have a clear signal for when a tracking discontinuity is due to a reset rather than a temporary tracking loss. As @kearwood amusingly pointed out, without a way to identify them as such developers might treat such jumps as real movement, leading to screwed up physics computations and possibly cheating opportunities where players abuse reset to "warp" into a physics-based object and apply unrealistic acceleration. (Kip's example was using a reset-warp to "hit" a tennis ball at superhuman speeds, which I think is amazing/hilarious and would _definitely_ be the type of thing gamers would try.)
  1. On platforms with 3DoF controllers, the reset typically aligns the head pose and hand pose. In some cases, especially in apps with artificial yaw control, it may make sense to preserve the user's current orientation during a reset so that, effectively, only the controller orientation is update.
  1. If an application allows artificial movement, the user may get into positions where doing a reset could potentially place them inside world geometry, and so the developer may want to take advantage of the visual discontinuity to nudge the user's position into a more sensible spot.
  1. Some platforms, such as Oculus's mobile devices, use resets as a signal for re-aligning UI, placing UI elements at the pitch of the user's head when they performed the reset. This is used, for example, to accommodate using the Oculus browser when laying down. This same effect can be achieved other ways, but it's certainly a use of the event worth considering.

Also, to reiterate a point that I brought up on the call, I do think that for a great many applications the appropriate response to a reset event will be to do nothing at all. Users will generally be familiar with the expected effects of a reset for their platform, and developers trying to get too clever with it is likely to be more confusing than helpful. As such we shouldn't encourage it.

toji:
This is superseded by the content in #409

Issue #120, opened 2016-10-15
Clarifcation of depthNear and depthFar behavior


NellWaliczek:
The comments on the VRDisplay.depthNear and VRDisplay.depthFar attributes do not state which units they are specified in.  I'm guessing it's meters?

They also aren't included at all in section 2.1.1

And looking at it further, it seems like it might be worth pointing out in the comments for these attributes that the values are used to create the projection matrices in VRFrameData.

Furthermore, it would probably be worth clarifying when changes to these values will be applied.  In other words, should the developer requery the frame data immediately or will the changes be pending until the next frame?

toji:
Hey! I actually wrote up a pull request for this! (Over a year and a half later, but still...) #359

PR #359, opened 2018-05-25
Add text explaining the depth range values.

toji:
Not much to say about this one. Resolves longstanding issue #120. Inclined to push this soon if there's no comments about it. The only possible concern I could see others having is the specification that changes to the depth values take effect on the next frame.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -159,11 +159,6 @@ function onSessionStarted(session) {
   // Store the session for use later.
   xrSession = session;
 
-  // The depth range of the scene should be set so that the projection
-  // matrices returned by the session are correct.
-  xrSession.depthNear = 0.1;
-  xrSession.depthFar = 100.0;
-
   // The `XRFrameOfReference` provides the coordinate system in which
   // `getViewMatrix()` and the `poseModelMatrix` are defined. For more
   // information on this see the `Advanced functionality` section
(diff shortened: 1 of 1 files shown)


toji:
@leweaver and the rest of the Microsoft crew asked for some time on the previous Immersive Web call to evaluate this change and ensure it works with their platform. If the answer is yes then I'll merge it because otherwise it seemed to be non-controversial.

leweaver:
Sorry for the delayed response @toji - we've been talking about this in depth (puns aside!) over the past few days. 

We are happy that this change will work on our platform and for it to be merged - but would also like to mention that in-frame changes to the near/far values are no longer restricted to taking effect on the next frame on our platform. This could possibly remove the need for text to that effect in this change?

toji:
For me the mid-frame restriction is actually more about predictability of content than a platform restriction. If you change the depth range mid-frame then anything you've rendered to that point will no longer depth-test properly, and it's possible that some reprojection techniques would begin to fail.

blairmacintyre:
Regarding @kearwood's comment:  what is the relationship to logarithmic depth?  In many of our AR demos with Argon, that used geospatial content, we used logarithmic depth so we could use "very large values" (e.g., put things on the other side of the planet).  If this isn't possible, it can be worked around, but it can be a pain.

Issue #308, opened 2017-12-14
Should the Feature Policy feature name be changed from "vr"?


ddorwin:
In https://github.com/WICG/feature-policy/issues/49, `"vr"` was added to control "access to VR displays." This string, rather than "webvr" was intentionally chosen to be extensible to other APIs. #86 tracks adding relevant text to this spec.

Now that the spec has been renamed and is intended to cover more user cases, we should decide whether to also change the feature name, and, if so, to what.

On one hand, it makes sense to more closely match the spec and apply to all devices accessed by this spec/`navigator.xr.requestDevice()`. On the other hand, there may someday be legitimate reasons to provide access to VR but not AR (or vice versa), and, as with the original name, the scope need not be limited to the specific API. Perhaps the strongest argument for a single name is the position some have stated that AR and VR are points on an XR/MR continuum. In other words, there may be scenarios and devices in between for which the correct value may not be obvious.

We should make a decision soon then file a new issue in Feature Policy similar to the one above and update #86 with the correct string.

/cc @raymeskhoury @clelland

mkungla:
@ddorwin 
> there may someday be legitimate reasons to provide access to VR but not AR

That is tough subject. In privacy perspective AR should never supersede VR. While API can be same, but getting users AR permissions should be separated.

anssiko:
We had a similar discussion for the Generic Sensor API in terms of its [Feature Policy integration](https://w3c.github.io/sensors/#feature-policy-api) and settled on reusing the low-level [permissions registry](https://w3c.github.io/permissions/#permission-registry) strings `accelerometer`, `gyroscope`, `magnetometer` etc. that directly map to the concrete sensors. The same strings are used in Feature Policy e.g. `<iframe allow="accelerometer">` and Permissions API e.g. `navigator.permissions.query({ name: "accelerometer" })`.

If you were to follow this pattern, you'd want to make a similar distinction between VR and AR.

The decision to use low-level rather than high-level catch-all strings (such as `motion-sensors` cf. `xr`) was motivated by security and privacy benefits that come with the principle of least privilege and [data minimization](https://www.w3.org/2001/tag/doc/APIMinimization.html). To mitigate possible user confusion (users might not understand the low-level concepts) implementations are able to coalesce the low-level concepts into appropriate higher level concepts familiar to the user if there's a need to surface UI, for example for user-facing permissions prompts.

cvan:
[...]

PR #360, opened 2018-05-25
Clarified when immersive sessions are rejected

### LINKED ISSUE #343: Avoiding race conditions in parallel exclusive session requests
Currently the spec does not define how race conditions between multiple exclusive session requests should be handled. 
https://immersive-web.github.io/webxr/#dom-xrdevice-requestsession

For example: If Request A starts and is held up in stage 7, then Request B starts while A is outstanding, what should the behavior of B be?

Proposal:
- Request A
- Request B
- B is blocked while A is outstanding
- If A succeeds reject B
- If A fails allow B to proceed

toji:
Fixes #343.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -130,7 +130,7 @@ In this sample, the `beginXRSession` function, which is assumed to be run by cli
 
 If `supportsSessionMode` resolved for a given mode, then requesting a session with the same mode should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.) The UA is ultimately responsible for determining if it can honor the request.
 
-Only one immersive session per XR hardware device is allowed at a time across the entire UA. All inline sessions are [suspended](#handling-suspended-sessions) when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
+Only one immersive session per XR hardware device is allowed at a time across the entire UA. If an immersive session is requested and the UA already has an active immersive session or a pending request for an immersive session, then the new request must be rejected. All inline sessions are [suspended](#handling-suspended-sessions) when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it. 
 
 Once the session has started, some setup must be done to prepare for rendering.
 - An `XRReferenceSpace` should be created to establish a space in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
--- a/index.bs
+++ b/index.bs
@@ -360,10 +360,14 @@ When an {{XRSession}} is shut down the following steps are run:
 
 <div class="algorithm" data-algorithm="end-session">
 
-The <dfn method for="XRSession">end()</dfn> method provides a way to manually shut down a session. When invoked, it MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
+The <dfn method for="XRSession">end()</dfn> method provides a way to manually shut down a session. When invoked, it MUST run the following steps:
 
+  1. Let |promise| be [=a new Promise=].
   1. [=Shut down the session|Shut down=] the target {{XRSession}} object.
-  1. [=/Resolve=] |promise|.
+  1. [=Queue a task=] to perform the following steps:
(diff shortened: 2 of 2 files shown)


NellWaliczek:
Looking at issue #343, it sounds like that's a solid approach for in-progress session creation requests.  What about existing sessions?  What's the reasoning behind blocking a new session from supplanting an existing session?  Off the top of my head that seems potentially unnecessarily restrictive...

RafaelCintron:
One of the reasons requestSession returns a promise (at least in desktop) is so that the browser can ask the user permission to enter exclusive mode for a given domain. If we allow new sessions to supplant existing sessions in all cases, that means we need to ask the user for this permission inside of the existing exclusive session, thus breaking the original experience.  I suppose we can allow it for same domain requests but, in that case, the developer might as well figure out a way to reuse the session objects themselves.

For the time being, I'm inclined to favor the "one at a time" proposal in this PR.

NellWaliczek:
I see where you're coming from.  Thinking this through a couple extra steps..here are two other things to consider:
1. What about iframes?  Do we want to enable a user to cause the active session to be supplanted by a session from another iframe?
2. Won't we need to address the "where is my permissions dialog?" question when we wrap up the page-to-page navigation design?

RafaelCintron:
> What about iframes? Do we want to enable a user to cause the active session to be supplanted by a session from another iframe?

Maybe but we should think through the ramifications.  

The user experience starts with the user saying 'yes' to a specific domain.  I'm willing to let other iframes from the same domain supplant with no permission and maybe even the top level document. After all, the user made the choice to navigate to the top domain in the first place.  But not so sure about some unrelated domain on the page that may not be visible.

Even if we get the user's permission, if we allow unrelated domains to supplant each other, we could end up with two pieces of code fighting for control for the HMD with blur events flying all over the place. We need to put the user in control of these situations as much as we can. 

> Won't we need to address the "where is my permissions dialog?" question when we wrap up the page-to-page navigation design?

I'm less worried about page-to-page navigation. Here, only the top-level domain will be navigating.  iFrames can't navigate the top level document, only the iframe.  Any dialogs for page to page navigation won't really be "permission" dialogs but dialogs that contain the name of the new place and a progress indicator.

ryzngard:
I think we should consider what type of experiences will be restricted by this decision.


Off the top of my head, I can think of at least one; A gallery of XR experiences. Do we require all experiences in the gallery to be hosted in the same origin to work as a continuous experience? Is it enough to require that such an experience use navigation to achieve it's goal rather than overtaking a session? 

The iframe issue is interesting. I looked at some of how fullscreen handles this paradigms, and it looks like there's an explicit attribute on iframe [allowfullscreen](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe). Will we require a similar attribute for xr?

toji:
Okay, finally (*finally*) coming back around to this issue. Rebased the commit and re-read the comments to figure out where we stand on this. My conclusion is that the original assertion that only one immersive session, whether pending or resolved, is allowed at a time still stands. In fact, it turns out that the [spec already had language to that effect](https://immersive-web.github.io/webxr/#dom-xr-requestsession) in it, so this is one of the few areas where the explainer is lagging behind.

To address a couple of other questions directly:

**Will we require an iframe permission attribute for xr?**
Yes, using the feature policy API. This is actually [already specified](https://immersive-web.github.io/webxr/#feature-policy), but there are [remaining questions](https://github.com/immersive-web/webxr/issues/308) about the feature naming and granularity.

**Will this block gallery-style pages that want to transition between isolated experiences hosted on the same origin?**
There's still a few ways to achieve that, the simplest answer being that if the page has enough visibility of the experiences to present them as a gallery in the immersive scene then it probably has enough context to share a single immersive session between those experiences. Otherwise I updated the spec in this PR to allow `requestSession()` to succeed immediately after `end()` was called on an existing immersive session, which would allow for new immersive sessions to be spawned in response to a single click event from the previous session.

**What about more general navigation?**
[We still need to spec more general navigation capabilities](https://github.com/immersive-web/webxr/issues/517), but that's beyond the scope of this specific PR and should not be affected by this limitation.

toji:
You forgot to invoke the name of @RafaelCintron with the appropriate runes so that the email imps deep within GitHub will summon him. 😉

Issue #317, opened 2018-01-10
Use Texture Arrays directly with WebGL 2


toji:
This idea originated with an email that @RafaelCintron sent me a couple of days ago which was then discussed on the last WebXR call. In short, he proposed having a XR layer type that was WebGL 2.0-only that provided Texture Array's rather than the opaque framebuffer of our current `XRWebGLLayer`. (`XRWebGLLayer` would stick around for WebGL 1.0 compatibility.) Then the WebGL multiview extensions could use the texture array directly, which would more directly mirror the underlying native extensions like OVR_multiview.

I'm in favor of this approach, as it solves a couple of problems. Primary among them is that it would give developers the ability to mix geometry that is rendered via multiview techniques with geometry that is rendered one eye at a time. I believe that this is necessary because otherwise any single technique that a library uses that has not been converted to be multiview compatible would disqualify an app from using multiview at all (this is likely to be common, as the web ecosystem delights in mashing code from a variety of sources together.)

> Note: Even if we introduce a new layer type the plan would still be to only allow one layer at a time to be presented in the initial version of WebXR.

One point of misunderstanding in our initial discussion: I had assumed that this proposal was meant to wholly replace the previously planned method of multiview compatibility which was via the opaque framebuffer and could potentially support WebGL 1.0. I'm personally OK with the tradeoff of saying that multiview is a WebGL 2.0-only feature (this is the state of native OpenGL ES anyway.) @RafaelCintron and others apparently weren't ready to go that far and appear to be in favor of supporting both the Texture Array method and the opaque framebuffer method. That would lead to broader compatibility at the expense of complexity, both in WebXR and WebGL. (And again, the opaque framebuffer route is an all-or-nothing affair as currently specced.) In any case, I would be very interested in hearing from a variety of sources how big of a hurdle a WebGL 2.0 requirement for this feature would represent.

To kick off the discussion, here's a quick stab at the IDL to support this feature, all subject to copious bikeshedding. This variant omits multiview on the `XRWebGLLayer`, but that's ultimately dependent on the groups feelings on the above points.

```webidl
// I think it's important (and easy) to retain WebGL 2.0 compatibility on the "basic" layer type.
typedef (WebGLRenderingContext or
         WebGL2RenderingContext) XRWebGLRenderingContext;

dictionary XRWebGLLayerInit {
  boolean antialias = true;
  boolean depth = true;
  boolean stencil = false;
  boolean alpha = true;
  boolean multiview = false;
[...]

PR #361, opened 2018-05-26
Removed multiview attribute from XRWebGLLayer

toji:
Based on some of the discussion in #317, the WebXR calls, and the WebGL face-to-face meetings, it appears that doing multiview rendering with the standard opaque framebuffer isn't going to work out. Instead, we'd like to make use of as many of the existing features of WebGL 2 as possible to align with the native multiview extension functionality.

I'm approaching this in two steps. First, in this pull request, I'm removing the multiview concept from the `XRWebGLLayer`. That's pretty straightforward.

The next step, then, would be to introduce a WebGL 2-only layer type that provides texture arrays instead of the opaque framebuffer. I'm not sure if that's something that we want to do in v1 release, though. My gut says no? For the sake of tracking where we'd like to go, however, I've written up the basic explainer text and IDL in the [webgl2_layer branch](https://github.com/immersive-web/webxr/commit/31341c8ab9dfea9971d4bf7abc95d8553b0b7725). Happy to have further discussion about that, but it's probably best done in #317.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -763,71 +763,6 @@ xrSession.requestFrameOfReference("stage", { stageEmulationHeight: 1.2 })
       // Will always succeed.
       xrFrameOfRef = frameOfRef;
     });
-```   
-
-### Multiview rendering
-
-Developers may optionally take advantage of the [WEBGL_multiview extension](https://www.khronos.org/registry/webgl/extensions/WEBGL_multiview/) to both WebGL 1.0 and WebGL 2.0 for optimized multiview rendering. The WEBGL_multiview extension must be successfully queried from the supplied context prior to the creation of the `XRWebGLLayer` or it will fall back to using a framebuffer that is not multiview-aware. Additionally, the UA may choose to not honor the multiview request for any reason, which will also fall back to using a framebuffer that is not multiview-aware. As such, developers must query the `XRWebGLLayer.multiview` property after the `XRWebGLLayer` is constructed and respond accordingly.
-
-When `XRWebGLLayer.multiview` is false:
-- The `XRWebGLLayer`'s `framebuffer` will be created in a side-by-side configuration.
-- Calling `XRWebGLLayer.getViewport()` will return a different `XRViewport` for each `XRView`.
-
-When `XRWebGLLayer.multiview` is true:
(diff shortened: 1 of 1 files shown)


Artyom17:
Just re-iterating here what I said on last WebXR call. My concern that removing the opaque FB approach may slowdown multiview adoption. It is already pretty "user-unfriendly" feature, meaning devs will be required to change their shaders to make it work. By requesting users to switch to WebGL 2 AND to manage texture arrays / framebuffers by themselves is another obstacle on multiview adoption way. 

Needless to say that most of the existing WebVR experiences are made using WebGL 1, including frameworks like three.js, aframe, react360, so on. Maybe, WebGL community wants to speed up the WebGL 2 adoption by limiting multiview to WebGL 2 only, right?

There are certain difficulties in adding multiview support to WebGL1, and the main one - shaders must be ES 3.00 (I mean, native shaders). This shader model exists only in WebGL 2, however, I already have a PoC modification to ANGLE shader transpiler, that takes WebGL1 shader and transform it to ES3.00; thus, multiview may work in WebGL 1 shaders (with ViewID support addition, of course).

But I also like the flexibility of WebGL2Layer approach, tbh. Like Brandon said, the opaque FB approach is "all or nothing", while WebGL2Layer will allow to add multiview for certain parts of the rendering and it just closer to native GL code, thus porting existing GL + MV experiences to WebGL 2 / WebXR would be much easier.

I don't have a solution to this, but I feel like I'd love to have both ways - the opaque FB and the WebGLLayer, but it is, probably, too much to ask.

I also forgot to respond on Brandon's question whether multiview support should or should not be in V1 of the spec: I feel like it must be there, otherwise multiview would never be used widely. If we don't want to spend too much time on it for V1, then the opaque FB approach (even limited to WebGL 2 only) sounds as a good enough compromise for V1, while WebGL2Layer way could be added in V2.

Just my 2c.

toji:
Coming back to this one after a 5 month lapse in attention. 😅

I'm inclined to merge this ASAP for the sake of cleaning up the explainer while we churn on some higher importance spec issues. I don't feel that the feature, as described here, could actually be usefully implemented by any UA at this point, and some of our discussions about using array textures, etc as alternative routes have yet to come to a close.

I'd definitely like to see this feature made available, and it sounds like there's probably some parties interested in pursuing it, but until we reach better consensus on how it works I think it's better left out of the explainer/spec.

toji:
#414 filed. Merging this PR now.

rolivo:
A real shit

PR #365, opened 2018-06-06
Unbind the framebuffer when the session ends.

darktears:
If you're reusing the same canvas with the same context to render
the content inside the page you want to tell to unbind it from
the XRWebGLLayer. Not doing triggers warning in Chrome :

WebGL : INVALID_FRAMEBUFFER_OPERATION: cannot render to a WebVR
layer outside of a frame callback.


toji:
LGTM, thanks for catching this!

PR #376, opened 2018-06-26
Add an explainer for session creation and configuration

ddorwin:
The initial version contains background, a proposal, open questions & issues, and a discussion of alternatives.

There is also an additional proposal to add a Start() method to XRSession.


toji:
I'm going to land the two explainer PRs at this point, but then immediately move them into a `designdocs/` folder with some light renaming. Reasoning is that while these documents contain valuable context and research in to the subjects in question they don't really fit what the W3C thinks of as "explainers" (For that matter, neither do our main explainers, but that's a bigger issue for another time...)

Also, I'd like the clear out some PRs that have stagnated so that it's easier to see what's under active development.

PR #383, opened 2018-07-20
Immersive navigation proposal

toji:
This pull request replaces #256 (though it was informed by much of the discussion in it) and compliments the explainer posted at #382.

Primary changes vs. previous attempts at this feature:

 - Explicit function call vs. an argument on an existing one to simplify logic and avoid weird edge cases.
 - Call directly on the `navigator.xr` object to avoid concerns about whether or not it's possible to make the call early enough in the page lifecycle.
 - Supplementary material in the form of [David's explainer](https://github.com/immersive-web/webxr/blob/navigation-explainer/navigation-explainer.md) to cover some of the wider considerations.
 - Simply not caring any more about the "activation" case.


arpu:
how can i use the Immersive navigation  with webXR and chrome now?

toji:
Closing this for now to reduce noise on the repo. This wouldn't be able to land as-is.

PR #387, opened 2018-08-02
Replace XRStageBoundsPoint with DOMPointReadOnly

toji:
Now that we've committed to using `DOMPointReadOnly` in `XRRay`, it looks a bit weird to have a separate "Point" type for stage bounds. Happily `DOMPointReadOnly` can work as a drop-in replacement with no changes to how the user would actually interact with the point array. The one thing that we'd want to specify is what the `y` value should be (I've explicitly made it 0 in this PR, since that sounds like a sensible default). `w` should be 1, identifying it as a point, not a direction, and staying consistent with our use in `XRRay`.

Any opposition to this change?

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -1,4 +1,4 @@
-﻿# WebXR Device API Explained
+# WebXR Device API Explained
 
 ## What is WebXR?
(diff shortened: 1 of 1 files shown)


toji:
Oops! Meant to leave a note on this: At this weeks call this was unanimously approved. Merging now.

PR #395, opened 2018-09-14
Add XRSession inputsourceschange event

jsantell:
Described in the explainer, but not in the spec text

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -355,6 +355,7 @@ enum XREnvironmentBlendMode {
   attribute EventHandler onresetpose;
   attribute EventHandler onend;
   attribute EventHandler onselect;
+  attribute EventHandler oninputsourceschange;
   attribute EventHandler onselectstart;
   attribute EventHandler onselectend;
 };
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1168,6 +1173,8 @@ A user agent MUST dispatch a <dfn event for="XRSession">resetpose</dfn> event on
 
 A user agent MUST dispatch a <dfn event for="XRSession">end</dfn> event on an {{XRSession}} when the session ends, either by the application or the user agent. The event MUST be of type {{XRSessionEvent}}.
 
+A user agent MUST dispatch a <dfn event for="XRSession">inputsourceschange</dfn> event on an {{XRSession}} when the session's [=list of active input sources=] has changed. The event MUST be of type {{XRSessionEvent}}.
```
jsantell: Possibly this should be an `XRInputSourceEvent` instead of `XRSessionEvent`, but it's unclear, as it is not related to any specific `XRInputSource`.

toji: No, I think this is correct. It's not intended to communicate the connection state of a specific input, only that the list has been updated.

toji:
LGTM, thanks!

Issue #403, opened 2018-09-21
Is an XRFrame object valid after the frame it represents has elapsed?


lincolnfrog:
How do XRFrame objects work with respect to previous frames? Is it possible for XRFrames to store historical frame data? If so, does this mean underlying implementations NEED to store historical tracking data?

To put it another way, can XRFrame objects be stored outside of the callback in which they are issued?
Note: if XRFrame object is not valid outside of the callback, it would discourage us from using promises in any API that resolves with XRFrame objects. This might make us want to avoid promises entirely in the API to avoid inconsistency. This is a potential downside of such a decision since promises are a very common and friendly way of doing async APIs.

NellWaliczek:
Hey Max, 
Is this issue covered by the discussion in https://github.com/immersive-web/webxr/issues/384 ?
Nell

lincolnfrog:
@NellWaliczek I don't think its the same issue. They are definitely related, but this issue could be decided in several different ways without impacting #384 . 384 is about how XRFrames interact with XRCoordinateSystems, this issue is more about the lifetime of an XRFrame object itself and whether underlying implementations have to track historical data to support queries on previous frames.

toji:
This was marked for discussion at TPAC, but [according to the minutes](https://www.w3.org/2018/10/25-immersive-web-minutes.html) it was deferred to the anchors discussion but then never revisited.

The opening comment mentioned that one of the primary reasons for wanting to keep `XRFrames` active outside the initial callback is so they can be used in promises, but as I've argued in #411 (which appears to have positive sentiment towards it) we shouldn't be using promises for timing sensitive data and I definitely view `XRFrame`s as pretty much the textbook definition of timing sensitve. As a result, if we subscribe to the guidelines I proposed in #411 they should be only handled in either callbacks or events. Happily that's already the case today!

[...]

Issue #396, opened 2018-09-14
Address confusion regarding XRFrameOfReference types


NellWaliczek:
A few weeks back, @toji and I discovered that we didn't have the same understanding of the definitions of the "eye-level", "stage", and "head-model" XRFrameOfReference types.  We did a poll on one of the weekly calls and it sounded like there were a handful of others that had various interpretations as well.  
This issue tracks the need to reach agreement the definitions of these types.  It also covers clarifying the explainer/spec text to reflect the expected behaviors on various devices such as 3DOF/6DOF or those which might need to emulate the floor offset.  It is also related to issue #389 filed by @Artyom17

RafaelCintron:
Thank you for bringing this us, @NellWaliczek . 

Reading through the current crop of frames of reference, I was confused how they're partially defined in terms of each other.  

For head-model, it says: `An XRFrameOfReference with a frame of reference type of "head-model" describes a coordinate system identical to an eye-level frame of reference, but where the device is always located at the origin. `

For eye-level, it says: `Describes a coordinate system with an origin that corresponds to the first device pose acquired by the XRSession after the "head-model" frame of reference is created`

Does this mean that you need to make a head-model frame of reference before you make an eye-level frame of reference?

lincolnfrog:
As per conversation in the f2f, maybe we can simplify this by splitting the role of FOR types apart into ~3 different concerns:

1) Getting a view matrix
Since the main use-case of these seems to be for generating a view matrix, we can just have a method XRSession.getViewMatrix(frameOfReferenceType, offsetTransform). This type parameter would likely not need to include "stage" as the stage bounds could be separated out and things like emulated height could be included in the offsetTransform (after having been queried separately - see below). Likely then we just need two options for frameOfReferenceType if we make it so that the "world" type where accuracy of tracking is best in the immediate vicinity of the headset is the standard behavior and we require ubiquitous anchoring of all virtual content (with emulated anchors for 3DOF or outside-in systems). The two options would basically be "head" and "world", where head has the translation zeroed out and world does not. Note: we might need a third setting for whether a neck-model should be used or not since we agreed we want to avoid people hacking the view matrix post-hoc to remove translations.

2) Stage bounds / emulated height
Just move these to XRSession as well - XRSession.getStageBounds() and XRSession.getEmulatedHeight(). The emulated height could then be passed into the getViewMatrix() function as an offset transform to differentiate between seated and standing modes. 

3) Feature detection
[...]

Issue #389, opened 2018-09-01
Add "local" XRFrameOfReference


Artyom17:
@casseveritt recently brought a good issue up. With all the positional tracking untethered devices, like HoloLens or Lenovo Solo you may need to render UI elements in such way that the user doesn't end up inside of them. Even if you initially render something on the distance from the current user's location, the user still may come "into" the UI element. Of course, you always may say that the experience could be responsible to recalculate the UI elements' positions in the world coordinate system; however, if we rely on each apps' (experience) logic then the behavior will be inconsistent between them.
So the idea is, we don't want apps to have to figure out where the user wants UI-related stuff. We want to have a system query where apps can ask for the current "local origin". And when the user does a long-press of a specific button on a controller to recenter or whatever, what it does is update that local origin.
Since this is a system-level setting, all apps would be working with the same location / orientation. So overlays like UI or panels in the VR space, would draw around where the user has said his / her center is. But that wouldn't change the coordinates that the tracking system sends you by default. So stuff you want to draw around the user, you do by drawing it around the "local" pose. But apps that want to draw stuff in the physical world will need to know about both local space and the room space. By providing a query for local space, we don't require apps to try to come up with their own idea of where the user wants local stuff drawn. And, in particular, we don't have to worry about every app (experience) coming up with different ways to communicate this.

So, the idea is to add 'local' XRFrameOfReference type, where the player's position will be (0,0,0) at the time of pose reset, for example, when user press and hold a specific button on the controller.

Any thoughts on that?

speigg:
Why not just use the eye-level frame, since the UA should be able to let the user reset that frame?

casseveritt:
Is the eye level frame always intended to be centered around the user?  If "eye-level" is synonymous with the local frame, that could work.  (Though the "local" name I think captures what is intended more clearly.)

speigg:
I agree, I think “local” is a better name for the “eye-level” frame—that’s what we used to call it in our Argon project, anyways. But yes, the “eye-level” frame should generally be “near” the user, while the “head-model” frame follows the user around.

casseveritt:
Ok, I think the name "local" will be a better match for OpenXR, if that matters.

Stage frames don't have an implicit correspondence to the local frame though.  Rather apps have to explicitly ask for their relation (the transform between them).  Stage space's origin is always at the center of the stage, with the edges axis aligned.

speigg:
[...]

Issue #384, opened 2018-07-25
Enable XRCoordinateSystem to support dynamic coordinate systems


thetuvix:
One gotcha in the base API at the moment is that all `XRCoordinateSystem` instances can only represent **logically static objects**, as the `XRCoordinateSystem.getTransformTo(coordinateSystem)` method is not time-indexed.

While the UA can adjust the relationships between two static `XRCoordinateSystem`s over time as the system learns about the world, missing time-indexing means an `XRCoordinateSystem` cannot represent a dynamic object that will physically move between now and a particular photon time. Instead, developers locate dynamic objects like a head or controller by asking a given time's `XRFrame` for an `XRDevicePose` or `XRInputPose`, respectively.

Limiting `XRCoordinateSystem` to static objects has worked OK for the current features in the spec, but it means that, as we build out our AR support, we don't have a durable currency that apps can use to represent a dynamically-moving object, such as the head or controller to track as part of a hit-test subscription.

Note that a common currency to represent a dynamic coordinate system will likely be useful for multiple WebXR APIs we are likely to need over time, for both VR and AR:
* A hit-test subscription method on `XRSession` that takes a base `XRCoordinateSystem` to track each frame (e.g. a device-locked `XRFrameOfReference`), along with an `XRRay` within that coordinate system ([hit-test #4](https://github.com/immersive-web/hit-test/issues/4))
* A coordinate system that tracks the root pose of detected AR world objects, some of which may require time-indexing (e.g. an ARKit ARFaceAnchor, etc.)
* A way to specify the reprojection basis for a second compositor layer to track, e.g. so the compositor can independently stabilize a handheld object to the motion of that hand during reprojection

### Proposal

The core proposal is to upgrade `XRCoordinateSystem` to represent either static or dynamic coordinate systems, by allowing time-indexing when relating two coordinate systems:
* Replace `XRCoordinateSystem.getTransformTo(coordinateSystem)` with **`XRFrame.getPose(coordinateSystem, baseCoordinateSystem)`**, getting the pose of `coordinateSystem` within `baseCoordinateSystem` at the time represented by that `XRFrame`.

This approach continues to support static coordinate systems, while enabling us to then introduce various dynamic coordinate systems over time, as we define VR and AR features that depend on durable tracking of dynamic objects.

Examples of features that will depend on dynamic coordinate systems:
[...]

PR #409, opened 2018-10-04
Rationalizing and clarifying XRFrameOfReference and XRCoordinateSystem

NellWaliczek:
Hey folks!

As promised this is a painfully long PR with the intention of addressing most of the confusion and concerns discussed at the Sept `18 F2F.  It also addresses issue #396, issue #389, issue #367, issue #355, and supersedes PR #358.  This PR does not address issue #384 and issue #403, though I imagine it will impact how we'll approach solving them as well.  

It's also worth pointing out that because we haven't officially agreed on whether or not `XRAnchor` should be part of the core of WebXR 1.0, I made a strong effort not to reference the concept in my descriptions.  However, it was fairly challenging to explain the appropriate usage of `XRUnboundedFrameOfReference`, the polyfill behavior for `XRBoundedFrameOfReference`, and the purpose of `XRCoordinateSystem` without doing so.  That's probably a pretty strong sign that we should be making that decision, one way or another, very soon.  Perhaps at TPAC?

Also, as Brandon mentioned on the Oct 2 call, it's been challenging for us to divide up the design work we've been doing into manageable size chunks.  Given the scope of this PR already, there are a handful of changes I have been contemplating but explicitly didn't include.  These changes would build on top of the main proposal and are candidates for additional PRs once this has been merged.  Please keep them in mind, but let's do our best to limit comments and feedback to the actual PR that's been posted for the time being.
1. Bikeshedding.  The name `XRFrameOfReference` seems appallingly bad, but is originally my fault sooooo.....  Recently, I'd been leaning strongly towards `XRStage`, but got feedback that this would be in conflict with the use of the term "stage" in OpenXR.  Other options that might be conflicting with OpenXR are `XRSpace` or `XRSpatialRoot`. Also, `XRStationaryFrameOfReferenceSubtype`.... eep.  In any case, because this doesn't seem to have a straightforward answer, I'd like handle bikeshedding the type names after we agree on an API shape.
1. In the past, we've toyed with adding an attribute for `XRSession.defaultFrameOfReference` which would allow the `getDevicePose` and `getInputPose` to make the `XRFrameOfReference` parameter optional.  Based on the current proposal, that seems like a solid idea.
1. There's an existing issue in the privacy-and-security repo regarding the safety of even returning 3DOF pose to an inline session without an indication of user intent.  When that issue has a concrete mitigation proposal, this document's section on Inline Sessions will need to be updated.
1. I have left out a mechanism for getting an explicitly body-locked coordinate system.  I explored the possibility of adding an additional subtype to `XRStationaryReferenceFrame`, but there were a few potentially compelling reasons to have it be a separate type that is not an `XRFrameOfReference`.  I feel strongly that this needs to be included in some form, but, looking at both approaches, it was clear that either approach was going to be additive to the current design.  As such, it seemed like it would be alright to address it in a subsequent PR.
1. It seems like we may be missing an additional frame of reference roughly called `XRDioramaFrameOfReference` to be available in inline sessions for use on devices like zspace or for 2D browsing within an immersive headset.  We need to resolve the open issues about privacy/security and pose data for inline sessions first.  Also, there's some quirks about how this relates to stereoscopy that make it not exactly a straightforward design.  I strongly suspect we're going to need this type, but didn't want to hold up the main PR on it.
1. I didn't include any details about how to report or handle tracking loss.  There seem to be some differing opinions on how, or even if, developers should be notified of this state.  It's absolutely something that we need closure on, but I didn't want to hold up the larger PR for it.
1. I'm hearing feedback that there is a desire to make `boundsGeometry` an axis-aligned XZ rectangle rather than arbitrary geometry.  For now, the text in the PR is mostly just a transfer from the main explainer with a few tweaks to reflect the differences in the shape of `XRFrameOfReference`
1. I suspect that the `requiredFrameOfReference` option in session creation may need to become an array.  I'm leaving it for now, but it seems likely that will be needed to allow developers to swap between `bounded` and `unbounded` frames of reference effectively.
1. I'm not a huge fan of the `XRInputPose.emulatedPosition` property, but I left it as-is.  And I'm not really sure how I'd replace it yet.  So....
1. As I mentioned on the Oct 2 call, the main explainer document is so long it's become unwieldy.  In addition to pulling the tracking, coordinate systems, frame of reference, etc. content into a new document, I'd also like to extract all input related content into its own explainer document.  Along with this, I think we're going to need to figure out a better way of merging/maintaining the partial webidl. @toji has graciously offered to look into how we'd do this.

Yep, that's a pretty long list of follow-up topics.  However, step one is first getting closure on whether or not this PR is the right overall approach.  I'm (obviously) hopeful that the answer is yes, especially since it's based on all the things I heard people mention (confusion, requirements, epiphanies, etc) at our last F2F in September.

Thanks, folks!

p.s. I'm REALLY sorry for dropping such a large document all at once.  I know that makes it challenging to review.  I just wasn't sure how else to scope it down...


explainer.md
```diff
@@ -1034,8 +913,6 @@ dictionary XRSessionCreationOptions {
   attribute EventHandler onselectend;
   attribute EventHandler oninputsourceschange;
 
-  Promise<XRFrameOfReference> requestFrameOfReference(XRFrameOfReferenceType type, optional XRFrameOfReferenceOptions options);
```
NellWaliczek: Missed removing `onresetpose`

TrevorFSmith:
This is solid work and I believe will avoid much of the confusion that the existing spec would have created.

One aspect that I'd like to consider is how avoid the pattern where devs nest a series of async calls just to find the best possible `XRFrameOfReference`. Most apps will know at the start which types of FoRs that are acceptable so we could let them make a single async call to fetch the best match.

We could use the nifty new [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) syntax to make `XRSession.requestFrameOfReference` capable of taking multiple parameters of acceptable type/subtype parameters like so:

xrSession.requestFrameOfReference(
  {
    type: 'unbounded', sub-type: 'floor-level'
  },
  {
    type: 'bounded', sub-type: 'floor-level'
  },
  {
    type: 'stationary', sub-type: 'eye-level'
  }
).then(frameOfReference => {
  /* app is responsible for checking the type and subtype of frameOfReference */
})

So, it would be the exact same API as in this PR but with the additional capability of indicating that more than one FoR is acceptable.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+
+[SecureContext, Exposed=Window] interface XRCoordinateSystem {
+  Float32Array? getTransformTo(XRCoordinateSystem other);
+};
+
+//
+// Frame of Reference
+//
+
+enum XRFrameOfReferenceType {
```
lincolnfrog: As alluded to in my previous comments, I feel that this is a huge amount of complexity of enumerations, types, subtypes, and inheritance that is unwarranted by the underlying concepts they are trying to represent.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+  readonly attribute XRRay targetRay;
+  readonly attribute Float32Array? gripMatrix;
+};
+
+//
+// Coordinate System
+//
+
+[SecureContext, Exposed=Window] interface XRCoordinateSystem {
+  Float32Array? getTransformTo(XRCoordinateSystem other);
```
lincolnfrog: Can we get a code example of how this is used?

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+  "bounded",
+  "unbounded"
+};
+
+dictionary XRFrameOfReferenceOptions {
+  XRFrameOfReferenceType type;
+};
+
+[SecureContext, Exposed=Window] interface XRFrameOfReference : EventTarget {
+  readonly attribute XRCoordinateSystem coordinateSystem;
```
lincolnfrog: I don't see why we need this - what would be an example use-case? My belief is that the only thing that we need to get out of XRFrameOfReference in the end is a view matrix. It ultimately represents the way that the device pose is being represented and calculated, which, when used to generate a camera in a virtual world, is not needed beyond that since the virtual objects in the world can exist in a simple substrate reference frame.

If the only reason for this added complexity is disconnected spaces in AR where we don't know how they relate to each other and need a way of knowing whether/how some system of objects/space should be rendered, then it seems worth exploring that as part of this explainer. As it stands, I don't see a lot of justification for this complexity and would like it to be described as a companion to this IDL.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Frames of Reference
+The WebXR Device API makes developers think upfront about the mobility needs of the experience they are building (rather than underlying tracking technology) by requring them to explicitly request an appropriate `XRFrameOfReference`.  The `XRFrameOfReference` object acts as a substrate for the XR experience being built by establishing guarantees about supported motion and providing a coordinate system in which developers can retrieve `XRDevicePose` and view matrices.  
+
+There are three types of frames of reference: `bounded`, `unbounded`, and `stationary`.  A bounded experience is one in which the user will move around their physical environment to fully interact, but will not need to travel beyond a fixed boundary defined by the XR hardware.  An unbounded experience is one in which a user is able to freely move around their physical environment and travel significant distances.  A stationary experience is one which does not require the user to move around in space, and includes "seated" or "standing" experiences.
```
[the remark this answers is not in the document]
toji: Nell actually had that ordering in a prior version of this document and was asked to change it by someone else. ¯\_(ツ)_/¯ I suspect at some level the order that feels more natural here likely depends on your prior experience in this space.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+  let pointB = xrFrameOfReference.boundsGeometry[0];
+  boundsMesh.addQuad(
+        pointA.x, 0, pointA.z, // Quad Corner 1
+        pointB.x, 2.0, pointB.z) // Quad Corner 2
+  }
+}
+```
+
+### Unbounded Frame of Reference
+A _unbounded_ experience is one in which the user is able to freely move around their physical environment. An example of this sort of experience would be a campus tour or a whole-house renovation preview. These experiences explicitly require that the user be unbounded in their ability to walk around, and the unbounded frame of reference will adjust its origin as needed to maintain optimal stability for the user, even if the user walks many meters from the origin. In doing so, the origin may drift from its original physical location.
```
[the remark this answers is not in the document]
toji: I would expect that large-scale, untethered VR experiences (like those that [The VOID](https://www.thevoid.com/) does) could be reasonably handled with an `unbounded` FoR. It's definitely easier to think about in terms of AR, though.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+    xrFrameOfReference = frameOfReference;
+  })
+  .then(setupWebGLLayer)
+  .then(() => {
+    xrSession.requestAnimationFrame(onDrawFrame);
+  });
+}
+```
+
+As developers start building experiences with the potential to span kilometers, they will also need to deal with the fact that floating point precision issues will be introduced.  When this is close to occurring, developers should create a new `XRUnboundedFrameOfReference`,and use the `coordinateSystem.getTransformTo()` method to reparent any nearby virtual object to the new frame of reference.  However, practically speaking, large _unbounded_ experiences will often segment app/game assets and logic into discrete map segments and, depending on map segment size, this may also be an opportune time to create a new `XRUnboundedFrameOfReference`.
```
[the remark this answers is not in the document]
toji: We talked about this at the F2F a bit, and I've discussed it with Nell further after that. It seems like in the majority of cases transitions between Frames of Reference would be dictated by content rather than floating point limitations?

Consider that with our system of 1 unit = 1 meter a 32 bit float can maintain accuracy of 0.5mm or better at up to 8192 meters (~5 miles). I find it hard to imagine any content developers out there who would be developing their content to stream in 5^2 mile blocks. You will almost certainly see developers breaking up their content into much smaller pieces and initializing new Frames of Reference to align with those streamed chunks rather than transitioning between systems based on precision alone.

That said it probably wouldn't be a bad idea to offer some guidance around what we feel is a recommended distance from the origin before the developer should think about moving to a new Frame of Reference. 2048 meters (1.2 miles) would be a pretty safe, conservative value IMO, given that precision at that point is still 0.1mm? Would appreciate some comments from platform owners on what they consider the minimum acceptable threshold for tracking accuracy.

(Floating point precision numbers taken from the [table on this page.](https://blog.demofox.org/2017/11/21/floating-point-precision/))

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: Understood, not an important change for me.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+
+### Stationary Frame of Reference
+A _stationary_ experience is one which does not require the user to move around in space.  This includes several categories of experiences that developers are commonly building today.  "Standing" experiences can be created by passing the `floor-level` subtype.  "Seated" experiences can be created by passing the `eye-level` subtype.  Orientation-only experiences such as 360 photo/video viewers can be created by passing the `position-disabled` subtype.
```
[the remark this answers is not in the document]
thetuvix: The approach in #396 proposes having `XRSession.getViewMatrix(frameOfReferenceType, offsetTransform)` instead, with just `"head"` and `"world"` (`"unbounded"` here) as the frame types.  In that model, an app would get their device poses relative to a `"bounded"` origin by fetching the stage bounds origin relative to the `"world"` origin separately, and then pass that offset in as the `offsetTransform` each frame.

One key gotcha for app developers in that model is that the stage bounds will have an ever-adjusting relationship to the `"unbounded"` origin on inside-out systems, as that origin adjusts to maintain stability near the user.  This is similar to how plane origins and mesh origins will continually adjust relative to the `"unbounded"` origin each frame as the system continues to learn.  Therefore, the only correct way to use `XRSession.getViewMatrix(frameOfReferenceType, offsetTransform)` for bounded experiences on inside-out devices would be to fetch the stage bounds origin transform every frame, and then keep updating the `offsetTransform` - that step is subtle and likely to be forgotten by many developers, especially those who test their apps on outside-in tracking systems.

Dealing with multiple independently-adjusting anchors is inherent complexity for unbounded experiences that rely on various planes and meshes, but this is not something that room-scale VR developers need to deal with today.  A key goal of the `"bounded"` frame is to give room-scale VR developers a set-it-and-forget-it frame that provides the same benefit of a single coordinate system in which to reason about device/input poses that developers have today in the Oculus SDK, OpenVR SDK and Windows MR SDK.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: See the discussion in https://github.com/immersive-web/anchors/issues/8#issuecomment-427473938.  As shipping today, developers doing desktop VR with Windows MR headsets with inside-out tracking (same as HoloLens) can benefit from independently-adjusting free-space anchors when their users are using backpack PCs to do `"unbounded"` experiences.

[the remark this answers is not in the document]
thetuvix: That said, that's more an argument not to bake "AR" as a term into the name of the frame, moreso than a reason to avoid mentioning AR examples in prose for clarity.  It seems perfectly fine to mention ARCore phones and HoloLens as evocative examples of devices/platforms where users can wander for many 10s/100s of meters.

spatial-tracking-explainer.md
NellWaliczek: Yeah, that's absolutely one example.  I'm not certain how common it will be, but it's worth adding to the explanation.  I'll probably put it in the section titled "Transitioning between XRFrameOfReferences"

(Another reason is actually that @toji is working on a design proposal that would make a single session capable of switching between `inline` and `immersive`.  In that situation, there will definitely need to be switching between frames of reference.)

spatial-tracking-explainer.md
NellWaliczek: Yeah, this is a tricky one!  I *almost* left out the `XRCoordinateSystem` type entirely in this PR because it is pretty meaningless until we have anchors and similar types.  (which is why that's next on my list!)  The main reason I left it was that I didn't want to give the impression that it was being removed permanently when most of the rest of the PR was aiming for parity (but with better clarity).

That said, since a few folks have missed it when reading...it's worth noting two things:
1. `XRFrameOfReference` no longer inherits from `XRCoordinateSystem`.  It has a member instead.  This pattern is one that seems like it will adapt better to the inclusion of anchors, meshes, and the like.  And 
1. The `getDevicePose()` and `getInputPose()` no longer take `XRCoordinateSystem` objects, but instead take `XRFrameOfReference` objects.  This more strongly correlates the mental model that these poses are only really relevant in the context of the `XRFrameOfReference` being used for rendering.

All that said... to answer your questions:
1. It should be possible so long as the two `XRFrameOfReference` objects are able to be related to one another at the moment the transition is desired to happen.  For example, if I put a bag over my head, walked down the hall, and created a new instance... well those probably won't be relatable until I walk back down the hallway bag-less.
1. I'm not entirely sure what you mean by "Would it make more sense to have the content from the old type be attached to a XRCoordinateSystem that came from the difference between the two types?" Are you suggesting an `XRCoordinateSystem` can be "new'd" up?
1. Yes ... but it will involve anchors ^_^;;;  And we're back to the problem of needing to put a cut line *somehwere* to get a PR out the door.  I'll be thrilled to get the anchors-related examples up because it'll be a much clearer explanation.

thetuvix: For question 3, the type that would represent a plane, mesh, free-space anchor or other tracked element would provide an (or be an) `XRCoordinateSystem`, allowing you to relate its pose to any frame of reference or any other such element.

We could alternatively give all such world entities a `.getPose` method that returns its current transform in a developer's preferred frame of reference, without having an `XRCoordinateSystem` for each such object.  However, once you give that method an `XRCoordinateSystem` parameter so the developer can specify the base frame they want, you've got the same signature as `.getTransformTo` again, just one type up.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+
+The `XRBoundedFrameOfReference` also reports geometry within which the application should try to ensure that all content the user needs to interact with can be reached. This polygonal boundary represents a loop of points at the edges of the safe space. The points are given in a clockwise order as viewed from above, looking towards the negative end of the Y axis. The shape it describes is not guaranteed to be convex. The values reported are relative to the frame of reference origin, and must have a `y` value of `0` and a `w` value of `1`.
+
+```js
+// Demonstrated here using a fictional 3D library to simplify the example code.
+function createBoundsMesh() {
+  boundsMesh.clear();
+  
+  // Visualize the bounds geometry as 2 meter high quads
+  let pointCount = xrFrameOfReference.boundsGeometry.length;
```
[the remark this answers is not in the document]
NellWaliczek: Yeah, I know there's an inclination to move bounds to the session, but that would create a different set of problems.  @thetuvix had posted a reply to a question that's no longer there, that I feel like touches on some of the weirdness:

> The approach in #396 proposes having XRSession.getViewMatrix(frameOfReferenceType, offsetTransform) instead, with just "head" and "world" ("unbounded" here) as the frame types. In that model, an app would get their device poses relative to a "bounded" origin by fetching the stage bounds origin relative to the "world" origin separately, and then pass that offset in as the offsetTransform each frame.
>
> One key gotcha for app developers in that model is that the stage bounds will have an ever-adjusting relationship to the "unbounded" origin on inside-out systems, as that origin adjusts to maintain stability near the user. This is similar to how plane origins and mesh origins will continually adjust relative to the "unbounded" origin each frame as the system continues to learn. Therefore, the only correct way to use XRSession.getViewMatrix(frameOfReferenceType, offsetTransform) for bounded experiences on inside-out devices would be to fetch the stage bounds origin transform every frame, and then keep updating the offsetTransform - that step is subtle and likely to be forgotten by many developers, especially those who test their apps on outside-in tracking systems.
>
> Dealing with multiple independently-adjusting anchors is inherent complexity for unbounded experiences that rely on various planes and meshes, but this is not something that room-scale VR developers need to deal with today. A key goal of the "bounded" frame is to give room-scale VR developers a set-it-and-forget-it frame that provides the same benefit of a single coordinate system in which to reason about device/input poses that developers have today in the Oculus SDK, OpenVR SDK and Windows MR SDK.

In general, my approach was that any design which decouples the bounds from an origin is going to make trouble for developers in the long run.

spatial-tracking-explainer.md
thetuvix: If you can locate a plane/mesh/etc.'s `XRCoordinateSystem` relative to your base frame at the moment, you can use that model transform to render the object this frame.  If not, you should hide the object this frame.

You're correct in noting that some of the concepts in the doc (`"unbounded"` frame and `XRCoordinateSystem`) are poorly motivated at the moment, as we have yet to introduce multiple independent anchors to the API.  Some of this complexity is scaffolding for those future APIs.

NellWaliczek: @thetuvix I think Max's question is related to the use of a single `XRUnboundedFrameOfReference`.  When you're in room A and move to room B by going down a long hallway... what is your transform when you get tracking back in room B.

thetuvix: > When you're in room A and move to room B by going down a long hallway... what is your transform when you get tracking back in room B.

There is a separate follow-up here where we need to decide for WebXR what common tracking loss/recovery behavior to establish as a developer promise for the various frames

At least in Windows MR, when you lose tracking, the WinMR equivalent of an `XRUnboundedFrameOfReference` will no longer provide you a device position.  When you regain tracking:
* If the system finds you somewhere back within the previous map fragment, your device position in the `XRUnboundedFrameOfReference` will jump there, keeping the origin in place.
* If the system _doesn't_ find you somewhere within the previous map fragment, your device position will stay at the same XYZ position in the `XRUnboundedFrameOfReference`, which has the effect of having "dragged" the unbounded origin along with you during the period of tracking loss.

Regardless, when you are in either map fragment, if you ask for a transform from a free-space anchor in the other fragment to your unbounded frame, you'll get back null.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: Understood, this explanation from @thetuvix is amazing. Thanks for providing clarity on this point.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: I don't have a strong preference here but my inclination is that more use-cases and examples help clarify these sorts of API decisions - I know they help me a lot!

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: Brandon - you bring up a good point about 5 miles being a lot, but just to be complete in our thinking I am going to ignore that for a second and assume that we want to support someone hiking across the country wearing AR glasses.

Maybe I am missing something here but I don't see how the runtime can prevent its own coordinate system from leaking to the app. If the app doesn't request a new frame of reference and the underlying system has to remap to avoid a floating point precision issue, then it is going to run into that same precision issue when it tries to convert its newly-remapped internal vectors into the app's frame-of-reference vectors, which have not been remapped. Those app vectors will be too large when being converted, right? I may be missing something...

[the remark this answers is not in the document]
NellWaliczek: Maybe this ends up being related to the representation of tracking loss?

spatial-tracking-explainer.md
NellWaliczek: Do you happen to have a table with the tracking loss behavior for many of the "major" systems?  

Also, as we dig into the handling of tracking loss, perhaps this is a case for adding the `onreset` event to the `XRUnboundedFrameOfReference`?

lincolnfrog: Does this mean you have to know that you need to request a new frame-of-reference when you get tracking back? Is there an event-driven API for this? I know Nell said she isn't covering tracking loss in this explainer, so maybe this detail doesn't belong here but there is definitely some overlap with respect to how coordinate systems are used to render virtual objects. It is still not clear to me how an app would use these APIs to deal with this situation.

NellWaliczek: Yeah, reading over all the comments this morning... I think we probably do need to address tracking loss before trying to merge this PR.  I'll dig into it

spatial-tracking-explainer.md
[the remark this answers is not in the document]
blairmacintyre: I'm curious how WindowsMR devices deal with this now, when the user creates a safe area?  I haven't looked at the SDK, but do the bounds change each frame?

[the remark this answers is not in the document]
blairmacintyre: Regarding the terminology, couldn't we use sub-types? XRFrameOfReference is generic; XRBoundedStageFrame is a subtype that adds the bounds, or anything else needed?

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+
+The origin of this type will be initialized at a position on the floor for which a boundary can be provided to the app, defining an empty region where it is safe for the user to move around. The y value will be 0 at floor level, while the exact x, z, and orientation values will be initialized based on the conventions of the underlying platform for room-scale experiences. Platforms where the user defines a fixed room-scale origin and boundary may initialize the remaining values to match the room-scale origin. Users with fixed-origin systems are familiar with this behavior, however developers may choose to be extra resilient to this situation by building UI to guide users back to the origin if they are too far away. Platforms that generally allow for unbounded movement may display UI to the user during the asynchronous request, asking them to define or confirm such a floor-level boundary near the user's current location.
```
blairmacintyre: How are we dealing with non-flat floors?  It seems reasonable to assume (perhaps) that bounded frames have a flat floor;  are we going to assume (perhaps telling users) that if they defined a bounded frame for an unbounded device, it should have a flat floor?  

It's probably "fine" to assume this;  the failure case is that users try to use a bounded experience somewhere where the floor isn't flat, and they can't reach some content (because it's too high, or it's under the real floor).

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: Indeed!  Even assuming a "perfect tracking system" that has 0 tracking error and rebases its internal coordinate system every frame to a nearby origin, the user will _still_ at some point walk far enough that a distant app-created `"unbounded"` frame origin will be a cause of floating-point error when device poses are expressed in that origin.

It is therefore up to the app to absorb some reset of its `"unbounded"` frame origin to mitigate floating-point error if the device moves some number of kilometers from the origin.  The key question is whether the runtime signals a forced reset, or if the app chooses when to reset.

Since this floating-point error manifests regardless of tracking system, WebXR middleware can be designed generically to handle it at the app layer:
* An app could use a naïve strategy, just resetting its `"unbounded"` frame when the user walks some fixed distance from the origin (e.g. 1024 meters).
* Alternatively, the app could have more involved reset behavior at app-meaningful chunk boundaries, as proposed by @toji above.  An app choosing this approach benefits from being in control of the origin reset, rather than having it also forced by the runtime at arbitrary moments that differ across UA or underlying XR platform.

Presuming the app is doing one of the above, it is then up to the UA or underlying XR platform to decide how often to rebase its own internal coordinate systems, so that such well-behaved apps will not observe floating-point error.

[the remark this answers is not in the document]
blairmacintyre: I would tend to agree with @thetuvix ... in unbounded frames, the coordinate system is not fixed, so the apps should be requesting to get values relative to some `XRFrameOfReference` each frame.  When the platform feels the system has moved "far enough", it can recenter.  

The whole point of requiring that all content be anchored to some FoR or CoordinateSystem is that the "current euclidean coordinate system" is only valid for a single frame;  aside from that, the coordinate system needs to be adjusted to whatever the platform says.

As has been discussed elsewhere (perhaps in the anchor repo?) this will have implications for things like managing physics.  But those can probably be abstracted once in an "XRPhysics" abstraction layer on a physics library.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: When designing APIs, we are all often compelled to be more forward-looking than appears required by current hardware, which can sometimes lead to confusion for today's developers targeting today's hardware.

I've generally found documentation to be a great release valve for that tension.  We can be very specific to today's hardware in how we describe things in docs, since we can trivially update that doc prose later as the hardware landscape changes - we can't make that kind of freeform change to the API shape itself.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
blairmacintyre: And, following @thetuvix's comment above mine (which came in as I was typing mine!), I will say that in Argon we did the reset for the user, and sent an event when the origin was reset.  We could do similarly here; have an event on an `unbounded` FoR that users could subscribe to if they wanted to know things had been rebased.  We used this to do just what he suggests: adjust all the anchored content.  The details inside the physics engine can also be adjusted there.

This will be important when geospatial content is used, as the geospatial frame should remain reasonably close to the user, since it's typically a tangent plane to the earth centered at some location.  As the user moves away from it, it gets less accurate an approximation to "the user".

spatial-tracking-explainer.md
thetuvix: @lincolnfrog:
> Does this mean you have to know that you need to request a new frame-of-reference when you get tracking back?

No - in Windows MR, the one unbounded frame of reference object you created at app startup continues to work after tracking loss, with the behavior described by the bullets above.  I've been presuming the same approach in WebXR.

We can perhaps separate two things here:
* Documenting the fact that all `XRFrameOfReference` instances are durable objects across tracking loss and recovery.
* Nailing down the precise details of that tracking loss/recovery behavior, such as the positions and orientations you get back after tracking recovery in various situations.

We could perhaps update this PR to document the former, while iterating on the latter in a follow-on PR.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+}
+```
+
+## Practical Usage Guidelines
+
+### Inline Sessions
+Inline sessions, by definition, do not require a user gesture or user permission to create, and as a result there must be strong limitations on the pose data that can be reported for privacy and security reasons.  Requests for an `XRBoundedFrameOfReference` or an `XRUnboundedFrameOfReference` will always be rejected on inline sessions.  Requests for `XRStationaryFrameOfReference` may succeed, but may also be rejected if the UA is unable provide any tracking information such as for an inline session on a desktop PC or a 2D browser window in a headset.
+
+Additionally, some inline experiences explicitly want tracking information disabled, such as a furniture viewer that will use pointer data to rotate the furniture.  Under both of these situations, there is no `XRFrameOfReference` needed.  Passing `null` as the `XRFrameOfReference` parameter to `getDevicePose()` or `getInputPose()` will return pose data whose values are based on the identity matrix.  Developers are then free to multiply in whatever transform they choose.
+
```
blairmacintyre: Given the threat vectors that caused the https://www.w3.org/TR/orientation-event/ API to never be approved, I am still not inclined to have access to any XR device tracking, even orientation, without some form of user approval or awareness.  I don't see how providing orientation data is going to be any different than the previous `device-orientation` data, in terms of threats;  if apps can get this without user permission, they can open WebXR on any web page and monitor the device's motion.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+  xrDevice.requestSession({ immersive: true,  requiredFrameOfReferenceType:'unbounded' })
+      .then(onSessionStarted)
+      .catch(err => {
+        // Error will indicate required frame of reference type unavailable
+      });
+}
+```
+
+### Floor Alignment Compatibility
+Some XR hardware with inside-out tracking has users establish "known spaces" that can be used to easily provide `XRBoundedFrameOfReference` and the `floor-level` subtype of `XRStationaryFrameOfReference`.  On inside-out XR hardware which does not intrinsically provide these known spaces, the User Agent may respond to the promise request by presenting UI for selecting a floor-aligned origin.  If the User Agent chooses not to do this, it will reject the request for floor aligned `XRFrameOfReference` objects with an error indicating the reason.  In response, a polyfill can be used to provide the same behavior by using another `XRFrameOfReference` type to draw UI for selecting a floor origin.  If anchors are available to the polyfill, it could then place an anchor at the user's selected floor point.
```
blairmacintyre: What is the rationale for not requiring a US to provide a floor-aligned origin?  Seems like this greatly complicates user code?  Perhaps there's a way for the UA to say "it's just a guess" but at least be required to give a guess, so simple apps can just "get on with it"?

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+xrFrameOfReference.addEventListener('reset', xrFrameOfReferenceEvent => {
+  // For an app that allows artificial Yaw rotation, this would be a perfect
+  // time to reset that.
+  resetYawTransform();
+
+  // For an app using the XRBoundedFrameOfReference, this would be a perfect time to
+  // reset the bounds and potentially re-layout content intended to be reachable within
+  // the bounds
+  createBoundsMesh();
+});
```
blairmacintyre: In the spirit of "onReset", should there be an event available on the FoR to tell programmers the state of the tracking system?  So, if the tracking system is temporarily in a bad state, the app can react reasonably?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Hey Blair, I'm not entirely sure what you're asking to see changed?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: I tried to strike a balance and have added a few examples to each section.  I'm wondering if it would also be helpful to make a flow chart that developers can use to determine which Frame Of Reference to use?  Maybe someone with better art skills than me wants to do that?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: So, I'm not quite following along... are we saying that the developers need to monitor and handle when the floating point error gets too high?  Or that WebXR will do it and issue an `onreset`?

FWIW, I just moved `onreset` to the `XRFrameOfReferenceType` for another reason.

spatial-tracking-explainer.md
NellWaliczek: Added an example that'll show up in my next push

spatial-tracking-explainer.md
NellWaliczek: @lincolnfrog Not sure if my question above got lost in the deluge of PR comments.  What were you meaning by 

>Would it make more sense to have the content from the old type be attached to a XRCoordinateSystem that came from the difference between the two types?

Are you suggesting an XRCoordinateSystem can be "new'd" up?

spatial-tracking-explainer.md
NellWaliczek: I've moved the `onreset` to the `XRFrameOfReference` base interface, but I still think the broader topic of tracking loss may need to be addressed before we can credibly merge this PR.

spatial-tracking-explainer.md
NellWaliczek: Yes, that's actually one of the simplifying assumptions developers will be able to make; `XRBoundedFrameOfReference` objects are expected to have a flat floor within the bounds.

spatial-tracking-explainer.md
NellWaliczek: Yeah, absolutely.  That said, the privacy-and-security repo is still working on guidance for what speedbumps, etc. are needed before providing tracking data (and at what granularity, say for diorama-style inline rendering) to the page.  As such, I've migrated the existing behavior into this new design as is.  One of the bullets in the PR description covers filing an issue to address this once the privacy-and-security document is completed.

spatial-tracking-explainer.md
NellWaliczek: We could do this, but I'd really like to get positive confirmation from all the hardware/UA folks that this is something they'd be willing to commit to.  Let's track this in a separate issue once the main PR is merged?

spatial-tracking-explainer.md
NellWaliczek: Yeah, originally I'd been hoping to merge this PR without needing to address tracking loss first (it's in the list of post-merge topics in the PR description).  The reason is that it's somewhat a thorny problem when you have to take into consideration all the different hardware behaviors.  However, I'm sufficiently convinced that the solution is too fundamental to do after-the-fact.  Poking at it now.

NellWaliczek:
I've addressed a bunch of comments locally, but am hesitant to push the commits to GitHub just yet since it doesn't always do a good job of not hiding active comments if it can't find them in the new version of files.
Once I have answers to the outstanding questions in the comments already posted, I'll push the new revision.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: It sounds like he is advocating having another class called XRBoundedStageFrameOfReference that inherits from XRFrameOfReference and adds bounds information

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: Yeah, you mentioned the flow-chart idea when we discussed this and it helped me understand what you were going for in terms of the decision-making required by the app developer. +1 to creating the actual flowchart as an image - my art-fu is weak otherwise I would volunteer.

spatial-tracking-explainer.md
lincolnfrog: I guess that part of my question does assume that XRCoordinateSystems could be created arbitrarily - and that you could get the transform between the new FOR and the old FOR and then create a XRCS that represents that transform. You could then attach all your content from the old FOR to that XRCS so that it renders correctly when you get the device pose using the new FOR.

Did that paragraph make ANY sense? I may be fundamentally misunderstanding something here, so hopefully, at the very least, my confusion can help.

The ultimate point I am trying to understand here is what the approach looks like to transition virtual content from an unbounded FOR situation to a bounded FOR situation and still have it render in "approximately" the right location. I believe it is important to either say "you can't do that" or to make sure it's something we support without requiring multiple render passes where the app developer has to get two FORs and use each one in a camera to render the appropriate scene.

spatial-tracking-explainer.md
lincolnfrog: @thetuvix how does the app know what virtual objects are no longer valid (i.e. belong to the old map fragment)?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: @blairmacintyre how would that type be different from `XRBoundedFrameOfReference`?

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+let xrSession = null;
+let xrFrameOfReference = null;
+
+function onSessionStarted(session) {
+  xrSession = session;
+  xrSession.requestFrameOfReference({ type:'unbounded' })
+  .then((frameOfReference) => {
+    xrFrameOfReference = frameOfReference;
+  })
+  .catch(() => {
```
tencircles: Is the preferred method of feature detection here the rejection of a promise? I'd love to see one of either of the following. 

1. The ability to pass an array of objects into `xrSession.requestFrameOfRefrence` which would then return an `xrSession` Object with attributes indicating type and subtype.

or 

2. The ability to iterate over the FoR modes the UA supports and choose mode from there. I get that it's not ideal to give away information about the user's device prior to starting a session, but it seems like this info can be gained by doing the following in any case.

```js
function onSessionStarted (session) {
    return getFrameOfReference(session)
        .catch(() => {
            // some fallback
        })
        .then(frameOfReference => ({
            frameOfReference,
            session
        }))
        .then(setupWebGLLayer) // Some side effects here?
        .then(state => ({
            rid : session.requestAnimationFrame(onDrawFrame),
            ...state
        }));
}

async function getFrameOfReference (session, modes = [
    {type: "bounded"},
    {type: "unbounded"},
    {type: "standing", subtype: "eye-level"}
    // ...etc
]) {
    let frameOfReference;

    for (const mode of modes) {
        try {
            frameOfReference = await session.requestFrameOfReference(mode);
        } catch (e) {}
    }

    return frameOfReference ?
        Promise.resolve(frameOfReference) :
        Promise.reject(Error("no frame of reference available"));
}
```

It just seems cleaner from a dev perspective to be able to access which modes the user's device supports directly rather than having to handle a series of promises which may reject, and I think, in practice at least, most devs looking to build progressively enhanced XR experiences will end up doing something like the above.

blairmacintyre: @TrevorFSmith had also suggest the "array of possible acceptable choices", it seems elegant.  May still need to deal with failure, but that's probably a UI that says "hey, sorry, you're out of luck" that that point.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
blairmacintyre: I personally think "developers need to monitor and handle when the floating point error gets too high" is a recipe for disaster.  I think developers will want to monitor and shift the coordinate frame for more application-specific reasons (e.g., based on the distribution and layout of their content, the constraints of the frameworks they are using, etc).  

But, having the platform shift the origin of the coordinate system when appropriate to the platform (not just because of javascript numerical precision) seems appropriate.  Especially when the platform is already adjusting the coordinate systems as the user moves around, anyway, so the application layers need to deal with that.

Having an `onreset` event is useful, but optional:  "recentering" is a really large move and may require major updates to content, so it may be useful to know explicitely that this has happened.  It's easy to detect, though:  monitor one anchor and see if it's pose relative to the camera moves a large amount between frames.

spatial-tracking-explainer.md
tencircles: A method similar to `video` and `audio` element's [canPlayType](https://www.w3.org/TR/2011/WD-html5-20110113/video.html#mime-types) method would be a good analogue for feature detection (in my mind at least). This is maybe a silly question, but are there any technical limitations what require feature detection to be asynchronous?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: @NellWaliczek I don't see that type in your IDL - you added bounds information to the base XRFrameOfReference type (which seems to be the only such type in your proposed IDL)

spatial-tracking-explainer.md
lincolnfrog: Only tangentially-related to this question, but RE: promises, there is some concern that we can't use promises for many of the async APIs in WebXR due to very strict timing concerns with the resolution of the promise. Our current thinking is we will use callbacks instead. I am wondering if we should avoid using promises entirely throughout WebXR to avoid the inconsistency of sometimes using promises and sometimes using callbacks... @toji ?

cwilso: That sounds like a question for the TAG, but I suspect the answer is use promises when possible.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: It should be on line 351?

spatial-tracking-explainer.md
tencircles: I'm definitely in favor of using promises over callbacks for async functionality, and especially in favor of avoiding mixing the two. However, using promises for feature detection presents some awkward challenges to developers since the UA is going to throw in cases where is arguably shouldn't, namely feature detection. 

Let's consider the case where an application wants to support multiple FoR and serve content to users tailored to each. Ideally, the application would feature detect which FoR the user's device supports, then do one of two things. In the case where multiple FoR are supported (and where it would make sense), allow the user to select (or confirm) the FoR they want to use via an DOM UI. In the case where only one FoR is supported, simply transition into the XR view.

Given the current design, developers are forced into actually requesting each FoR, then catching an error if it isn't supported. It's certainly something that can be done easily enough, but it definitely feels a little clunky. The code I included above I think illustrates this in that there is an empty catch block, any time I see that it's a bit of a red flag for me. If the correct behavior in most cases for a catch statement is an empty block, then it really begs the question: why is the API throwing?

NellWaliczek: @tencircles Yes, I've already incorporated Trevor's suggestion to request more than one FoR at at a time and you'll see that in the next push. https://github.com/immersive-web/webxr/pull/409#issuecomment-427435288

@lincolnfrog if you want to file a generic issue to discuss the approach WebXR is taking with promises vs. callbacks, please go ahead :)  However, let's not use this PR as the place to discuss it?

@tencircles Regarding the desire to feature detect... The reason this isn't supported had to do with avoiding fingerprinting concerns.  Knowing which FoR are supported is only something that should be queryable after the user has indicated intent to use the functionality, or in the case of immersive actually given permission. That said!  @toji and I are working on a subsequent PR to this one that may actually have a path to addressing this issue.  I'll be sure to get your feedback when we have it posted.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: I somehow had different IDL than I see now - it looks like you do indeed have this type and LGTM!

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: If we want WebXR to trigger an automatic reset event to the `"unbounded"` space at some defined point (e.g. 1024 meters), we could spec that out and do that consistently across all UAs.

I want to address one key point though:

@blairmacintyre:
> Especially when the platform is already adjusting the coordinate systems as the user moves around, anyway, so the application layers need to deal with that.

An inside-out tracking system will be continuously and smoothly adjusting the `"unbounded"` origin as needed to keep coordinates stable near the user.  Every time there is a tiny shift of that coordinate system, we would not trigger a `"onreset"`.  This would be hard to define anyway - when the `"unbounded"` origin shifts relative to its anchors, did the anchors move or did the frame origin move?

Given that, there will be specific known cases where various frames of reference would trigger `"onreset"`:
* The `"stationary"` frame would trigger `"onreset"` when the user engages some platform UI to reset their "local zero" origin.  Some platforms have no such UI, and apps there would never see an `"onreset"`.
* The `"bounded"` frame would trigger `"onreset"` when the user redraws their room setup, or the platform observes the user walking closer to some second stage they've defined.  Some platforms support only a single stage and do not allow for room redefinition while an app is running, and apps there would never see an `"onreset"`.
* The `"unbounded"` frame would trigger `"onreset"` when the user walks past some spec-defined boundary where we forcibly reset the app to help avoid floating-point error, or when the origin has a large discontinuity due to tracking recovery in a new map fragment.  Some platforms may not allow users to walk far enough for floating-point error, or may only support a single map fragment, and apps there would never see an `"onreset"`.

Given the above, floating-point error caused by the app being far from the `"unbounded"` origin is unlikely to be due to platform-specific tracking behavior, since it is the job of the `"unbounded"` frame to smoothly integrate that drift over time to avoid large discontinuities.  If we do want to have UAs automatically reset the `"unbounded"` frame due to floating-point error, it therefore seems that we could help keep things consistent for apps and make that a specified feature of WebXR itself.

spatial-tracking-explainer.md
thetuvix: @lincolnfrog 

Each frame, for any given anchor:
* If a pose **is available** for that anchor in your frame of reference, you should render the anchor's content there.
* If a pose **is not available** for that anchor in your base frame of reference, you should hide the anchor's content until a pose is available again.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: @blairmacintyre, I'm going to go ahead "resolve conversation".  Please reopen if you feel this hasn't been addressed adequately!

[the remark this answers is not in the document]
blairmacintyre: fine

spatial-tracking-explainer.md
thetuvix: Agreed - I can't see a UA doing any _worse_ than if an app ran with the default emulated floor height.  A UA should have access to either its own user-provided, platform-provided (or worst case, UA default) value for user height, which should be better than the app-provided emulated height, except in special cases where the app knows the user is sitting/kneeling.

It also benefits the WebXR ecosystem to get all standing apps that will guess at the floor anyway to just use a real `"floor-level"` frame of reference, since that enables best results on devices with real floor data or a convention to shift the XZ origin for standing experiences to an alternate location.

NellWaliczek: @lincolnfrog Is this something Chrome would be comfortable doing on top of ARCore?
@rcabanier How about Helio?
@Artyom17 Oculus Browser?

(Also, general feedback from @rcarbanier and @artyom17 would be very appreciated!~)

spatial-tracking-explainer.md
NellWaliczek: Ok, I've been talked into (and reasonably so) dividing the question of tracking loss handling in half.  In this PR I'll be updating the onreset to fire when a tracking loss recovery causes a disconnected map fragment.  However, there is still a general question of WebXR's behavior *during* tracking loss.  I'll file that as a separate issue as originally planned.

spatial-tracking-explainer.md
NellWaliczek: After the IMCG call this morning, it sounds like I should update the language regarding floor-aligned FoR's to clearly state that the creation cannot cause UI elements to appear.

spatial-tracking-explainer.md
NellWaliczek: There are definitely questions about anchor culling that we'll need to take into consideration, but that should be part of a different PR ^_^

NellWaliczek: This seems like we've come to a good conclusion, and the next iteration will have (hopefully) addressed this adequately.  Please reopen if that turns out not to be the case.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
lincolnfrog: I am happy with the detail you added. Go forth and resolve!

spatial-tracking-explainer.md
rcabanier: Helio would have no problem providing a "floor-level" frame of reference

spatial-tracking-explainer.md
NellWaliczek: Chatted with Max and have adjusted some language to address this confusion.

spatial-tracking-explainer.md
NellWaliczek: Ok, i'm going to make the change.  @artyom17, yell if that's a problem!

NellWaliczek:
Pushed the latest set of changes

NellWaliczek:
FYI I rebased before pushing the latest set of changes, so it may be a bit confusing as to what's changed... To just see my changes, go here: https://github.com/immersive-web/webxr/pull/409/files/75164fc9ab1735bf7a61ba57958a7570b2631a28..f3e627db9f95e42abe72405166d4990ffe654720

NellWaliczek:
(bother... the link was missing two commits.  I've edited the link in the comment to be correct now)

NellWaliczek:
Apologies for all the git noise folks.  I fixed the earlier mistake and things should be sensible now.

spatial-tracking-explainer.md
rcabanier: @NellWaliczek I've talked to some people internally on the unbounded proposal and they believe this is a good approach and something we can provide. 
They did express some concern that this might be difficult to use for authors

explainer.md
NellWaliczek: Fixed locally

spatial-tracking-explainer.md
NellWaliczek: Hey Rik,
Good to hear that the design is implementable on Magic Leap!  I'm curious if you can share more details about the usability concern for authors?  I know that the lack of anchors in the current document iteration makes it a bit challenging to intuit the recommended unbounded usage pattern. Or is it something else?

NellWaliczek:
Hey folks!  General feedback seems to be pretty good so far.  And, assuming no new blocking concerns come up today, my inclination is to merge the PR **tomorrow by EOD** .

Obviously there's still a bunch of smaller topics that need working out, and we can track those as GitHub issues.  As part of merging the PR. I'll also file the ones I'm aware of.

spatial-tracking-explainer.md
```diff
@@ -0,0 +1,382 @@
...
+    xrFrameOfReference = frameOfReference;
+  })
+  .then(setupWebGLLayer)
+  .then(() => {
+    xrSession.requestAnimationFrame(onDrawFrame);
+  });
+}
+```
+
+#### Eye-level Subtype
```
Artyom17: I previously posted a comment about the lack of 'local' FoR and then realized that I haven't finished my reading and removed the comment :) 
So, yeah, the 'eye-level' FoR should work for user-centered UI case I described in issue https://github.com/immersive-web/webxr/issues/389. The only issue I have is the name: I like 'local' better and, according to @casseveritt it is more inlined with OpenXR names.

NellWaliczek: Sounds good!  I'll be filing a separate issue to bikeshed the type names and we can all arm wrestle over it in that thread.

spatial-tracking-explainer.md
rcabanier: I suspect it is because we are used to hiding this complexity from the user.

PR #415, opened 2018-10-12
Move input-related content into separate explainer document

NellWaliczek:
As I mentioned on the Oct 2 call, the main explainer document is so long it's become unwieldy. Now that spatial-tracking content is pulled into a separate document, it’s also time to extract all input related content into its own explainer document.

As an aside, we're going to need to figure out a better way of merging/maintaining the partial webidl. @toji has graciously offered to look into how we'd do this, but we decided not to block on that for right now.


toji:
LGTM. Thanks for splitting this out. Gonna merge now since there's no functional changes that this introduces.

Issue #394, opened 2018-09-14
WebXR Modes


toji:
Putting aside entirely for a moment the question of how `XRSession`s are created and managed, recent conversations with Chrome's Privacy and Security team have shuffled Google's perspective on what contexts AR data should be accessible from. Fortunately, in conversations thus far, most other teams have found the shift agreeable and/or closer to what they were hoping to do anyway, so there does not appear to be much controversy. It does mean, however, that we need to re-examine the various modes an `XRSession` can operate in. This issue attempts to lay out one proposal for how it might work.

XR Mode
-------

Sessions will operate with a specific “XR Mode” at any given time. The XR mode represents the way that WebXR content from that session is presented to the user. It's tempting to call this a "display mode" but I've actually come to think about this more in terms of defining how users will interact with the content rather than dictating how and where it's displayed. (I believe that [Blair has advocated for that approach](https://github.com/immersive-web/webxr/issues/388) in the past.)

The necessary modes as I see them today (names subject to bikeshedding) are:

 - **Inline**: Content shows as an embedded element on a page, and is denied access to any environmental data. This should be the “default” mode if the API shape allows for it.

    In keeping with previous discussions regarding a Unified Rendering Path (aka: [Poseless sessions](https://github.com/immersive-web/webxr/issues/367)) I would propose that inline sessions also do not expose any motion tracking data by default. Sessions with that configuration should be guaranteed to be consent-free by the UA so that developers have a predictable baseline to start from, ensuring that content can consistently be rendered with a single draw loop no matter how the `XRSession` is configured. Inline sessions should be accessible without user activation or frame focus.

    Motion tracking data (such as orientation for phones or actual head-relative-to-page tracking for zSpace-like devices) should still be requestable, likely as a “feature”, which would allow the UA to request user consent if desired. Frame focus would likely be required. Even in the absence of head tracking, the nature of WebXR's render loop allows this content to be rendered in stereo if applicable, providing another benefit to using this mode over a simple WebGL canvas.

    This mode should be restricted to Simple Input Model only, with no controller tracking. (Interactions with an in-XR browser using a controller would be treated as traditional pointer input against the canvas, which is then converted into simple “select” events)

 - **Immersive VR**: Effectively the same as the immersive mode we have today. The term is under discussion but I'll continue using it here since everyone by now understands how it's used in the spec. This mode would enable controller tracking and the Advanced Input Model in addition to the Simple Input Model. Only accessible with user activation and frame focus.

[...]

PR #419, opened 2018-10-13
Introduce XRSessionMode enum, describe immersive AR mode.

toji:
This PR switches the inline/immersive mode selection from a boolean to a full enum and in the process introduces "Immersive AR" mode, which covers both phone and headset AR.

I fully expect that the naming I've chosen here will be a source of some controversy, but after a lot of thought I really do feel like something along these lines, that explicitly uses the terms "VR/AR" (or at least "virtual/augmented"), is the best for users. It's certainly true that it's a gross simplification from our standpoint as API creators and platform owners, but the truth is that nobody is going to start looking at our documentation saying "I want to build a app with an optically transparent headset that has access to environmental data!" No, they're gonna say "I want to build an AR app. Where do I start?" and as such it's in our and their best interest that we have something that is readily recognizable as the thing they want regardless of the minutia of how our specific API defines that.

Otherwise this PR is pretty straightforward with a bunch of related cleanups and search-and-replace tweaks to verbiage. (Down with Magic Window! All hail Inline!)

**Personal note:** I would love to see this start collecting feedback, but due to family plans I'm unlikely to be able to respond to any of it until late next week. Please don't assume I'm ignoring your comments!

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -61,7 +61,7 @@ The basic steps most WebXR applications will go through are:
 
 The UA will identify an available physical unit of XR hardware that can present imagery to the user, referred to here as an "XR device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung Gear VR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices. Any queries for XR capabilities or functionality are implicitly made against this device.
 
-> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to expose. The UA is allowed to use any criteria it wishes to select which device is used, including settings UI that allows users to manage device priority. Calling `navigator.xr.supportsSession` or `navigator.xr.requestSession` with `{ immersive: false }` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
+> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to expose. The UA is allowed to use any criteria it wishes to select which device is used, including settings UI that allows users to manage device priority. Calling `navigator.xr.supportsSession` or `navigator.xr.requestSession` with `{ mode: 'inline' }` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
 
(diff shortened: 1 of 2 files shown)


toji:
And here I figured I'd at least be comments on the enum names. 🤷‍♀️ 

I do encourage anyone that will be participating in TPAC this week to at least read through this PR prior to the F2F meetings on Thursday/Friday. I would definitely like to get this issue wrapped up by end of week.

tvp-jwheeler:
v

toji:
Forgot to mention: this partially addresses #394.

Discussed during today's IWWG call. There's some concerns about making sure this interacts naturally with the need for feature permission requests, which are still in-progress, and there's still the potential for some bikeshedding. That aside, though, the sentiment expressed was that we could merge this for now to keep things moving forward with the understanding that it may still need some revision as the rest of the spec finalizes. Doing so now.

sjpt:
I'm not sure if this is the correct place to post this; apologies if not.  When using WebVR in exhibitions it is often needed to show the best possible VR experience to a VR viewer, and also show on a monitor (typically projection) a good as possible a derivative view for the other visitors.  This was straightforward in WebVR, but it is very unclear how this can be achieved with the current WebXR spec.

toji:
@sjpt: It's not really the "right" place, but I don't have a better one to point you at for the moment and I'm happy to answer here, so no worries!

In order to simplify the spec and make it easier for developers to understand in the initial spec we dropped the in-page mirroring feature. It may return at some point, but it'll be more of an advanced feature that developers will have to go to some effort to enable, rather than the mirroring "side effect" that made it almost accidental in WebVR.

One of the driving factors behind that decision, though, was the fact that almost all VR platforms offer a native way (and frequently better way) of providing mirroring themselves. Oculus Rift, Windows Mixed Reality, and Steam VR all have mirroring utilities you can use to display the in-headset view. For mobile or standalone devices, WebVR-style mirroring never worked anyway but they also frequently provide casting capabilities to either a phone or a device like Chromecast. There's pros and cons to each approach, but in general we felt it would be better to allow each device to make use of it's native mirroring capabilities rather than have developers rely on an API mechanism which has the potential to reduce performance and add complexity while not actually providing any functionality on a significant number of devices.

sjpt:
@toji Thank you for the response.  I have been shown by klausw on Slack the correct way to do it (by appropriate use of **renderer.setFramebuffer(NULL)**; also with comments from you).  I accept there are potential performance hits, but relatively easy if you know how.  I haven't seen that we can use the Steam mirroring for exactly what we need.  We render an oval for each eye to save wasted render time. and our spectator mirror cuts a rectangle from within that oval; loss of resolution but adequate.

Specs are written to define exactly what will happen, but are usually (for good reason) quite incomprehensible as to how to make things particular things happen.  WebXR doesn't seem to be quite at the stage yet where there are comprehensive comprehensible guides

Issue #412, opened 2018-10-09
XRFrameOfReference needed for Projection Matrices on some backends?


toji:
From the OpenXR SIGGRAPH presentation (https://www.khronos.org/assets/uploads/developers/library/2018-siggraph/04-OpenXR-SIGGRAPH_Aug2018.pdf) on Page 43 - "Viewport Projections"

In this slide we can see that OpenXR (as it stands now) has a function called `xrGetViewportProjections` that returns more-or-less everything that we report in our `XRFrame`. That includes the information necessary to build view and projection matrices. Because it's returning view information it's easy to surmise that the equivalent of an `XRFrameOfReference` (what OpenXR calls an `XRSpace`) needs to be provided to the function in order for it to return the right information. This also means, however, that _projection matrices_ aren't available until the point the `XRFrameOfReference` is passed in, either.

This is a point of incompatibility with WebXR as it stands today. Currently the `XRFrame` reports an array of `XRView`s prior to any frame of reference being specified, which in turn contains projection matrices as a property. With a lot of native backends that seems fine, as projection parameters are frequently treated as static. Despite that, it's not too surprising to see some APIs may want to combine it into the same function call that gets the other space-dependent frame rendering data. In any case, we definitely want to address a known incompatibility with a native API.

There's two straightforward ways that I can see to address this:

Move view info into the call that takes a FrameOfRef
--------------------

Probably would justify a method rename, but in essence we'd simply be moving the views array under what is today the `getDevicePose` call:

IDL
```webidl
interface XRFrame {
  readonly attribute XRSession session;
  // No more views array here

  XRDevicePose? getDevicePose(XRFrameOfReference frameOfRef);
  XRInputPose? getInputPose(XRInputSource inputSource, XRFrameOfReference frameOfRef);
};

interface XRView {
  readonly attribute XREye eye;
  readonly attribute Float32Array projectionMatrix;
  readonly attribute Float32Array viewMatrix; // View matrix just becomes another property of the XRView
};

interface XRDevicePose {
  readonly attribute boolean emulatedPosition;
  readonly attribute Float32Array poseModelMatrix;
  readonly attribute FrozenArray<XRView> views; // Views are now here
};
```

And the render loop we describe in the explainer doesn't change too much:

```js
function onDrawFrame(timestamp, xrFrame) {
  let pose = xrFrame.getDevicePose(xrFrameOfRef);
[...]

PR #422, opened 2018-10-30
Move Views array to Pose

toji:
Addresses the core issue of #412, though there's room for a follow up if we decide that having a default frame of reference is a good idea.

This PR makes the relatively simple change of moving the `views` array off of the `XRFrame` and under the queried pose, so that projection matricies that take into account the frame of reference if necessary. In my opinion this also simplifies the common case code a little bit and leaves less questions about why the various matrices are spread across the various interfaces instead of all in once place.

I've also taken the opportunity to rename the pose from `XRDevicePose` to `XRViewerPose`, taking a page from [@thetuvix's suggestion](https://github.com/immersive-web/webxr/issues/412#issuecomment-429583278). I'm happy to take additional naming suggestions, but "Viewer" feels appropriate because it doesn't ascribe the pose to a specific piece of hardware, as that may not be accurate in some cases (CAVE/zSpace, for example). Instead it suggests that in all cases the pose is associated with the person/thing that is viewing the scene. "Camera" may be another alternative, but seems like it has the potential to be confused for the cameras that are actually doing the tracking/passthrough on some devices.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -133,7 +133,7 @@ If `supportsSessionMode` resolved for a given mode, then requesting a session wi
 Only one immersive session per XR hardware device is allowed at a time across the entire UA. All inline sessions are suspended when an immersive session is active. Inline sessions are not required to be created within a user activation event unless paired with another option that explicitly does require it.
 
 Once the session has started, some setup must be done to prepare for rendering.
-- A `XRFrameOfReference` should be created to establish a coordinate system in which `XRDevicePose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
+- A `XRFrameOfReference` should be created to establish a coordinate system in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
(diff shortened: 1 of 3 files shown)


TrevorFSmith:
+1 for `XRViewerPose`

toji:
Thanks for the review, @NellWaliczek! Rebased so the commit list is sane again.

We've discussed the placement of `getViewport()` in the past, and at one point it actually was on the `XRView` taking in a layer as you said. We moved it in #329 because the viewport requirements of layer types will vary. (And it ended up being cleaner to implement that way too.)

NellWaliczek:
Looking over issue #329, I don't quite follow the logic on that one.  I was in between jobs at the time it was changed, so perhaps I missed something there?

The reason I raised the question is that now all the frame-to-frame rendering parameters are all grouped on a single object.  There's something a bit odd about having one hanging off to the side somewhere else.

toji:
It's hanging off to the side primarily because it's a value that's unique to that layer, and in fact may not be view dependent at all. For example: In the case of multiview rendering to a texture array you may just have a single viewport for the entire stack. As such tying the viewport to the view incurs unnecessary busywork to set the viewport.

NellWaliczek:
Ah, that's a good point about multiview.  Thx!

toji:
PR has been up a week without dissent and the responses have been favorable, so merging.

PR #427, opened 2018-11-05
Remove XRDevice from the spec text

toji:
Update the existing spec text to take into account the changes made in #405, removing XRDevice.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -151,7 +151,7 @@ Context Isolation {#contextisolation-security}
 
 The trusted UI must be drawn by an independent rendering context whose state is isolated from any rendering contexts used by the page. (For example, any WebGL rendering contexts.) This is to prevent the page from corrupting the state of the trusted UI's context, which may prevent it from properly rendering a tracked environment. It also prevents the possibility of the page being able to capture imagery from the trusted UI, which could lead to private information being leaked.
 
-Also, to prevent CORS-related vulnerabilities each page will see a new instance of objects returned by the API, such as {{XRDevice}} and {{XRSession}}. Attributes such as the {{XRWebGLLayer/context}} set by one page must not be able to be read by another. Similarly, methods invoked on the API MUST NOT cause an observable state change on other pages. For example: No method will be exposed that enables a system-level orientation reset, as this could be called repeatedly by a malicious page to prevent other pages from tracking properly. The user agent MUST, however, respect system-level orientation resets triggered by a user gesture or system menu.
+Also, to prevent CORS-related vulnerabilities each page will see a new instance of objects returned by the API, such as {{XRSession}}. Attributes such as the {{XRWebGLLayer/context}} set by one page must not be able to be read by another. Similarly, methods invoked on the API MUST NOT cause an observable state change on other pages. For example: No method will be exposed that enables a system-level orientation reset, as this could be called repeatedly by a malicious page to prevent other pages from tracking properly. The user agent MUST, however, respect system-level orientation resets triggered by a user gesture or system menu.
 
 Fingerprinting {#fingerprinting-security}
 --------------
@@ -160,7 +160,7 @@ Given that the API describes hardware available to the user and its capabilities
 
 Issue: Discuss use of sensor activity as a possible fingerprinting vector.
 
-Device Enumeration {#deviceenumeration}
+Initialization {#initialization}
 ===================
 
 XR {#xr-interface}
@@ -169,7 +169,8 @@ XR {#xr-interface}
 <pre class="idl">
 [SecureContext, Exposed=Window] interface XR : EventTarget {
   // Methods
-  Promise&lt;XRDevice?&gt; requestDevice();
+  Promise&lt;void&gt; supportsSession(optional XRSessionCreationOptions options);
+  Promise&lt;XRSession&gt; requestSession(optional XRSessionCreationOptions parameters);
 
   // Events
   attribute EventHandler ondevicechange;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -181,96 +182,66 @@ partial interface Navigator {
...
+NOTE: The user agent is allowed to use any criteria it wishes to [=select an XR device=] when the [=list of XR devices=] contains multiple devices. For example, the user agent may always select the first item in the list, or provide settings UI that allows users to manage device priority. Ideally the algorithm used to select the default device is stable and will result in the same device being selected across multiple browsing sessions.
 
-<div class="algorithm" data-algorithm="request-device">
+<div class="algorithm" data-algorithm="ensure-device-selected">
 
-The page can request a device by calling the <dfn method for="XR">requestDevice()</dfn> method on the {{Navigator/xr}} object. When invoked it MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
+Any time an [=XR device=] is needed by an algorithm it can <dfn>ensure an XR device is selected</dfn> by running the following steps:
```
NellWaliczek: This seems to imply that it might change from call to call even after a session has been established..

index.bs
```diff
@@ -181,96 +182,66 @@ partial interface Navigator {
...
-Each {{XRDevice}} has a <dfn>supports immersive</dfn> value, which is a boolean which MUST be set to <code>true</code> if the device can support [=immersive sessions=] and <code>false</code> if it cannot.
-
-Each {{XRDevice}} has an <dfn>active immersive session</dfn>, which MUST be initially <code>null</code>, and a <dfn>list of non-immersive sessions</dfn>, which MUST be initially empty.
+Each [=XR device=] has a <dfn>supports immersive</dfn> value, which is a boolean which MUST be set to <code>true</code> if the device can support [=immersive sessions=] and <code>false</code> if it cannot.
 
-<div class="algorithm" data-algorithm="supports-session">
+<div class="algorithm" data-algorithm="supports-session-mode">
 
-When the <dfn method for="XRDevice">supportsSession(|options|)</dfn> method is invoked, it MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
+When the <dfn method for="XR">supportsSession(|options|)</dfn> method is invoked, it MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
```
NellWaliczek: Is this really "in parallel"?

index.bs
```diff
@@ -181,96 +182,66 @@ partial interface Navigator {
...
 </div>
 
+Calling {{XR/supportsSession()}} MUST NOT trigger device-selection UI as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
+
+The {{XR}} object has an <dfn>active immersive session</dfn>, which MUST be initially <code>null</code>, and a <dfn>list of non-immersive sessions</dfn>, which MUST be initially empty.
+
 <div class="algorithm" data-algorithm="request-session">
 
-When the <dfn method for="XRDevice">requestSession(|options|)</dfn> method is invoked, the user agent MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
+When the <dfn method for="XR">requestSession(|options|)</dfn> method is invoked, the user agent MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
```
NellWaliczek: Also, "in parallel"?

index.bs
```diff
@@ -181,96 +182,66 @@ partial interface Navigator {
...
 
-  1. Let |device| be the target {{XRDevice}} object.
-  1. If the |options| are not [=supported by the device=] |device|, [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
+  1. [=Ensure an XR device is selected=].
```
NellWaliczek: Refer to the algorithm above?

index.bs
```diff
@@ -181,96 +182,66 @@ partial interface Navigator {
...
+  1. If [=XR device=] is <code>null</code>, [=reject=] |promise| with <code>null</code>.
```
NellWaliczek: ><code>null</code>

Is that standard behavior?

index.bs
```diff
@@ -181,96 +182,66 @@ partial interface Navigator {
...
+  1. If the |options| are not [=supported by the XR device=], [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
   1. Let |immersive| be the {{XRSessionCreationOptions/immersive}} attribute of the |options| argument.
-  1. If |immersive| is <code>true</code> and |device|'s [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
+  1. If |immersive| is <code>true</code> and the [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
```
NellWaliczek: Really? I thought this was supposed to take over for any existing active immersive session.

index.bs
```diff
@@ -287,6 +258,7 @@ xrDevice.requestSession({ immersive: true }).then((session) => {
 </pre>
 </div>
 
+ISSUE: Take permissions into account when calling {{XR/supportsSession()}} or {{XR/requestSession()}}.
```
NellWaliczek: Is it worth having this here right now?

index.bs
```diff
@@ -982,42 +950,41 @@ WebGL Context Compatiblity {#contextcompatibility}
 
 <pre class="idl">
 partial dictionary WebGLContextAttributes {
-    XRDevice compatibleXRDevice = null;
```
NellWaliczek: Etiquette question: given that these are WebGL changes, can they actually be specified in this document or do they need to be done differently?

index.bs
```diff
@@ -982,42 +950,41 @@ WebGL Context Compatiblity {#contextcompatibility}
...
 
-The [=compatible XR device=] can be set either at context creation time or after context creation, potentially incurring a context loss. To set the [=compatible XR device=] at context creation time, an {{XRDevice}} is supplied to the {{compatibleXRDevice}} context creation attibute dictionary when requesting a WebGL context.
+The [=XR compatible=] boolean can be set either at context creation time or after context creation, potentially incurring a context loss. To set the [=XR compatible=] boolean at context creation time, the {{xrCompatible}} context creation attibute must be set to <code>true</code> when requesting a WebGL context.
```
NellWaliczek: Not specifically related to your change, but I can't recall.... did we evaluate the fingerprinting risks associated with setting a context compatibility?

index.bs
```diff
@@ -1051,7 +1018,7 @@ When the <dfn method for="WebGLRenderingContextBase">setCompatibleXRDevice(|devi
 
 Additionally, when any WebGL [=handle the context loss|context is lost=] run the following steps prior to firing the "webglcontextlost" event:
 
-  1. Set the context's [=compatible XR device=] to <code>null</code>.
+  1. Set the context's [=XR compatible=] boolean to <code>false</code>.
```
NellWaliczek: We need to explicitly call out the behavior of the compatible webgl context when a device change happens.

index.bs
```diff
@@ -1162,15 +1129,13 @@ Event Types {#event-types}
...
 
 A user agent MAY dispatch a <dfn event for="XRSession">focus</dfn> event on an {{XRSession}} to indicate that presentation to the {{XRSession}} by the page has resumed after being suspended. The event MUST be of type {{XRSessionEvent}}.
 
-A user agent MUST dispatch a <dfn event for="XRSession">resetpose</dfn> event on an {{XRSession}} when the system resets the {{XRDevice}}'s position or orientation. The event MUST be of type {{XRSessionEvent}}.
+A user agent MUST dispatch a <dfn event for="XRSession">resetpose</dfn> event on an {{XRSession}} when the system resets the [=XR device=]'s position or orientation. The event MUST be of type {{XRSessionEvent}}.
```
NellWaliczek: Thanks for making this PR scoped to just the XRDevice changes!

index.bs
```diff
@@ -181,96 +182,66 @@ partial interface Navigator {
...
-  1. Let |oldDefaultDevice| be the [=default device=].
-  1. If the [=list of XR devices=] is empty, set the [=default device=] to <code>null</code>.
-  1. If the [=list of XR devices=] contains one device set the [=default device=] to that device.
-  1. If the [=list of XR devices=] contains multiple devices set the [=default device=] to a device of the user agent's choosing.
-  1. If this is not the first time devices have been enumerated and |oldDefaultDevice| does not equal the [=default device=] queue a task that fires a simple event named {{devicechange}} on the {{XR}} object.
+  1. Let |oldDevice| be the current [=XR device=].
+  1. If the [=list of XR devices=] is empty, set the [=XR device=] to <code>null</code>.
+  1. If the [=list of XR devices=] contains one device set the [=XR device=] to that device.
+  1. If the [=list of XR devices=] contains multiple devices set the [=XR device=] to a device of the user agent's choosing.
+  1. If this is not the first time devices have been enumerated and |oldDevice| does not equal the [=XR device=] queue a task that fires a simple event named {{devicechange}} on the {{XR}} object.
```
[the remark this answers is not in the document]
NellWaliczek: Ah, I meant to delete this comment.  Ignore it.  Sorry!

index.bs
toji: It's a promise, so yes.

index.bs
toji: To clarify the "In parallel" text is a standard web spec thing that means "in parallel to the main thread". Good resource for this kind of minutia: https://garykac.github.io/procspec/ (Search for "in parallel" in that doc)

index.bs
NellWaliczek: But each of the steps aren't in parallel, yeah?

index.bs
toji: That's what the [=Brackets=] do, links this to the <dfn></dfn> tag above with the same text.

index.bs
toji: I don't believe we did, but I also can't think of much in the way of risk? Is there a specific fingerprinting avenue that you had in mind? (The fact that you MUST set the context compatibility bit whether or not the context would have been created on the right GPU anyway seems like it mitigates most issues that could have come up with it.)

index.bs
toji: That wasn't my understanding. 🙁

index.bs
NellWaliczek: The scenario I was thinking of is: 
1. Creates a canvas and get the context
1. Peep at the GPU info
1. Sets the context for XR compatibility
1. Peep at the GPU info again

On desktop machines with the XR device plugged into a secondary GPU, this would enable a developer to know that there is more than one GPU without actually engaging the XR hardware.

Are we worried about this?

index.bs
NellWaliczek: Can you point me to the line number?  I'm not seeing where we explain that the XR compatible boolean gets reset to false when the device change happens.

NellWaliczek: Ah, you meant you *just* added it. ^_^

index.bs
NellWaliczek: I followed it, though yes... a bit convoluted :)
As a sanity check, are we certain about the ordering of those last three steps?  As in, can we think of any reason why we might want to fire the devicechange before shutting down the session?

toji: I gave some thought to that, and the primary reason I came up with for the current order is that some developers may try and create a session in response to the event. We wouldn't want to end it immediately after, so this order simply makes the logic cleaner.

index.bs
toji: There are existing scenarios where one spec implies additions or related functionality to another one. (That's more-or-less why we have the `partial` keyword.) So I think this is appropriately scoped. For example, search for "WebGL" here: https://www.w3.org/TR/2017/WD-html52-20170718/single-page.html

index.bs
toji: Done. #429

PR #428, opened 2018-11-05
Describe XRFrame lifetime

toji:
Adding text to address #403, making the lifetime of `XRFrame` objects explicit. CC: @lincolnfrog, @thetuvix


thetuvix:
Spending Thanksgiving Eve finally catching up on a bunch of PRs 🦃

Given the wide range of platforms WebXR will run on, it seems like a reasonable place to start for apps to stash away any data within the input event callback that gave them the historical `XRFrame` instance.  We can explore loosening up this contract in the future if we need to.

This limitation may actually help apps do the right thing in regard to throwing interactions.  We've found that a common app mistake with throwing in VR is to use the controller's predicted position/velocity for the current frame's future photon time to aim the throw - this often ends up throwing the object down into the ground at a slower speed than the user intended, and so apps add fudge factors to the angle and magnitude of the throw.  [Throwing feels way better](https://docs.microsoft.com/en-us/windows/mixed-reality/gestures-and-motion-controllers-in-unity#throwing-objects) when the position/velocity used are from the historical time when the user actually released the button - we've found that apps that make this switch can often just remove their fudge factors.

Making it more annoying in an input event to get the predicted data than the historical data may help improve WebXR app quality 😄

PR #430, opened 2018-11-07
Updating spec to account for recent XRView and XRFrame changes

toji:
Nothing surprising here, just trying to reduce the delta between the spec and the explainer. Did some handwaving over a couple of algortihms to keep PR focused on just the updated parts.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -43,6 +43,7 @@ urlPrefix: https://www.w3.org/TR/html5/semantics-scripting.html
     type: interface; text: HTMLCanvasElement
 urlPrefix: https://www.w3.org/TR/html5/
     type: interface; text: Document
+    type: dfn; text: dispatch
 urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.0/
     type: typedef; text: uniformMatrix4fv
     type: interface; text: WebGLFramebuffer
@@ -553,7 +554,7 @@ When the <dfn method for="XRFrame">getViewerPose(|frameOfRef|)</dfn> method is i
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
   1. If |frameOfRef|'s [=XRFrameOfReference/session=] does not equal |session|, return <code>null</code> and abort these steps.
   1. If the [=viewer=]'s pose cannot be determined relative to |frameOfRef|, return <code>null</code>
-  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to |frameOfRef|.
+  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |frameOfRef|.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -531,22 +533,43 @@ XRFrame {#xrpresentationframe-interface}
...
-<dfn attribute for="XRFrame">views</dfn>
+<div class="algorithm" data-algorithm="get-viewer-pose">
+
+When the <dfn method for="XRFrame">getViewerPose(|frameOfRef|)</dfn> method is invoked, the user agent MUST run the following steps:
+
+  1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw a {{NotAllowedError}} and abort these steps.
+  1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
+  1. If |frameOfRef|'s [=XRFrameOfReference/session=] does not equal |session|, return <code>null</code> and abort these steps.
+  1. If the [=viewer=]'s pose cannot be determined relative to |frameOfRef|, return <code>null</code>
+  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to |frameOfRef|.
```
NellWaliczek: How do you feel about this wording instead?
> Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |frameOfRef|.

index.bs
```diff
@@ -1106,7 +1132,7 @@ dictionary XRInputSourceEventInit : EventInit {
 </pre>
 
 <dfn attribute for="XRInputSourceEvent">frame</dfn>
-An {{XRFrame}} that corresponds with the time that the event took place. This frame's {{XRFrame/views}} array MUST be empty.
+An {{XRFrame}} that corresponds with the time that the event took place. Any {{XRViewerPose}} queried from this frame MUST have an empty {{XRViewerPose/views}} array.
```
NellWaliczek: Did I miss where we are specifying when an XRFrame's active boolean is set to true before invoking the input event callback?

index.bs
```diff
@@ -730,24 +757,23 @@ WebXR provides various transforms in the form of <dfn lt="matrix">matrices</dfn>
...
 
-NOTE: The {{XRDevicePose/poseModelMatrix}} can be used to position graphical representations of the [=XR device=] for spectator views of the scene or multi-user interaction.
+NOTE: The {{XRViewerPose/poseMatrix}} can be used to position graphical representations of the [=viewer=] for spectator views of the scene or multi-user interaction.
 
-The <dfn method for="XRDevicePose">getViewMatrix(view)</dfn> method returns a [=matrix=] describing the view transform to be used when rendering the passed {{XRView}}. The matrices represent the inverse of the model matrix of the associated viewpoint.
+The <dfn attribute for="XRViewerPose">views</dfn> array is a sequence of {{XRView}}s describing the viewpoints of the XR scene, relative to the {{XRFrameOfReference}} the {{XRViewerPose}} was queried with, which must be rendered in order to display correctly on the [=XR device=]. Each {{XRView}} includes view and projection matrices, and can be used to query {{XRViewport}}s from layers when needed.
```
NellWaliczek: nit: this feels like a really long sentence.  Can we break it up somehow?

index.bs
toji: Added a section to describe that, thanks!

toji:
Thanks for the review! Updated to address your comments. PTAL!

PR #432, opened 2018-11-08
Updated spec with XRSessionMode changes

toji:
Focused here primarily on getting the IDL and algorithms updated. There's a fair amount of new text that probably should be added to cover what, exactly, "immersive-ar" mode implies but I'm deferring that work for the moment.

Also threw in a few additional fixes like warning corrections for the bikeshed `make` process and the odd spelling error that I stumbled on.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -251,21 +251,26 @@ navigator.xr.supportsSessionMode('immersive-vr').then(() => {
 </pre>
 </div>
 
-The {{XR}} object has an <dfn>active immersive session</dfn>, which MUST be initially <code>null</code>, and a <dfn>list of inline sessions</dfn>, which MUST be initially empty.
+The {{XR}} object has a <dfn>pending immersive session</dfn> boolean, which MUST be initially <code>false</code>, an <dfn>active immersive session</dfn>, which MUST be initially <code>null</code>, and a <dfn>list of inline sessions</dfn>, which MUST be initially empty.
 
 <div class="algorithm" data-algorithm="request-session">
 
 When the <dfn method for="XR">requestSession(|options|)</dfn> method is invoked, the user agent MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
 
+  1. Let |mode| be the {{XRSessionCreationOptions/mode}} attribute of the |options| argument.
+  1. Let |immersive| be a boolean set to <code>true</code> if |mode| is {{immersive-vr}} or {{immersive-ar}} and <code>false</code> otherwise.
+  1. If |immersive| is <code>true</code>: 
+    1. If [=pending immersive session=] is <code>true</code> or [=active immersive session=] in not <code>null</code>, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
+    1. Set [=pending immersive session=] to be <code>true</code>.
   1. [=Ensure an XR device is selected=].
   1. If [=XR device=] is <code>null</code>, [=reject=] |promise| with <code>null</code>.
-  1. Let |mode| be the {{XRSessionCreationOptions/mode}} attribute of the |options| argument.
-  1. If |mode| is in not in [=XR device=]'s [=list of supported modes=], [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
-  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} and the [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
-  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} and the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a {{SecurityError}} and abort these steps.
+  1. Else if |mode| is in not in [=XR device=]'s [=list of supported modes=], [=reject=] |promise| with a {{NotSupportedError}}.
+  1. Else If |immersive| is <code>true</code> and the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a {{SecurityError}} and abort these steps.
+  1. If |promise| was [=rejected=] and |immersive| is <code>true</code>, set [=pending immersive session=] to be <code>false</code>.
+  1. If |promise| was [=rejected=], abort these steps.
   1. Let |session| be a new {{XRSession}}.
   1. [=Initialize the session=] |session| with the [=session description=] given by |options|.
-  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} set the [=active immersive session=] to |session|.
+  1. If |immersive| is <code>true</code> set the [=active immersive session=] to |session| and set [=pending immersive session=] to <code>false</code>.
   1. Else append |session| to the [=list of inline sessions=].
   1. [=/Resolve=] |promise| with |session|.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -217,48 +217,59 @@ Any time an [=XR device=] is needed by an algorithm it can <dfn>ensure an XR dev
...
-  1. If |immersive| is <code>true</code> and the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a {{SecurityError}} and abort these steps.
+  1. Let |mode| be the {{XRSessionCreationOptions/mode}} attribute of the |options| argument.
+  1. If |mode| is in not in [=XR device=]'s [=list of supported modes=], [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
+  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} and the [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
+  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} and the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a {{SecurityError}} and abort these steps.
   1. Let |session| be a new {{XRSession}}.
   1. [=Initialize the session=] |session| with the [=session description=] given by |options|.
-  1. If |immersive| is <code>true</code> set the [=active immersive session=] to |session|.
-  1. Else append |session| to the [=list of non-immersive sessions=].
+  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} set the [=active immersive session=] to |session|.
```
NellWaliczek: Not specific to this PR, but we need to define the behavior when calling requestSession() twice in rapid succession.  Right now, it looks like the second call wouldn't be rejected unless the first call had already reached this step.

index.bs
```diff
@@ -268,33 +279,40 @@ xrDevice.requestSession({ immersive: true }).then((session) => {
...
+  "immersive-vr",
+  "immersive-ar"
 };
 </pre>
 
-The {{XRSessionCreationOptions}} dictionary provides a <dfn>session description</dfn>, indicating the desired properties of a session to be returned from {{requestSession()}}.
+The {{XRSessionMode}} enum defines the modes that an {{XRSession}} can operate in.
 
-<div class="algorithm" data-algorithm="supported-by-device">
+  - A session mode of <dfn enum-value for="XRSessionMode">inline</dfn> indicates that the session's output will be shown as an element in the HTML document. {{inline}} session content MAY be displayed in mono or stereo and MAY allow for [=viewer=] tracking. User agents MUST allow {{inline}} sessions to be created for any [=XR device=].
```
NellWaliczek: Should it be a requirement that an outputContext be provided?

index.bs
```diff
@@ -268,33 +279,40 @@ xrDevice.requestSession({ immersive: true }).then((session) => {
...
+  - A session mode of <dfn enum-value for="XRSessionMode">immersive-vr</dfn> indicates that the session's output will be shown on the [=XR device=] hardware in a manner that occupies a significant portion of the user's vision. The {{environmentBlendMode}} MAY be {{opaque}} for {{immersive-vr}} sessions.
```
NellWaliczek: > The {{environmentBlendMode}} MAY be {{opaque}} for {{immersive-vr}} sessions.

I'm inclined to see this be a bit stronger a statement.  For example, we don't want to allow {{alpha-blend}}, right?  We might want to take it further and say the only time {{additive}} should be allowed is if the hardware cannot support {{opaque}}.

index.bs
```diff
@@ -268,33 +279,40 @@ xrDevice.requestSession({ immersive: true }).then((session) => {
...
+  -  A session mode of <dfn enum-value for="XRSessionMode">immersive-ar</dfn> indicates that the session's output will be shown either on the [=XR device=] hardware in a manner that occupies a significant portion of the user's vision or fullscreen on hardware that does not occupy a significant portion of the user's vision. The {{environmentBlendMode}} MUST NOT be {{opaque}} for {{immersive-ar}} sessions.
```
NellWaliczek: oof.  I had a hard time understanding that first sentence... Instead, can we talk about it having exclusive access to the xr device display?

index.bs
toji: Uuuuuuggghhhh. Totally agree that this needs to be addressed (I've been saying I'd do it since May, according to #343), but my goodness the the spec-y algorithm get ugly. 😝 Why can't we just write specs in actual code? 😉

index.bs
toji: Yes, at this point, but I'm hoping to strip that requirement out pretty quickly.

index.bs
toji: Yeah, sorry about that one. Taking another crack at it. 🙄

toji:
Okay, update pushed. Tried to address everything, hopefully at least the immersive mode descriptions are clearer. (I know the `requestSession` algorthim is not...)

index.bs
```diff
@@ -251,21 +251,26 @@ navigator.xr.supportsSessionMode('immersive-vr').then(() => {
...
-  1. Let |mode| be the {{XRSessionCreationOptions/mode}} attribute of the |options| argument.
-  1. If |mode| is in not in [=XR device=]'s [=list of supported modes=], [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
-  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} and the [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
-  1. If |mode| is {{immersive-vr}} or {{immersive-ar}} and the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a {{SecurityError}} and abort these steps.
+  1. Else if |mode| is in not in [=XR device=]'s [=list of supported modes=], [=reject=] |promise| with a {{NotSupportedError}}.
+  1. Else If |immersive| is <code>true</code> and the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a {{SecurityError}} and abort these steps.
+  1. If |promise| was [=rejected=] and |immersive| is <code>true</code>, set [=pending immersive session=] to be <code>false</code>.
+  1. If |promise| was [=rejected=], abort these steps.
   1. Let |session| be a new {{XRSession}}.
```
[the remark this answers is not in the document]
toji: This is a good question that I don't really know the answer to? I'm gonna push this PR for now and then look into how other specs handle algorithmic steps that boil down to "UA-specific behavior here, if needed", because that seems like it could apply almost anywhere? But I agree that there's benefit to saying "consent prompts should go here."

index.bs
```diff
@@ -301,10 +306,12 @@ enum XRSessionMode {
...
+  -  A session mode of <dfn enum-value for="XRSessionMode">immersive-ar</dfn> indicates that the session's output will be given [=exclusive access=] to the [=XR device=] display and that content <b>is</b> intended to be integrated with the user's environment. The {{environmentBlendMode}} MUST NOT be {{opaque}} for {{immersive-ar}} sessions.
+
+An <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or an {{immersive-ar}} session. [=Immersive sessions=] MUST provide some level of [=viewer=] tracking, and content MUST be shown at the proper scale relative to the user and/or the surrounding environment. Additionally, [=Immersive sessions=] MUST be given <dfn>exclusive access</dfn> to the [=XR device=], meaning that while the [=immersive session=] is not [=blurred=] the HTML document is not shown on the [=XR device=]'s display, nor is content from other applications shown.
```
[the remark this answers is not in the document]
toji: I tweaked the verbiage a little bit to indicate that other applications should not be shown *on the XR device*, but maybe that's not what your concern here was?

toji:
Thanks Nell! Merged!

index.bs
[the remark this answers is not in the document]
NellWaliczek: Sounds good.  Can we file an issue against the repo to track this?

Issue #417, opened 2018-10-12
Requiring multiple frames of reference


NellWaliczek:
In PR #409 an optional session creation argument was added that allows developers to block a session from being created if it will be unable to provide a specific type of frame of reference.   The current design only allows pages to block on a single frame of reference type, but it’s possible that a scene may wish to block on more than one.  For example an unbounded experience which wants to switch to a bounded experience and back again.

Naively, the simple solution would be to change the `requiredFrameOfReference` to accept an array rather than a single item.  The question is, should the `requestSession()` call reject if none of the required frames of reference are available or if at least one is unavailable.

My inclination is to reject if none are available since the `requestFrameOfReference()` call accepts an array of desired types.

Any objections?

johnpallett:
Nell is one outcome of this that the UA could enforce privacy/security requirements for all potential session types at creation time, instead of on-the-fly? For example, if the UA hypothetically applied different policies for bounded vs. unbounded, those policies could be evaluated at session creation?

NellWaliczek:
>For example, if the UA hypothetically applied different policies for bounded vs. unbounded, those policies could be evaluated at session creation?

I'm not entirely sure what you mean by this part... do you mean different policies based on immersive vs. inline sessions?

NellWaliczek:
Also, to follow up, this issue was discussed at TPAC and the conclusion was that the single value should be expanded to an array in which only one type needs to be supported for session creation to succeed.  I'll get a PR posted shortly with this change.

NellWaliczek:
Resolving this issue is on hold while we work on solutions for #423 and #424

NellWaliczek:
Closed by #739

Issue #424, opened 2018-10-30
Define how to request features which require user consent


toji:
We've identified that there will be multiple features that WebXR will want to expose over time that will require some form of user consent. This subject has gone back and forth a LOT, and I'm not going to try to capture all of the discussion again here, but instead present my current thinking and iterate from there. 

We want the request for these features to have certain properties (Partially pulled from #330):

 - We want to avoid a scenario where some browser are forced into presenting the user with a sequential series of dialogs for common scenarios. And example of this "forced" behavior is if the creation of a session requires a permission prompt, and then calling a method on the session also requires a permission prompt. The UA can't do anything to make the presentation less jarring because the session has to fully resolve before the next permission is even requested.
 - Ideally there's a way to allow the browser to bundle all the necessary permissions into a single dialog if desired/applicable.
 - The system should also allow for bespoke feature requests at the time they're actually necessary. (ie: Not requesting access to RGB camera data until the user clicks a "take a photo" button.)

The most straightforward way I see for addressing that is to simply pass the requested features into the session request call, either as dictionary keys or possibly and array of strings. For the moment I'm going to suggest the dictionary key route:

```js
navigator.xr.requestSession({
  mode: 'immersive-ar',
  lightingEstimation: true,
  environmentMeshing: true
}).then(/* ... */);
```

And for bespoke feature requests afterwards have an explicit function that takes in the same dictionary args (minus the session mode):

```js
xrSession.requestFeatures({
  lightingEstimation: true,
  environmentMeshing: true
}).then(/* ... */);
```

The UA should _not_ reject the session if any of these features are either not supported or the user does not consent to giving access to them. Instead the session should still be created like normally and the mechanism for accessing the feature once enabled should be null/reject/report an error/whatever failure mode is appropriate. No differentiation should be made between the feature not being supported and the user denying consent to avoid leaking information about the users system without their permission. That way we're encouraging developers to be responsive to varying feature sets. If a feature is deemed to be absolutely critical for developers to identify prior to starting a session it should be handled by the mechanism discussed in #423. 

[...]

Issue #423, opened 2018-10-30
Allow session creation to be blocked on required features


toji:
We've previously talked about the need to have a way of doing basic feature detection before creating a session. This is needed because it's a poor user experience to launch into an immersive session, which may involve user-visible transitions and setup steps, only to be immediately told "Oops! Sorry, you can't actually do this." We must balance the desire to avoid that type of user inconvenience, however, with the need to suppress as much fingerprintable information as possible prior to session creation.

The proposed solution so far is to allow the user to pass in a set of "required features" when requesting the session. That way the UA can, prior to performing any user-visible actions, opaquely determine if the requirements can be met and reject early if not. Functionally, this would mean that sites with hard requirements beyond "Do you support AR or VR?" would still add a button to enter immersive mode to their page but upon clicking it the users could ~immediately be presented with a dialog from the page stating "Your system doesn't meet the requirements for this experience."

This can be done in a privacy sensitive way because:

- Immersive session creation requires user activation, so features cannot be checked prior to user interaction.
- The session rejection should NOT state which features of the list were not available, making it difficult to determine what was missing in a list of more than one requirement.
- Because of the user activation requirement, session requests cannot be rapidly spammed with different combinations of features to eventually see what works and what doesn't.
- The consequences of a successful request are high, in that it will incur user-visible side effects, and thus developers cannot do a "silent" feature check.

We've also discussed in the past that we want to encourage developers to be as responsive as possible to different systems with XR content, and this feature would seem to discourage that in favor of allowing developers to easily block off content to all but a small set of users, but I believe that it has some built-in deterrents:

- The user experience for getting rejected when you click an "Enter XR" button, while better than if it started incurring full screen transitions and the like, is still relatively poor. Nobody likes to be teased with a feature and then told "Nope! Not for you!" As such, developers are likely to only use this as a gate if they know their experience simply cannot work without a given feature.
- We can control the list of features that can be tested for, so let's maybe not add a `requiresOculusRift` enum, yeah? 

This general concept has already been introduced in limited form to the spatial tracking explainer with the [`requiredFrameOfReferenceType` dictionary argument](https://github.com/immersive-web/webxr/blob/master/spatial-tracking-explainer.md#ensuring-hardware-compatibility). #417 is also highly related. I feel, however, like it should be handled using a more generic mechanism so that any type of requirement we deem necessary in the future (for example: Presence of 6DoF input?) can be tested the same way.

[...]

PR #433, opened 2018-11-13
Added session feature handling to the explainer.

toji:
Addresses #423 and #424.

A couple of key points that aren't explicitly stated here:

 - Checking with @bfgeek, it sounds like passing the "desired" feature list as string and mapping them to the feature enum is a reasonable way of avoiding errors on unrecognized enums. Hence the difference in the IDL between the desired and required sets.
 - Since the desired behavior for requiring frames of reference is "or" logic, which is different than every other expected feature, this PR leaves the existing `requiredFrameOfReferenceTypes` dictionary key alone.

Finally, this PR is obviously using a "filler feature" to demonstrate the API, since we don't have any other real features decided on yet. I would like to establish this as the expected method for handling these features now, but effectively mark it as "unimplemented" until we have features that will actually make use of it. (I'd need to update the PR to show that prior to landing.)

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -251,6 +251,67 @@ function drawScene(view) {
 }
 ```
 
+### Enabling device features
+
+The use of some XR device features may introduce user privacy concerns, and as a result must gain the users consent before they are exposed to the page. In order to make use of these features an XRSession must explicitly request them by passing the `desiredFeatures` session creation option into a `requestSession` call.
+
+The `desiredFeatures` session creation option takes an array of strings. The `requestSession` call will iterate through the array and map each string to an `XRSessionFeature` enum. Items from the array that do not have a matching entry in the `XRSessionFeature` enum are ignored. (This is to ensure that newly added features do not break older UAs that don't recognize them.) Similarly, if the feature is not supported by the UA or XR device it is ignored. If the UA requires user consent for any of the supported features in the array, it must request that consent prior to the `requestSession` call resolving. Though UAs may have differing opinions on which features require explicit user consent, developers should expect that any entries in the `desiredFeatures` array will trigger a consent request visible to the user.
+
+An `XRSession` should still be returned even if the user denies consent for the desired features, and the feature's availability should be detected in a feature-specific manner once the session has been created. No differentiation is made at runtime between features that have not been requested, features that are not supported, and features for which consent has been denied.
+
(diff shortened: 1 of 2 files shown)


toji:
Updated this PR so that it also fixes #417 while I'm at it, since it's closely related.

explainer.md
```diff
@@ -251,6 +251,67 @@ function drawScene(view) {
...
+
+An `XRSession` should still be returned even if the user denies consent for the desired features, and the feature's availability should be detected in a feature-specific manner once the session has been created. No differentiation is made at runtime between features that have not been requested, features that are not supported, and features for which consent has been denied.
+
+This samples shows how a developer would request a theoretical feature (here called `feature-name`) that exposes new data on the `XRFrame` and test for it's availability once the session has been created:
+
+```js
+// TODO: Make the example feature not theoretical.
+navigator.xr.requestSession({
+    mode: 'immersive-vr',
+    desiredFeatures: ['feature-name']
```
blairmacintyre: I've been suggesting "geoAligned" as a feature.  Perhaps that would be a reasonable example?  It matches the pattern almost perfectly.

explainer.md
```diff
@@ -251,6 +251,67 @@ function drawScene(view) {
...
+  }).then(onSessionStarted);
+
+function onDrawFrame(timestamp, xrFrame) {
+  if (xrFrame.featureName) {
```
blairmacintyre: see above:

`if (xrFrame.geoAligned) { `

explainer.md
```diff
@@ -251,6 +251,67 @@ function drawScene(view) {
...
+
+Finally, there are some occasions where developers may want to defer a feature request until a more contextually appropriate time. For example **(FIXME: Is there a better short-term example?)**, an `immersive-ar` application may not need access to the camera stream until the point that the user clicks a "take a photo" button, at which point a request for camera access will be less concerning for the user. For these types of situations, new features can be requested after session creation with the `requestAdditionalFeatures` method. `requestAdditionalFeatures` can only be called in response to a user activation event, and accepts an array of strings that are processed identically to the `desiredFeatures` session creation option. It returns a promise that resolves when the user has consented to or denied use of any supported features. Individual features must still be checked to see if they are available after the promise resolves.
```
blairmacintyre: Are there a list of "candidate features" anywhere?  Could choose from that.

blairmacintyre:
Overall, I like this.

toji:
Closing because when we do revisit this issue it's likely to not have the exact solution proposed here.

PR #434, opened 2018-11-16
Updated XRFrameOfReference/XRCoordinateSystem in spec

toji:
This doesn't cover 100% of the material related to the recent tracking system changes, but I don't want this PR to get any more complicated than it already is. I've added `Issue:` blocks to the doc for known omissions.

This PR does build off of #432, so that'll need to land first.


toji:
Updated this PR to use the new names introduced in #435, assuming that'll land before this does.

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
 
-  XRViewerPose? getViewerPose(XRFrameOfReference frameOfRef);
-  XRInputPose? getInputPose(XRInputSource inputSource, XRFrameOfReference frameOfRef);
+  XRViewerPose? getViewerPose(optional XRReferenceSpace referenceSpace);
+  XRInputPose? getInputPose(XRInputSource inputSource, optional XRReferenceSpace referenceSpace);
 };
 </pre>
 
-An {{XRFrame}} provides all the values needed to render a single frame of an XR scene to the [=XR device=]'s display. Applications can only acquire an {{XRFrame}} by calling {{XRSession/requestAnimationFrame()}} on an {{XRSession}} with an {{XRFrameRequestCallback}}. When the callback is called it will be passed an {{XRFrame}}.
+An {{XRFrame}} provides all the values needed to render a single <dfn>animation frame</dfn> of an XR scene to the [=XR device=]'s display. Applications can only acquire an {{XRFrame}} by calling {{XRSession/requestAnimationFrame()}} on an {{XRSession}} with an {{XRFrameRequestCallback}}. When the callback is called it will be passed an {{XRFrame}}.
```
NellWaliczek: (I wonder... should we consider renaming `XRFrame` to `XRAnimationFrame`?)

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
+When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw a {{NotAllowedError}} and abort these steps.
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
-  1. If |frameOfRef|'s [=XRFrameOfReference/session=] does not equal |session|, return <code>null</code> and abort these steps.
-  1. If the [=viewer=]'s pose cannot be determined relative to |frameOfRef|, return <code>null</code>
-  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |frameOfRef|.
+  1. If |referenceSpace|'s [=XRReferenceSpace/session=] does not equal |session|, return <code>null</code> and abort these steps.
+  1. If the [=viewer=]'s pose cannot be determined relative to |referenceSpace|, return <code>null</code>
+  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |referenceSpace|.
```
NellWaliczek: nit: Maybe this instead?
>1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |referenceSpace| at the timestamp of the {{XRFrame}}.

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
 
-<b>eye-level</b>
+An {{XRReferenceSpace}} is obtained by calling {{XRSession/requestReferenceSpace()}}, which creates an instance of an interface extending {{XRReferenceSpace}}, determined by the {{XRReferenceSpaceOptions/type}} value of the {{XRReferenceSpaceOptions}} dictionary passed into the call:
 
-An {{XRFrameOfReference}} with a [=frame of reference type=] of <dfn enum-value for="XRFrameOfReferenceType">"eye-level"</dfn> describes a coordinate system
-with an origin that corresponds to the first device pose acquired by the {{XRSession}} after the <code>"head-model"</code> frame of reference is created, with the device yaw (rotation around the Y axis) at that point defining the coordinate system's forward direction.
+  - Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRReferenceSpaceType">stationary</dfn> creates an {{XRStationaryReferenceSpace}} instance.
```
NellWaliczek: This isn't always guaranteed for an inline session, right?

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
 
-<b>head-model</b>
+  - Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRReferenceSpaceType">bounded</dfn> creates an {{XRBoundedReferenceSpace}} instance if supported by the [=XR device=].
```
NellWaliczek: This should probably say:
>if supported by the [=XR device=] and the {{XRReferenceSpace/session}}.

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
+  1. Let |referenceSpace| be set to <code>null</code>.
+  1. If |type| is {{stationary}}, let |referenceSpace| be a new {{XRStationaryReferenceSpace}}.
+  1. Else if |type| is {{bounded}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
+  1. Else if |type| is {{unbounded}}, let |referenceSpace| be a new {{XRUnboundedReferenceSpace}}.
+  1. Return |referenceSpace|.
 
-Note: A [=stage=] is not intended to describe multi-room spaces, areas with uneven floor levels, or very large open areas. Future iterations of this specification may add more detailed support for tracking in those scenarios.
+</div>
+
+Issue: Describe circumstances when a reference space type might be rejected.
```
NellWaliczek: We might want to explain the reset event a bit more here?

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
 
-<b>stage</b>
+When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a reference space</dfn> by running the following steps:
 
-An {{XRFrameOfReference}} with a [=frame of reference type=] of <dfn enum-value for="XRFrameOfReferenceType">"stage"</dfn> describes a space known as a <dfn for="XRStageBounds">stage</dfn>. A [=stage=] is a bounded, floor-relative play space that the user can be expected to safely be able to move within. Other XR platforms sometimes refer to this concept as "room scale" or "standing space".
+  1. Initialize [=XRReferenceSpace/session=] be the {{XRSession}} object that requested creation of a reference space.
+  1. Let |options| be the {{XRReferenceSpaceOptions}} passed to {{requestReferenceSpace()}}.
+  1. Let |type| be set to |options| {{XRReferenceSpaceOptions/type}}.
+  1. Let |referenceSpace| be set to <code>null</code>.
+  1. If |type| is {{stationary}}, let |referenceSpace| be a new {{XRStationaryReferenceSpace}}.
```
NellWaliczek: Subtypes should be checked here, yeah?

NellWaliczek: And actually, the dictionary type is different too...

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
 
-The [=stage bounds=] are described by a {{XRFrameOfReference}}'s <dfn attribute for="XRFrameOfReference">bounds</dfn> attribute. If the frame of reference is not a [=stage=] or the [=stage bounds=] cannot be determined the {{bounds}} attribute MUST be <code>null</code>.
+[SecureContext, Exposed=Window]
+interface XRStationaryReferenceSpace : XRReferenceSpace {
+  readonly attribute XRStationaryReferenceSpaceSubtype subtype;
+};
+</pre>
 
-Note: When the {{bounds}} for a [=stage=] are <code>null</code> the user should not be required to physically move around their environment in order to interact with content. The device may still support [=6DoF=] tracking, but it can't be assumed that the user will know where they are relative to their environment and as such content that encourages movement beyond leaning is discouraged.
+An {{XRStationaryReferenceSpace}} represents a tracking space that the user is not expected to move around within. Tracking in a {{stationary}} reference space is optimized for keeping the reference space origin stable relative to the environment.
```
NellWaliczek: It feels a bit weird to say this about the stationary reference space since the hardware may be orientation only...
>optimized for keeping the reference space origin stable relative to the environment.

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
-  1. Let |session| be the {{XRSession}} object that requested |frameOfRef|'s creation.
-  1. Let |type| be the {{XRFrameOfReferenceType}} passed to {{XRSession/requestFrameOfReference()}}.
-  1. Initialize |frameOfRef|'s [=XRFrameOfReference/session=] to |session|.
-  1. Initialize |frameOfRef|'s [=frame of reference type=] to |type|.
-  1. If |type| is not {{XRFrameOfReferenceType/"stage"}}, abort these steps.
-  1. // Initialize stage-specific properties
+Note: The {{position-disabled}} subtype is primarily intended for use with pre-rendered media such as panoramic photos or videos. It should not be used for most other media types due to user discomfort associated with the lack of a neck model or full positional tracking.
 
-</div>
+The {{XRStationaryReferenceSpace}}'s <dfn attribute for="XRBoundedReferenceSpace">subtype</dfn> attribute is the {{XRStationaryReferenceSpaceSubtype}} that the {{XRStationaryReferenceSpace}} was created with.
```
NellWaliczek: XRBoundedReferenceSpace?

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
-  readonly attribute FrozenArray&lt;DOMPointReadOnly&gt; geometry;
-};
+[SecureContext, Exposed=Window]
+interface XRBoundedReferenceSpace : XRReferenceSpace {
+  readonly attribute FrozenArray&lt;DOMPointReadOnly&gt; boundsGeometry;
+};    
 </pre>
 
-The {{XRStageBounds}} interface describes a border around a [=stage=], known as the <dfn>stage bounds</dfn> which the user can be expected to safely be able to move within. 
+An {{XRBoundedReferenceSpace}} represents a floor-relative tracking space where the user is expected to move within a pre-established boundary. In order to ensure user safety tracking in an {{bounded}} reference space is optimized for keeping the reference space origin and bounds geometry stable relative to the environment. Other XR platforms sometimes refer to this as "room scale" tracking.
```
NellWaliczek: >In order to ensure user safety tracking in an {{bounded}} reference space is optimized for keeping the reference space origin and bounds geometry stable relative to the environment.

Something about this feels a bit awkward?

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
+The <dfn attribute for="XRBoundedReferenceSpace">boundsGeometry</dfn> attribute describes the border around the {{XRBoundedReferenceSpace}}, which the user can expect to safely move within. 
 
-Note: Content should not require the user to move beyond the [=stage bounds=]; however, if their physical surroundings allow for it, it is possible for the user to ignore the bounds resulting in position values outside of the polygon they describe. This is not an error condition and should be handled gracefully by page content.
+The polygonal boundary is given as an array of {{DOMPointReadOnly}}s, which represents a loop of points at the edges of the safe space. The points describe offsets from the {{XRReferenceSpace}} origin in meters, but MAY not contain the origin. Points MUST be given in a clockwise order as viewed from above, looking towards the negative end of the Y axis. The {{DOMPointReadOnly/y}} value of each point MUST be <code>0</code> and the {{DOMPointReadOnly/w}} value of each point MUST be <code>1</code>. The bounds can be considered to originate at the floor and extend infinitely high. The shape it describes MAY not be convex.
```
NellWaliczek: >but MAY not contain the origin.

Didn't we change this?

index.bs
```diff
@@ -568,153 +564,187 @@ XRFrame {#xrframe-interface}
...
 
-Note: Content generally should not provide a visualization of the [=stage bounds=], as it's the user agent's responsibility to ensure that safety critical information is provided to the user.
+Note: Content should not require the user to move beyond the {{boundsGeometry}}. It is possible for the user to move beyond the bounds if their physical surroundings allow for it, resulting in position values outside of the polygon they describe. This is not an error condition and should be handled gracefully by page content.
```
NellWaliczek: May be worth saying that content wishing to larger than this space can manage their own teleportation offset?  Or should we just wait until we merge the other PR regarding that?

index.bs
```diff
@@ -1234,7 +1267,7 @@ A user agent MUST dispatch a <dfn event for="XRSession">selectend</dfn> event on
 
 A user agent MUST dispatch a <dfn event for="XRSession">select</dfn> event on an {{XRSession}} when one of its {{XRInputSource}}s has fully completed a [=primary action=]. The event MUST be of type {{XRInputSourceEvent}}.
 
-A user agent MUST dispatch a <dfn event for="XRFrameOfReference">boundschange</dfn> event on an {{XRFrameOfReference}} when the [=stage bounds=] change. This includes changes to the {{geometry}} point array or the {{bounds}} attribute changing to or from <code>null</code>. The event MUST be of type {{XRCoordinateSystemEvent}}.
+A user agent MUST dispatch a <dfn event for="XRReferenceSpace">reset</dfn> event on an {{XRReferenceSpace}} when a discontinuity of the origin occurs, or when the {{boundsGeometry}} changes for a {{XRBoundedReferenceSpace}}. The event MUST be of type {{XRReferenceSpaceEvent}}, and MUST be dispatched prior to the execution of any [=animation frame=]s that make use of the new origin.
```
NellWaliczek: I wonder if we need to elaborate more on what "discontinuity" means or the circumstances which cause it.

index.bs
toji: Ah, thanks. You are correct.

index.bs
toji: I'd prefer to wait on that one, just to avoid work we're reasonably sure will be discarded in the near future.

index.bs
toji: This is a weird pattern that appears in most web standards I've seen when the events themselves are described not at the handler attribute but as a separate events section (which in this doc is further down). Not my favorite layout, but I'd prefer not to break precedence. 🤷‍♀️

index.bs
toji: I'd like to have a separate conversation about how that particular scenario is best handled. I'm mostly worried about ensuring the timing of consent dialogs is consistent with other features.

toji:
Thanks for the feedback! Updated or left comments on everything, please take another look when you have a moment.

index.bs
NellWaliczek: I'm not sure I follow... can you elaborate a bit on that?

index.bs
NellWaliczek: Fair enough.

index.bs
```diff
@@ -1252,7 +1285,7 @@ A user agent MUST dispatch a <dfn event for="XRSession">selectend</dfn> event on
 
 A user agent MUST dispatch a <dfn event for="XRSession">select</dfn> event on an {{XRSession}} when one of its {{XRInputSource}}s has fully completed a [=primary action=]. The event MUST be of type {{XRInputSourceEvent}}.
 
-A user agent MUST dispatch a <dfn event for="XRFrameOfReference">boundschange</dfn> event on an {{XRFrameOfReference}} when the [=stage bounds=] change. This includes changes to the {{geometry}} point array or the {{bounds}} attribute changing to or from <code>null</code>. The event MUST be of type {{XRCoordinateSystemEvent}}.
+A user agent MUST dispatch a <dfn event for="XRReferenceSpace">reset</dfn> event on an {{XRReferenceSpace}} when the position or orientation of the origin changes relative to the user's environment. It also fires when the {{boundsGeometry}} changes for a {{XRBoundedReferenceSpace}}. The event MUST be of type {{XRReferenceSpaceEvent}}, and MUST be dispatched prior to the execution of any [=XR animation frame=]s that make use of the new origin.
```
NellWaliczek: Erk.  That's a bit too far in the other direction.  The way this is written, the event would fire practically every frame on a device with inside-out tracking ^_^

toji: 🤔 I get what you're saying, but I'm not sure I have better phrasing in mind? (Aside from maybe saying "*large* changes"?) Any suggestions?

index.bs
toji: Sure: https://github.com/immersive-web/editor-collab/issues/15

index.bs
toji: Alright, gave it another try:

> MUST dispatch a `reset` event on an `XRReferenceSpace` when discontinuities of the origin occur. (That is, significant changes in the origin's position or orientation relative to the user's environment.)

If you still feel it's not accurate I'd really appreciate some suggested verbiage.

NellWaliczek: I suspect that's as good as we're going to get

Issue #431, opened 2018-11-07
Provide a way for developer to supply their own transform offset.


darktears:
Here is a context in this demo : https://darktears.github.io/demos/xr/

In that example there are two cases where the user is moved by an offset : 

- Moving the user in space during in an inline/magic window scenario. In this demo it is achieved by a virtual navpad/joystick. Say you would want to explore your hotel room without going immersive.
- Teleporting/locomotion when in immersive mode. For example moving further than the stage bounds or in 3DoF being able to move in the world.

I'm sure there will be more use cases of artificial locomotion but I believe #2 will be extremely common to see amongst WebXR experiences.

One of the current problem with the API is that it's up to the developer to basically update the view matrix before the rendering to account for that "moving" offset. The math is not impossible but it's something that we don't want developers to have to do for each frames ([see here](https://github.com/darktears/darktears.github.io/blob/master/demos/xr/demo.js#L830)). Also it slightly different in the case of a magic window vs immersive mode ([see here](https://github.com/darktears/darktears.github.io/blob/master/demos/xr/demo.js#L780)). An additional burden for the developer is that he/she has to update the controller position as well (and that's yet another math operation on each frame, for e.g. [here](https://github.com/darktears/darktears.github.io/blob/master/demos/xr/demo.js#L999) and [here](https://github.com/darktears/darktears.github.io/blob/master/demos/xr/demo.js#L880))

I think we could simplify the life of developers by providing them a way to specify an offset matrix before they retrieve the view matrix. The UA will then be able to do all these math operations efficiently and transparently for the developer.

One possibility on the spec side would be to let the developer provide the offset inside the `XRFrameOfReference` as a member (most likely a Float32Array). When the developer would query the view matrix using `xrFrame.getViewerPose(xrFrameOfRef);` it's already adjusted.

This would work for the inputs as well when calling `xrFrame.getInputPose(inputSource, xrFrameOfRef);`. The only caveat here is in the IDL : `optional XRFrameOfReference frameOfReference`. This may create confusion is the developer doesn't provide the frame of reference.

@thetuvix mentioned to me at TPAC that OpenXR may have the concept of offset provided by the developer as well.

@toji @NellWaliczek thoughts?

modulesio:
The question I have is:
- is offset a property of the session?
- a property of the content/contexts in the session?
- or a thing passed in by the app's renderloop for easier math (what is suggested here)?

[...]

PR #440, opened 2018-12-11
Introduce XRRigidTransform

toji:
Part of a solution to address #431. We've also heard from several developers that they end up decomposing the view matrices we supply so they can apply a translation and rotation to their engine's camera primitive, which feels annoying and error prone. This PR proposes that we start representing most transforms in the API as an actual transform interface, which includes a translation vector and orientation quaternion. (Both as `DOMPointReadOnly`) It also provides a Float32Array matrix attribute (which can be lazily evaluated for performance reasons) as a convenience.

Using a transform interface like this will allow transform inputs into the API (such as for teleportation in the follow-up PR) to enforce constraints that a simple matrix would not, such as ensuring that there is no shearing or other weirdness that wouldn't work well in that context.

The XRView providing view and projection matrices direction is not changed in this PR. The view matrix is because this is about the only situation where the API would need to provide an inverted matrix, so it doesn't feel too bad as a one-off and avoids cluttering the `XRRigidTransform` with a more confusing, general case attribute like `invertedMatrix`. The projection matrix is still provided directly because it is still my understanding that some platforms need to express projections that don't fit cleanly into, say, an FOV struct.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -622,6 +622,7 @@ interface XRRigidTransform {
   readonly attribute XREye eye;
   readonly attribute Float32Array projectionMatrix;
   readonly attribute Float32Array viewMatrix;
+  readonly attribute XRRigidTransform transform;
 };
 
 [SecureContext, Exposed=Window] interface XRViewerPose {
--- a/explainer.md
+++ b/explainer.md
@@ -209,7 +209,7 @@ The `XRWebGLLayer`s framebuffer is created by the UA and behaves similarly to a 
 
 Once drawn to, the XR device will continue displaying the contents of the `XRWebGLLayer` framebuffer, potentially reprojected to match head motion, regardless of whether or not the page continues processing new frames. Potentially future spec iterations could enable additional types of layers, such as video layers, that could automatically be synchronized to the device's refresh rate.
 
-To get view matrices or the `poseMatrix` for each `XRFrame`, developers must call `getViewerPose()` and provide an `XRReferenceSpace` in which these matrices should be returned. Due to the nature of XR tracking systems, this function is not guaranteed to return a value and developers will need to respond appropriately.  For more information about what situations will cause `getViewerPose()` to fail and recommended practices for handling the situation, refer to the [Spatial Tracking Explainer](spatial-tracking-explainer.md).
+To get view matrices or the viewer's `transform` for each `XRFrame`, developers must call `getViewerPose()` and provide an `XRReferenceSpace` in which these matrices should be returned. Due to the nature of XR tracking systems, this function is not guaranteed to return a value and developers will need to respond appropriately.  For more information about what situations will cause `getViewerPose()` to fail and recommended practices for handling the situation, refer to the [Spatial Tracking Explainer](spatial-tracking-explainer.md).
 
 ```js
 function onDrawFrame(timestamp, xrFrame) {
@@ -240,7 +240,7 @@ function drawScene(view) {
   let viewMatrix = null;
   let projectionMatrix = null;
   if (view) {
-    viewMatrix = pose.viewMatrix;
+    viewMatrix = view.viewMatrix;
     projectionMatrix = view.projectionMatrix;
   } else {
     viewMatrix = defaultViewMatrix;
@@ -609,6 +609,15 @@ enum XREye {
   "right"
 };
 
+[SecureContext, Exposed=Window,
+ Constructor(optional DOMPointInit translation, optional DOMPointInit orientation)]
+interface XRRigidTransform {
+  readonly attribute DOMPointReadOnly translation;
+  readonly attribute DOMPointReadOnly orientation;
+
+  readonly attribute Float32Array matrix;
+};
+
 [SecureContext, Exposed=Window] interface XRView {
   readonly attribute XREye eye;
   readonly attribute Float32Array projectionMatrix;
@@ -616,7 +625,7 @@ enum XREye {
 };
 
 [SecureContext, Exposed=Window] interface XRViewerPose {
-  readonly attribute Float32Array poseMatrix;
+  readonly attribute XRRigidTransform transform;
   readonly attribute FrozenArray<XRView> views;
 };
 
--- a/explainer.md
+++ b/explainer.md
@@ -209,7 +209,11 @@ The `XRWebGLLayer`s framebuffer is created by the UA and behaves similarly to a 
 
 Once drawn to, the XR device will continue displaying the contents of the `XRWebGLLayer` framebuffer, potentially reprojected to match head motion, regardless of whether or not the page continues processing new frames. Potentially future spec iterations could enable additional types of layers, such as video layers, that could automatically be synchronized to the device's refresh rate.
 
-To get view matrices or the viewer's `transform` for each `XRFrame`, developers must call `getViewerPose()` and provide an `XRReferenceSpace` in which these matrices should be returned. Due to the nature of XR tracking systems, this function is not guaranteed to return a value and developers will need to respond appropriately.  For more information about what situations will cause `getViewerPose()` to fail and recommended practices for handling the situation, refer to the [Spatial Tracking Explainer](spatial-tracking-explainer.md).
+### Viewer tracking
+
+Each `XRFrame` the scene will be drawn from the perspective of a "viewer", which is the user or device viewing the scene, described by an `XRViewerPose`. Developers retrieve the current `XRViewerPose` by calling `getViewerPose()` on the `XRFrame` and providing an `XRReferenceSpace` for the pose to be returned in. Due to the nature of XR tracking systems, this function is not guaranteed to return a value and developers will need to respond appropriately.  For more information about what situations will cause `getViewerPose()` to fail and recommended practices for handling the situation, refer to the [Spatial Tracking Explainer](spatial-tracking-explainer.md).
+
+The `XRViewerPose` contains a `views` attribute, which is an array of `XRView`s. Each `XRView` has a `viewMatrix` and a `projectionMatrix` that should be used when rendering with WebGL. The `XRView` is also passed to an `XRWebGLLayer`'s `getViewport()` method to determine what the WebGL viewport should be set to when rendering. This ensures that the appropriate perspectives of scene are rendered to the correct portion on the `XRWebGLLayer`'s `framebuffer` in order to display correctly on the XR hardware.
 
 ```js
 function onDrawFrame(timestamp, xrFrame) {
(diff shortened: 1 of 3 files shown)


explainer.md
```diff
@@ -607,14 +607,24 @@ enum XREye {
   "right"
 };
 
+[SecureContext, Exposed=Window,
+ Constructor(optional DOMPointInit translation, optional DOMPointInit orientation)]
+interface XRRigidTransform {
+  readonly attribute DOMPointReadOnly translation;
+  readonly attribute DOMPointReadOnly orientation;
```
thetuvix: For these two attributes, should we pick a consistent naming pair?  To me, the two consistent pairs would be the following:
* `position`/`orientation`
* `translation`/`rotation`

The former implies a location within a base space, while the latter implies an explicit mutation of an entity's original transform.  Of these, the former makes a lot more sense to me as used in `...Pose` types like `XRViewerPose` and `XRInputPose`, as well as general world geometry types like `XRRay` - the developer will likely think of an entity's pose as a position and orientation within the general coordinate space in which they operate.

The key place that naming _could_ feel awkward is `XRSpace.getTransformTo`.  However, I've found that method name to often be confusing anyway - I, at least, need to do the mental exercise each time to remember which space should be the "from" space and which space should be the "to" space.  If we rename that method to `XRSpace.locateWithin(baseSpace)` or such, it could both make which space is which more clear and align even better with general `position`/`orientation` naming on `XRRigidTransform`.

explainer.md
```diff
@@ -607,14 +607,24 @@ enum XREye {
...
+
+  readonly attribute Float32Array matrix;
+};
+
 [SecureContext, Exposed=Window] interface XRView {
   readonly attribute XREye eye;
   readonly attribute Float32Array projectionMatrix;
   readonly attribute Float32Array viewMatrix;
+  readonly attribute XRRigidTransform transform;
```
thetuvix: Is `transform.matrix` always the same as `viewMatrix` here?

thetuvix: If so, perhaps we just combine these two attributes as `viewTransform`?

toji: No, that was not my intent (because otherwise it's just data duplication.) The transform here would not be inverted, and therefore the matrix wouldn't be either. So that makes it effectively the model matrix for the camera, and the view matrix that sits outside the interface would be the inverted version of the matrix.

This definitely needs to be explained better.

input-explainer.md
```diff
@@ -27,15 +27,15 @@ The properties of an XRInputSource object are immutable. If a device can be mani
...
+The `grip` is an `XRRigidTransform` into a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
 
-If the input source has only 3DOF, the grip matrix may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripMatrix` will be `null` if the input source isn't trackable. 
+If the input source has only 3DOF, the grip transform may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `grip` will be `null` if the input source isn't trackable.
```
thetuvix: Any given tracked entity (both viewers and input sources) can also _temporarily_ lose its 6DOF position.  For example, an inside-out tracked motion controller can move outside of the headset's tracking volume, but can still use its IMU to provide orientation updates.

For any given `XRRigidTransform`, we'll likely need to indicate if the position is a true detected position or some sort of "last known" position.  We could do this as either a boolean `positionTracked` attribute or by making `position` itself nullable - the former feels more like a "pit of success" for developers who don't consider such edge cases.

explainer.md
toji: Thanks for the detailed explanation. I don't think I've ever given it much thought, but when laid out this way I do agree with your interpretation of the terms. 'position/orientation' feel like appropriate names, so I'll make that change.

As for `XRSpace.getTransformTo()` we already know we need to refactor that so that it takes the frame into account somehow, so it shouldn't be a big deal to tweak the naming to be more fitting.

input-explainer.md
```diff
@@ -235,19 +233,19 @@ function onSelectStart(event) {
...
   let hitResult = scene.hitTest(ray.origin, ray.direction);
   if (hitResult && hitResult.draggable) {
-    // Use the gripMatrix translation to drag the intersected object, rather than the target ray.
+    // Use the grip translation to drag the intersected object, rather than the target ray.
     activeDragInteraction = {
       target: hitResult,
       targetStartPosition: hitResult.position,
       inputSource: event.inputSource,
-      inputSourceStartPosition: getTranslation(inputPose.gripMatrix);
+      inputSourceStartPosition: inputPose.grip.translation;
```
thetuvix: This line is a smell that `XRRigidTransform.position` naming would be more natural than `translation`, as this is being immediately assigned to an app attribute named `inputSourceStartPosition` - the app thinks of the return value here as a position.

input-explainer.md
```diff
@@ -342,7 +342,7 @@ interface XRInputSource {
 interface XRInputPose {
   readonly attribute boolean emulatedPosition;
   readonly attribute XRRay targetRay;
-  readonly attribute Float32Array? gripMatrix;
+  readonly attribute XRRigidTransform? grip;
```
thetuvix: `gripTransform` for better symmetry with `targetRay`?  This also makes it consistent that every attribute returning you a transform has `transform` in the name.

input-explainer.md
thetuvix: `XRInputPose.emulatedPosition` could handle this, but then we need an equivalent attribute on `XRViewerPose`.  Apps can also generically get into precisely the same situation with any relation between two `XRSpace`s.  It may make sense to just promote `emulatedPosition` onto `XRRigidTransform` in general, now that we have a richer transform type that could hold it.

explainer.md
toji: Updated some of the text around this (and ended up fleshing out some underexplained XRView concepts in the process.) Thanks for calling it out!

toji:
Update the PR to take into account feedback so far. Biggest changes are renaming `translation`->`position` and adding a fair bit of new explanatory text around both `XRRigidTransform` and `XRViewerPose`/`XRViews` because I found I was lacking a good contextual point to build the transform text off of.

input-explainer.md
blairmacintyre: Feels like this this is a simplified special case, to me.  The fact that this particular transform is used as something that feels like an absolute value (i.e., "position") versus something that feels like a relative value (i.e., "translation") shouldn't dictate the term.  

To me, a RigidTransform feels relative (i.e., translation / rotation) because they can be composed, even if it's not something that gets done in these special cases.

blairmacintyre:
See my comments above:  if these are called "transforms", I feel like the components should be the names that imply the composable attributes (translation / orientation).

That said, it probably doesn't matter.  Three.js, for example, uses position and orientation/quaternion.  So, those terms would at least be parallel with that common case.  

On the other hand, babylonjs also seems to mix up these terms (it uses (rotation, translation) as params to decompose)

toji:
Thanks for the thumbs up, Nell, and thanks for the feedback Blair and Alex!

Blair: See, now I'm frustrated because I understand both you and Alex's point of view on the names, and I don't have a principled disagreement with either. 😕

We can at least look at parallels in other native APIs to get a sense for if there's a popular term, though:
 - **Oculus:** `[ovrPosef::Position](https://developer.oculus.com/reference/libovr/1.30/structovr_posef/)`
 - **OpenVR:** Pose is delivered as a matrix, but elsewhere in the API they have `[VRBoneTransform_t::position](https://github.com/ValveSoftware/openvr/blob/master/headers/openvr.h#L1390)`
 - **WinMR:** Pose is delivered as a matrix.
 - **Daydream:** Pose delivered as a matrix.

So mixed opinions with a lot of APIs avoiding the question all together, but `position` does seem to come up more frequently in both the API surface and documentation.

toji:
Oh, and I guess I could also mention that OpenXR seems to favor the term "position" as well.

blairmacintyre:
I'm fine either way, so position/orientation is good.

input-explainer.md
thetuvix: Trying to give a timestamp may expose more nuance than we want to tackle right now, however.  For example:
* There may be a timestamp for the last sensor reading, but then the position is forward-predicted to this `XRFrame`'s photon time.  In that case, is the timestamp the timestamp of the last actual reading or the target timestamp of the prediction?
* If an inside-out-tracked motion controller stops being observed long enough (e.g. is behind your back), it might stop getting logical controller-to-head position updates, and so the developer would expect to check that by observing the timestamp is fixed from that point forward.  However, when asking where the controller is relative to the _world_, the UA should continue to incorporate updated head-to-world position updates, which means that the transform would continue to update every frame, even though it is now logically "estimated".  Would an app that gets back the same old timestamp again expect that it also gets precisely the same value for its transform?

For WebXR 1.0, the best balance here may just be a bool with some name like "estimatedPosition", "approximatePosition" or "inferredPosition".

explainer.md
```diff
@@ -251,6 +255,10 @@ function drawScene(view) {
 }
 ```
 
+The `XRViewerPose` also contains a `transform`, which is an `XRRigidTransform` describing the position and orientation of the viewer as a whole relative to the `XRReferenceSpace` origin. This is primarily useful for rendering a visual representation of the viewer for spectator views or multi-user environments. Each `XRView` has a `transform` as well that can be used in lieu of the `viewMatrix` to position virtual cameras in the scene if the rendering library being used prefers.
```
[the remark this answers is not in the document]
toji: I think this is good point for further discussion, but I would like to perhaps do so in a different issue so that the core of this PR can land.

I would like to clarify something to make sure we're talking about the same thing: If we were to make your proposed change the `viewTransform` would still effectively represent the "camera" position/orientation, correct? Therefore if someone wanted to get a `viewMatrix` they'd have to invert the `viewTransform.matrix`. That feels weird to me, but I do expect that the assertion about most apps simply using the transform component directly is probably accurate, so maybe it's a moot point.

input-explainer.md
toji: This is a really good line of discussion and I agree we should address it, but I've actually been working on something that feels like it would give us a cleaner place to put the `estimated` flags that I want to propose in this repo Real Soon Now™️, so I'd prefer to land this PR first and then work on this specific problem (which fortunately should be an additive thing either way.)

Filed issue #445 to ensure it gets tracked.

blairmacintyre: Sounds good to me.

explainer.md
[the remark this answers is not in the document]
thetuvix: > I would like to clarify something to make sure we're talking about the same thing: If we were to make your proposed change the `viewTransform` would still effectively represent the "camera" position/orientation, correct?

Yep!  The WebXR API then would not directly deliver an inverted left/right matrix for use as a view matrix - apps would instead construct their view matrices just as they would if no XR was involved at all: calculating them based on the app's camera position relative to the app's scene origin.  I have seen other new XR APIs increasingly take a similar approach to `viewTransform` data.

I'll create a new issue to discuss this.

PR #441, opened 2018-12-11
Add application-defined Reference space transforms ("Teleportation")

toji:
Follow up to #440, second part of a solution to #431. This PR adds the ability for developers to specify a transform to any `XRReferenceSpace` that will be applied to any sensor-controlled poses, effectively giving developers a simple method of "teleporting" the user around a scene. (That's a bit over-specific, since this also allows for things like touch/mouse rotation of magic window content and continuous stick motion if you have the stomach for it.)

Technically this is something that any dev can do themselves, but it means ensuring that you're consistently applying the transform to all your inputs, manually reversing it out for things like hit testing, and inverting, applying, and re-inverting in order to apply to view matrices. Thus handling it in the API is a significant convenience (that developers can still ignore if they are so inclined). It's also something that can be handled entirely at the API implementation level and won't need specialized logic for every native backend, so realistically the code will need to be written ~3 times and then everyone benefits forever. 

As mentioned in #440, the fact that this uses an `XRRigidTransform` rather than a raw `Float32Array` matrix is important because it allows us to impose constraints/validation on the types of transforms being done. No weird shearing, for example.

### DIFF
--- a/spatial-tracking-explainer.md
+++ b/spatial-tracking-explainer.md
@@ -178,11 +178,9 @@ function onSessionStarted(session) {
 ```
 
 ### Application-supplied transforms
-Frequently developers will want to provide an additional, artificial transform on top of the user's tracked motion to allow the user to navigate larger scenes than their tracking systems or physical space allows. An examples of this would be the common "teleportation" mechanic, where the user selects a point in the virtual scene to "jump" to, after which the selected point is treated as the new origin which all tracked motion is relative to. This effect is traditionally accomplished by mathematically combining the API-provided transform with the desired additional application transforms. It is useful to allow the API do handle it, however, because it ensures that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.
+Frequently developers will want to provide an additional, artificial transform on top of the user's tracked motion to allow the user to navigate larger scenes than their tracking systems or physical space allows. This effect is traditionally accomplished by mathematically combining the API-provided transform with the desired additional application transforms. It is useful to allow the API do handle it, however, because it ensures that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.
 
-Developers can specify application-specific transforms by setting the `baseTransform` attribute of any `XRReferenceSpace`. When the `transform` is not null, any values queried using the `XRReferenceSpace` will be offset by the `translation` and `orientation` the `baseTransform` describes. The `XRReferenceSpace`'s `baseTransform` can be updated at any time and will immediately take effect, meaning that any new values that are queried with the `XRReferenceSpace` will take into account the new `baseTransform`. Previously queried values will not be altered. Changing the `baseTransform` between pose queries in a single frame is not advised, since it will cause inconsistencies in the tracking data and rendered output.
-
-When interpreting an `XRRigidTransform` the `orientation` is applied prior to the `translation`. This means that, for example, a transform that indicates a quarter rotation to the right and a 1 meter translation along -Z would place a transformed object at `[0, 0, -1]` facing to the right.
+Developers can specify application-specific transforms by setting the `originTransform` attribute of any `XRReferenceSpace`. When the `transform` is not null, any values queried using the `XRReferenceSpace` will be offset by the `position` and `orientation` the `originTransform` describes. The `XRReferenceSpace`'s `originTransform` can be updated at any time and will immediately take effect, meaning that any new values that are queried with the `XRReferenceSpace` will take into account the new `originTransform`. Previously queried values will not be altered. Changing the `originTransform` between pose queries in a single frame is not advised, since it will cause inconsistencies in the tracking data and rendered output.
 
 The following example demonstrates how to create a `floor-level` reference space to account for a scene where the user is intended to start at a location 2 meters to the right and 3 meters back from the scene's origin.
 
@@ -195,12 +193,26 @@ function onSessionStarted(session) {
   xrSession.requestReferenceSpace({ type:'stationary', subtype:'floor-level' })
   .then((referenceSpace) => {
     xrReferenceSpace = referenceSpace;
-    xrReferenceSpace.baseTransform = new XRRigidTransform({ x: 2, z: 3 });
+    xrReferenceSpace.originTransform = new XRRigidTransform({ x: 2, z: 3 });
   })
   .then(setupWebGLLayer)
   .then(() => {
     xrSession.requestAnimationFrame(onDrawFrame);
   });
+}
+```
+
+Another common use case for this attribute would be for a "teleportation" mechanic, where the user "jumps" to a new point in the virtual scene, after which the selected point is treated as the new origin which all tracked motion is relative to.
+
+```js
+// Teleport the user a certain number of meters along the X, Y, and Z axes
+function teleport(deltaX, deltaY, deltaZ) {
+  let currentOrigin = xrReferenceSpace.originTransform;
+  xrReferenceSpace.originTransform = new XRRigidTransform(
+      { x: currentOrigin.position.x + deltaX,
+        y: currentOrigin.position.y + deltaY,
+        z: currentOrigin.position.z + deltaZ },
+      currentOrigin.orientation);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


spatial-tracking-explainer.md
```diff
@@ -177,6 +177,33 @@ function onSessionStarted(session) {
...
+  .then((referenceSpace) => {
+    xrReferenceSpace = referenceSpace;
+    xrReferenceSpace.baseTransform = new XRRigidTransform({ x: 2, z: 3 });
+  })
+  .then(setupWebGLLayer)
+  .then(() => {
+    xrSession.requestAnimationFrame(onDrawFrame);
+  });
+}
+```
```
thetuvix: If this static scene start offset was the only thing in play, we'd likely just suggest that scene authors shift their content around so that the static starting location is at (0, 0, 0).  The motivating bit comes in later on, when the user teleports within the scene.

It may be worth also showing a `function teleport(transform)` that later mutates the transform on `xrReferenceSpace`, reading it back, applying the offset and then setting it again.

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,33 @@ function onSessionStarted(session) {
...
+
+Developers can specify application-specific transforms by setting the `baseTransform` attribute of any `XRReferenceSpace`. When the `transform` is not null, any values queried using the `XRReferenceSpace` will be offset by the `translation` and `orientation` the `baseTransform` describes. The `XRReferenceSpace`'s `baseTransform` can be updated at any time and will immediately take effect, meaning that any new values that are queried with the `XRReferenceSpace` will take into account the new `baseTransform`. Previously queried values will not be altered. Changing the `baseTransform` between pose queries in a single frame is not advised, since it will cause inconsistencies in the tracking data and rendered output.
+
+When interpreting an `XRRigidTransform` the `orientation` is applied prior to the `translation`. This means that, for example, a transform that indicates a quarter rotation to the right and a 1 meter translation along -Z would place a transformed object at `[0, 0, -1]` facing to the right.
```
thetuvix: This is yet another benefit to `position`/`orientation` naming - developers have a natural instinct around what it means for an entity to be positioned and oriented within a coordinate system and thus would be unlikely to get this backwards.

thetuvix: Note that the explanation you provide here for how to interpret the two attributes describes an object that is placed (i.e. positioned) at (0, 0, -1) with a given orientation.  If the attributes had those two names directly, that would likely be immediately obvious to developers.

spatial-tracking-explainer.md
```diff
@@ -364,7 +391,9 @@ dictionary XRReferenceSpaceOptions {
   required XRReferenceSpaceType type;
 };
 
-[SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {
+[SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {  
+  attribute XRRigidTransform? baseTransform;
```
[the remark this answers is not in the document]
blairmacintyre: I can't recall who suggested offset, but I liked that too.  `originOffset` 

I do not see offset as implying translation only, but I can see how some might, though.

I'm not a fan of "base".

Or, `originOffset`?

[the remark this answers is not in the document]
toji: I suspect you intended to not suggest `originOffset` twice, so now I'm curious what the other suggestion was supposed to be? 😉 

In general, though, I'm seeing a lot of support for the term `origin` and more support than expected for some variant of `offset`, which likely means that my concern about it's implied meaning is overblown. Going to move for the moment to `originTransform` since it gets away from the `base` term that doesn't seem too popular and I've now heard it suggested from a couple of different sources (thanks @darktears!) but I could be persuaded to go with `originOffset` if that has more support.

spatial-tracking-explainer.md
toji: Good idea. Done!

spatial-tracking-explainer.md
toji: #440 has been updated to use `position/orientation` now. Thanks again for the suggestion.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
darktears: As it was highlighted in IRC yesterday (and thus my suggestion in the first place) origin is consistent with the web platform, see https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,45 @@ function onSessionStarted(session) {
 }
 ```
 
+### Application-supplied transforms
+Frequently developers will want to provide an additional, artificial transform on top of the user's tracked motion to allow the user to navigate larger scenes than their tracking systems or physical space allows. This effect is traditionally accomplished by mathematically combining the API-provided transform with the desired additional application transforms. It is useful to allow the API do handle it, however, because it ensures that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.
```
NellWaliczek: nit: navigate larger scenes  -> navigate larger virtual scenes

NellWaliczek: nit: API do handle it -> API to handle it

NellWaliczek: Nit: this is a bit confusing sounding to me...

>It is useful to allow the API do handle it, however, however, because it ensures that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.

How about this instead:

Nit: this is a bit confusing sounding to me...

> WebXR offers developers a simplification to ensure that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.

spatial-tracking-explainer.md
```diff
@@ -364,7 +403,9 @@ dictionary XRReferenceSpaceOptions {
   required XRReferenceSpaceType type;
 };
 
-[SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {
+[SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {  
+  attribute XRRigidTransform? originTransform;
```
NellWaliczek: Is there usefulness in this being nullable?  Or can we get away with just making it be identity by default?

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,45 @@ function onSessionStarted(session) {
...
+    xrReferenceSpace.originTransform = new XRRigidTransform({ x: 2, z: 3 });
+  })
+  .then(setupWebGLLayer)
+  .then(() => {
+    xrSession.requestAnimationFrame(onDrawFrame);
+  });
+}
+```
+
+Another common use case for this attribute would be for a "teleportation" mechanic, where the user "jumps" to a new point in the virtual scene, after which the selected point is treated as the new origin which all tracked motion is relative to.
```
NellWaliczek: nit: the new origin -> the new virtual origin

spatial-tracking-explainer.md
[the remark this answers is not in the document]
toji: I missed that in IRC yesterday, thanks for repeating it here! I definitely think consistency with the rest of the web platform is a positive thing!

spatial-tracking-explainer.md
toji: Seems reasonable to me!

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,45 @@ function onSessionStarted(session) {
 }
 ```
 
+### Application-supplied transforms
+Frequently developers will want to provide an additional, artificial transform on top of the user's tracked motion to allow the user to navigate larger virtual scenes than their tracking systems or physical space allows. This effect is traditionally accomplished by mathematically combining the API-provided transform with the desired additional application transforms. WebXR offers developers a simplification to ensure that all tracked values are transformed consistently, allowing the user's viewpoint and inputs to stay in sync.
```
NellWaliczek: nit: maybe give an example of two types of tracked values in the last sentence

spatial-tracking-explainer.md
```diff
@@ -177,6 +177,45 @@ function onSessionStarted(session) {
...
+
+Developers can specify application-specific transforms by setting the `originTransform` attribute of any `XRReferenceSpace`. When the `transform` is not null, any values queried using the `XRReferenceSpace` will be offset by the `position` and `orientation` the `originTransform` describes. The `XRReferenceSpace`'s `originTransform` can be updated at any time and will immediately take effect, meaning that any new values that are queried with the `XRReferenceSpace` will take into account the new `originTransform`. Previously queried values will not be altered. Changing the `originTransform` between pose queries in a single frame is not advised, since it will cause inconsistencies in the tracking data and rendered output.
```
NellWaliczek: >  When the `transform` is not null

This sentence needs to be fixed now that this matrix isn't nullable

NellWaliczek: > meaning that any new values that are queried with the `XRReferenceSpace` will take into account the new `originTransform`.

perhaps could be
> meaning that any subsequent poses queried with the `XRReferenceSpace` will take into account the new `originTransform`.

spatial-tracking-explainer.md
NellWaliczek: I'm actually not sure the previous example is super helpful.  The teleportation one below I'm all for!

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: I do somewhat prefer `originOffset` to `originTransform`, as it more strongly suggests that there is a natural origin that is meaningful in some way, but which you can offset relative to that natural location if you need to, e.g. to accomplish teleportation.

Most other attributes in the API named `transform` or `...Transform` can be though of fairly simply as "the absolute position and orientation of the thing within the space", which is not really what this relative offset is, and so perhaps it's good that the attribute name is a little different.

toji:
Addressed the latest round of feedback. Biggest change is that I have switched from `originTransform` to `originOffset` due to some of the well reasoned arguments that have been presented in it's favor, as well as simple numbers. (Nell, Alex, and Blair have voiced a preference for `offset`, while Alexis and myself have been the ones favoring `transform`)

PR #443, opened 2018-12-12
Use InvalidStateError for inactive XRFrame

### LINKED ISSUE #442: Inactive XRFrame should use InvalidStateError
Currently operations called on an inactive XRFrame throw "NotAllowedError" DOMException. I think that's more of a situation where the operation is not allowed due to a user denial in the context. It seems "InvalidStateError" would fit better.

jungkees:
This change makes an inactive XRFrame object throw an "InvalidStateError"
DOMException instead of a "NotAllowedError" DOMException. The rationale is we
throw an exception when an object is in an invalid state rather than when the
operation is denied by the user for example.

Fixes #442.


toji:
This makes perfect sense to me, and I'm honestly not sure why I didn't use `InvalidStateError` to begin with now that you've pointed it out. Thanks!

Gonna merge this now, as it seems pretty clearly like the right thing to do and it's easy enough to revert ig anyone had a really good reason for thinking otherwise.

PR #446, opened 2018-12-14
Added XRRigidTransform to the spec

toji:
Very straightforward addition of the XRRigidTransform type to the spec text. PTAL!


index.bs
```diff
@@ -613,7 +613,7 @@ An {{XRSpace}} describes an entity that is tracked by the [=XR device=]'s tracki
 When the <dfn method for="XRSpace">getTransformTo(|other|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. Let |current| be the target {{XRSpace}} object.
-  1. If a known transform exists from the space described by |current| to the space described by |other|, return it as a [=matrix=].
+  1. If a known transform exists from the space described by |current| to the space described by |other|, return it as an {{XRRigidTransform}}.
```
NellWaliczek: Ooof... this really makes me want to land dynamic spaces fix...

index.bs
```diff
@@ -879,19 +921,23 @@ XRRay {#xrray-interface}
...
 
-The <dfn attribute for="XRRay">transformMatrix</dfn> is a [=matrix=] which represents the transform from a ray originating at <code>[0, 0, 0]</code> and extending down the negative Z axis to the ray described by the {{XRRay}}'s {{XRRay/origin}} and {{direction}}.
+The <dfn attribute for="XRRay">matrix</dfn> attribute is a [=matrix=] which represents the transform from a ray originating at <code>[0, 0, 0]</code> and extending down the negative Z axis to the ray described by the {{XRRay}}'s {{XRRay/origin}} and {{direction}}.
+
+NOTE: The {{XRRay}}'s {{XRRay/matrix}} can be used to easily position graphical representations of the ray when rendering.
 
-NOTE: The {{XRRay}}'s {{XRRay/transformMatrix}} can be used to easily position graphical representations of the ray when rendering.
+ISSUE: Define {{XRRay}} constructor behaviors
```
NellWaliczek: Why postpone this?  Is it complicated or contentious?

toji: Ah, no, I intended this as effectively a "Note to self" that I was going to address before committing. Which, as strategies go, apparently leaves something to be desired. 😜

toji:
Thanks for the review! Addressed the issues and merged.

index.bs
```diff
@@ -815,21 +820,113 @@ WebXR provides various transforms in the form of <dfn lt="matrix">matrices</dfn>
...
+  readonly attribute Float32Array matrix;
+};
+</pre>
+
+<div class="algorithm" data-algorithm="construct-ray-origin-direction">
+
+The <dfn constructor for="XRRay">XRRay(|origin|, |direction|)</dfn> constructor MUST perform the following steps when invoked:
+
+  1. Let |ray| be a new {{XRRay}}.
+  1. If |origin| is not a {{DOMPointInit}} initialize |ray|'s {{XRRay/origin}} to <code>{ x: 0.0, y: 0.0, z: 0.0, w: 1.0 }</code>.
```
NellWaliczek: Actually... thinking on this again, should it be:
> 1. If |origin| is <code>null</code> initialize |ray|'s {{XRRay/origin}} to <code>{ x: 0.0, y: 0.0, z: 0.0, w: 1.0 }</code>.
> 1. If |origin| is not a {{DOMPointInit}} initialize throw an {{InvalidStateError}}

And same for {{XRRigidTransform}}?

toji: I will double check, but to know knowledge when you specify a parameter as optional it actually comes through as `undefined` instead of `null`, which is why I phrased it the way that I did (because I would want `undefined` and `null` to both cause an init to the default values.)

That says, your point still stands I think. It ought to be whatever the spec-language equivalent of:
```
if (!origin) { // default }
else if (origin instanceOf DOMPointInit) { // Set value }
else { throw }
```

But event then, now that I"m writing it out like that, `DOMPointInit` is actually a dictionary, so doing "instance of" doesn't really work. :( Lemme think about this some more.

PR #448, opened 2018-12-19
Added originOffset to XRReferenceSpace

toji:
Relatively simple addition of a single attribute. In the process found that we were previously describing a `XRReferenceSpace` attribute (`session`) that should have just been an internal variable. Re-arranged that to make more sense.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -572,7 +572,7 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method 
 
   1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw a {{InvalidStateError}} and abort these steps.
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -647,15 +650,19 @@ An {{XRReferenceSpace}} is obtained by calling {{XRSession/requestReferenceSpace
 
   - Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRReferenceSpaceType">unbounded</dfn> creates an {{XRUnboundedReferenceSpace}} instance if supported by the [=XR device=] and the {{XRSession}}.
 
-The <dfn for="XRReferenceSpace">session</dfn> attribute is the {{XRSession}} that created the {{XRReferenceSpace}}.
+Each {{XRReferenceSpace}} has a
```
NellWaliczek: Was this sentence supposed to have more in it?

toji: I... honestly have no idea what this was intended to say. 😅 Removed.

PR #450, opened 2018-12-19
baseLayer is nullable

Manishearth:
baseLayer is null after construction, so the attribute must be nullable


toji:
Great catch, thank you!

PR #451, opened 2018-12-20
Encapsulate session/frame render state into it's own interface

toji:
This change introduces an `XRRenderState` object that encapsulates any values that get picked up on frame boundaries, which hopefully makes it easier for developers to reason about how and when those values are updated. Updates happen with a `updateRenderState()` function that takes in a dictionary and returns a promise that resolves once the updated values have actually been applied.

One extra bit of feedback I'd like on this: I feel like "RenderState" is the right term for these values, but @NellWaliczek has expressed that "RenderOptions" feels like a better name, saying that "State" feels like something you'd have to re-set every frame. I could live with either name, but given the difference of opinion between the editors I'd love to hear from the rest of the WG if one or the other feels more appropriate to you!

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -134,7 +134,7 @@ Only one immersive session per XR hardware device is allowed at a time across th
 
 Once the session has started, some setup must be done to prepare for rendering.
 - An `XRReferenceSpace` should be created to establish a space in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
-- An `XRLayer` must be created and assigned to the `XRSession`'s `baseLayer` attribute. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
+- An `XRLayer` must be created and set as the `XRSession`'s `renderState.baseLayer`. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
 - Then `XRSession.requestAnimationFrame` must be called to start the render loop pumping.
 
 ```js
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -437,7 +440,8 @@ If the `framebufferScaleFactor` is set to a number higher or lower than `1.0` th
 function setupWebGLLayer() {
   return gl.makeXRCompatible().then(() => {
     // Create a WebGL layer with a slightly lower than default resolution.
-    xrSession.baseLayer = new XRWebGLLayer(xrSession, gl, { framebufferScaleFactor: 0.8 });
+    let glLayer = new XRWebGLLayer(xrSession, gl, { framebufferScaleFactor: 0.8 });
+    xrSession.updateRenderState({ baseLayer: glLayer });
```
NellWaliczek: Return the promise?

explainer.md
```diff
@@ -448,7 +452,8 @@ function setupNativeScaleWebGLLayer() {
   return gl.makeXRCompatible().then(() => {
     // Create a WebGL layer that matches the device's native resolution.
     let nativeScaleFactor = XRWebGLLayer.getNativeFramebufferScaleFactor(xrSession);
-    xrSession.baseLayer = new XRWebGLLayer(xrSession, gl, { framebufferScaleFactor: nativeScaleFactor });
+    let glLayer = new XRWebGLLayer(xrSession, gl, { framebufferScaleFactor: nativeScaleFactor });
+    xrSession.updateRenderState({ baseLayer: glLayer });
```
NellWaliczek: Return the promise?

explainer.md
```diff
@@ -600,6 +605,18 @@ enum XREnvironmentBlendMode {
...
+dictionary XRRenderStateOptions {
+  double depthNear;
+  double depthFar;
+  XRLayer baseLayer;
+};
+
+[SecureContext, Exposed=Window] interface XRRenderState {
+  readonly attribute double depthNear;
+  readonly attribute double depthFar;
+  readonly attribute XRLayer baseLayer;
```
NellWaliczek: Per the comment on the PR that was just filed by @Manishearth, this should be nullable, yes?

toji:
Thanks for the quick feedback! Addressed all of the above.

toji:
I haven't, primarily due to my web ergo muse (@bfgeek) being OOO recently.

Issue #453, opened 2018-12-20
Specify when we enter/exit immersive mode and what the baseLayer setter does


Manishearth:
It's unclear to me exactly when we're supposed to enter or exit immersive mode.

`requestSession()` returns a session, but without a `baseLayer` can it be in immersive mode? We should specify this.

Similarly, what happens when you set `baseLayer` to null? Or change it? The baseLayer setter needs to be filled out

kearwood:
I agree we should define what happens when baseLayer is null.  Perhaps a reasonable expectation would be to render black (VR) or transparent (AR)?

Manishearth:
Worth noting: baseLayer will always be null on initialization, ideally `requestSession` has a way to provide a baseLayer?


(If I'm assuming correctly that requestSession is what starts immersive mode)

NellWaliczek:
@toji this is related to the changes we've been working on to batch render state into a single object.  When that PR is posted, let's link it to this issue and make sure it addresses the behavior of the rendering when baseLayer is null.

Manishearth:
Is there any reason for being able to mutate baseLayer at all (as opposed to setting it during initialization the way WebVR handles it)?

Being able to change baseLayer complicates the RAF code.

asajeffrey:
In the current spec (https://immersive-web.github.io/webxr/#xrrenderstate-interface), `baseLayer` is `readonly`, and is initialized to `null`. I don't see anything that subsequently sets it to a non-null value.

Manishearth:
Right, the user currently manually sets it.

asajeffrey:
Er, how? The attribute is `readonly`.

Manishearth:
Oh, sorry, stuff changed. `updateRenderState` queues up new baseLayers now.

asajeffrey:
Ah, that's where it's hidden!

I fell down this rabbit-hole from #528, trying to work out whether the `XRSession` is allowed to provide a `XRWebGLRenderingContext` (which it could do in whatever GL context it pleases) or have to use one provided by the user (which will be in the browser's GL context).

toji:
`baseLayer` is mutatable (via `updateRenderState`, as pointed out by @Manishearth) because we know that manipulating the layers being used will be an important part of the API in the future, especially when used multiple layers can be used at once.

[...]

PR #458, opened 2019-01-03
Add XRRenderState to the spec

toji:
Also fixed a minor inconsistency in the explainer regarding the name of the associated dictionary.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -608,7 +608,7 @@ enum XREnvironmentBlendMode {
 dictionary XRRenderStateInit {
   double depthNear;
   double depthFar;
-  XRLayer baseLayer;
+  XRLayer? baseLayer;
 };
 
 [SecureContext, Exposed=Window] interface XRRenderState {
--- a/explainer.md
+++ b/explainer.md
@@ -587,7 +587,7 @@ dictionary XRSessionCreationOptions {
   attribute EventHandler onfocus;
   attribute EventHandler onend;
 
-  Promise<void> updateRenderState(XRRenderStateInit state);
+  Promise<void> updateRenderState(optional XRRenderStateInit state);
 
   long requestAnimationFrame(XRFrameRequestCallback callback);
   void cancelAnimationFrame(long handle);
@@ -605,7 +605,7 @@ enum XREnvironmentBlendMode {
   "alpha-blend",
 };
 
-dictionary XRRenderStateOptions {
+dictionary XRRenderStateInit {
   double depthNear;
   double depthFar;
   XRLayer baseLayer;
--- a/explainer.md
+++ b/explainer.md
@@ -159,7 +159,7 @@ function onSessionStarted(session) {
 
 ### Setting up an XRLayer
 
-The content to present to the device is defined by an `XRLayer`. In the initial version of the spec only one layer type, `XRWebGLLayer`, is defined and only one layer can be used at a time. This is set via the `XRSession`'s `updateRenderState()` function. `updateRenderState()` takes a dictionary containing new values for a variety of options affecting the session's rendering, including `baseLayer`. Only the options specified in the dictionary are updated, and a `renderstatechange` event is fired on the session when the new values take effect.
+The content to present to the device is defined by an `XRLayer`. In the initial version of the spec only one layer type, `XRWebGLLayer`, is defined and only one layer can be used at a time. This is set via the `XRSession`'s `updateRenderState()` function. `updateRenderState()` takes a dictionary containing new values for a variety of options affecting the session's rendering, including `baseLayer`. Only the options specified in the dictionary are updated.
 
 Future iterations of the spec will define new types of `XRLayer`s. For example: a new layer type would be added to enable use with any new graphics APIs that get added to the browser. The ability to use multiple layers at once and have them composited by the UA will likely also be added in a future API revision.
 
@@ -586,7 +586,6 @@ dictionary XRSessionCreationOptions {
   attribute EventHandler onblur;
   attribute EventHandler onfocus;
   attribute EventHandler onend;
-  attribute EventHandler onrenderstatechange;
 
   void updateRenderState(optional XRRenderStateInit state);
 
--- a/explainer.md
+++ b/explainer.md
@@ -159,7 +159,7 @@ function onSessionStarted(session) {
 
 ### Setting up an XRLayer
 
-The content to present to the device is defined by an `XRLayer`. In the initial version of the spec only one layer type, `XRWebGLLayer`, is defined and only one layer can be used at a time. This is set via the `XRSession`'s `updateRenderState()` function. `updateRenderState()` takes a dictionary containing new values for a variety of options affecting the session's rendering, including `baseLayer`. Only the options specified in the dictionary are updated, and a promise is returned that resolves when the new values take effect.
+The content to present to the device is defined by an `XRLayer`. In the initial version of the spec only one layer type, `XRWebGLLayer`, is defined and only one layer can be used at a time. This is set via the `XRSession`'s `updateRenderState()` function. `updateRenderState()` takes a dictionary containing new values for a variety of options affecting the session's rendering, including `baseLayer`. Only the options specified in the dictionary are updated, and a `renderstatechange` event is fired on the session when the new values take effect.
 
 Future iterations of the spec will define new types of `XRLayer`s. For example: a new layer type would be added to enable use with any new graphics APIs that get added to the browser. The ability to use multiple layers at once and have them composited by the UA will likely also be added in a future API revision.
 
@@ -180,7 +180,7 @@ function setupWebGLLayer() {
   return gl.makeXRCompatible().then(() => {
     // The content that will be shown on the device is defined by the session's
     // baseLayer.
-    return xrSession.updateRenderState({ baseLayer: new XRWebGLLayer(xrSession, gl) });
+    xrSession.updateRenderState({ baseLayer: new XRWebGLLayer(xrSession, gl) });
   });
 }
 ```
@@ -203,7 +203,7 @@ The WebXR Device API provides information about the current frame to be rendered
 
 A new `XRFrame` is created for each batch of `requestAnimationFrame()` callbacks or for certain events that are associated with tracking data. `XRFrame` objects act as snapshots of the state of the XR device and all associated inputs. The state may represent historical data, current sensor readings, or a future projection. Due to it's time-sensitive nature, an `XRFrame` is only valid during the execution of the callback that it is passed into, and once control is returned to the browser any active `XRFrame` objects are marked as inactive. Calling any method of an inactive `XRFrame` will throw an [`InvalidStateError`](https://heycam.github.io/webidl/#invalidstateerror).
 
-The `XRFrame` also makes a copy of the  `XRSession`'s `renderState`, such as `depthNear/Far` values and the `baseLayer`, at the time the first `requestAnimationFrame()` call in the current batch was made. This captured `renderState` is what will be used when computing view information like projection matrices and when the frame is being composited by the XR hardware. Any subsequent calls the developer makes to `updateRenderState()` will not be applied until the the next `XRFrame` is created.
+The `XRFrame` also makes a copy of the  `XRSession`'s `renderState`, such as `depthNear/Far` values and the `baseLayer`, just prior to the `requestAnimationFrame()` callbacks in the current batch being called. This captured `renderState` is what will be used when computing view information like projection matrices and when the frame is being composited by the XR hardware. Any subsequent calls the developer makes to `updateRenderState()` will not be applied until the next `XRFrame`'s callbacks are processed.
 
 The timestamp provided is acquired using identical logic to the [processing of `window.requestAnimationFrame()` callbacks](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#run-the-animation-frame-callbacks). This means that the timestamp is a `DOMHighResTimeStamp` set to the current time when the frame's callbacks begin processing. Multiple callbacks in a single frame will receive the same timestamp, even though time has elapsed during the processing of previous callbacks. In the future if additional, XR-specific timing information is identified that the API should provide it is recommended that it be via the `XRFrame` object.
 
@@ -441,7 +441,7 @@ function setupWebGLLayer() {
   return gl.makeXRCompatible().then(() => {
     // Create a WebGL layer with a slightly lower than default resolution.
     let glLayer = new XRWebGLLayer(xrSession, gl, { framebufferScaleFactor: 0.8 });
-    return xrSession.updateRenderState({ baseLayer: glLayer });
+    xrSession.updateRenderState({ baseLayer: glLayer });
   });
 ```
 
@@ -453,7 +453,7 @@ function setupNativeScaleWebGLLayer() {
     // Create a WebGL layer that matches the device's native resolution.
     let nativeScaleFactor = XRWebGLLayer.getNativeFramebufferScaleFactor(xrSession);
     let glLayer = new XRWebGLLayer(xrSession, gl, { framebufferScaleFactor: nativeScaleFactor });
-    return xrSession.updateRenderState({ baseLayer: glLayer });
+    xrSession.updateRenderState({ baseLayer: glLayer });
   });
 ```
 
@@ -586,8 +586,9 @@ dictionary XRSessionCreationOptions {
   attribute EventHandler onblur;
   attribute EventHandler onfocus;
   attribute EventHandler onend;
+  attribute EventHandler onrenderstatechange;
 
-  Promise<void> updateRenderState(optional XRRenderStateInit state);
+  void updateRenderState(optional XRRenderStateInit state);
 
   long requestAnimationFrame(XRFrameRequestCallback callback);
   void cancelAnimationFrame(long handle);
--- a/index.bs
+++ b/index.bs
@@ -304,12 +304,10 @@ enum XREnvironmentBlendMode {
   readonly attribute XRSessionMode mode;
   readonly attribute XRPresentationContext? outputContext;
   readonly attribute XREnvironmentBlendMode environmentBlendMode;
-
-  attribute double depthNear;
-  attribute double depthFar;
-  attribute XRLayer? baseLayer;
+  readonly attribute XRRenderState renderState;
 
   // Methods
+  Promise&lt;void&gt; updateRenderState(optional XRRenderStateInit state);
   Promise&lt;XRReferenceSpace&gt; requestReferenceSpace(XRReferenceSpaceOptions options);
 
   FrozenArray&lt;XRInputSource&gt; getInputSources();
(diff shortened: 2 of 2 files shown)


index.bs
```diff
@@ -368,7 +364,32 @@ The <dfn method for="XRSession">end()</dfn> method provides a way to manually sh
...
+
+<div class="algorithm" data-algorithm="apply-pending-render-states">
+
+When an {{XRSession}} has finished processing a batch of [=XR animation frames=], it MUST <dfn>apply pending render states</dfn> by running the following steps:
+
+  1. Let |session| be the target {{XRSession}}.
+  1. Let |renderStates| be a list of the entries in |session|'s [=list of pending render states=], in the order in which they were added to the list.
+  1. Set |session|'s [=list of pending render states=] to the empty list.
+  1. For each entry |newState| in |renderStates|, in order:
+    1. [=Update the render state=] with |newState|
```
NellWaliczek: On my first read-through I got confused between `[=Update the render state=]` which is defined as `update-renderstate` and the section above defining the behavior of `updateRenderState()` which is defined as `update-render-state`.  Can we rename one of them?  Perhaps the former should be `[=Apply individual render state=]` with the algorithm name of `apply-individual-render-state`?  Or something?

index.bs
```diff
@@ -455,6 +476,46 @@ dictionary XRSessionCreationOptions {
...
+
+</div>
+
+<div class="algorithm" data-algorithm="update-renderstate">
+
+When changes to an {{XRRenderState}} object |state| with a {{XRRenderStateInit}} dictionary |newState| are requested, the user agent MUST <dfn>update the render state</dfn> by running the following steps:
+
+  1. If |newState|'s {{XRRenderStateInit/depthNear}} value is set, set |state|'s {{XRRenderState/depthNear}} to |newState|'s {{XRRenderStateInit/depthNear}}.
+  1. If |newState|'s {{XRRenderStateInit/depthFar}} value is set, set |state|'s {{XRRenderState/depthFar}} to |newState|'s {{XRRenderStateInit/depthFar}}.
+  1. If |newState|'s {{XRRenderStateInit/baseLayer}} is set, set |state|'s {{XRRenderState/baseLayer}} to |newState|'s {{XRRenderStateInit/baseLayer}}.
```
NellWaliczek: Just to be clear "is set" includes if it's set to null, yes?

index.bs
```diff
@@ -503,10 +564,10 @@ When the <dfn method for="XRSession">cancelAnimationFrame(|handle|)</dfn> method
     1. If an exception is thrown, [=report the exception=].
   1. Set |session|'s [=processing frame=] to <code>false</code>.
   1. Set |frame|'s [=active=] boolean to <code>false</code>.
+  1. If |session|'s [=list of pending render states=] is not empty, [=apply pending render states=].
```
NellWaliczek: Hmm, I think this should probably be the first step so that each frame has the freshest data.

index.bs
toji: Yes, but I'm not clear on if there's a better way to communicate that? This verbiage is taken from another specs that had similar-ish behavior (though I'm embarrassed to admit at the moment that I don't recall which one?)

index.bs
toji: Yeah, the naming was more confusing than I realized. Thanks for pointing it out, new commit cleans that up a lot by reducing the number of separate algorithms used.

toji:
Nell mentioned in #453 that this PR is related.

index.bs
```diff
@@ -368,7 +364,27 @@ The <dfn method for="XRSession">end()</dfn> method provides a way to manually sh
...
+<div class="algorithm" data-algorithm="update-render-state">
+
+When the <dfn method for="XRSession">updateRenderState(|newState|)</dfn> method is invoked, the user agent MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
+
+  1. Let |session| be the target {{XRSession}}.
+  1. Let |pendingState| be |session|'s [=pending render state=].
+  1. If |newState|'s {{XRRenderStateInit/depthNear}} value is set, set |pendingState|'s {{XRRenderState/depthNear}} to |newState|'s {{XRRenderStateInit/depthNear}}.
+  1. If |newState|'s {{XRRenderStateInit/depthFar}} value is set, set |pendingState|'s {{XRRenderState/depthFar}} to |newState|'s {{XRRenderStateInit/depthFar}}.
+  1. If |newState|'s {{XRRenderStateInit/baseLayer}} is set, set |pendingState|'s {{XRRenderState/baseLayer}} to |newState|'s {{XRRenderStateInit/baseLayer}}.
+  1. If |session|'s [=processing frame=] flag is <code>false</code> and |session|'s [=list of animation frame callbacks=] is empty, copy |session|'s [=pending render state=] to |session|'s [=active render state=] and [=/Resolve=] |promise|.
```
RafaelCintron: I am confused by the need for this step.  Does it imply that doing the following:

> session.requestAnimationFrame();
session.updateRenderState(…);
let depthNear = session.renderState.depthNear;

would return a different value for depthNear than: 

> session.updateRenderState(…);
session.requestAnimationFrame();
let depthNear = session.renderState.depthNear;

If so, it seems unexpected to say the least.

To make things more consistent, it seems like we want the requested render state to "lock in" right before Session.RAF callbacks are executed, and therefore always after the current JavaScript execution is finished.

index.bs
```diff
@@ -368,7 +364,27 @@ The <dfn method for="XRSession">end()</dfn> method provides a way to manually sh
...
+  1. Else append |promise| to |session|'s [=list of pending render state promises=].
+
+</div>
+
+NOTE: The [=active render state=] of an {{XRSession}} is constant for the full duration of an [=XR animation frame=], and is defined as the [=pending render state=] when the first {{XRSession/requestAnimationFrame()}} call that initiates the [=XR animation frame=] is made.
```
RafaelCintron: What does this note imply about calling updateRenderState while inside of a windows.RAF, or other JavaScript callback?  Why is an XR animation frame special with regard to active render state being constant?

index.bs
```diff
@@ -491,10 +537,13 @@ When the <dfn method for="XRSession">cancelAnimationFrame(|handle|)</dfn> method
...
 <div class="algorithm" data-algorithm="run-animation-frames">
 
-  When an {{XRSession}} |session| receives updated [=viewer=] state from the [=XR device=], it runs an <dfn>XR animation frame</dfn> with a timestamp |now| and an {{XRFrame}} |frame|, which MUST run the following steps:
+When an {{XRSession}} |session| receives updated [=viewer=] state from the [=XR device=], it runs an <dfn>XR animation frame</dfn> with a timestamp |now| and an {{XRFrame}} |frame|, which MUST run the following steps:
 
   1. Let |callbacks| be a list of the entries in |session|'s [=list of animation frame callback=], in the order in which they were added to the list.
   1. Set |session|'s [=list of animation frame callbacks=] to the empty list.
+  1. Let |newRenderState| be a copy of |session|'s [=pending render state=].
+  1. Let |renderStatePromises| be a list of the entries in |session|'s [=list of pending render state promises=], in the order in which they were added to the list.
+  1. Set |session|'s [=list of pending render state promises=] to the empty list.
```
RafaelCintron: Does setting the session's list of pending render state promises to an empty list mean we resolve the promise at this step?  If not, when do we resolve the promise?

How does the ordering of the promise resolution interleave with the RAF callbacks?  Could we end up in a state where the `updateRenderState` promise resolves after the next frame's Session.RAF callback?  If so, this seems unnatural as the web developer would likely expect that the "render state accepted" promise would resolve BEFORE they receive the RAF callback with the render state applied.  

What would be the downside of dispensing with the promise altogether and have `updateRenderState` overwrite the pending render state when called?  Pending render state becomes active render state right before RAF callbacks are serviced.

kearwood:
This generally LGTM.  In addition to comments by @NellWaliczek, I might suggest that we define explicitly what happens if "baseLayer" is null.  By extension, this would define what the behavior would be prior to the first call to updateRenderState.

May I propose that we simply define the behavior of this initial state as "rendering blackness" for VR and "rendering fully transparent" for AR?

explainer.md
```diff
@@ -605,7 +605,7 @@ enum XREnvironmentBlendMode {
   "alpha-blend",
 };
 
-dictionary XRRenderStateOptions {
+dictionary XRRenderStateInit {
   double depthNear;
   double depthFar;
```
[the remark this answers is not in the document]
thetuvix: Even then, the API could accept `depthNear`/`depthFar` at the end of the frame, rather than during the previous frame.

index.bs
```diff
@@ -455,6 +471,36 @@ dictionary XRSessionCreationOptions {
...
+  readonly attribute double depthFar;
+  readonly attribute XRLayer? baseLayer;
+};
+</pre>
+
+<div class="algorithm" data-algorithm="initialize-renderstate">
+
+When an {{XRRenderState}} object is created, the user agent MUST <dfn>initialize the render state</dfn> by running the following steps:
+
+  1. Let |state| be the newly created {{XRRenderState}} object.
```
Manishearth: We can just set defaults in the webidl here

index.bs
```diff
@@ -455,6 +471,36 @@ dictionary XRSessionCreationOptions {
...
+[SecureContext, Exposed=Window] interface XRRenderState {
+  readonly attribute double depthNear;
+  readonly attribute double depthFar;
+  readonly attribute XRLayer? baseLayer;
+};
+</pre>
+
+<div class="algorithm" data-algorithm="initialize-renderstate">
+
+When an {{XRRenderState}} object is created, the user agent MUST <dfn>initialize the render state</dfn> by running the following steps:
```
Manishearth: I don't think we use this algorithm?

Also, this is better achieved by setting these defaults in the dictionary's webidl directly imo

index.bs
```diff
@@ -455,6 +471,36 @@ dictionary XRSessionCreationOptions {
...
+XRRenderState {#xrrenderstate-interface}
+-------------
+
+There are multiple values that developers can configure which affect how the session's output is composited. These values are tracked by an {{XRRenderState}} object.
+
+<pre class="idl">
+dictionary XRRenderStateInit {
+  double depthNear;
+  double depthFar;
+  XRLayer baseLayer;
```
Manishearth: Is this supposed to be non-nullable?

index.bs
NellWaliczek: The reason this isn't set on the dictionary is that the dictionary elements are optional.  When `updateRenderState()` is invoked, only the values that are present are applied.

explainer.md
```diff
@@ -587,7 +587,7 @@ dictionary XRSessionCreationOptions {
   attribute EventHandler onfocus;
   attribute EventHandler onend;
 
-  Promise<void> updateRenderState(XRRenderStateInit state);
+  Promise<void> updateRenderState(optional XRRenderStateInit state);
```
NellWaliczek: Sorry if I'm not remembering something... why is this optional?

explainer.md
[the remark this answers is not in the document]
NellWaliczek: Are you asking for a change or this is a +1 or something else?

index.bs
```diff
@@ -368,6 +364,26 @@ The <dfn method for="XRSession">end()</dfn> method provides a way to manually sh
 
 </div>
 
+Each {{XRSession}} has an <dfn>active render state</dfn> which is a new {{XRRenderState}}, a <dfn>pending render state</dfn> which is a new {{XRRenderState}}, and a <dfn>list of pending render state promises</dfn>, which is initially empty.
```
NellWaliczek: > is a new {{XRRenderState}}

new?

index.bs
```diff
@@ -368,6 +364,26 @@ The <dfn method for="XRSession">end()</dfn> method provides a way to manually sh
...
+<div class="algorithm" data-algorithm="update-render-state">
+
+When the <dfn method for="XRSession">updateRenderState(|newState|)</dfn> method is invoked, the user agent MUST return [=a new Promise=] |promise| and run the following steps [=in parallel=]:
+
+  1. Let |session| be the target {{XRSession}}.
+  1. Let |pendingState| be |session|'s [=pending render state=].
+  1. If |newState|'s {{XRRenderStateInit/depthNear}} value is set, set |pendingState|'s {{XRRenderState/depthNear}} to |newState|'s {{XRRenderStateInit/depthNear}}.
+  1. If |newState|'s {{XRRenderStateInit/depthFar}} value is set, set |pendingState|'s {{XRRenderState/depthFar}} to |newState|'s {{XRRenderStateInit/depthFar}}.
+  1. If |newState|'s {{XRRenderStateInit/baseLayer}} is set, set |pendingState|'s {{XRRenderState/baseLayer}} to |newState|'s {{XRRenderStateInit/baseLayer}}.
+  1. If |session|'s [=processing frame=] flag is <code>false</code> and |session|'s [=list of animation frame callbacks=] is empty, copy |session|'s [=pending render state=] to |session|'s [=active render state=] and [=/Resolve=] |promise|.
```
NellWaliczek: I'm inclined to have the spec always put the promise in the queue and defer switching the pending render state to be the active render state to the algorithm for invoking the next batch of frame callbacks. wdyt?

index.bs
```diff
@@ -364,30 +364,25 @@ The <dfn method for="XRSession">end()</dfn> method provides a way to manually sh
...
-
-  1. Let |session| be the target {{XRSession}}.
-  1. Let |renderStates| be a list of the entries in |session|'s [=list of pending render states=], in the order in which they were added to the list.
-  1. Set |session|'s [=list of pending render states=] to the empty list.
-  1. For each entry |newState| in |renderStates|, in order:
-    1. [=Update the render state=] with |newState|
-    1. [=/Resolve=] the promise associated with |newState|.
-
-</div>
+NOTE: The [=active render state=] of an {{XRSession}} is constant for the full duration of an [=XR animation frame=], and is defined as the [=pending render state=] when the first {{XRSession/requestAnimationFrame()}} call that initiates the [=XR animation frame=] is made.
```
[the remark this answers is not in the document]
NellWaliczek: Now that you've changed the algorithm above, this note doesn't seem quite accurate.

index.bs
NellWaliczek: > Does setting the session's list of pending render state promises to an empty list mean we resolve the promise at this step? If not, when do we resolve the promise?

I second this question from Rafael :)

NellWaliczek: >What would be the downside of dispensing with the promise altogether and have updateRenderState overwrite the pending render state when called? Pending render state becomes active render state right before RAF callbacks are serviced.

The purpose of the promise is just to let developers know when the changes have been applied.  It's entirely possible no one will use the promise returned, but I'm not certain. That said, I think Brandon's latest changes mostly address the rest of your point.

index.bs
Manishearth: Ah, I misinterpreted this, my bad.

We should explicitly mention that this is invoked during XRSession construction

index.bs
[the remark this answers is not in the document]
Manishearth: > I propose we wait to apply render state until just before the first RAF callback of the next frame is invoked.

FWIW this is how I'm implementing it in Servo, it seemed to be the approach that made the most sense

explainer.md
toji: It's a silly WebIDL thing. It has to be optional because none of the members of the dictionary are required. 🤷‍♀️

index.bs
toji: Sorry, I realized that I had some uncommitted changes with my last code push that change this logic up a bit. You'll want to take a quick look again.

As for always deferring the state change till the next rAF callback, I didn't feel like that was appropriate because you may have a session that you're not actively pumping the render loop for (maybe it's suspended?) that you don't want pending state changes to stack up on forever. The latest algorithm changes I've made state that the changes always immediately apply unless you're in the middle of a batch of frame callbacks, at which point the changes are deferred until the frame callbacks have finished.

Manishearth: > that you don't want pending state changes to stack up on forever

You kinda need this anyway, an implementation will need a separate thread adding rAF events to the event loop, and when the session is suspended (e.g. the layer is null) it probably doesn't want to keep triggering rAF and needs some wakeup signaling mechanism.

explainer.md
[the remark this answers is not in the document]
thetuvix: Just setting context as we revisit `XRRenderState*` here that we may be able to do away with `depthNear`/`depthFar`, if that simplifies the reshuffle here.  Some platforms had previously needed the near/far planes to be set on the previous frame, but this is no longer the case for Windows MR.

index.bs
RafaelCintron: I understand that the purpose is to let developers know when the changes have been applied.  However, as the spec is currently written, the developer would receive the promise resolution after the RAF callback with the new values has been fired.  This is true even in the latest revision.

Put another way, a naïve developer would call updateRenderState and then render with the old values until they received a promise resolution signifying the 'all clear'. With the current spec language, they would end up rendering one or more frames with the wrong values.

toji:
Made (hopefully) one last simplification to the render state logic based on Nell's most recent feedback offline. Render state changes are now delayed until just prior to the frame's callbacks being run, at which point they get flushed out in order and all new render state changes get deferred to the next frame. To facilitate state changes not piling up forever it's specified that the animation frame algorithm should run for every VR frame regardless of whether or not any rAF callbacks have been registered.

index.bs
```diff
@@ -304,12 +304,10 @@ enum XREnvironmentBlendMode {
...
   readonly attribute XRPresentationContext? outputContext;
   readonly attribute XREnvironmentBlendMode environmentBlendMode;
-
-  attribute double depthNear;
-  attribute double depthFar;
-  attribute XRLayer? baseLayer;
+  readonly attribute XRRenderState renderState;
 
   // Methods
+  Promise&lt;void&gt; updateRenderState(optional XRRenderStateInit state);
```
RafaelCintron: Not super critical for this change but we should have spec language that states when this function throws an exception on bad input. For instance, if the baseLayer passed in does not belong to the current session, that should throw an exception.

index.bs
```diff
@@ -1197,7 +1251,7 @@ function onXRSessionStarted(xrSession) {
 
   loadWebGLResources();
 
-  xrSession.baseLayer = new XRWebGLLayer(xrSession, gl);
+  xrSession.updateRenderState({ baseLayer: new XRWebGLLayer(xrSession, gl) });
```
RafaelCintron: Sample code should include a .then handler with at least a comment describing when it would be good to do something on the resolve.

toji:
Thanks for your comments, @RafaelCintron, sorry I've been bad about addressing them directly as the language here is being iterated on. Your questions about the usage of promises here, though, caused myself and Nell to realize that there's at least one known case coming up where we want the render state to be changed by the system, not the user. (Specifically, we want to make the `outputContext` part of the render state and allow it to move between sessions.) As a result, and due to the fact that as stated now render state changes are always guaranteed to apply before the next batch of rAF callbacks, it seemed appropriate to move from a promise-based to an event-based notification of when changes are applied. Happily, this actually simplifies the spec logic even further!

Hoping the latest changes feel good to everyone that's been tracking this PR!

index.bs
NellWaliczek: @RafaelCintron has a really great point on this.  Whether we are using an event or a promise, if we can't come up with reasonable sample code for how it would be used, then that's a pretty good sign that we shouldn't include it right now.  It's much easier to add things to a standard than to remove them ;)
And TBH, I expect most people will just look at the frame's render state to decide what to do.  So until we hear otherwise, perhaps we should actually remove the event/promise altogether.

NellWaliczek: (sorry this PR has pingponged so much!)

toji: tenor

toji: (Okay. Yes. That a very good point and I'm always happy to reduces API surface... :grimacing:)

explainer.md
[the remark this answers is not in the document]
RafaelCintron: In the latest iteration, I noticed that all of the sample code only updated the `baseLayer` of the render state and never the `depthNear/depthFar`.  Was this done for sample code brevity or because you're proposing we remove depthNear and depthFar.  If it is the latter, then you need to remove the two variables from the explanatory text and IDL as well.  If the former, then you're good.

RafaelCintron:
Thank you for the clarification, @toji .  The latest iteration LGTM.  I look forward to reviewing the upcoming event based PR.

explainer.md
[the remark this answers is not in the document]
toji: Just for code brevity.

toji:
Thanks Rafael! Merged so that we can keep the focus on other items at the F2F. 😀

PR #459, opened 2019-01-04
 Describe the identity reference space

toji:
This adds a new reference space type that takes the place of passing `null` as the mechanism to return identity poses. Move to handle it with a new space like this was due the desire to allow the `originOffset` to still be used even in situations with no tracking.

Nell and I had a question about whether or not the included code sample is something that belongs in the spec and/or explainer, since it's a pretty specific case and the purpose of this doc is not to teach anyone quaternion math. Still, it seems like a good demonstration of why this new type matters. Thoughts?

### DIFF
--- a/spatial-tracking-explainer.md
+++ b/spatial-tracking-explainer.md
@@ -199,7 +199,7 @@ function teleport(deltaX, deltaY, deltaZ) {
 ### Identity reference spaces
 An _identity_ reference space is one which provides no tracking. This type of reference space is used for creating inline experiences with tracking information explicitly disabled, such as a furniture viewer that will use [click-and-drag controls](#click-and-drag-view-controls) to rotate the furniture. It also supports cases where the developer wishes to avoid displaying any type of tracking consent prompt to the user prior while displaying inline content.
 
-This type of reference space is requested with a type of `identity` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`. The pose will only contain a single `XRView`, which must also have a transform equal to the `originOffset`. `identity` references spaces cannot be created by immersive sessions, since a lack of tracking in an immersive experience could cause severe discomfort for the user.
+This type of reference space is requested with a type of `identity` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`. The pose will only contain a single `XRView`, which must also have a transform equal to the `originOffset`.
 
 ```js
 let xrSession = null;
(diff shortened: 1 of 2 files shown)


NellWaliczek:
It was pointed out to me by a colleague that there could be valid reasons to use an Identity Reference Space along with one of the other types in an Immersive Session.  For example, as a 3D hud.  How would you feel about switching to allowing this?

toji:
That makes sense, and it's a pretty minor change, so sure! I'll update the text in just a moment.

Artyom17:
Looks good to me!

toji:
I highly doubt that we're going to get any comments voicing concerns at this point, so I'm pulling the trigger on the merge. Thanks everyone!

spatial-tracking-explainer.md
```diff
@@ -196,12 +196,63 @@ function teleport(deltaX, deltaY, deltaZ) {
 }
 ```
 
+### Identity reference spaces
+An _identity_ reference space is one which provides no tracking. This type of reference space is used for creating inline experiences with tracking information explicitly disabled, such as a furniture viewer that will use [click-and-drag controls](#click-and-drag-view-controls) to rotate the furniture. It also supports cases where the developer wishes to avoid displaying any type of tracking consent prompt to the user prior while displaying inline content.
+
+This type of reference space is requested with a type of `identity` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`. The pose will only contain a single `XRView`, which must also have a transform equal to the `originOffset`.
```
thetuvix: > `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`.

Isn't this the reverse of what we defined `originOffset` to do for all other reference spaces?

**For the `"bounded"` space:**
Imagine that the viewer is located at position (2, 2, 2) within the default `"bounded"` reference space.  If I then "offset" the bounded reference space's "origin" by (2, 2, 2), I would expect the viewer to now be at position (0, 0, 0), since I shifted that space's origin over by (2, 2, 2) to now coincide with the viewer.

This means that **applying an `originOffset` of (2, 2, 2)** to the `"bounded"` reference space causes the viewer's pose to **translate by (-2, -2, -2)**.

**For the `"identity"` space:**
However, per this PR, if I apply an `originOffset` of (2, 2, 2) to the `"identity"` reference space, the viewer's position changes from (0, 0, 0) to (2, 2, 2).

This means that **applying an `originOffset` of (2, 2, 2)** to the `"identity"` reference space causes the viewer's pose to **translate by (2, 2, 2)**, the opposite behavior as with the `"bounded"` reference space.

Issue #445, opened 2018-12-13
Better indicate when elements of a pose are estimated


toji:
In #440 there was some discussion about the need to better flag when certain elements of a pose, such as the `position` are based on estimations rather than observed sensor data. Specifically, see [these comments](https://github.com/immersive-web/webxr/pull/440#discussion_r240743420) between @thetuvix and @blairmacintyre.

We already have an `emulatedPosition` bool in `XRInputPose`, but as Alex points out we don't have an equivalent for head poses and we should be more consistent about it. The suggestion within the thread is to promote the estimated flag into `XRRigidTransform`, which is workable. I've got an alternate idea for where the flags could be hosted that would make the ergonomics of creating transform for things like teleportation cleaner, which I'll try to put up in this repo soon for discussion. Either way, it's a good point that should be addressed.

NellWaliczek:
Fixed by #460

PR #460, opened 2019-01-04
Handle pose queries in a more generalized fashion

toji:
This PR is intended to pave the way for making it easier to introduce new types of pose tracking functionality in the future by making how we query poses for almost everything more generalized. (The viewer pose is still special because it's the only thing that needs the `views` array, but even there this change makes it less of an outlier.) Ideally this change will make it much easier to reason about adding things like anchors in the future, and also serves to set the stage for #384 as well as fixing the frame-bound concern brought up in that issue.

### DIFF
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -25,15 +25,17 @@ The properties of an XRInputSource object are immutable. If a device can be mani
 
 ### Input poses
 
-Each input source can query a `XRInputPose` using the `getInputPose()` function of any `XRFrame`. Getting the pose requires passing in the `XRInputSource` you want the pose for, as well as the `XRReferenceSpace` the pose values should be given in, just like `getViewerPose()`. `getInputPose()` may return `null` in cases where tracking has been lost (similar to `getViewerPose()`), or the given `XRInputSource` instance is no longer connected or available.
+Each input source provides two `XRSpace`s, which can be used to query an `XRPose` using the `getPose()` function of any `XRFrame`. Getting the pose requires passing in the `XRSpace` you want the pose for, as well as the `XRSpace` the returned pose should be relative to (which may be an `XRReferenceSpace`). Just like `getViewerPose()`, `getPose()` may return `null` in cases where tracking has been lost, or the `XRSpace`'s `XRInputSource` instance is no longer connected or available.
 
-The `gripTransform` is an `XRRigidTransform` into a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
+The `gripSpace` represents a space where if the user was holding a straight rod in their hand it would be aligned with the negative Z axis (forward) and the origin rests at their palm. This enables developers to properly render a virtual object held in the user's hand. For example, a sword would be positioned so that the blade points directly down the negative Z axis and the center of the handle is at the origin.
 
-If the input source has only 3DOF, the `gripTransform` may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripTransform` will be `null` if the input source isn't trackable. 
+If the input source has only 3DOF, the grip pose may represent only a translation or rotation based on tracking capability. An example of this case is for physical hands on some AR devices which only have a tracked position. The `gripSpace` will be `null` if the input source isn't trackable. 
 
-An input source will also provide its preferred pointing ray, given by the `XRInputPose`'s `targetRay`. The ray, which is an `XRRay` object, includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. A matrix can also be queried from the `XRRay` with the `matrix` attribute which represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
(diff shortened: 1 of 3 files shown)


NellWaliczek:
This PR fixes issues #445 and #384.

input-explainer.md
```diff
@@ -230,22 +235,18 @@ function onSelectStart(event) {
...
+
+  let targetRayPose = event.frame.getPose(event.inputSource.targetRaySpace, xrReferenceSpace);
 
   // Use the input source target ray to find a draggable object in the scene
-  let ray = inputPose.targetRay;
-  let hitResult = scene.hitTest(ray.origin, ray.direction);
+  let hitResult = scene.hitTest(new XRRay(targetRayPose.transform));
   if (hitResult && hitResult.draggable) {
-    // Use the grip position to drag the intersected object, rather than the target ray.
+    // Use the targetRayPose position to drag the intersected object.
```
thetuvix: Is the change from dragging at the grip pose to dragging at the target ray origin intentional here?  Is the idea to support dragging with a broader set of input sources that may not have grip poses?

In practice, I wonder if 3DoF VR controllers that provide an emulated "shoot from the hip" target ray origin would supply an emulated grip pose as well, if only to allow for rendering of a controller model that matches the emulated target ray.

thetuvix: Note that for a "shoot from the hip" 3DoF controller that only has a target ray and no grip pose, the target ray origin is not going to be very useful for dragging either, since it will basically be at the hip as well, and the furthest the user could drag an object by moving their hand would be the radius of the controller.

In that case, I'd suggest that conditioning on the availability of `gripSpace` here is more correct for dragging - if a controller doesn't have a grip pose, it's almost certainly not going to give you enough range of target ray origin 6DoF motion to actually perform a drag operation.  The app would need to implement some alternate dragging interaction for that case, perhaps based on the end of the target ray rather than its origin.

spatial-tracking-explainer.md
```diff
@@ -349,24 +349,21 @@ partial interface XRSession {
...
-  readonly attribute XRRigidTransform? gripTransform;
 };
 
 //
 // Space
 //
 
 [SecureContext, Exposed=Window] interface XRSpace : EventTarget {
-  XRRigidTransform? getTransformTo(XRSpace other);
+  // Intentionally an opaque type, used by XRInputSource.
```
thetuvix: "used by XRFrame"?

NellWaliczek:
I've got a PR coming (hopefully this week) that cleans up this sample code pretty significantly.  If you'd prefer we switch it back in this particular PR, I'm not opposed.

thetuvix:
This is a great generalization to how apps reason over poses and spaces, setting up WebXR to support the hit-test, anchors and compositor layer scenarios discussed in #384.

thetuvix:
> I've got a PR coming (hopefully this week) that cleans up this sample code pretty significantly. If you'd prefer we switch it back in this particular PR, I'm not opposed.

I'm fine either way.  Just pointed out these particular lines since they were changed as part of this PR.

input-explainer.md
```diff
@@ -25,38 +25,40 @@ The properties of an XRInputSource object are immutable. If a device can be mani
...
+An input source will also provide its preferred pointing ray, given by the `XRInputSource`'s `targetRaySpace`. This can be used to get the `XRPose` for the input's target rays.
 
-The `targetRay` will never be `null`. It's value will differ based on the type of input source that produces it, which is represented by the `targetRayMode` attribute:
+An `XRRay` can be constructed from the `XRPose`'s `transform`. Rays includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. The `XRRay`'s `matrix` represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
```
kearwood: nit: "Rays include" or "A ray includes"

spatial-tracking-explainer.md
toji: No, this was meant to imply that the type was used as an attribute on XRInputSource, but I can see why there was some confusion. I'll simplify it to say just that it's intentionally opaque.

toji:
Feedback from Kip and Alex has been addressed and the PR has been rebased. Ready to merge?

input-explainer.md
```diff
@@ -33,7 +33,7 @@ If the input source has only 3DOF, the grip pose may represent only a translatio
...
 
-An `XRRay` can be constructed from the `XRPose`'s `transform`. Rays includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. The `XRRay`'s `matrix` represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
+An `XRRay` can be constructed from the `XRPose`'s `transform`. A rays includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. The `XRRay`'s `matrix` represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
```
NellWaliczek: typo: A rays includes

spatial-tracking-explainer.md
```diff
@@ -416,7 +416,7 @@ interface XRPose {
 //
 
 [SecureContext, Exposed=Window] interface XRSpace : EventTarget {
-  // Intentionally an opaque type, used by XRInputSource.
+  // Interface is intentionally opaque.
```
NellWaliczek: nit: maybe no period at the end?

Issue #392, opened 2018-09-12
Advanced WebXR Controller Input


toji:
_From @toji on July 11, 2018 21:53_

Given feedback from multiple developers citing concerns about our initial plans for a limited input system, the WebXR community group wanted to re-open a discussion about how to handle more advanced input. We agreed on a recent call to put together a proposal for how such a system might work so that we can iterate on the design in public and gather feedback from relevant parties (such as the OpenXR working group and the W3C Gamepad CG).

With WebVR we had exposed VR controller state with extensions to the gamepad API. We feel that building this new advanced input on top of the Gamepad API as it exists today is problematic, though, for a couple of reasons:
 - Developers frequently complained about the difficulty mapping multiple controllers using that method, and as a result we saw a lot of experiences that excluded all controllers that didn't have a specific name string.
 - OpenXR's proposed input model makes it effectively impossible to enumerate all available inputs on a controller (or, for that matter, all available controllers)
 - XR controllers appear to be converging on a small set of common layouts relatively quickly, which should allow us to expose something more limited in scope than the generic gamepad API with higher confidence than usual that it'll be relevant for multiple years to come.

That said, we don't want to re-invent wheels that we don't have to, so we're open to further discussing this proposal with the individuals maintaining the Gamepad API at the moment to see if there's a common ground that can be reached that isn't deterimental to this use case.

## Proposed IDL

```webidl
interface XRInputState {
  readonly attribute boolean pressed;
  readonly attribute boolean touched;
  readonly attribute FrozenArray<double> value;
};

interface XRControllerState {
  readonly attribute XRInputState? trigger;
  readonly attribute XRInputState? joystick;
  readonly attribute XRInputState? touchpad;
  readonly attribute XRInputState? grip;
  readonly attribute FrozenArray<XRInputState> buttons;
};

partial interface XRInputSource {
  readonly attribute XRControllerState? controllerState;
};
```

And some really brief sample code:

```js
let inputSource = xrSession.getInputSources()[0];

if (inputSource.controllerState) {
  // Is a controller with buttons and stuff!

  let joystick = inputSource.controllerState.joystick;
  if (joystick && joystick.value.length == 2) {
    // Has a 2-axis joystick!
    PlayerMove(joystick.value[0], joystick.value[1]);
  }

  if (inputSource.controllerState.buttons.length > 0) {
    let button = inputSource.controllerState.buttons[0];
    if (button.pressed) {
      PlayerJump();
    }
  }
  
  // etc.
}
[...]

PR #462, opened 2019-01-04
Custom interface for controller button/axis state (Variant A)

toji:
This is a proposal to fix the long-running #392, as well as address #336. (And honestly those two could be separated if one part or the other of this proves controversial, but for now I think it's useful to see them together.)

The PR is mostly a more formalized version of the sketched out proposal in #392, with some of the edge cases thought through a bit more. If there was any particular piece of this that I'm not really comfortable with, it's the name 'XRTrackedControllerState`, because I don't feel like that really communicates what it is very well. The term that comes to mind for me immediately is "XRTrackedControllerInput", but that makes for a bit of a namespace conflict with things like "XRInputSource". Problem is that neither Nell or I have come up with a particularly good alternative. ([Twitter has opinions on this](https://twitter.com/Tojiro/status/1075463933119483904), FWIW. Most of which are "Input!" 😉)

So anyway, if you have a better name then please please please PLEASE share it! Any beyond that we'd love feedback on whether or not this feels like a system that will enable whatever particular input usecase you and your app/browser/platform/etc. have in mind.

### DIFF
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -280,6 +280,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
 
 For each of these events the `XRInputSource`'s target ray must be updated to originate at the point that was interacted with on the canvas, projected onto the near clipping plane (defined by the `depthNear` attribute of the `XRSession`) and extending out into the scene along that projected vector.
 
+## Controller State
+
+Some applications need more than point-and-click style interaction provided by the `select` events. For devices that make use of controllers, more complete information about the state of those controller's inputs (buttons, triggers, touchpads, etc) can be observed via the `XRInputSource`'s `controller` attribute.
+
+The `controller` attribute is an instance of the `XRTrackedController` interface if the `XRInputState` represents an XR controller device with a subset of common inputs, or `null` otherwise. Examples of controllers that may expose their state this way include Oculus Touch, Vive wands, Oculus Go and Daydream controllers, or other similar devices. More traditional gamepads, such as XBox or Playstation controllers, should not be exposed using this interface, nor should tracked devices without discreet inputs, such as optical hand tracking. The intent of the `XRTrackedController` interface is not to capture the complete range of inputs for all possible current and future XR controllers, but instead to expose the most commonly observed inputs in an easy-to-consume fashion.
+
+A tracked controller reports the state of several input types that are commonly found on XR controllers. These are:
+
+ - Joystick
+ - Touchpad
+ - Trigger
+ - Grip button
+
+The `XRTrackedController` also exposes a list of "Extended" inputs to capture controller elements who's semantic meanings are not captured in the list above, such as face buttons.
+
+Aside from the `extended` list it is assumed that each controller either has zero or one of the above inputs. If the controller does not have one of the inputs listed above the corresponding attribute in the `controller` object must be `null`.
+
+Each input, if present, is represented as an `XRTrackedControllerState` instance. How the  values are interpreted depends on the inputs in question, but there are common guidelines to their use.
+
+  * `pressed` is a boolean that indicates if the input is physically pressed down. If the input cannot detect when it is pressed this value must be false. If the input represents an analog button or trigger `pressed` should be `true` if the analog value is greater than a system-recommended threshold.
+  * `touched` is a boolean that indicates if the input is currently being touched by the user. If the input is not capacitive (cannot detect touch) this value should mirror the `pressed` value. If the input represents an analog button or trigger `touched` should be `true` any time the analog value is non-zero.
+  * `value` is an array of doubles that represents the analog value(s) of the input. If the input represents an analog button or trigger `values` should have a single element who's value represents the full range of input normalized to a `[0, 1]` range, with `1` meaning "100% engaged". If the input represents a joystick, touchpad, or other logical group of axes `values` should have one element for each axis and report their movement range normalized to a `[-1, 1]` range, with `0` representing the neutral position of the axis. If the input is a touchpad all `value`s should report `0` when the user is not touching the input.
+
+Each input also has a `name` which is a human readable label that should correspond to the name the platform gives for the input if possible. For example, inputs in the `extended` array may have names like "A", "B", "X", and "Y". If a platform-specific name cannot be determined the name should reflect the input's semantic type, such as "Touchpad" or "Button 0".
+
+```js
+function onXRFrame(timestamp, frame) {
+  let inputSource = xrSession.getInputSources()[0];
+
+  // Check to see if the input source is a controller.
+  if (inputSource.controller) {
+    
+    // Determine if the controller has a joystick with at least 2 axes.
+    let joystick = inputSource.controller.joystick;
+    if (joystick && joystick.value.length >= 2) {
+      MoveUser(joystick.value[0], joystick.value[1]);
+    }
+
+    // Determine if the controller has at least one extended input, such as a
+    // face button.
+    if (inputSource.controller.extended.length > 0) {
+      let extended0 = inputSource.controller.extended[0];
+      if (extended0.pressed) {
+        EmitPaint();
+      }
+    }
+    
+    // etc.
+  }
+
+  // Do the rest of typical frame processing...
+}
+```
+
+If the application includes interactions that require user activation (such as starting media playback), the application can listen to the `XRInputSource`s `select` events, which fire for every pressed and released input on the controller. When triggered by a controller input, the `XRInputSourceEvent` will include a non-null `XRTrackedControllerState` to indicate which input triggered the event.
+
+The UA may update the controller state at any point, which the restriction that it must remain constant during the running of each batch of `XRSession` `requestAnimationFrame` callbacks.
+
+> Note: You might wonder why `XRTrackedController` isn't tied to an `XRFrame` like the `XRInputPose` is. The answer is that data exposed by `XRFrame` is intended to be highly timing dependent, and have a limited useful lifespan. Poses are continuously changing, demand low latency, and are typically projected forward to a specific point in time. The also must be queried relative to a specific `XRFrameOfReference`. By contrast, the values reported by a `XRTrackedControllerState` change with comparatively lower frequency, are less sensitive to latency requirements, and can be reasonably inspected outside the core frame loop without concerns of the data becoming unusably stale, and carry no spatial data so there's no need to associate them with an `XRFrameOfReference`. As such, it's preferable to aim fot the better API ergonomics offered by making `XRTrackedController` more persistently available. 
+>
+> Despite this, it's still expected that most implementations will update `XRTrackedController`s in sync with the rest of the data delivered by an `XRFrame`.
+
+### Exposing standard controllers with an action maps
+
+Some native APIs rely on what's commonly referred to as an "action mapping" system to handle controller input. In action map systems the developer creates a list of application-specific actions (such as "undo" or "jump") and suggested input bindings (like "left hand touchpad") that should trigger the related action. Such systems may allow users to re-bind the inputs associated with each action, and may not provide a mechanism for enumerating or monitoring the inputs outside of the action map.
+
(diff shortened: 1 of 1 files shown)


toji:
Communicating a piece of offline feedback that @cwilso gave to me: The naming of the `.controller` attribute on the `XRInputSource` caused some confusion on his first readthrough because he felt it implied that the attributed represented the entirety of the controller device rather than communicating a subset of the input source's state (specifically the button/axis state). An attribute name along of the lines of (but less terrible than) `.buttonAxisState` rather than `.controller` would have better implied the right relationship between objects in his mind.

input-explainer.md
```diff
@@ -280,6 +294,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+  if (inputSource.controller) {
+    
+    // Determine if the controller has a joystick with at least 2 axes.
+    let joystick = inputSource.controller.joystick;
+    if (joystick && joystick.value.length >= 2) {
+      MoveUser(joystick.value[0], joystick.value[1]);
+    }
+
+    // Determine if the controller has at least one extended input, such as a
+    // face button.
```
thetuvix: If we pursue an `extended` approach, we'll need to explicitly think through how buttons/axes will transition from `extended` to full named attributes.  Otherwise, promoting a button or axis will break existing content.

It may be more compatible to change `extended` to `elements` or such, including both the core and extended elements.  That way, there is no temptation for UAs to remove elements as they are promoted to core.

input-explainer.md
```diff
@@ -177,13 +177,27 @@ Most applications will want to visually represent the input sources somehow. The
...
+
+ - A `renderId` of `''` (empty string) indicates the input source should not be rendered. Cursors and pointing rays should still be rendered as described above if needed by the application. Input sources with a `targetRayMode` of `'gaze'` or `'screen'` MUST have a `renderId` of `''`. `'tracked-pointer'` input sources should typically only have a `renderId` of `''` when using displays that allow the user to see their environment and the physical input source.
+ - The `renderId` may be `'unknown'` if the type of input source cannot be reliably identified or the UA determines that the input source type must be masked for any reason.
+ - Otherwise the `renderId` should be a lower-case string that describes the physical input source. It should not include and indication of the handedness of the input source, as that should be implied by the `handedness` attribute.
+   - If the input source is a tracked hand without any held controller, the `renderId` MUST be `hand`
+   - For most controllers this should ideally be of the format `<vendor>-<product-id>`. For example: `oculus-touch`. UAs should make an effort to align on the strings that are returned for any given device.
```
thetuvix: > UAs should make an effort to align on the strings that are returned for any given device.

We should explicitly dig in on this at the F2F to work out some clever approach for going about this even without lots of cross-UA coordination before the first UA ships support for a device.

input-explainer.md
thetuvix: From the call, it sounds like there are concerns about `element`, `component` or `part`.  However, we should agree on the structure we determine that we need here, and then find the least-bad name.

kearwood:
Some initial feedback:
- Value ranges might need to reflect values on both sides of "engaged".  Eg, some devices can be "more engaged".  Pressure sensitive buttons or grips could fall into this category.
- "name" of inputs perhaps should be split into multiple values
(One for localization, one as fixed string, One index identifier for skinned mesh models)
- If same device is available in Gamepad API, should it be referenced?

input-explainer.md
NellWaliczek: I'm open to renaming `extended` to `all` or something similar with the appropriate behavior.  
That said, IIRC, webidl doesn't allow for a dictionary of values as an attribute.  At one point we considered a function like `getInputPart(string partName)`.  Perhaps that's an option?

NellWaliczek:
@kearwood you mentioned
>If same device is available in Gamepad API, should it be referenced?

That shouldn't ever be the case.  This design is really just intended for tracked controllers, and it's intended they would never show up in the gamepads array.

NellWaliczek:
Also at the risk of saying something that's either shooting myself in the foot... or making @toji stare at me like -_-.... Looking at the `XRTrackedController` design right now.... it occurs to me that there really isn't anything tracked about it.  It's really just a nicer `Gamepad` attached an `XRInputSource`.  Which brings up two thoughts:
1. Should this idea of named buttons/trackpads and strongly typed data be proposed as a separate spec from WebXR?  Perhaps as a gamepad 2.0?
2. If that's the case, perhaps the right thing for the WebXR Level1 spec is to do what it was @kearwood  thought we were suggesting...and instead just have a `Gamepad` member.  That means the Gamepad spec could update independently to add the features described above and, so long as it's done in a way that can be feature detected, developers wouldn't be negatively impacted.

>note: I am totally NOT suggesting we keep the gamepad extensions.

*whispers* please don't hate me.....

NellWaliczek:
Oh, and one small aside... a colleague of mine noted that the `grip` and the `gripSpace` are a bit confusing due to their similar names... My inclination would be to change the name of `gripSpace` to something more like `renderSpace`

thetuvix:
> If that's the case, perhaps the right thing for the WebXR Level1 spec is to do what it was @kearwood  thought we were suggesting...and instead just have a `Gamepad` member.  That means the Gamepad spec could update independently to add the features described above and, so long as it's done in a way that can be feature detected, developers wouldn't be negatively impacted.

This is interesting to explore, given where we've landed with input source spaces.  The key thing we'd have to make sure is that whatever Gamepad 1.0 mapping we come up with avoids all the pitfalls we hit with WebVR's use of the Gamepad API:
* Well-known elements like "menu button" and "touchpad X" would need well-known button and axis indices in some kind of `"xr"` Gamepad `mapping`.  UAs would need to support gaps in both lists when a controller is missing a given element.
* For future extensibility of the core mapping as we standardize additional elements, we could have `"xr2"`/`"xr3"` or such.  This allows an app still using `"xr"` to get the old mapping, where non-standard buttons for that controller show up with their old non-standard indices.
* We'll have to work out some way to represent touch states for now - we could fall back to treating each button's touch amount as an additional sibling "button", but that'd be unfortunate.  Maybe it's a small enough tweak to add `touched` next to `pressed` on `GamepadButton`.

daoshengmu:
> 
> interface XRTrackedController {
>   readonly attribute XRTrackedControllerStateDouble? trigger;
>   readonly attribute XRTrackedControllerStateVector2? joystick;
>   readonly attribute XRTrackedControllerStateVector2? touchpad;
>   readonly attribute XRTrackedControllerStateDouble? grip;
>   readonly attribute XRTrackedControllerStateDouble? menu;
>   readonly attribute XRTrackedControllerStateDouble? a;
>   readonly attribute XRTrackedControllerStateDouble? b;
> };
> 
I love this way, at least, we don't need to care about the button and axis order as `Gamepad` API. However, we can't cover all of buttons in world. For example, Oculus Touch has x, y buttons. If we wanna add new buttons, we have to modify this `.webidl` and update it to the spec.

The other thing is I am thinking about handness inputs, like SteamVR Knuckles. They would expose hands as bone matrices, so Float32Array is an another type of `XRTrackedControllerState` that we should consider.

So, I was thinking to use action-base approach (https://github.com/immersive-web/webxr/issues/392#issuecomment-431200201), and Firefox has shipped OpenVR action-based input API in 65. My proposal of action-based like API is not really webby, and I am also not good at defining Web APIs, but this is just my two cents that might be helpful.

thetuvix:
> Oh, and one small aside... a colleague of mine noted that the `grip` and the `gripSpace` are a bit confusing due to their similar names... My inclination would be to change the name of `gripSpace` to something more like `renderSpace`

I've noticed a similar confusion - I often refer now to a handheld controller's **"palm pose"** rather than "grip pose" in other contexts to help reduce that confusion.  That would lead to `palmSpace`.

If we didn't want to presume a "palm" for generic renderable input sources, I'd rather just refer to the source's `rootSpace` rather than `renderSpace`, as I'd expect other source-relative data (e.g. Leap Motion hand joint positions) to make use of that same origin in the future.

input-explainer.md
ddorwin: https://www.w3.org/wiki/Repositories appears to be the latest thinking on such things and seems like a good fit.

I think this is a large enough topic that we should break the renderId part out of this PR so we can focus on controller state ergonomics and the logistics around this ID separately.

toji:
Lots of feedback to sort through, but in the meantime I've rebased the issue now that the pose refactor has landed. I've also removed the `renderId` attribute (since that's now covered in #479).

input-explainer.md
```diff
@@ -280,6 +294,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+The UA may update the controller state at any point, which the restriction that it must remain constant during the running of each batch of `XRSession` `requestAnimationFrame` callbacks.
+
+> Note: You might wonder why `XRTrackedController` isn't tied to an `XRFrame` like the `XRInputPose` is. The answer is that data exposed by `XRFrame` is intended to be highly timing dependent, and have a limited useful lifespan. Poses are continuously changing, demand low latency, and are typically projected forward to a specific point in time. The also must be queried relative to a specific `XRFrameOfReference`. By contrast, the values reported by a `XRTrackedControllerState` change with comparatively lower frequency, are less sensitive to latency requirements, and can be reasonably inspected outside the core frame loop without concerns of the data becoming unusably stale, and carry no spatial data so there's no need to associate them with an `XRFrameOfReference`. As such, it's preferable to aim fot the better API ergonomics offered by making `XRTrackedController` more persistently available. 
+>
```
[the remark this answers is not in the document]
toji: "must remain constant during Javascript execution" is basically the default assumption of Javascript, isn't it? Being single threaded, it would be highly unusual if values did change mid-execution without an obvious call to update them. rAF is special simply in that during the processing of rAF callbacks control is returned to the browser multiple times, once per callback, and yet we want the controller state to be consistent across all of them.

I'm looking at an alternative way of approaching this functionality, but I'll try to update all of this language to be clearer and take into account the rest of your feedback if it carries through to the next update.

input-explainer.md
cwilso: I'd caution against taking the thinking in that wiki as a current, direct solution.  It was a work item of the Advisory Board (while I was on it), hasn't been completed (or worked on in a while?) and it's not an adopted part of the W3C Process (translation: this hasn't been tested by any other repository/schema-like thing in W3C-space that I know of).

input-explainer.md
```diff
@@ -280,6 +280,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
 
+## Controller State
+
+Some applications need more than point-and-click style interaction provided by the `select` events. For devices that make use of controllers, more complete information about the state of those controller's inputs (buttons, triggers, touchpads, etc) can be observed via the `XRInputSource`'s `controller` attribute.
+
+The `controller` attribute is an instance of the `XRTrackedController` interface if the `XRInputState` represents an XR controller device with a subset of common inputs, or `null` otherwise. Examples of controllers that may expose their state this way include Oculus Touch, Vive wands, Oculus Go and Daydream controllers, or other similar devices. More traditional gamepads, such as XBox or Playstation controllers, should not be exposed using this interface, nor should tracked devices without discreet inputs, such as optical hand tracking. The intent of the `XRTrackedController` interface is not to capture the complete range of inputs for all possible current and future XR controllers, but instead to expose the most commonly observed inputs in an easy-to-consume fashion.
```
dmarcos: typo? discreet -> discrete

dmarcos: Microsoft uses `Xbox` capitalization in [its materials](https://account.microsoft.com/account/xboxlive) vs. `XBox`

dmarcos: [Sony uses](https://www.sony.com/electronics/playstation) `PlayStation` vs `Playstation`

toji:
This was discussed at the Jan 2019 F2F and pursuing a gamepad-based solution (as described in #499) was overwhelmingly preferred to this approach. Closing as a result in favor of devoting all input-related energy to PR #499.

PR #486, opened 2019-01-16
Remove all Bikeshed issues

cwilso:
Put in placeholders for definitions which refer to their Github issue.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -412,13 +412,13 @@ The <dfn attribute for="XRSession">onselectend</dfn> attribute is an [=Event han
 
 The <dfn attribute for="XRSession">onselect</dfn> attribute is an [=Event handler IDL attribute=] for the {{XRSession/select}} event type.
 
-Issue: Document what happens when we <dfn>end the session</dfn>.
+We still need to document what happens when we <dfn>end the session</dfn>. (This is <a href="https://github.com/immersive-web/webxr/issues/467">filed</a>.)
 
-Issue: Document effects when we <dfn lt="blur all sessions">blur the session</dfn>.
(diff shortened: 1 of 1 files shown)


ddorwin:
While I fully support ensuring we have GitHub issues for every issue box in the spec, I'm concerned that the other parts of this change may not be positive for readers. First, why not continue to use issue boxes for the text that identifies and issue and references the GitHub issue? IOW, why not leave `Issue: ` in the new text?

Second, I think it's better to be clear in the spec where things aren't fully defined or where there is work to be done. For example, if the frame lifecycle needs more explanation. This will help readers understand the current state, what might not be stable, and that the thing they're trying to find is indeed missing without having to cross-reference all the open issues.

Suggestion: It might be useful to include the issue number in the text rather than just "This is filed." Something like `<a>Issue 123</a>: This needs to be defined.` has worked well in the past. That's searchable both by number and `issue`, which is not otherwise searchable due to the way the issue boxes work.

cwilso:
In short, this was done as a reminder to not simply put in Bikeshed issues (i.e., we want to ensure everything is tracked in Github).  The only reason for leaving remnants at all was because some of the issues were missing definitions that are referred to elsewhere in the spec, and removing them would cause Bikeshed errors.  I followed the process I did after discussing with Brandon what he would prefer, and frankly, whatever the Editors decide here is okay with me.  I would not personally want them to spend all their time inserting placeholders in the spec saying "this isn't quite done yet", and without doing so it would seem uneven (that is, if a placeholder isn't there, one might assume that section is done), but it is entirely their decision.

Issue #400, opened 2018-09-18
DOM Overlays (for Phone AR support and more)


toji:
As described in #394, when we move away from allowing AR content inline there's a desire to still enable 2D UI to be built using DOM for cases like phone AR, where support for displaying DOM elements and the AR stream together should be pretty trivial. (Previously the same effect would have been supported by using the fullscreen API on a parent element that contained both the AR canvas and some overlaid DOM elements.) Given that we want to utilize the advantages of the web as our platform whenever possible, it would be incredibly unfortunate to lose this ability in pursuit of an explicit AR mode.

For reference, consider this image of Pokemon Go:

![pokemon-go](https://user-images.githubusercontent.com/805273/45705946-cbb08380-bb2f-11e8-84b8-d09a23d12ab0.jpg)

If created with WebXR, ideally the animal and likely the ball at the bottom of the screen (because it has an associated throwing animation) would be rendered with WebGL as part of the core session rAF loop, but the other UI elements would ideally be handled as standard DOM elements that are simply composited over the AR content by the UA. (The name of the animal floating over it's head is a special case that I'll address more in a second.)

However, while the core need is to retain DOM support for phone AR there's also a desire to potentially enable that DOM content to be surfaced on headsets, with the idea being that the AR content would be fully immersive while the DOM portion is composited in by the UA somehow to ensure that the user can still access it. Exactly how the would appear is an open question, and one that we probably would want to leave up to the UA to avoid prescribing unproven UX patterns. Some possibilities I could see are:

 - Having the overlay DOM be a floating, moveable window in space
 - Attaching it to your wrist
 - Pinning it to a wall
 - Showing/hiding it with the push of a button

Doing so would likely be considered a "compatibility" mode, and would definitely not be a path we'd encourage for developers explicitly targeting headset AR. The big benefit being that it would enable AR content built for the more common devices (phone AR) to still be accessible on more advanced devices, thus immediately increasing the content that's accessible to them.

That said, there's also been some concerns voiced that supporting DOM like this in headsets could be difficult for some platforms, or would be hard to make a good user experience. As such, I'm reluctant to say that supporting a DOM overlay should be required for all devices that support AR. And certainly I believe that we should offer the right signals and tools in all cases to allow developers to explicitly create experiences optimized for any given devices they choose to support.

So this issue is simply to talk about how we should go about supporting those overlays and what guarantees of availability that mode should have. 

Some other considerations:
[...]

PR #493, opened 2019-01-22
Adding an explainer for real-world hit testing

NellWaliczek:
This PR contains an explainer with a proposal for adding real-world hit testing to WebXR.  It builds strongly on the great work done by @lincolnfrog, @judax, @blairmacinyre, and others to clearly define the problem space. While it is expected that hit-testing will often be used to create anchors, this PR does not include a proposal for an anchors API; that will come in a subsequent PR but still needs a bit more design work.

Rather than a long-winded PR description, I'm hoping that the explainer stands on its own.  Please don't hesitate to call out areas that are unclear.

For clarity's sake, the base for this PR is the input-explainer-cleanup branch in PR #492


hit-testing-explainer.md
```diff
@@ -0,0 +1,275 @@
...
+* `offsetRay` is optional and, if provided, is the `XRRay` from which the hit test should be performed. The ray's coordinates are defined with `space` as the origin. If an `offsetRay` is not provided, hit testing will be performed using a ray with coincident with the `space` origin and pointing in the "forward" direction. For more information on constructing an `XRRay`, see the [Spatial Tracking explainer](spatial-tracking-explainer.md#rays).
+
+In this example, an `XRHitTestSource` is created slightly above the center of the `viewerSpace`. This is because the developer has chosen to leave space for buttons along the bottom of the hand-held AR device screen while still giving the perception of a centered cursor. For more information, see [Rendering cursors and highlights](#rendering-cursors-and-highlights) in the Input Explainer.
```
klausw: > the developer has chosen to leave space for buttons along the bottom of the hand-held AR device screen

I think this is a bit confusingly phrased. Buttons sound like DOM elements, and "leaving space" could be interpreted as shrinking the camera view to make room for them, and both aren't currently supported in `immersive-ar` mode, right? Maybe something like _"the developer has chosen to draw user interface elements in the bottom part of the immersive view"_ would make it clearer that this would still be part of the fullscreen immersive mode.

klausw:
About "Combining virtual and real-world hit testing", the intent appears to be that WebXR API takes care of real-world understanding, while the application is responsible for virtual objects. For consistency, should  real-world information provided by VR headsets also be handled by the hit test API? I've created issue #494 "Using hit test API for floor/bounds in VR mode" with some more thoughts on this, and would welcome feedback there.

hit-testing-explainer.md
```diff
@@ -0,0 +1,275 @@
...
+
+WebXR addresses this challenge through the use of the `XRHitTestSource` type which acts somewhat like a subscription mechanism. The presence of an `XRHitTestSource` signals to the user agent that the developer intends to query `XRHitTestResult`s in subsequent `XRFrame`s. The user agent can then precompute `XRHitTestResult`s based on the `XRHitTestSource` properties such that each `XRFrame` will be bundled with all "subscribed" hit test results. When the last reference to the `XRHitTestSource` has been released, the user agent is free to stop computing `XRHitTestResult`s for future frames.
+
+### Requesting a hit test source
+To create an `XRHitTestSource` developers call the `XRSession.requestHitTestSource()` function. This function accepts an `XRHitTestOptionsInit` dictionary with the following values:
+* `space` is required and is the `XRSpace` to be tracked by the hit test source. As this `XRSpace` updates its location each frame, the `XRHitTestSource` will move with it. If the supplied `space` is an `XRReferenceSpace`, the `offsetRay` of the resulting `XRHitTestSource` will be updated each time the `XRReferenceSpace.originOffset` is updated, even mid-frame.
```
NellWaliczek: Should we block the use of an 'identity' reference space as the `space` in the dictionary to create `XRHitTestSource`?

hit-testing-explainer.md
```diff
@@ -0,0 +1,275 @@
...
+  let hitTestResults = xrFrame.getHitTestResults(hitTestSources[preferredInputSource], xrReferenceSpace);
+  if (hitTestResults && hitTestResults.length > 0) {
+    // Do something with the results
+  }
+
+  // Other scene update logic ...
+}
+```
+
+On occasion, developers may want hit test results for the current frame even if they have not already created an `XRHitTestSource` to subscribe to the results. For example, when a virtual object needs to bounce off a real-world surface, a single hit-test result can be requested. The results will be delivered asynchronously, though they will be accurate for the frame on which the request was made. Otherwise, `requestAsyncHitTestResults()` shares the behavior of `getHitTestResults()` as described above.
```
NellWaliczek: Probably want to block the use of 'identity' reference spaces here too...

hit-testing-explainer.md
thetuvix: Is there any reason to block it?

If I want to carry on targeting 360-degree video hotspots while manually tracked with my finger, shouldn't that work just as well as everything else in the `"identity"` space?

hit-testing-explainer.md
```diff
@@ -0,0 +1,275 @@
...
+  // Other scene update logic ...
+}
+
+function getHitCombinedHitTestResult(frame, inputSource, hitTestSource) {
+  // Try to get virtual hit test result
+  if (inputSource) {
+    let inputSourcePose = frame.getPose(inputSource.source, xrReferenceSpace);
+    if (inputSourcePose) {
+      var virtualHitTestResult = scene.virtualHitTest(new XRRay(inputSource.transform));
+      return { result:virtualHitTestResult, virtualTarget:virtualHitTestResult.target }
```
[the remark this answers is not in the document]
thetuvix: You mentioned at the F2F that this is due to not having an occlusion solution, and so virtual content will always be visible anyway.  However, it doesn't sound like AR apps will produce intuitive experiences for users if things always end up available through walls.

Does this suggest that perhaps a simple on/off occlusion solution is needed to meet the credible "AR min-bar" for WebXR?

hit-testing-explainer.md
thetuvix: Is the behavior of "request now, receive results later, but results are based on data from previous frame" what apps will actually want?  Do we have a concrete scenario for wanting that?

I see two coherent paths for the following scenario:
> At time A, the user presses a button.  After some amount of wireless latency, the UA receives the press and sets controller state, available to the app for an `XRFrame` with forward-predicted time B.  At the frame for forward-predicted time C, the app gets results that match either...
* Time A, when the press actually occurred.  This would be useful for hit-testing what a controller was pointing at back when the press occurred, (available through `XRInputSourcesChangeEvent.getHitTestSource()`)  _or_
* Time C, for the forward-predicted time at which the app will target its current frame.  This would be useful for a view-relative reticle. (available through `XRFrame.getHitTestResults(hitTestSource)`)

Time B seems arbitrary relative to the user's actions, and so going out of our way to return data for Time B but not Time A seems of low value.  This makes me question if there is a strong value now for the arbitrary hit-test. (`XRFrame.requestAsyncHitTestResults`)

thetuvix: A relevant nuance is that what XR APIs generally allow time-indexing of is **dynamic** data, for things like heads and controllers that are actually moving.  In contrast, for **static** and **semi-static** data, such as planes and meshes, APIs generally only expose the latest snapshot of their poses and geometry.

Unless a plane itself is being posed smoothly every frame in a forward-predicted manner, the following may be a reasonable balance that gets apps what they need while imposing a minimal constraint on UA architecture:
* Raycast **from** the pose of the `XRSpace` at the exact historical or future time requested in the `hitTest` call.
* Raycast **against** the world geometry that is current at the moment when the data is delivered.

NellWaliczek: Per Alex's comment above, no we will not block this.

hit-testing-explainer.md
NellWaliczek: Yep, that makes sense.

hit-testing-explainer.md
NellWaliczek: That's exactly what's being described ^_^  We know a solution is required for some form of DOM  overlay on hand-held AR, and that's being tracked in issue #400.  That said, until we have a concrete solution, I can understand the hesitation about overly specific language and will fix it.

hit-testing-explainer.md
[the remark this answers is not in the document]
blairmacintyre: I commented over there, but I'll put the comment here too:  if we are providing one simple example, I would think @thetuvix's first suggestion of returning the closer one is correct.  An app developer would only call this hypothetical method if they wanted to test both "as if" they were shooting a ray through the combined geometry, so it should return the closer one.  If a dev wants to preferentially test some subset (e.g., "I'm interacting with a virtual object that's marked as SELECTED, and if it goes behind physical stuff, I still want to interact with it"), they can fairly obviously implement that.  But THIS sample method makes no sense except in the "behave as if the geometry is combined" sense.

hit-testing-explainer.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Real-world hit testing
+A key challenge with enabling real-world hit testing in WebXR is that computing real-world hit test results can be performance-impacting and dependant on secondary threads in many of the underlying implementations. However from a developer perspective, out-of-date asynchronous hit test results are often, though not always, less than useful. 
+
+WebXR addresses this challenge through the use of the `XRHitTestSource` type which acts somewhat like a subscription mechanism. The presence of an `XRHitTestSource` signals to the user agent that the developer intends to query `XRHitTestResult`s in subsequent `XRFrame`s. The user agent can then precompute `XRHitTestResult`s based on the `XRHitTestSource` properties such that each `XRFrame` will be bundled with all "subscribed" hit test results. When the last reference to the `XRHitTestSource` has been released, the user agent is free to stop computing `XRHitTestResult`s for future frames.
```
blairmacintyre: FWIW (sorry for being absent last month), I think how this turned out @NellWaliczek

PR #496, opened 2019-01-22
Added XRPose and related refactors to the spec

toji:


index.bs
```diff
@@ -578,48 +578,39 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
 
-When the <dfn method for="XRFrame">getInputPose(|inputSource|, |referenceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
+When the <dfn method for="XRFrame">getPose(|space|, |relativeTo|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw a {{InvalidStateError}} and abort these steps.
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
-  1. If |referenceSpace|'s [=XRSpace/session=] does not equal |session|, return <code>null</code> and abort these steps.
-  1. If |inputSource|'s pose cannot be determined relative to |referenceSpace|, return <code>null</code>
-  1. Return a new {{XRInputPose}} describing |inputSource|'s pose relative to the origin of |referenceSpace|.
+  1. If |space|'s [=XRSpace/session=] does not equal |session|, return <code>null</code> and abort these steps.
```
NellWaliczek: This seems like it should throw rather than return null.  wdyt?

NellWaliczek: Same for the next line.

index.bs
```diff
@@ -578,48 +578,39 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
+  1. If |relativeTo|'s [=XRSpace/session=] does not equal |session|, return <code>null</code> and abort these steps.
+  1. If |space|'s pose cannot be determined relative to |relativeTo|, return <code>null</code>
+  1. Return a new {{XRPose}} describing |space|'s pose relative to the origin of |relativeTo|.
 
 </div>
 
-Issue: Describe behavior for passing <code>null</code> {{XRReferenceSpace}}s to {{getViewerPose()}} and {{getInputPose()}}
+Issue: Describe behavior for passing <code>null</code> {{XRReferenceSpace}}s to {{getViewerPose()}} and {{getPose()}}
```
NellWaliczek: This seems like it would be super easy to fix.  Would you be willing to do that in this PR?

index.bs
```diff
@@ -919,24 +910,37 @@ NOTE: The {{XRRay}}'s {{XRRay/matrix}} can be used to easily position graphical
...
   readonly attribute XRRigidTransform transform;
-  readonly attribute FrozenArray&lt;XRView&gt; views;
+  readonly attribute boolean emulatedPosition;
 };
 </pre>
 
-The <dfn attribute for="XRViewerPose">transform</dfn> is the {{XRRigidTransform}} of the [=viewer=] relative to the origin of the {{XRReferenceSpace}} the {{XRViewerPose}} was queried with.
+The <dfn attribute for="XRPose">transform</dfn> attribute describes the {{XRRigidTransform}} between two {{XRSpace}}.
+
+The <dfn attribute for="XRPose">emulatedPosition</dfn> attribute indicates the accuracy of the {{XRPose/transform}}.{{XRRigidTransform/position}} value. {{XRPose/emulatedPosition}} MUST be set to <code>true</code> if the given positional values are software estimations, such as those provided by a neck or arm model. {{XRPose/emulatedPosition}} MUST be set to <code>false</code> if the positional values are based on sensor readings.
```
NellWaliczek: >indicates the accuracy of the

This phrasing has always somewhat bugged me.  There's something about it that implies that there's a gradient of accuracy being communicated.

index.bs
```diff
@@ -977,29 +983,11 @@ The <dfn attribute for="XRInputSource">targetRayMode</dfn> attribute describes t
...
+The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XRSpace}} that tracks the pose of the preferred pointing ray of the {{XRInputSource}}, as defined by the {{targetRayMode}}.
 
-The {{gripTransform}} MUST be `null` if the input source isn't trackable. 
+The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace}} that tracks the transform that should be used to render virtual objects such that they appear to be held in the user's hand. If the user were to hold a straight rod, this {{XRSpace}} places the origin at the centroid of their curled fingers and where the <code>-Z</code> axis points along the length of the rod towards their thumb. The <code>X</code> axis is perpendicular to the back of the hand being described, with back of the users right hand pointing towards <code>+X</code> and the back of the user's left hand pointing towards <code>-X</code>. The <code>Y</code> axis is implied by the relationship between the <code>X</code> and <code>Z</code> axis, with <code>+Y</code> roughly pointing in the direction of the user's arm.
```
NellWaliczek: >that tracks the transform

Let's be consistent with this language when describing pose behavior.  Other places we say "that tracks the pose" which seems more accurate given that the pose includes information about position emulation.

index.bs
```diff
@@ -977,29 +983,11 @@ The <dfn attribute for="XRInputSource">targetRayMode</dfn> attribute describes t
...
 
-The <dfn attribute for="XRInputPose">emulatedPosition</dfn> attribute indicates the accuracy of the {{XRRay/origin}} of the {{XRInputPose/targetRay}} and {{XRRigidTransform/position}} of the {{XRInputPose/gripTransform}}. {{XRInputPose/emulatedPosition}} MUST be set to <code>true</code> if positional values are software estimations, such as those provided by a neck or arm model. {{XRInputPose/emulatedPosition}} MUST be set to <code>false</code> if the positional values are based on sensor readings.
+The {{gripSpace}} MUST be `null` if the input source isn't trackable.
```
NellWaliczek: Maybe "isn't inherently trackable"?

index.bs
```diff
@@ -977,29 +983,11 @@ The <dfn attribute for="XRInputSource">targetRayMode</dfn> attribute describes t
...
-The <dfn attribute for="XRInputPose">targetRay</dfn> describes the preferred pointing ray of the {{XRInputSource}}, as defined by the {{targetRayMode}}.
-
-The <dfn attribute for="XRInputPose">gripTransform</dfn> MUST describe a transform to a space which, if the user were to hold a straight rod in their hand their hand, places the origin at the centroid of their curled fingers and where the <code>-Z</code> axis points along the rod towards their thumb. The <code>X</code> axis is perpendicular to the back of the hand being described, with back of the users right hand pointing towards <code>+X</code> and the back of the user's left hand pointing towards <code>-X</code>. The <code>Y</code> axis is implied by the relationship between the <code>X</code> and <code>Z</code> axis, with <code>+Y</code> roughly pointing in the direction of the user's arm. 
-
-The {{gripTransform}} MAY represent an emulated translation or rotation if the input source's cannot supply full [=6DoF=] tracking.
+The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XRSpace}} that tracks the pose of the preferred pointing ray of the {{XRInputSource}}, as defined by the {{targetRayMode}}.
```
NellWaliczek: Might be worth calling out some examples of what this space can represent.

toji: I feel like that's addressed pretty well by the reference at the end of this paragraph stating "as defined by the {{targetRayMode}}", which is detailed only a few lines above this.

toji:
Addressed your feedback, Nell. Please take another look when you get a chance!

index.bs
```diff
@@ -718,13 +707,13 @@ interface XRStationaryReferenceSpace : XRReferenceSpace {
...
+- Passing a {{XRReferenceSpaceOptions/subtype}} of <dfn enum-value for="XRStationaryReferenceSpaceSubtype">floor-level</dfn> creates an {{XRStationaryReferenceSpace}} with it's origin positioned at the floor in a safe position for the user to stand. The `y` axis equals `0` at floor level, with the `x` and `z` position and orientation initialized based on the conventions of the underlying platform. If the floor level isn't known it will be estimated.
 
-- Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRStationaryReferenceSpaceSubtype">position-disabled</dfn> creates an {{XRStationaryReferenceSpace}} where orientation is tracked but the [=viewer=]s position is always reported as being at the origin.
+- Passing a {{XRReferenceSpaceOptions/subtype}} of <dfn enum-value for="XRStationaryReferenceSpaceSubtype">position-disabled</dfn> creates an {{XRStationaryReferenceSpace}} where orientation is tracked but the [=viewer=]s position is always reported as being at the origin.
```
toji: The above few changes are unrelated to this PR, but they're causing some bikeshed build warnings and were straight up incorrect in a couple of cases, so I'm slipping them into here with the expectation that this PR is almost ready to land. I can split it out if the review proves more contentious than expected.

toji:
Addressed feedback. Please take another look, @NellWaliczek!

index.bs
```diff
@@ -1050,7 +1052,7 @@ The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XR
 
 The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace}} that tracks the pose that should be used to render virtual objects such that they appear to be held in the user's hand. If the user were to hold a straight rod, this {{XRSpace}} places the origin at the centroid of their curled fingers and where the <code>-Z</code> axis points along the length of the rod towards their thumb. The <code>X</code> axis is perpendicular to the back of the hand being described, with back of the users right hand pointing towards <code>+X</code> and the back of the user's left hand pointing towards <code>-X</code>. The <code>Y</code> axis is implied by the relationship between the <code>X</code> and <code>Z</code> axis, with <code>+Y</code> roughly pointing in the direction of the user's arm.
 
-The {{gripSpace}} MUST be `null` if the input source isn't inherently trackable.
+The {{gripSpace}} MUST be <code>null</code> if the input source isn't inherently trackable or for input sources with a {{targetRayMode}} of {{XRTargetRayMode/gaze}} or {{XRTargetRayMode/screen}}.
```
NellWaliczek: nit: "such as for input sources"?

PR #499, opened 2019-01-29
Gamepad-based button/axis state

toji:
This is a proposal for how we might use the existing Gamepad API to expose controller button and axis state.

### DIFF
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -220,11 +220,11 @@ function updatePointingRay(inputSourcePose, virtualHitTestResult) {
 ```
 
 ### Renderable models
-For `tracked-controller` input sources, it is often appropriate for the application to render a contextually appropriate model (such as a racket in a tennis game). Other times it's desirable for the application to render a device that matches what the user is holding, especially when relaying instructions about it's use. Finally, there are times when it's best to not render anything at all, such as when the XR device uses a transparent display and the user can see their hands and/or any tracked devices without app assistance. See [Handling non-opaque displays](explainer.md#Handling-non-opaque-displays) in the main explainer for more details.
+For `tracked-pointer` input sources, it is often appropriate for the application to render a contextually appropriate model (such as a racket in a tennis game). Other times it's desirable for the application to render a device that matches what the user is holding, especially when relaying instructions about it's use. Finally, there are times when it's best to not render anything at all, such as when the XR device uses a transparent display and the user can see their hands and/or any tracked devices without app assistance. See [Handling non-opaque displays](explainer.md#Handling-non-opaque-displays) in the main explainer for more details.
 
 
 #### Choosing renderable models
-The majority of `tracked-controller` input sources will have a non-null `gamepad` attribute on the `XRInputSource` object. The `Gamepad`'s `id` is used to determine what should be rendered if the app intends to visualize the input source itself, rather than an alternative virtual object. (See the section on [Button and Axis State](#button-and-axis-state) for more details.)
+The majority of `tracked-pointer` input sources will have a non-null `gamepad` attribute on the `XRInputSource` object. The `Gamepad`'s `id` is used to determine what should be rendered if the app intends to visualize the input source itself, rather than an alternative virtual object. (See the section on [Button and Axis State](#button-and-axis-state) for more details.)
 
 The WebXR Device API currently does not offer any way to retrieve renderable resources that represent the input devices from the API itself, and as such the `Gamepad`'s `id` must be used as a key to load an appropriate resources from the application's server or a CDN. The example below presumes that the `getInputSourceRenderableModel` call would do the required lookup and caching.
 
@@ -276,57 +276,41 @@ function updateRenderableInputModels(xrFrame) {
 
 ## Button and Axis State
 
-Some applications need more than point-and-click style interaction provided by the `select` events. For devices that make use of controllers with buttons and axes, more complete information about the state of those inputs can be observed via the `XRInputSource`'s `gamepad` attribute.
+Some applications need more than point-and-click style interaction provided by the `select` events. For input sources with buttons and axes, more complete information about the state of those inputs can be observed via the `XRInputSource`'s `gamepad` attribute. `gamepad` is an instance of the [`Gamepad`](https://w3c.github.io/gamepad/#gamepad-interface) interface if the input source has buttons and axes to report, and `null` otherwise.
 
-If the `XRInputState` represents an XR controller device with buttons or axes the `gamepad` attribute is an instance of a [`Gamepad`](https://w3c.github.io/gamepad/#gamepad-interface) interface, and `null` otherwise. Examples of controllers that may expose their state this way include Oculus Touch, Vive wands, Oculus Go and Daydream controllers, or other similar devices. Controllers not directly associated with the XR device, such as the majority of traditional gamepads, should not be exposed using this interface. Tracked devices without discreet inputs, such as optical hand tracking, should also have a `null` `gamepad` attribute.
+Examples of input sources that may expose their state this way include Oculus Touch, Vive wands, Oculus Go and Daydream controllers, or other similar devices. Input devices not directly associated with the XR device, such as the majority of traditional gamepads, and tracked devices without discreet inputs, such as optical hand tracking, must not be exposed using this interface. 
 
 `Gamepad` instances reported in this way have several notable behavioral changes vs. the ones reported by `navigator.getGamepads()`:
 
   - `Gamepad` instances connected to an `XRInputSource` must not be included in the array returned by `navigator.getGamepads()`.
   - The `Gamepad`'s `index` attribute must be `-1`.
   - The `Gamepad`'s `connected` attribute must always be `true`.
-  - The `Gamepad`'s `mapping` attribute must be `xr-standard`.
 
 Finally, the `id` attribute for `Gamepad`s surfaced by the WebXR API are more strictly formatted than those of traditional gamepads in order to make them a more appropriate key for determining rendering assets.
 
   - The `id` MAY be `'unknown'` if the type of input source cannot be reliably identified or the UA determines that the input source type must be masked for any reason. Applications should render a generic input device in this case.
   - Inline sessions MUST only expose `id`s of `'unknown'`.
-  - Otherwise the `id` should be a lower-case string that describes the physical input source.
+  - Otherwise the `id` must be a lower-case string that describes the physical input source.
     - For most devices this SHOULD be of the format `<vendor>-<product-id>`. For example: `oculus-touch`. UAs SHOULD make an effort to align on the strings that are returned for any given device.
-    - It should not include an indication of the handedness of the input source, as that should be implied by the `handedness` attribute.
+    - It should not include an indication of the handedness of the input source (such as `oculus-touch-left`), as that can be determined from the `handedness` attribute.
 
 All other attributes behave as described in the [Gamepad](https://w3c.github.io/gamepad/) specification.
-
-The WebXR Device API introduces a new standard controller layout indicated by the `mapping` value of `xr-standard`. (Additional mapping variants may be added in the future if necessary.) This defines a specific layout for the inputs most commonly found on XR controller devices today. The following table describes the buttons/axes and their physical locations:
-
-| Button/Axis  | Location                          |
-| ------------ | --------------------------------- |
-| buttons[0]   | Primary trigger                   |
-| buttons[1]   | Primary Touchpad/Joystick click   |
-| buttons[2]   | Grip/Secondary trigger            |
-| buttons[3]   | Secondary Touchpad/Joystick click |
-| axes[0]      | Primary Touchpad/Joystick X       |
-| axes[1]      | Primary Touchpad/Joystick Y       |
-| axes[2]      | Secondary Touchpad/Joystick X     |
-| axes[3]      | Secondary Touchpad/Joystick Y     |
-
-Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array. If a device has both a touchpad and a joystick the UA should designate one of them to be the primary axis-based input and expose the other at axes[2] and axes[3] with an associated button at button[3].
 
 ```js
 function onXRFrame(timestamp, frame) {
   let inputSource = primaryInputSource;
 
-  // Check to see if the input source has buttons/axes and is using the standard
-  // mapping.
-  if (inputSource && inputSource.gamepad &&
-      inputSource.gamepad.mapping == 'xr-standard') {
+  // Check to see if the input source has gamepad data.
+  if (inputSource && inputSource.gamepad) {
     let gamepad = inputSource.gamepad;
     
     // Use joystick or touchpad values for movement.
-    MoveUser(gamepad.axes[0], gamepad.axes[1]);
+    if (gamepad.axes.length >= 2) {
+      MoveUser(gamepad.axes[0], gamepad.axes[1]);
+    }
 
-    // Use the trigger or joystick/touchpad click for painting.
-    if (gamepad.buttons[0].pressed || gamepad.buttons[1].pressed) {
+    // If the first gamepad button is pressed, perform an action.
+    if (gamepad.buttons.length >= 1 && gamepad.buttons[0].pressed) {
       EmitPaint();
     }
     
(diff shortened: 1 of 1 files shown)


input-explainer.md
```diff
@@ -280,6 +280,78 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+The WebXR Device API introduces a new standard controller layout indicated by the `mapping` value of `xr-standard`. (Additional mapping variants may be added in the future if necessary.) This defines a specific layout for the inputs most commonly found on XR controller devices today. The following table describes the buttons/axes and their physical locations:
+
+| Button/Axis  | Location                |
+| ------------ | ----------------------- |
+| buttons[0]   | Primary trigger         |
+| buttons[1]   | Touchpad/Joystick click |
+| buttons[2]   | Grip/Secondary trigger  |
+| axes[0]      | Touchpad/Joystick X     |
+| axes[1]      | Touchpad/Joystick Y     |
```
thetuvix: I very much like the idea of an `"xr-standard"` mapping with inactive slots for missing parts.  Even if we then pursue an app-level mapping mechanism as @NellWaliczek described, we'll be taking far less forward-compat and risk if that's built on top of a consistent web API mapping like this.

As we think through the details of this proposal vs. the #462 proposal, there are two separate design questions that we should tease apart:
* Do we make a new "tracked controller" API or reuse the `Gamepad` API? _(the vote at the F2F here was quite decisive)_
* Do we represent `trigger`/`touchpad`/`joystick`/`grip` separately, or semantically cluster them into "primary trigger", "touchpad/joystick" and "grip/secondary trigger" buckets? _(this was not yet addressed)_

In designing such an`"xr-standard"` mapping, an alternative here is to just go all the way and create a more detailed mapping that skips this semantic clustering.  For example, here, axes 0 and 1 are multiplexed across both touchpad and joystick, depending on what the device has.  Beyond the fact that we ship a device with both :smile:, I generally don't see app logic that expects either touchpad or joystick end up accidentally doing something reasonable when confronted with the opposite control.  For example, if apps are tested only on joysticks, they will often build interactions that expect the XY to start at (0, 0) for each interaction, which is not how touchpads work during use and often users end up blocked.

In the spirit of the original #462 proposal, we could instead establish explicit trigger, touchpad, joystick and primary controller button/axis IDs here, with UAs just leaving slots inactive when a control is not present on a device.  This creates a very predictable mapping and helps reduce the risk of per-UA bugs, since there are less judgment calls for each UA to make in deciding what is "primary" and "secondary".

One other risk of putting the semantic bucketing here is that it encourages app use of `buttons[0].pressed` rather than the `onselect` event.  This would be strictly worse for app compatibility, as it would exclude users of devices using hand-tracking or other non-controller XR input forms from proceeding.  If we want to establish an additional semantic bucket for secondary press, it may be better to just add `onsecondary` or such as a sibling to `onselect`, which also gives us an extra user-initiated event.

The counterbalancing risk with this more explicit approach is that an app developer who codes to this layer directly really does just think about one controller, mapping a key action to the "trigger" button ID even though the user's controller has no trigger.  However, if these low-level button/axis IDs did semantically cluster as proposed, but then a layer like @NellWaliczek described unpacks those values out anyway to strongly-named `trigger`, `touchpad` and `joystick` attributes, developers using that higher-level input API would fall into the same situation anyway.  It may be cleaner as everyone experiments with higher-level input abstractions to just make the low-level `"xr-standard"` mapping here very explicit about well-known parts.

input-explainer.md
```diff
@@ -177,22 +177,34 @@ Most applications will want to visually represent the input sources somehow. The
...
+
+Controller devices should generally only be rendered if the `XRSession`'s `environmentBlendMode` is `'opaque'`, as any other mode implies that the user can see any physical device they may be holding. Rendering cursors or target rays may still be desirable.
+
+If the input source is a handheld device, the `XRInputSource` object should have a non-null `gamepad` attribute. The `Gamepad`'s `id` is used to determine what should be rendered if the app intends to visualize the input source itself, rather than an alternative virtual object. (See the section on [Button and Axis State](#button-and-axis-state) for more details.)
```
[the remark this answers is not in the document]
NellWaliczek: Also do we want to state that there must be a non-null `gripSpace` if the gamepad isn't null?

input-explainer.md
```diff
@@ -177,22 +177,34 @@ Most applications will want to visually represent the input sources somehow. The
...
 
-  // Controller meshes should not be rendered on transparent displays (AR), so
-  // only render a controller mesh if the XREnvironmentBlendMode is 'opaque' and
-  // we have a valid gripPose to transform it with.
-  if (gripPose && xrFrame.session.environmentBlendMode == 'opaque') {
-    let controllerMesh = getControllerMesh(inputSource);
+  // Only render a controller mesh if we have a valid gripPose.
+  let gripPose = xrFrame.getPose(inputSource.gripSpace, xrFrameOfRef);
+  if (gripPose) {
     renderer.drawMeshAtTransform(controllerMesh, gripPose.transform);
```
NellWaliczek: When you rebase this, can we switch from straight-up rendering the mesh to just saving it off for the imaginary renderer to do?  I think there may be an old commit hanging around somewhere that I'd done this if you want me to dig it up...

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
 
+## Button and Axis State
+
+Some applications need more than point-and-click style interaction provided by the `select` events. For devices that make use of controllers with buttons and axes, more complete information about the state of those inputs can be observed via the `XRInputSource`'s `gamepad` attribute.
+
+If the `XRInputState` represents an XR controller device with buttons or axes the `gamepad` attribute is an instance of a [`Gamepad`](https://w3c.github.io/gamepad/#gamepad-interface) interface, and `null` otherwise. Examples of controllers that may expose their state this way include Oculus Touch, Vive wands, Oculus Go and Daydream controllers, or other similar devices. Controllers not directly associated with the XR device, such as the majority of traditional gamepads, should not be exposed using this interface. Tracked devices without discreet inputs, such as optical hand tracking, should also have a `null` `gamepad` attribute.
```
NellWaliczek: > should not be exposed using this interface.

I've come around to agree this should say "must not"

NellWaliczek: And at the very least, if we do report a traditional gamepad here, it should not be present in the `navigator.getGamepads()` array at the same time.

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+`Gamepad` instances reported in this way have several notable behavioral changes vs. the ones reported by `navigator.getGamepads()`:
+
+  - `Gamepad` instances connected to an `XRInputSource` must not be included in the array returned by `navigator.getGamepads()`.
+  - The `Gamepad`'s `index` attribute must be `-1`.
+  - The `Gamepad`'s `connected` attribute must always be `true`.
```
NellWaliczek: What about on `XRInputSources` from sessions that have ended?  What about during the blur state?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+  - The `Gamepad`'s `mapping` attribute must be `xr-standard`.
+
+Finally, the `id` attribute for `Gamepad`s surfaced by the WebXR API are more strictly formatted than those of traditional gamepads in order to make them a more appropriate key for determining rendering assets.
+
+  - The `id` MAY be `'unknown'` if the type of input source cannot be reliably identified or the UA determines that the input source type must be masked for any reason. Applications should render a generic input device in this case.
+  - Inline sessions MUST only expose `id`s of `'unknown'`.
```
NellWaliczek: This is an interesting question... are we going to allow `tracked-pointer` input sources for `inline` sessions?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+  - Otherwise the `id` should be a lower-case string that describes the physical input source.
+    - For most devices this SHOULD be of the format `<vendor>-<product-id>`. For example: `oculus-touch`. UAs SHOULD make an effort to align on the strings that are returned for any given device.
```
NellWaliczek: Just to confirm you mean motion controller vendor and not headset vendor, yes?  In other words would that mean "valve-knuckles" for the Valve Knuckles controller?  What does that mean for the Windows Motion controller that comes with the Samsung Odyssey?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+  - `Gamepad` instances connected to an `XRInputSource` must not be included in the array returned by `navigator.getGamepads()`.
+  - The `Gamepad`'s `index` attribute must be `-1`.
+  - The `Gamepad`'s `connected` attribute must always be `true`.
+  - The `Gamepad`'s `mapping` attribute must be `xr-standard`.
+
+Finally, the `id` attribute for `Gamepad`s surfaced by the WebXR API are more strictly formatted than those of traditional gamepads in order to make them a more appropriate key for determining rendering assets.
+
+  - The `id` MAY be `'unknown'` if the type of input source cannot be reliably identified or the UA determines that the input source type must be masked for any reason. Applications should render a generic input device in this case.
+  - Inline sessions MUST only expose `id`s of `'unknown'`.
+  - Otherwise the `id` should be a lower-case string that describes the physical input source.
```
NellWaliczek: Should or must?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+    - For most devices this SHOULD be of the format `<vendor>-<product-id>`. For example: `oculus-touch`. UAs SHOULD make an effort to align on the strings that are returned for any given device.
+    - It should not include an indication of the handedness of the input source, as that should be implied by the `handedness` attribute.
```
NellWaliczek: Should not or must not?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+
+All other attributes behave as described in the [Gamepad](https://w3c.github.io/gamepad/) specification.
+
+The WebXR Device API introduces a new standard controller layout indicated by the `mapping` value of `xr-standard`. (Additional mapping variants may be added in the future if necessary.) This defines a specific layout for the inputs most commonly found on XR controller devices today. The following table describes the buttons/axes and their physical locations:
+
+| Button/Axis  | Location                          |
```
NellWaliczek: Nit: Might be a good idea to have these be two separate tables?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+  let inputSource = primaryInputSource;
+
+  // Check to see if the input source has buttons/axes and is using the standard
+  // mapping.
+  if (inputSource && inputSource.gamepad &&
+      inputSource.gamepad.mapping == 'xr-standard') {
+    let gamepad = inputSource.gamepad;
+    
+    // Use joystick or touchpad values for movement.
+    MoveUser(gamepad.axes[0], gamepad.axes[1]);
```
NellWaliczek: I swear I saw a version in which you'd fixed this sample so it didn't assume there were axes when that may not necessarily be true...?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+    }
+    
+    // etc.
+  }
+
+  // Do the rest of typical frame processing...
+}
+```
+
+If the application includes interactions that require user activation (such as starting media playback), the application can listen to the `XRInputSource`s `select` events, which fire for every pressed and released button on the controller. When triggered by a controller input, the `XRInputSourceEvent` will include a `buttonIndex` other than `-1` to indicate which button on the gamepad triggered the event.
```
NellWaliczek: How do we plan to rationalize with the design work happening around events in the gamepad spec? If we allow those events to fire, we're somewhat breaking the mental model that all XR input fires one event type.  We'd also need to deal with the fact that those events wouldn't be aligned to XRFrame timing necessarily.  Also, we should find out the timeframe in which the design might actually be available to be used? https://github.com/w3c/gamepad/pull/15

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+|-------------|------------------|---------------------------|
+| button[0]   | "trigger"        | "[device]/trigger"        |
+| button[1]   | "touchpad-click" | "[device]/touchpad/click" |
+| button[2]   | "grip"           | "[device]/grip"           |
+| axis[0]     | "touchpad-x"     | "[device]/touchpad/x"     |
+| axis[1]     | "touchpad-y"     | "[device]/touchpad/y"     |
+
+If the API does not provided a way to enumerate the available input devices, the UA should provide bindings for the left and right hand instead of a specific device and expose a `Gamepad` for any hand that has at least one non-`null` input.
+
+The UA should not make any attempt to circumvent user remapping of the inputs.
```
NellWaliczek: Should not or must not?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+      EmitPaint();
+    }
+    
+    // etc.
+  }
+
+  // Do the rest of typical frame processing...
+}
+```
+
```
NellWaliczek: Just to be explicit, we're not talking about firing events for axis changes or button value changes, yes?  Just on changes to GamepadButton.pressed?

input-explainer.md
```diff
@@ -280,6 +292,91 @@ When the canvas receives a `pointerdown` event an `XRInputSource` is created wit
...
+| buttons[0]   | Primary trigger                   |
+| buttons[1]   | Primary Touchpad/Joystick click   |
+| buttons[2]   | Grip/Secondary trigger            |
+| buttons[3]   | Secondary Touchpad/Joystick click |
+| axes[0]      | Primary Touchpad/Joystick X       |
+| axes[1]      | Primary Touchpad/Joystick Y       |
+| axes[2]      | Secondary Touchpad/Joystick X     |
+| axes[3]      | Secondary Touchpad/Joystick Y     |
+
+Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array. If a device has both a touchpad and a joystick the UA should designate one of them to be the primary axis-based input and expose the other at axes[2] and axes[3] with an associated button at button[3].
```
NellWaliczek: What's the expected behavior of a motion controller with 2+ buttons and no touchpad/thumbstick style inputs?

input-explainer.md
toji: Hey! This just so happens to be what I did anyway! Yay!

input-explainer.md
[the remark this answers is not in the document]
toji: Hm............. What about a GearVR-style device where there's a touchpad on the device itself? Seems natural to want to expose that state via the `gamepad` attribute. Especially because that's already what happens with WebVR. (Probably wouldn't use the `xr-standard` mapping, though.)

input-explainer.md
thetuvix: > And at the very least, if we do report a traditional gamepad here, it should not be present in the `navigator.getGamepads()` array at the same time.

Bingo: It's less that we need to be super-prescriptive about where you enumerate a given device than that we ensure devices are not enumerated in two different APIs, creating an unbounded deduplication requirement that we offload to apps.

input-explainer.md
toji: Probably not, but as indicated above I don't know that we should restrict this to only `tracked-pointer` input sources. Admittedly, though, that's mostly because there's a couple of examples of `gaze` devices that have buttons/axes. I have a hard time seeing how this would be applicable to `screen` input sources.

input-explainer.md
toji: I hedged here slightly because it sounded feasible that if we did something like Alex's suggestion of just USB vendorId-productId they might have different product ids for the left and right hands, and I don't know if we want to force that to be scrubbed?

input-explainer.md
toji: The lack of XRFrame alignment is the biggest issue I would have with trying to use the gamepad events. I think we'd probably just state that gamepads not exposed by `navigator.getGamepads()` don't fire standard gamepad events?

input-explainer.md
toji: Correct, that was my thought.

input-explainer.md
toji: Per our call today, that should be exposed as a gamepad with 2+ buttons and 0 axes, but should not report a mapping of 'xr-standard'. I'm adding some text to the explainer to that effect.

toji:
Some new updates to this PR to address the feedback thus far. PTAL!

toji:
This PR fixes #392 and #336

input-explainer.md
NellWaliczek: Should we be talking to them about including an optional XRFrame on their events?

NellWaliczek:
FYI the mappings proposal from the F2F has been cleaned up and added here: https://github.com/immersive-web/xr-gamepad-mappings

input-explainer.md
toji: I'm not sure I feel like blur state should play into it, but I did update the PR to indicate that it should be false when the input source is removed or the session is ended.

input-explainer.md
toji: I feel like this probably deserves a bit more discussion, so I punted it to #550 (which I linked in the explainer). I'd like to get the rest of this sorted out independently.

input-explainer.md
toji: I don't know that I feel that's the right solution. Maybe I'm just a bit gunshy, though, because we previously tried to pile a bunch of VR stuff on top of Gamepads and that didn't work out very well.

toji:
Updated one more time. I think all feedback should be addressed in some fashion, now.

input-explainer.md
NellWaliczek: I'm ok to file an issue in the WebXR repo to track a unification plan with those proposed events.

toji:
Issue to investigate the events has been filed at #551. I'm going in for the merge!

PR #502, opened 2019-01-30
Merge XRStationaryReferenceSpaceOptions into XRReferenceSpaceOptions

Manishearth:
We currently have `dictionary XRStationaryReferenceSpaceOptions: XRReferenceSpaceOptions`, and `requestReferenceSpace(XRReferenceSpaceOptions options)`.

WebIDL doesn't allow accepting a subtype dictionary as one of its parent types in a method argument, it will be upcasted first. This commit moves the `subtype` field to `XRReferenceSpaceOptions` and removes `XRStationaryReferenceSpaceOptions`, instead having some dynamic checks during `requestReferenceSpace()` to ensure the options are well-formed.

r? @toji cc @kearwood

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -683,8 +683,15 @@ enum XRReferenceSpaceType {
   "unbounded"
 };
 
+enum XRStationaryReferenceSpaceSubtype {
+  "eye-level",
+  "floor-level",
+  "position-disabled"
+};
+
 dictionary XRReferenceSpaceOptions {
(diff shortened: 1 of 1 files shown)


toji:
Thanks @Manishearth! We discussed this briefly at the F2F and it's pretty obvious this was behavior that isn't supported by WebIDL, so I'm merging this now to fix the bug.

PR #504, opened 2019-02-01
Make outputContext part of the XRRenderState

toji:
This was discussed at the Jan 2019 F2F as part of a broader topic of refactoring output contexts. Making output contexts mutable as described in the PR was seen as non-controversial while other aspects of the discussion did not reach an agreement.

Change moves setting the `outputContext` from a session-creation time event that was then immutable to part of the `XRRenderState`, which allows it to be changed at any point. `outputContext`'s can still only be associated with a single session at a time, so it will be removed from any previous session when reassigned.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -624,14 +624,14 @@ dictionary XRRenderStateInit {
   double depthNear;
   double depthFar;
   XRLayer? baseLayer;
-  XRPresentationContext outputContext
+  XRPresentationContext? outputContext
 };
 
 [SecureContext, Exposed=Window] interface XRRenderState {
   readonly attribute double depthNear;
   readonly attribute double depthFar;
   readonly attribute XRLayer? baseLayer;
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -609,12 +622,14 @@ dictionary XRRenderStateInit {
   double depthNear;
   double depthFar;
   XRLayer? baseLayer;
+  XRPresentationContext outputContext
```
NellWaliczek: This should be nullable, yes?

NellWaliczek: And on `XRRenderState`, too.

NellWaliczek:
When you make this change to the spec, let's add the text actually explaining the steps to fill in the context with pixels?

darktears:
This looks like a good change and will simplify experiences written with WebXR. I'm supportive.

toji:
I'm gonna go ahead and merge this then, thanks!

PR #505, opened 2019-02-04
Fix and improve links to external definitions

dontcallmedom:


toji:
Very nice, thank you!

PR #508, opened 2019-02-05
Update explainer.md

sressler:
taken the liberty of updating a bunch of relatively minor grammatical changes and minor working updates, such as splitting up very long sentences ...only 1 or 2 actual edits of any semantics ..really nice explainer! just requesting pull for these edits


toji:

toji:
Thank you!

As it turns out these kind of edits fall under an exception where, being editorial in nature, they are considered non-substantive and can avoid the W3C auth, so we can proceed with the merge either way, but I appreciate you working to get the auth right regardless. :)

sressler:
thanks! I'm github challenged ;-)

PR #510, opened 2019-02-06
Narrow application of the 'unstable' markup in the spec

toji:
Previously the entire spec was contained in one big "unstable" section. This update changes that to only apply to sections of the spec that the editors (conservatively) feel may still see breaking changes. 

Sections not marked "unstable" are **not** considered "done". They are expected to still see updates to prose, algorithm improvements, and new additive functionality, but are not likely to have backwards-compatibility-breaking changes made to them going forward.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -23,7 +23,7 @@ Ignored Vars: layer
 Warning: custom
 Custom Warning Title: Unstable API
 Custom Warning Text:
-  <b>The version of the WebXR Device API represented in this document is incomplete and may change at any time.</b>
+  <b>Parts of the API represented in this document are incomplete and may change at any time.</b>
   <p>While this specification is under development some concepts may be represented better by the <a href="https://github.com/w3c/webvr/blob/master/explainer.md">WebXR Device API Explainer</a>.</p>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -23,7 +23,7 @@ Ignored Vars: layer
 Warning: custom
 Custom Warning Title: Unstable API
 Custom Warning Text:
-  <b>The version of the WebXR Device API represented in this document is incomplete and may change at any time.</b>
+  <b>Parts of API represented in this document are incomplete and may change at any time.</b>
```
NellWaliczek: "Parts of the"

index.bs
```diff
@@ -451,6 +462,7 @@ dictionary XRSessionCreationOptions {
   XRPresentationContext? outputContext = null;
 };
 </pre>
+</section>
```
NellWaliczek: It's probably a silly nit... but I'd kinda like the box not to span across top level headings...

toji: It's a fair point! I've updated this for everything *but* the privacy and security section, because that feels like it should all be taken as a whole anyway.

PR #518, opened 2019-02-11
Introduce spec internal concept and start to use Infra spec definitions

jungkees:
This change defines a spec internal concept, XR device, distinguished from the
XR object's XR device internal slot. The XR device concept is not exposed to the
script surface but just used by spec internal algorithms. The XR device concept
represents an XR device and the XR object's XR device internal slot can hold an
XR device.

This change moves the definition of the list of supported modes from XR object's
XR device internal slot to XR device concept. This makes us be able to define XR
device's states under the concept rather than the platform object's internal
slot and write algorithm steps that reference those states more easly.

This change specifies navigator.xr attribute's getter and the creation of XR
object associated with a Navigator object.

This change introduces the Infra spec defintions to define XR object's
properties and algorithm steps: select an XR device, supportsSeesionMode()
method, requestSession() method for this particular PR. (More to come.)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1183,10 +1183,7 @@ The <dfn method for="XRWebGLLayer">requestViewportScaling(|scaleFactor|)</dfn> m
 
   1. If |scaleFactor| is greater than 1.0 set |scaleFactor| to 1.0.
   1. If |scaleFactor| is less than the [=minimum viewport scale factor=] set |scaleFactor| to the [=minimum viewport scale factor=].
-  1. If the [=/XR device=] places additional device-specific restrictions on viewport size, adjust |scaleFactor| accordingly.
-
-    Issue: Clarify which XR device the XR device in this step references to.
-
+  1. If the [=XR/XR device=] places additional device-specific restrictions on viewport size, adjust |scaleFactor| accordingly.
   1. Set the [=viewport scale factor=] to |scaleFactor|.
 
 </div>
@@ -1246,10 +1243,7 @@ The [=XR compatible=] boolean can be set either at context creation time or afte
 
 When the {{HTMLCanvasElement}}'s getContext() method is invoked with a {{WebGLContextAttributes}} dictionary with {{xrCompatible}} set to <code>true</code>, run the following steps:
 
-  1. [=Create the WebGL context=] as usual, ensuring it is created on a [=compatible graphics adapter=] for the [=/XR device=].
-
-    Issue: Clarify which XR device the XR device in this step references to.
-
+  1. [=Create the WebGL context=] as usual, ensuring it is created on a [=compatible graphics adapter=] for the [=XR/XR device=].
   1. Let |context| be the newly created WebGL context.
   1. Set |context|'s [=XR compatible=] boolean to true.
   1. Return |context|.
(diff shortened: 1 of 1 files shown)


jungkees:
@toji, @NellWaliczek, I tried to separate the XR device concept out from the XR object's XR device property. I am aware of the history that we removed the XR device platform object exposed to the script surface. This XR device spec concept is internal to the spec algorithms and makes us be able to model what we access and write spec algorithms more easily. Also I suggest we use [Infra](https://infra.spec.whatwg.org) spec definitions. Please take a look.

index.bs
```diff
@@ -1160,13 +1183,16 @@ The <dfn method for="XRWebGLLayer">requestViewportScaling(|scaleFactor|)</dfn> m
 
   1. If |scaleFactor| is greater than 1.0 set |scaleFactor| to 1.0.
   1. If |scaleFactor| is less than the [=minimum viewport scale factor=] set |scaleFactor| to the [=minimum viewport scale factor=].
-  1. If the [=XR device=] places additional device-specific restrictions on viewport size, adjust |scaleFactor| accordingly.
+  1. If the [=/XR device=] places additional device-specific restrictions on viewport size, adjust |scaleFactor| accordingly.
+
+    Issue: Clarify which XR device the XR device in this step references to.
```
toji: This would be [=XR/XR device=]

index.bs
```diff
@@ -1212,23 +1238,26 @@ partial interface mixin WebGLRenderingContextBase {
...
 The [=XR compatible=] boolean can be set either at context creation time or after context creation, potentially incurring a context loss. To set the [=XR compatible=] boolean at context creation time, the {{xrCompatible}} context creation attribute must be set to <code>true</code> when requesting a WebGL context.
 
 <div class="algorithm" data-algorithm="create-with-compatible-xr-device">
 
 When the {{HTMLCanvasElement}}'s getContext() method is invoked with a {{WebGLContextAttributes}} dictionary with {{xrCompatible}} set to <code>true</code>, run the following steps:
 
-  1. [=Create the WebGL context=] as usual, ensuring it is created on a [=compatible graphics adapter=] for the [=XR device=].
+  1. [=Create the WebGL context=] as usual, ensuring it is created on a [=compatible graphics adapter=] for the [=/XR device=].
+
+    Issue: Clarify which XR device the XR device in this step references to.
```
toji: This is also [=XR/XR device=]

index.bs
```diff
@@ -1251,13 +1280,19 @@ When the <dfn method for="WebGLRenderingContextBase">makeXRCompatible()</dfn> me
   1. Let |context| be the target {{WebGLRenderingContextBase}} object.
   1. If |context|'s [=WebGL context lost flag=] is set, [=reject=] |promise| with an {{InvalidStateError}} and abort these abort these steps.
   1. If |context|'s [=XR compatible=] boolean is <code>true</code>, [=/resolve=] |promise| and abort these steps.
-  1. If |context| was created on a [=compatible graphics adapter=] for the [=XR device=]:
+  1. If |context| was created on a [=compatible graphics adapter=] for the [=/XR device=]:
+
+    Issue: Clarify which XR device the XR device in this step references to.
```
toji: [=XR/XR device=]

index.bs
```diff
@@ -1251,13 +1280,19 @@ When the <dfn method for="WebGLRenderingContextBase">makeXRCompatible()</dfn> me
...
       1. Set |context|'s [=XR compatible=] boolean to <code>true</code>.
       1. [=/Resolve=] |promise| and abort these steps.
-  1. Queue a task to perform the following steps:
+  1. [=Queue a task=] to perform the following steps:
       1. Force |context| to be lost and [=handle the context loss=] as described by the WebGL specification.
       1. If the [=canceled flag=] of the "webglcontextlost" event fired in the previous step was not set, [=reject=] |promise| with an {{AbortError}} and abort these steps.
-      1. [=Restore the context=] on a [=compatible graphics adapter=] for the [=XR device=].
+      1. [=Restore the context=] on a [=compatible graphics adapter=] for the [=/XR device=].
+
+        Issue: Clarify which XR device the XR device in this step references to.
```
toji: [=XR/XR device=]

index.bs
jungkees: Yes, this shouldn't be [=/XR device=] but needs to be a concrete reference that the step can use. Here, [=XR/XR device=] would still not be entirely a correct handle as we don't have (specification) ways to reference to the XR object's state within this algorithm. Let's get it sorted going forward. For now, I addressed your comment by using [=XR/XR device=], which is closer to what we mean here over [=/XR device=].

index.bs
jungkees: Same above

index.bs
jungkees: Same above

index.bs
jungkees: Same above

toji:
Thanks for the update!

PR #519, opened 2019-02-12
 Add inlineVerticalFieldOfView to XRRenderState

### LINKED ISSUE #272: Default FoV for magic window canvases
Something I ran into while sketching out some sample code for 2.0: The new magic window structure has the UA providing the projection matrix for magic window canvases based largely on the properties of the canvas itself, the environment, and a small amount of developer supplied scene information. The classic components needed for a simple projection matrix are:

- near/far bounds: VRSession attributes
- aspect ratio: Computed from canvas dimensions
- Horizontal or vertical FoV: Varies

The FoV is a funny one, because ideally we want it to align to several things very precisely to retain the feeling of it being an actual window. In a phone AR scenario it should be precisely matched to the properties of the camera that's providing the video feed (All AR APIs have a way of communicating this.) In the case of an in-VR browser or a device like a ZSpace system you'll want to compute the FoV to match the user's eye positions in relation to the canvas position on screen. (Well, you'll probably be dealing with frusta at that point rather than FoV, but that's besides the point.) And obviously in presentation mode we want to report values that align to the device optics.

However, if you don't have any information about where the user's eyes are in relation to the canvas, and you're not matching to an external data source then we still need to some up with a field of view somehow. And it needs to be consistent across UAs so that carefully aligned content doesn't clip on one device but not on others. The question is: how do we determine that number?

Easiest way to handle this is to roll some dice, declare that to be the default value, and move along. Being slightly more practical we can plop down another attribute next to `depthNear`/`depthFar` called `defaultVerticalFieldOfView` or something similar and call it good. We'd still need our dice roll to come up with a default for the default attribute, but otherwise it just uses whatever the developer sets _unless_ there's a canonical source of FoV. We could also try to get really technical about it and say "Projection matrices should be computed assuming the user's eye is one meter away from the center of the device perpendicularly blah blah blah" and come up with a number based on that, but if we're being honest with ourselves that's not any more useful than a dice roll number and a lot more work.

So the *TL;DR* is: We probably need a  `defaultVerticalFieldOfView` (or horizontal) on `VRSession` that describes the Field of View used when computing magic window matrices that don't need to match external data. We also need to collectively agree on what the default value for that attribute should be. If anyone has a more clever/useful idea of how to handle it, or thinks there's a better way of expressing the necessary values than a simple FoV, I'd appreciate hearing it!

toji:
Fixes #272.

Pretty simple feature, based on the feedback from the F2F. It's apparent to me that we're not going to magically come up with a default value that has a good reason behind it, so I plugged 90deg in there just because it's a clean number that's easy to guess and gives reasonable results. Also left it as a single direction that can be specified (vertical) because you can pretty trivially do the math to convert from one to the other and vertical seems to be the value that most developers/game players are used to setting. (This because monitors are traditionally wider than tall and it's easier to get predictable results out of a single vertical FoV when jumping between 4:3 and 16:9 aspect ratios).

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -555,6 +555,23 @@ function drawScene() {
 }
 ```
 
+### Changing the Field of View for inline sessions
+
+Whenever possible projection matrices delivered by the API should make use of physical properties such as the headset optics or camera lens to determine the field of view to use. However, for most inline content there isn't any physically based values to draw from. In order to provide a unified render pipeline for inline content an arbitrary field of view must be selected.
+
+By default a vertical field of view of 0.5 radians (90 degrees) is used when no physically-based field of view is provided. The horizontal field of view is computed from the vertical field of view based on the width/height ratio of the `outputContext`'s canvas.
+
+If a different default field of view is desired, it can be specified by passing a new `defaultVerticalFieldOfView` value, in radians, to the `updateRenderState` method:
+
+```js
+// This changes the default vertical field of view to 0.4 radians (72 degrees).
+xrSession.updateRenderState({
+  defaultVerticalFieldOfView: 0.4 * Math.PI,
+});
+```
+
+The UA is allowed to clamp the value, and it is expected that if any physically-based values are available they should always be used in favor of the default.
+
 ## Appendix A: I don’t understand why this is a new API. Why can’t we use…
 
 ### `DeviceOrientation` Events
@@ -646,6 +663,7 @@ enum XREnvironmentBlendMode {
 dictionary XRRenderStateInit {
   double depthNear;
   double depthFar;
+  double defaultVerticalFieldOfView;
   XRLayer? baseLayer;
   XRPresentationContext? outputContext
 };
@@ -653,6 +671,7 @@ dictionary XRRenderStateInit {
 [SecureContext, Exposed=Window] interface XRRenderState {
   readonly attribute double depthNear;
   readonly attribute double depthFar;
+  readonly attribute double defaultVerticalFieldOfView;
   readonly attribute XRLayer? baseLayer;
   readonly attribute XRPresentationContext? outputContext;
 };
--- a/explainer.md
+++ b/explainer.md
@@ -557,20 +557,22 @@ function drawScene() {
 
 ### Changing the Field of View for inline sessions
 
-Whenever possible projection matrices delivered by the API should make use of physical properties such as the headset optics or camera lens to determine the field of view to use. However, for most inline content there aren't any physically based values from which to infer a field of view. In order to provide a unified render pipeline for inline content an arbitrary field of view must be selected.
+Whenever possible the matrices given by `XRView`'s `projectionMatrix` attribute should make use of physical properties, such as the headset optics or camera lens, to determine the field of view to use. Most inline content, however, won't have any physically based values from which to infer a field of view. In order to provide a unified render pipeline for inline content an arbitrary field of view must be selected.
 
 By default a vertical field of view of 0.5 radians (90 degrees) is used for inline sessions. The horizontal field of view can be computed from the vertical field of view based on the width/height ratio of the `outputContext`'s canvas.
 
-If a different default field of view is desired, it can be specified by passing a new `verticalFieldOfView` value, in radians, to the `updateRenderState` method:
+If a different default field of view is desired, it can be specified by passing a new `defaultFieldOfView` value, in radians, to the `updateRenderState` method:
 
 ```js
 // This changes the vertical field of view to 0.4 radians (72 degrees).
 xrSession.updateRenderState({
-  verticalFieldOfView: 0.4 * Math.PI,
+  defaultFieldOfView: 0.4 * Math.PI,
 });
 ```
 
-The UA is allowed to clamp the value, and if any physically-based values are available they must always be used in favor of developer supplied values. Attempting to set a `verticalFieldOfView` value on an immersive session will cause `updateRenderState()` to throw an `InvalidStateError`.
+The UA is allowed to clamp the value, and if a physically-based field of view is available it must always be used in favor of the default value.
+
+Attempting to set a `defaultFieldOfView` value on an immersive session will cause `updateRenderState()` to throw an `InvalidStateError`. `XRRenderState.defaultFieldOfView` must return `null` on immersive sessions.
 
 ## Appendix A: I don’t understand why this is a new API. Why can’t we use…
 
@@ -663,7 +665,7 @@ enum XREnvironmentBlendMode {
 dictionary XRRenderStateInit {
   double depthNear;
   double depthFar;
-  double verticalFieldOfView;
+  double defaultFieldOfView;
   XRLayer? baseLayer;
   XRPresentationContext? outputContext
 };
@@ -671,7 +673,7 @@ dictionary XRRenderStateInit {
 [SecureContext, Exposed=Window] interface XRRenderState {
   readonly attribute double depthNear;
   readonly attribute double depthFar;
-  readonly attribute double verticalFieldOfView;
+  readonly attribute double? defaultFieldOfView;
   readonly attribute XRLayer? baseLayer;
   readonly attribute XRPresentationContext? outputContext;
 };
--- a/explainer.md
+++ b/explainer.md
@@ -557,20 +557,20 @@ function drawScene() {
 
 ### Changing the Field of View for inline sessions
 
-Whenever possible projection matrices delivered by the API should make use of physical properties such as the headset optics or camera lens to determine the field of view to use. However, for most inline content there aren't any physically based values to draw from. In order to provide a unified render pipeline for inline content an arbitrary field of view must be selected.
+Whenever possible projection matrices delivered by the API should make use of physical properties such as the headset optics or camera lens to determine the field of view to use. However, for most inline content there aren't any physically based values from which to infer a field of view. In order to provide a unified render pipeline for inline content an arbitrary field of view must be selected.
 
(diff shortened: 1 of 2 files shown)


explainer.md
```diff
@@ -532,6 +532,23 @@ function drawScene() {
 }
 ```
 
+### Changing the Field of View for inline sessions
+
+Whenever possible projection matrices delivered by the API should make use of physical properties such as the headset optics or camera lens to determine the field of view to use. However, for most inline content there aren't any physically based values to draw from. In order to provide a unified render pipeline for inline content an arbitrary field of view must be selected.
```
NellWaliczek: nit: "to draw from" is a bit too close to sounding like a double entendre.  Perhaps "from which to infer a field of view".

explainer.md
```diff
@@ -532,6 +532,23 @@ function drawScene() {
...
+
+By default a vertical field of view of 0.5 radians (90 degrees) is used for inline sessions. The horizontal field of view can be computed from the vertical field of view based on the width/height ratio of the `outputContext`'s canvas.
+
+If a different default field of view is desired, it can be specified by passing a new `verticalFieldOfView` value, in radians, to the `updateRenderState` method:
+
+```js
+// This changes the default vertical field of view to 0.4 radians (72 degrees).
```
NellWaliczek: nit: Remove the word "default" from the comment

explainer.md
```diff
@@ -532,6 +532,23 @@ function drawScene() {
...
+xrSession.updateRenderState({
+  verticalFieldOfView: 0.4 * Math.PI,
+});
+```
+
+The UA is allowed to clamp the value, and it is expected that if any physically-based values are available they should always be used in favor of the default. Attempting to set a `verticalFieldOfView` value on an immersive session will cause `updateRenderState()` to throw an `InvalidStateError`.
```
NellWaliczek: > The UA is allowed to clamp the value, and it is expected that if any physically-based values are available they should always be used in favor of the default.

Two things here:
1. I'm inclined to have this be a stronger statement; if physically-based values are available, they MUST always be used in favor of the default.  Though perhaps I'm not recalling a reason for allowing this?
2. That word "default" again ;)

explainer.md
```diff
@@ -623,13 +640,15 @@ enum XREnvironmentBlendMode {
...
   double depthFar;
+  double verticalFieldOfView;
   XRLayer? baseLayer;
   XRPresentationContext? outputContext
 };
 
 [SecureContext, Exposed=Window] interface XRRenderState {
   readonly attribute double depthNear;
   readonly attribute double depthFar;
+  readonly attribute double verticalFieldOfView;
```
NellWaliczek: Should this be nullable and only valid for sessions without physically-driven values?

toji: Ewe, yeah. I see what you're getting at now. If it's just stated as "verticalFieldOfView" developers may look at it thinking it's reporting the physically based field of view in cases where there is one, which we don't want.

So the behavior that you're suggesting that we would report it as `null` any time a physical FoV is present, but the dictionary would not accept a nullable value? That feels a little quirky to me, but workable.

toji:
Updated to take all feedback into account except the last question, which I want to have a bit more discussion on.

explainer.md
klausw: I think the core point of confusion is having to know if this is a recommended value that can be overridden, or a physically-driven value that must be used as-is. Would it help to make this explicit in naming?

For example, for the nullable approach, using `suggestedVerticalFieldOfView` as the field name would make it clearer that the application doesn't have to use it as-is.

Alternatively, if the goal is to simplify applications by always supplying a field of view, there could be a separate `allowFieldOfViewModification` boolean that's true for devices where it's just a suggestion. That seems a bit clunky though.

blairmacintyre: This is an important point.  Providing a default value without indicating that it's arbitrary is a bad idea.

Would it be reasonable to simply omit the value if it's not "real"?  developers can then do something like

`var fov = rendState.verticalFieldOfView || 90;`

klausw: I think there's a minor theoretical advantage to having the device and/or UA provide a default, for example they could adjust the suggested value up or down based on the screen dimensions and typical usage, but I'm not aware of any UA doing this or planning to do so.

How about `var fov = rs.fixedVerticalFieldOfView || rs.suggestedVerticalFieldOfView`, where the UA would supply one or the other but never both? I think this would make the difference fairly clear to avoid accidental misuse, and the burden on apps that don't care is still fairly small since they can just use this conditional as-is.

toji:
Okay, thought long and hard about this and have a way forward that I'm not 100% thrilled with, but I think it's workable. PR has been updated to take it into account, but here's the TL;DR:

 - attrib name is now `defaultFieldOfView`. `defaultInlineVerticalFieldOfView` would be more accurate but... no.
 - The dictionary value is non-nullable, but the read only attribute on the `XRRenderState` *is* nullable. This is slightly weird, but makes sense in terms of usage.
 - If the session is immersive, the `defaultFieldOfView` must be `null`, because it should always be using a physical value of some sort. Attempting to set a `defaultFieldOfView` will throw an exception.
 - If the session is inline the `defaultFieldOfView` will report the **default** value, not the one currently being used. You can try and derive that from the projection matrices if you really need it.

So... does that work for everyone who voiced concerns above? Pinging @NellWaliczek, @klausw, and @blairmacintyre.

NellWaliczek:
I'm struggling with this approach... What if we added an optional nested "inline" attribute to the `XRRenderState`...  So for an immersive session `XRRenderState.inline` would be null, but for an inline session it would be non-null with a child `verticalFieldOfView`?

NellWaliczek:
To clarify... something like this:
```
dictionary XRRenderStateInit {
  double depthNear;
  double depthFar;
  XRInlineRenderStateInit? inlineRenderStateInit;
  XRLayer? baseLayer;
  XRPresentationContext? outputContext
};

dictionary XRInlineRenderStateInit {
  double verticalFieldOfView;
};

[SecureContext, Exposed=Window] interface XRRenderState {
  readonly attribute double depthNear;
  readonly attribute double depthFar;
  readonly attribute XRInlineRenderState? inlineRenderState;
  readonly attribute XRLayer? baseLayer;
  readonly attribute XRPresentationContext? outputContext;
};


[SecureContext, Exposed=Window] interface XRInlineRenderState {
  readonly attribute double verticalFieldOfView;
};
```

NellWaliczek:
By the by, it's worth noting that there are a *number* of places in the WebXR APIs that we pick an arbitrary default value that developers are allowed to attempt to change with no guarantee of success.  And in none of those places do we put the word "default".  For example, a few off the top of my head.
* `XRWebGLLayerInit.ignoreDepthValues` defaults to `false` and `true` may not be accepted.
* `XRWebGLLayerInit.framebufferScaleFactor` defaults to `1` and it's possible other values won't be accepted.
* `XRWebGLLayer.requestViewportScaling()` defaults to `1` and there's no guarantee other values will be accepted.

All this to say.... if we're worried about unnecessarily introducing additional dictionaries or types, then I'd vote for `verticalFieldOfView` and have the promise reject if it's not allowed.

toji:
Okay, back to `verticalFieldOfView` then! PTAL

toji:
Okay, after our last-second "wait, that still doesn't make sense" exchange today, I've changed the name once more to `inlineVerticalFieldOfView`.

Also, to be clear:
 - Name includes "inline" now because it will only ever apply to inline sessions.
 - The interface value is nullable because we don't want to report anything on immersive sessions where it could be misinterpreted as the actual FOV.
 - The dictionary value is *not* nullable, because there's no circumstance under which you should ever need to pass a null. (You will get an exception for passing one on immersive sessions and for inline you shouldn't set it to null ever.)
 - Yes, this is slightly weird.
 - One workaround would be to state that setting it to null for an inline session is equivalent to setting it back to the default (90deg), but given that the default is static I don't see much use for that mechanism.

toji:
Pinging @NellWaliczek to re-review

toji:
Yeah, I'll happily add the spec change to this PR. I've been doing that more frequently recently with issues that are more straightforward, but given that I expected (and got) some significant feedback on this one I had wanted to focus on the simplest language first.

(As a general rule I'd actually like to start getting into a pattern where the bulk of the text is spec and the explainer becomes more of "Here's a high level overview of the concept and go to `this link in the spec` for more details.")

toji:
Okay, spec text is added.

index.bs
```diff
@@ -571,6 +581,8 @@ The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the dis
 
 {{XRRenderState/depthNear}} and {{XRRenderState/depthFar}} is used in the computation of the {{XRView/projectionMatrix}} of {{XRView}}s and determines how the values of an {{XRWebGLLayer}} depth buffer are interpreted. {{XRRenderState/depthNear}} MAY be greater than {{XRRenderState/depthFar}}.
 
+The <dfn attribute for="XRRenderState">inlineVerticalFieldOfView</dfn> attribute defines the default vertical field of view in radians used when computing projection matrices for {{XRSessionMode/inline}} {{XRSession}}s. The projection matrix calculation also takes into account the aspect ratio of the {{XRRenderState/outputContext}}'s {{XRPresentationContext/canvas}}. This value MUST be <code>null</code> for {{XRSessionMode/immersive-vr}} and {{XRSessionMode/immersive-ar}} {{XRSession}}s.
```
NellWaliczek: >This value MUST be <code>null</code> for {{XRSessionMode/immersive-vr}} and {{XRSessionMode/immersive-ar}} {{XRSession}}s.

Let's just use the immersive session property instead of calling out both types.

toji: Done

PR #524, opened 2019-02-14
Indicate the preferred ergonomics of a tracked-pointer ray

### LINKED ISSUE #516: 9.2 Input.XRInputPose.targetRay : pointing source position & direction relative to gripTransform
If the targetRay is the default interaction technic of the WebXR API, we need to ensure that the pointing source position and direction is ergonomic by default.

For the moment, the ray of oculus touch is much more logical from a human hand pose perspective that the vive. The vive controller ray follow the object geometry of the controller (like a remote control) and points (more) "up" while oculus touch points "toward" (more like a gun) with the same "relaxed" (default) hand pose. There is the same problem with a lot of 3dof remote controllers.

The probably best pointing direction may be seen as when you handle the controller, points with your index finger. It's more natural, don't involve wrist annoying rotation and follow more the arm direction that when you points with the thumb up.

I don't know if I'm clear, but don't hesitate to comment! (I will probably add images to ensure everyone can understand).

Maybe this issue is much more related to hardware controller api...

toji:
Fixed #516.


index.bs
```diff
@@ -1049,7 +1049,7 @@ The <dfn attribute for="XRInputSource">handedness</dfn> attribute describes whic
...
 
   - <dfn enum-value for="XRTargetRayMode">gaze</dfn> indicates the target ray will originate at the user's head and follow the direction they are looking (this is commonly referred to as a "gaze input" device).
-  - <dfn enum-value for="XRTargetRayMode">tracked-pointer</dfn> indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing.
+  - <dfn enum-value for="XRTargetRayMode">tracked-pointer</dfn> indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing. The orientation of the target ray relative to the tracked object SHOULD follow platform-specific ergonomics guidelines when available. In the absence of platform-specific guidance, the target ray SHOULD point in the same direction as the user's index finger if it was outstretched.
```
NellWaliczek: > The orientation of the target ray relative to the tracked object SHOULD follow platform-specific ergonomics guidelines when available.

Is there a reason we wouldn't use "MUST" in this sentence?

toji: Nope, SGTM.

toji:
Review ping: Anything else you'd like to see changed before landing?

index.bs
```diff
@@ -1067,7 +1067,7 @@ The <dfn attribute for="XRInputSource">handedness</dfn> attribute describes whic
...
 
   - <dfn enum-value for="XRTargetRayMode">gaze</dfn> indicates the target ray will originate at the user's head and follow the direction they are looking (this is commonly referred to as a "gaze input" device).
-  - <dfn enum-value for="XRTargetRayMode">tracked-pointer</dfn> indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing.
+  - <dfn enum-value for="XRTargetRayMode">tracked-pointer</dfn> indicates that the target ray originates from either a handheld device or other hand-tracking mechanism and represents that the user is using their hands or the held device for pointing. The orientation of the target ray relative to the tracked object MUST follow platform-specific ergonomics guidelines when available. In the absence of platform-specific guidance, the target ray SHOULD point in the same direction as the user's index finger if it was outstretched.
```
NellWaliczek: > the target ray SHOULD point in the same direction

Any reason why this isn't a MUST either?

toji: It's feasible that the UA has a tracked device that it knows the position and orientation of but no sense of how it's actually shaped, in which case it's difficult to comply with a MUST here.

PR #526, opened 2019-02-15
Changed XRHandedness enum to use 'none' instead of ''

### LINKED ISSUE #515: 9.1 Input.XRInputSource.XRHandedness enum : replace empty string by none
The XRHandedness enum have currently 3 values : "", "left" and "right".
I think that "" (the empty string) is not a proper way to describe something, that's why I propose to replace it with "none".

toji:
Fixed #515.

Thanks to @HyroVitalyProtago for the suggestion! Empty string was definitely a weird enum, and so I do feel like it's appropriate to change it to a more representative string. I had suggested `neutral` in the issue, but upon thinking it through further I feel like @HyroVitalyProtago's original suggestion of `none` is more appropriate. This is because the input in question may not actually be associated with any hand at all, such as is the case with gaze input, whereas `neutral` implies to me that the input is actually associated with a hand but doesn't know or care which. i.e: A Vive wand that hasn't established it's spatial relationship yet strikes me as a "neutral" input device, while a cardboard device has no hand association whatsoever. In both cases `none` feels like an appropriate descriptor of the handedness.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1022,7 +1022,7 @@ An {{XRInputSource}} represents any input mechanism which allows the user to per
 
 <pre class="idl">
 enum XRHandedness {
-  "",
+  "none",
   "left",
   "right"
 };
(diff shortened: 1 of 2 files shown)


NellWaliczek:
As an aside, I'm inclined to change the gamepad mapping stuff I've been working on to say "none" instead of "neutral" to keep in sync.  Yell if you object.

PR #531, opened 2019-02-20
Remove viewMatrix and add XRTransform.inverse()

### LINKED ISSUE #447: Should we just remove `viewMatrix` and rename `transform` to `viewTransform`?
In looking through #440, I wonder if any non-trivial engine with an actual scene graph will ever make use of `viewMatrix`, or if `transform` is _actually_ all they will evaluate each frame for all purposes, including positioning the camera in the scene relative to the app's chosen reference space origin.

As a case study, in the Windows MR native APIs, we exposed a similar member to `viewMatrix`: [`HolographicCameraPose.TryGetViewTransform()`](https://docs.microsoft.com/en-us/uwp/api/windows.graphics.holographic.holographiccamerapose.trygetviewtransform), returning you a `Left` and `Right` view matrix, each as a `Matrix4x4`.  However, when engines like Unity get a hold of this each frame, they immediately decompose it into a position and orientation for the view, which they then apply as the world transform for their `Camera`'s `GameObject` in their scene graph.  At the end of the frame, the engine then just calculates its own view matrix from the `Camera`'s final position and orientation, identically to how it would for a non-XR game where all camera movement is app-driven.

My own mental model here is that WebXR is not so much the first 10% of an opinionated XR render engine (e.g. dealing directly with view matrices, etc.), as it is **a source of low-policy poses for tracked objects**, such as viewers, input sources, planes, meshes and more.  UAs can consistently deliver those poses (including viewer poses) as positions and orientations within app-created reference spaces, with engines then picking up from there to position their scene cameras appropriately, with the engine then calculating view matrices relative to their scene origin precisely as they've done for many years.

In that model, do we see a clear use case for engines to use `viewMatrix` without just decomposing it into `transform` and manually recomposing it before rendering anyway?  If not, perhaps we should just remove the `viewMatrix` attribute and rename `transform` to `viewTransform`, defining it simply as you have here.

_Originally posted by @thetuvix in https://github.com/immersive-web/webxr/pull/440_

toji:
Fixes #447.

Ooh, hey! I'm gonna try GitHub's shiny new "Draft Pull Request" thingy here. ('Cuz this PR is clearly not quite there yet, but I feel it's worth discussing anyway.)

This was discussed on the IW call today, but I'll recap here: There's been some ongoing discussion about removing `XRView.viewMatrix` because it's computable from the values given in the `XRView.transform`. I've been a bit reluctant to do that simply because if someone *is* using the matrix then this makes it significantly more involved to get the same value back.

Separately, we had a discussion around the `originOffset` at the January 2019 Face to Face and one of the outcomes of that was that no matter how we chose to interpret the value given there we'd end up feeling "backwards" about 50% of the time.

This CL attempts to make both of those issues easier by introducing an `inverse()` method to the `XRRigidTransform` interface, which simply returns another `XRRigidTransform` that represents the opposite transform. ("inverse" rather than "invert" because the latter sounds like it modifies the object in place to me. Happy to take opposing opinions on that one.) The math to do this isn't terrible (though as you can tell from the CL I haven't fully worked out the algorithm yet) but it will be a big developer convenience if there's a mechanism to do it in the API.

With this function in place it means that developers that find that their use of `originOffset` is backwards from what they were expecting can correct it with a simple

```
referenceSpace.originOffset = transform.inverse();
```

Similarly, for developers that would really prefer to use a `viewMatrix` directly, the call now becomes

```
// Previously view.viewMatrix
view.transform.inverse().matrix;
```

Which is admittedly more verbose but also makes use of a more broadly applicable API mechanism rather than introducing data duplication that we presume most developers won't need.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -833,7 +833,6 @@ enum XREye {
 [SecureContext, Exposed=Window] interface XRView {
   readonly attribute XREye eye;
   readonly attribute Float32Array projectionMatrix;
-  readonly attribute Float32Array viewMatrix;
   readonly attribute XRRigidTransform transform;
 };
 </pre>
@@ -842,12 +841,9 @@ The <dfn attribute for="XRView">eye</dfn> attribute describes which eye this vie
 
 The <dfn attribute for="XRView">projectionMatrix</dfn> attribute provides a [=matrix=] describing the projection to be used when rendering the [=view=]. It is <b>strongly recommended</b> that applications use this matrix without modification. Failure to use the provided projection matrices when rendering may cause the presented frame to be distorted or badly aligned, resulting in varying degrees of user discomfort.
 
-The <dfn attribute for="XRView">viewMatrix</dfn> attribute provides a [=matrix=]
-describing the view transform to be used when rendering the [=view=]. The [=matrix=] represents the inverse of the {{XRView/transform}}'s {{XRRigidTransform/matrix}}. It is <b>strongly recommended</b> that applications use this matrix without modification. Failure to use the provided view matrices when rendering may cause the presented frame to be distorted or badly aligned, resulting in varying degrees of user discomfort.
+The <dfn attribute for="XRView">transform</dfn> attribute is the {{XRRigidTransform}} of the viewpoint.
 
-The <dfn attribute for="XRView">transform</dfn> attribute is the {{XRRigidTransform}} of the viewpoint. 
-
-NOTE: The {{XRView/transform}} can be used to position camera objects in many rendering libraries instead of using the {{XRView/viewMatrix}} directly if the library is more naturally set up to consume data in that format.
+NOTE: The {{XRView/transform}} can be used to position camera objects in many rendering libraries. If a more traditional view matrix is needed by the application one can be retrieved by calling `view.transform.inverse().matrix`.
 </section>
 
 XRViewport {#xrviewport-interface}
@@ -925,6 +921,8 @@ interface XRRigidTransform {
   readonly attribute DOMPointReadOnly position;
   readonly attribute DOMPointReadOnly orientation;
   readonly attribute Float32Array matrix;
+
+  XRRigidTransform inverse();
 };
 </pre>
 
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -939,6 +937,18 @@ The <dfn attribute for="XRRigidTransform">matrix</dfn> attribute returns the tra
 
 An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y: 0, z: 0 w: 1 }</code> and an {{XRRigidTransform/orientation}} of <code>{ x: 0, y: 0, z: 0, w: 1 }</code> is known as an <dfn>identity transform</dfn>.
 
+The {{XRRigidTransform/inverse()}} method returns an {{XRRigidTransform}} which, if applied to an object that had previously been transformed by the original {{XRRigidTransform}}, would undo the transform and return the object to it's initial pose.
+
+<div class="algorithm" data-algorithm="rigid-transform-inverse">
+
+When the <dfn method for="XRRigidTransform">inverse()</dfn> method is invoked, the user agent MUST run the following steps:
+
+  1. Figure out WTF the math that goes here should be.
```
NellWaliczek: I spent a couple seconds trying to figure out which math algorithm had that TLA before realizing.... ^_^

blairmacintyre:
@NellWaliczek I am concerned about performance implications (not just GC, although that's probably more of a concern) also the wasted effort actually _doing_ the matrix inversion.

It seems weird that if (internally) we have (or might have) the inverse matrix, that we don't supply it here, but for apps to do wasted work.

NellWaliczek:
@blairmacintyre I'm not sure I'm following your concern ^_^;  I think what Brandon is suggesting is to provide that matrix, but to do it via this additional API?  My concern is that it means that some 50% of the time we end up allocating two`XRRigidTransform` objects for each `getViewerPose` call.

blairmacintyre:
@NellWaliczek ah, I wondered if this was a shortcut or if he was proposing that people can just invert it (using JS math) when needed.  Shortcut == good.

The good part about this is it's lazy so we don't need to allocate the second object unless it's needed, right?

NellWaliczek:
Yes, that's right, but since it's not just inverting the matrix (it's creating a whole new XRRigidTransform) that means the the 50% of folks who want the matrix in view matrix form will always be allocating two... perhaps it's not a big deal though?  I know @RafaelCintron and I used to talk about that a lot back in the WebVR days, but I can't recall why it dropped off as a conversation topic...

toji:
Did some thinking on this and decided that maybe it's not appropriate to have an algorithm here that dictates how the inversion must be done. The primary reason for that is that there's a myriad of methods for doing the transform inversion online and, depending on the UA's internals, the optimal version may be different in different situations. Some browser may choose to compute the inverse based on the vector/quaternion, others may already be representing it internally as a matrix and thus find that easier to invert, and others may already have an inverted version cached for various reasons. As such I've pulled the algorithm out and just described the expected output of the function, implicitly leaving the UA to decide the implementation details. Thoughts?

@blairmacintyre: Yeah, this is an API method, not a developer suggestion. And yes, it's lazily evaluated.

PR #535, opened 2019-02-21
Specify that getViewerPose throws an error for non-rAF XRFrames

toji:
Previous explainer text indicated that this scenario should have returned `null`, but that seems inconsistent with the behavior of other similar restrictions, so I've updated it to throw an `InvalidStateError`. Thanks to Jacob DeWitt (for whom I don't yet know a GitHub handle) for helping highlight the inconsistency!


toji:
Thanks! Added the recommended sentence.

PR #536, opened 2019-02-21
Spec: Move outputContext to XRRenderState

toji:
Also documents the algorithm for removing the ouputContext from previously bound sessions when a new session has been associated.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -305,7 +305,6 @@ enum XREnvironmentBlendMode {
 [SecureContext, Exposed=Window] interface XRSession : EventTarget {
   // Attributes
   readonly attribute XRSessionMode mode;
-  readonly attribute XRPresentationContext? outputContext;
   readonly attribute XREnvironmentBlendMode environmentBlendMode;
   readonly attribute XRRenderState renderState;
   readonly attribute XRSpace viewerSpace;
@@ -339,7 +338,6 @@ When an {{XRSession}} is created, the user agent MUST <dfn>initialize the sessio
   1. Let |session| be the newly created {{XRSession}} object.
   1. Let |options| be the {{XRSessionCreationOptions}} passed to {{requestSession()}}.
   1. Initialize |session|'s {{XRSession/mode}} to |options| {{XRSessionCreationOptions/mode}} value.
-  1. Initialize |session|'s {{XRSession/outputContext}} to |options| {{XRSessionCreationOptions/outputContext}} value.
   1. [=Initialize the render state=].
   1. If no other features of the user agent have done so already, perform the necessary platform-specific steps to initialize the device's tracking and rendering capabilities.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1349,15 +1350,31 @@ Canvas Rendering Context {#canvas-rendering-context}
...
+
+<div class="algorithm" data-algorithm="presentation-context-session-update">
+
+When another algorithm indicates it will <dfn>update the XRPresentationContext session</dfn> to {{XRSession}} |session|, run the following steps:
+
+  1. Let |context| be the target {{XRPresentationContext}}.
+  1. Let |prevSession| be |context|'s [=XRPresentationContext/session=].
+  1. If |prevSession| is equal to |session| abort these steps.
+  1. If |prevSession| is not <code>null</code>:
+    1. If |prevSession|'s {{XRSession/renderState}}'s {{XRRenderState/outputContext}} is equal to |context| set |prevSession|'s {{XRSession/renderState}}'s {{XRRenderState/outputContext}} to <code>null</code>.
```
NellWaliczek: Does this need to be queued up for processing by |prevSession|?

index.bs
```diff
@@ -1349,15 +1350,31 @@ Canvas Rendering Context {#canvas-rendering-context}
...
+  1. Set |context|'s [=XRPresentationContext/session=] to |session|.
```
NellWaliczek: If removing the output context from |prevSession| does need to be queued, does this last step also need to be waited on?

index.bs
```diff
@@ -1213,7 +1214,7 @@ The [=native WebGL framebuffer resolution=] is determined by running the followi
...
   1. If |session|'s {{XRSession/mode}} value is not <code>"inline"</code>, set the [=native WebGL framebuffer resolution=] to the resolution required to have a 1:1 ratio between the pixels of a framebuffer large enough to contain all of the session's {{XRView}}s and the physical screen pixels in the area of the display under the highest magnification and abort these steps. If no method exists to determine the native resolution as described, the [=recommended WebGL framebuffer resolution=] MAY be used.
-  1. If |session|'s {{XRSession/mode}} value is <code>"inline"</code>, set the [=native WebGL framebuffer resolution=] to the size of the |session|'s {{XRSession/outputContext}}'s {{XRPresentationContext/canvas}} in physical display pixels and reevaluate these steps every time the size of the canvas changes.
+  1. If |session|'s {{XRSession/mode}} value is <code>"inline"</code>, set the [=native WebGL framebuffer resolution=] to the size of the |session|'s {{XRSession/renderState}}'s {{XRRenderState/outputContext}}'s {{XRPresentationContext/canvas}} in physical display pixels and reevaluate these steps every time the size of the canvas changes.
```
NellWaliczek: Given that the output context can be changed, where are we specifying "native-webgl-framebuffer-resolution" be invoked?

index.bs
toji: It should take effect immediately, IMO. Otherwise you could feasibly go for a full frame with two different sessions both reporting that they are using the same context.

index.bs
toji: Added a line to the `native-webgl-framebuffer-resolution` algorithm indicating it should be re-evaluated when the `outputContext` changes. As for application of that algorithm, that'll probably be something that's detailed whenever we get around to detailing the compositor/swap chain more rigorously.

index.bs
NellWaliczek: I agree that a frame in which two sessions both believe they're using the same context would be bad.... my bigger concern is that two different sessions might not be on the same thread if OffscreenCanvas is used and that as a result this change could cause a problem for `|prevSession|`

toji: Ha! Okay! I knew if I procrastinated on this long enough the answer would come to me... 😅

You can't use an object in more than one worker at a time. Javascript only allows for them to be transferred between workers. (Exception is shared arrays, which only contain binary data, not objects.) AND FURTHERMORE you have to explicitly mark an object as transferable in order to move it between workers and we have not done so with this context type, SO......

It's a non-issue! YAY! 🎉

NellWaliczek: Hah!  Ok, that's fair.

index.bs
NellWaliczek: (small favor to ask... can you hold off on squashing commits until the final merge?  It would have been a lot easier to review this fix if I could look at the additive commit ^_^)

Issue #533, opened 2019-02-20
Clarify that all XRStationaryReferenceSpaceSubtypes must be supported if "stationary" is supported


ddorwin:
If the `"stationary"` `XRReferenceSpaceType` is supported, it is expected that all of the `XRStationaryReferenceSpaceSubtype`s are supported. (If an implementation has enough data for one, it can emulate the others.) This should be clarified in the spec text.

PR #537, opened 2019-02-21
Spec: Stationary subtype support is all-or-nothing

### LINKED ISSUE #532: XRStationaryReferenceSpaceSubtype values are defined in XRStationaryReferenceSpace section
Clicking one of the enum values in the XRStationaryReferenceSpaceSubtype IDL definition (https://immersive-web.github.io/webxr/#enumdef-xrstationaryreferencespacesubtype) goes to text in the XRStationaryReferenceSpace section.

More generally, all the content this section from "There are several subtypes ..." to the Note on `position-disabled` really applies to use in `XRReferenceSpace` or `requestReferenceSpace()`.

Related to that, `XRReferenceSpaceOptions` and its enums are only actually used in `XRSession::requestReferenceSpace()`, so it's maybe a bit odd for them to be defined in `XRReferenceSpace`. `XRReferenceSpaceOptions` and `XRReferenceSpaceType` aren't used by `XRReferenceSpace` or any of its child interfaces, and `XRStationaryReferenceSpaceSubtype` shouldn't be (#513), so it probably doesn't make sense to define them there.

toji:
Fixes #532 and #533

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -693,12 +693,6 @@ enum XRReferenceSpaceType {
   "unbounded"
 };
 
-enum XRStationaryReferenceSpaceSubtype {
-  "eye-level",
-  "floor-level",
-  "position-disabled"
-};
-
 dictionary XRReferenceSpaceOptions {
   required XRReferenceSpaceType type;
   XRStationaryReferenceSpaceSubtype subtype;
@@ -749,6 +743,12 @@ XRStationaryReferenceSpace {#xrstationaryreferencespace-interface}
 An {{XRStationaryReferenceSpace}} represents a tracking space that the user is not expected to move around within. Tracking in a {{stationary}} reference space is optimized for the assumption that the user will not move much beyond their starting point, if at all. For devices with [=6DoF=] tracking, {{stationary}} reference spaces should emphasize keeping the origin stable relative to the user's environment.
 
 <pre class="idl">
+enum XRStationaryReferenceSpaceSubtype {
+  "eye-level",
+  "floor-level",
+  "position-disabled"
+};
+
 [SecureContext, Exposed=Window]
 interface XRStationaryReferenceSpace : XRReferenceSpace {
   readonly attribute XRStationaryReferenceSpaceSubtype subtype;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -765,6 +765,8 @@ There are several subtypes of {{XRStationaryReferenceSpace}}, determined by the
 
 Note: The {{position-disabled}} subtype is primarily intended for use with pre-rendered media such as panoramic photos or videos. It should not be used for most other media types due to user discomfort associated with the lack of a neck model or full positional tracking.
 
+Devices which support {{XRReferenceSpaceType/stationary}} reference spaces MUST support all {{XRStationaryReferenceSpace/subtype}}s.
```
ddorwin: nit: s/which/that/

index.bs
```diff
@@ -749,6 +743,12 @@ XRStationaryReferenceSpace {#xrstationaryreferencespace-interface}
 An {{XRStationaryReferenceSpace}} represents a tracking space that the user is not expected to move around within. Tracking in a {{stationary}} reference space is optimized for the assumption that the user will not move much beyond their starting point, if at all. For devices with [=6DoF=] tracking, {{stationary}} reference spaces should emphasize keeping the origin stable relative to the user's environment.
 
 <pre class="idl">
+enum XRStationaryReferenceSpaceSubtype {
```
ddorwin: Regarding #532:
This change moves the IDL closer to the definition it will link to. However, these values are most relevant in `XRSession`, and the definitions actually describe the behavior of values passed to `XRSession::requestReferenceSpace()` (as part of `XRReferenceSpaceOptions`). As noted in #532, it probably makes sense to move the definitions/prose and IDL (including other things like `XRReferenceSpaceOptions` to `XRSession`.

index.bs
ddorwin: I think you want to replace
  `{{XRStationaryReferenceSpace/subtype}}s`
with
  `{{XRStationaryReferenceSpaceSubtype}} values`

toji:
Thanks David and Nell! Addressed both of David's nits and moved the subtype description up in the file. PTAL!

PR #538, opened 2019-02-25
added a few AR device examples

blairmacintyre:
minor suggestion.  Added a few AR device examples to the list of VR device examples.


toji:
LGTM, thanks! Given that this list keeps growing I think I'll jump in an reformat it into a list, but I'm good with the newly added devices in the meantime.

PR #548, opened 2019-02-28
Add ignoreDepthValues attribute to the XRWebGLLayer

### LINKED ISSUE #523: Encourage use of the depth buffer for the XR WebGL context
For good headpose prediction, it helps if the author provides a depth buffer.
The spec is currently silent on this requirement.

Can we add some language to the current spec to encourage authors to create a valid depth buffer for the scene? Maybe there could also be a boolean on the `XRSessionCreationOptions` if the depth buffer should be used by the graphics backend.

toji:
Fixes #523.

This value instructs the UA to prevent the values left in the depth buffer after a frame is rendered for compositing. Slightly concerned about potential for confusion regarding the `depth` and `ignoreDepthValues` attributes being on the same dictionary, so clarifying bikeshedding is welcome there if you have suggestions.

Given the scope of the feature it felt appropriate to handle both the spec text and explainer text in the same PR. Please take a look!

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -520,6 +520,18 @@ xrSession.updateRenderState({
 });
 ```
 
+### Preventing the compositor from using the depth buffer
+
+By default the depth attachment of an `XRWebGLLayer`'s `framebuffer`, if present, may be used to assist the XR compositor. For example, the scene's depth values may be used by advanced reprojection techniques or to help avoid depth disparity when rendering platform/UA interfaces. This assumes, of course, that the values in the depth buffer are representative of the scene content.
+
+Some applications may violate that assumption, such as when using certain deferred rendering techniques or rendering stereo video. In those cases if the depth buffer's values are used by the compositor it may result in objectionable artifacts. To avoid this, the compositor can be instructed to ignore the depth values of an `XRWebGLLayer` by setting the `ignoreDepthValues` option to `true` at layer creation time:
+
+```js
+let webglLayer = new XRWebGLLayer(xrSession, gl, { ignoreDepthValues: true });
+```
+
+If `ignoreDepthValues` is not set to `true` the The UA is allowed (but not required) to use depth buffer as it sees fit. As a result, barring compositor access to the depth buffer in this way may lead to certain platform or UA features being unavailable or less robust.
+
 ### Handling non-opaque displays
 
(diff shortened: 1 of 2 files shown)


explainer.md
```diff
@@ -520,6 +520,18 @@ xrSession.updateRenderState({
 });
 ```
 
+### Preventing the compositor from using the depth buffer
+
+By default the depth attachment of an `XRWebGLLayer`'s `framebuffer`, if present, may be used to assist the XR compositor. For example, the scene's depth values may be used by advanced reprojection techniques or to help avoid depth disparity when rendering platform/UA interfaces. This assumes, of course, that the values in the depth buffer are representative of the scene content.
```
NellWaliczek: > depth disparity

Is that an official term?  I've never heard it though I have heard "depth inversion" used before...

explainer.md
```diff
@@ -689,6 +701,7 @@ dictionary XRWebGLLayerInit {
   boolean depth = true;
```
NellWaliczek: Per our conversation earlier about naming confusion... Perhaps these become "depthBuffer" or "alphaEnabled"?

toji: I'm highly reluctant to do so because currently these match, in both verbiage and function, the arguments from WebGL context creation exactly.

explainer.md
toji: I'm reasonably certain that I've heard it from multiple different VR developers in one form or another, but doing a quick Google search on the phrase doesn't yield the results I would have thought, so perhaps this isn't as common as I thought. "Depth inversion" isn't a term I've ever heard applied to this specific effect, but I did come across the term "depth conflict" and that seems like a winner all around, with the term inherently suggesting "this is a bad thing that you don't want!" ;)

explainer.md
NellWaliczek: Yeah.  That makes sense...

explainer.md
cabanier: > For example, the scene's depth values may be used by advanced reprojection techniques or to help avoid depth disparity when rendering platform/UA interfaces. This assumes, of course, that the values in the depth buffer are representative of the scene content.

I think this example doesn't need to be part of the spec.

toji:
Updated this so that it also fixes #555, given that everybody seems to agree on it's interpretation. PTAL.

thetuvix:
LGTM!

Per https://github.com/immersive-web/webxr/issues/555#issuecomment-470352605, if we also expect some WebXR content to start trying out reversed-Z, I wonder if we should also specify that UAs **must** allow `depthNear` to be greater than `depthFar`, to avoid some UAs failing to render some scenes.

toji:
Added line to the spec allowing `depthNear` to be greater than `depthFar`. And with that, and Nell's approval, I'm going to merge this.

PR #553, opened 2019-03-04
Spec: Document Gamepad integration

toji:
Work-in-progress PR for gamepad integration spec text. Mostly still need to address the `gamepad.id` formatting. It's far enough along though that some feedback would be useful. I'd especially like feedback on the new mapping diagram and whether it helps or hurts understanding.


index.bs
```diff
@@ -1081,6 +1115,104 @@ The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XR
...
 
 The {{gripSpace}} MUST be <code>null</code> if the input source isn't inherently trackable such as for input sources with a {{targetRayMode}} of {{XRTargetRayMode/gaze}} or {{XRTargetRayMode/screen}}.
+
+The <dfn attribute for="XRInputSource">gamepad</dfn> attribute is a {{Gamepad}} that describes the state of any buttons and axes on the input device. If the {{XRInputSource}} does not represent a device with either multiple buttons or at least one axis the {{gamepad}} attribute MUST be <code>null</code>.
```
NellWaliczek: > either multiple buttons

Is the multiple buttons thing supposed to differentiate it from a mouse? If so ...uh... mice often have more than one button.  Is it explicitly excluding the HoloLens clicker?  What about cardboard?

index.bs
```diff
@@ -1081,6 +1115,104 @@ The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XR
...
+Gamepad API Integration {#gamepad-api-integration}
+-----------------------
+
+{{Gamepad}} instances returned by an {{XRInputSource}}'s {{XRInputSource/gamepad}} attribute behave as described by the <a href="https://w3c.github.io/gamepad/">Gamepad API</a>, with several additional behavioral restrictions:
+
+  - {{XRInputSource/gamepad}} MUST NOT be included in the array returned by {{navigator.getGamepads()}}.
+  - {{XRInputSource/gamepad}}'s {{Gamepad/index}} attribute MUST be <code>-1</code>.
+  - {{XRInputSource/gamepad}}'s {{Gamepad/connected}} attribute MUST be <code>true</code> until the {{XRInputSource}} is removed from the array returned by {{XRSession/getInputSources()}} or the {{XRSession}} is ended.
+
+ISSUE: Document {{Gamepad/id}} restrictions and formatting.
```
NellWaliczek: Should we reference the Github issue #?

index.bs
```diff
@@ -1081,6 +1115,104 @@ The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XR
...
+</section>
+
+<section class="unstable">
+"xr-standard" Gamepad Mapping {#xr-standard-gamepad-mapping}
```
NellWaliczek: Do we want to add a restriction that xr-standard mappings can only happen when the Gamepad is associated with an `XRInputSource`?

index.bs
```diff
@@ -1081,6 +1115,104 @@ The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XR
...
+-----------------------------
+
+The WebXR Device API introduces a new {{GamepadMappingType}} to describe the mapping of common XR controller devices.
+
+<pre class="idl">
+  enum GamepadMappingType {
+    "xr-standard",
```
NellWaliczek: Do we need to also list "standard" since this isn't explicitly partial?

index.bs
```diff
@@ -1081,6 +1115,104 @@ The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XR
...
+  };
+</pre>
+
+The <dfn enum-value for="GamepadMappingType">xr-standard</dfn> mapping indicates that the layout of the buttons and axes of the {{XRInputSource/gamepad}} corresponds as closely as possible to the tables below. In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} the device MUST report a {{XRInputSource/targetRayMode}} of {{XRTargetRayMode/tracked-pointer}}, MUST have at least one touchpad or joystick and MUST have at least one primary button, often a trigger, separate from the touchpad or joystick. If a device does not meet the requirements for the {{GamepadMappingType/xr-standard}} mapping it may still be exposed on an {{XRInputSource}} with the <code>""</code> (empty string) mapping.
```
NellWaliczek: Can we say it must also have a non-null `gripSpace`?

NellWaliczek: > If a device does not meet the requirements for the {{GamepadMappingType/xr-standard}} mapping it may still be exposed on an {{XRInputSource}} with the <code>""</code> (empty string) mapping.

I assume this is for compatibility with Gamepad?  What about "xr-unknown" instead?

index.bs
```diff
@@ -1081,6 +1115,104 @@ The <dfn attribute for="XRInputSource">targetRaySpace</dfn> attribute is an {{XR
...
+    </tr>
+    <tr>
+      <td>axis[3]</td>
+      <td>Secondary Touchpad/Joystick Y</td>
+      <td>No</td>
+    </tr>
+  </tbody>
+</table>
+
+Additional buttons or axes may be exposed after these reserved indices, and SHOULD appear in order of decreasing importance. Devices that lack one of the inputs listed in the tables above MUST still preserve their place in the {{Gamepad/buttons}} or {{Gamepad/axes}} array. If a device has both a touchpad and a joystick the UA MAY designate whichever it chooses to be the primary axis-based input. Buttons reserved by the UA or platform MUST NOT be exposed.
```
NellWaliczek: Can we call out that additional X/Y axis things should always report the X axis before the Y?

index.bs
```diff
@@ -1396,18 +1528,22 @@ XRInputSourceEvent {#xrinputsourceevent-interface}
 interface XRInputSourceEvent : Event {
   readonly attribute XRFrame frame;
   readonly attribute XRInputSource inputSource;
+  readonly attribute long buttonIndex;
```
NellWaliczek: Should this be nullable?

index.bs
toji: Mostly explicitly excluding things like the Cardboard button or HoloLens clicker while leaving room for something like the GearVR touchpad.

index.bs
toji: Yeah, probably. Really annoying that you can't declare partial enums. 😠

index.bs
toji: That would exclude using this as a mechanism for exposing GearVR-style inputs. We can make that choice, but I think we'll get some pushback from it.

toji: Yes, it's for gamepad compatability. I'm not sure there's a benefit to defining a new enum here for "not mapped".

index.bs
toji: That's probably a better pattern than -1, yeah. Thanks!

toji:
Updated to address Nell's comments. PTAL!

index.bs
NellWaliczek: I'm not sure i follow this logic?  That was true when we had xr-standard applying to all xr related gamepads, but that's not the case now, yes?

toji: Ah, I see what you're getting at now! Yes, I'll add that as a restriction.

toji:
Okay, added a requirement for `gripSpace` before you can use `xr-standard`. (Also fixed a minor thing that was bugging me about the mapping diagram.) One more look?

index.bs
toji: I believe so, because this statement limits what should be exposed as a gamepad, as opposed to what is exposed as an `xr-standard` gamepad. I just don't see it being useful to expose "gamepads" for something that is functionally a binary button. The implication is that if the `gamepad` attribute is non-null you know that it means there's at least some level of more advanced input accessible.

I'm willing to reassess that if we feel it's not productive.

NellWaliczek: Ah, I see what you're getting at.  Are you saying that an `XRInputSource` with a `targetRayMode` of 'tracked-pointer` that is a physical object with only a single button would simply fire "select" rather than have a gamepad object.  However, how would someone poll the value of the button if it is capable of reporting analog data?

toji: Yes, that was exactly the intent. Thanks for pointing out the analog button case, though. That *should* be a case where a gamepad is appropriate. Update the PR to account for that and added a non-normative note to explain the single-button select mechanic.

toji:
Review ping for @NellWaliczek. Any further changes you'd like to see? I'm feeling pretty good that this is ready-to-land.

Issue #534, opened 2019-02-21
Clarify when XRPose.emulatedPosition should be true


toji:
A question that came up while reviewing some incoming Chrome code: If a viewer device only offers 3DoF tracking, does that imply that every `XRPose` queried by the system must report `emulatedPosition = true`?

A hypothetical scenario: You have a 3DoF VR headset, which obviously reports `emulatedPosition = true` when getting head poses. I create a `floor-level` and `eye-level` stationary frame of reference, where the `floor-level` is obviously using a height estimate to offset the headset pose. So lots of emulation going on all around. However, if I query `xrFrame.getPose(eyeLevelRefSpace, floorLevelRefSpace)` the relationship between those two spaces is fixed and can be reported with perfect accuracy... so does `emulatedPosition` apply or not? (You can also apply similar logic to two of the same type of reference space with different `originOffset`s or the relationship between a `gripSpace` and `tartgetRaySpace` on a controller.)

My instinct is to say that we shouldn't flag as `emulatedPosition` if the relationship between the spaces in question is fixed or otherwise known to a high degree of accuracy. I wanted to sanity check that here first, though, before adding any clarifying text to the spec.

NellWaliczek:
Paging @thetuvix!  Any update on getting your proposal from the F2F into a PR?

thetuvix:
Apologies for the delay!  As you can imagine, last week was quite busy for us 😄

These are interesting corner cases for `emulatedPosition`, especially given that apps will likely end up using it most of the time to reason about tracking state:
* For a 3DoF device, the distance from `eye-level` to `floor-level` may be **fixed**, in that it won't change its value.  However, being fixed is separate from being **tracked** vs. emulated.  For example, a 6DoF controller that leaves the FOV and drops to 3DoF may lock in place relative to the user's head.  An app that relates the `eye-level` space and the controller's grip space might also get a quite fixed answer at that moment, but the emulated guess being fixed is precisely what the app wants to be warned about when they check whether the position is now `emulated`.
[...]

PR #559, opened 2019-03-12
Tracking loss and tracking recovery

### LINKED ISSUE #243: Handling tracking loss
Handling tracking loss is an important aspect of VR/MR app development.

In current version explainer, the tracking loss handling is only mentioned in [Main Render Loop](https://github.com/w3c/webvr/blob/master/explainer.md#main-render-loop):
> Unless the "headModel" VRFrameOfReference is being used, this function is not guaranteed to return a value. For example, the most common frame of reference, "eyeLevel", will fail to return a viewMatrix or a poseModelMatrix under tracking loss conditions. In that case, the page will need to decide how to respond. It may wish to re-render the scene using an older pose, fade the scene out to prevent disorientation, fall back to a "headModel" VRFrameOfReference, or simply not update. For more information on this see the Advanced functionality section.

There might be several potential improvements for tracking loss handling:

### 1. getDevicePose may fail even with "headModel"
Even for "headModel" VRFrameOfReference, it is still possible to experience "tracking loss".

For example, on OpenVR/Vive, if headset moves out of the tracking area, OpenVR will stops updating `mDeviceToAbsoluteTracking` of  [`TrackedDevicePose_t`](https://github.com/ValveSoftware/openvr/blob/master/headers/openvr.h#L180), sets `bPoseIsValid` to false and `eTrackingResult` to `TrackingResult_Calibrating_OutOfRange`.  It means no any pose data update in this case. Similar for WinMR, app needs to be prepared for [`SpatialLocatability.Unavailable`](https://docs.microsoft.com/en-us/uwp/api/windows.perception.spatial.spatiallocatability) for no pose data. So the sentence "_Unless the "headModel" VRFrameOfReference is being used, this function is not guaranteed to return a value._ " might need to be changed.

### 2. Add handling tracking loss section with sample

It would be better to add code samples of tracking loss handling for different tracking experiences.

For example, in current API design, developer should know how to use a "headModel" VRFrameOfReference as fallback when "eyeLevel" or "stage" VRFrameOfReference fails to return pose data. e.g.

```js
let pose = vrFrame.getDevicePose(eyeLevelFrameOfRef);
if (pose) {
  // Head position and orientation available, render the seated-experience.
} else {
  // Fail to get pose in eye level frame of reference, try head model frame of reference.
  pose = vrFrame.getDevicePose(headModelFrameOfRef);
  if (pose) {
    // Fallback to render orientation-only experience.
  } else {
    // No tracking data is available
    // Fade out content and render the head-locked warning message.
  }
}
```

### 3. Add tracking loss event
When tracking loss happening, it might be useful to expose the reason to app. So app is able to guide user to recover. For outside-in tracking, the reason might be "out-of-range". For inside-out tracking, the reason might be "vision-inhibited". Or "unavailable" for other hardware errors.

```js
partial interface VRSession : EventTarget {
  attribute EventHandler ontrackingloss;
};

enum VRTrackingLossReason {
  "out-of-range",
  "vision-inhibited",
  "unavailable"
};

[Constructor(DOMString type, VRTrackingLossEventInit eventInitDict)]
interface VRTrackingLossEvent : Event {
  readonly attribute VRTrackingLossReason reason;
};

dictionary VRTrackingLossEventInit : EventInit {
   required VRTrackingLossReason reason;
};
```

Would love to hear community's thoughts on this. Thanks!

thetuvix:
This resolves #243 by addressing the Jan 2019 F2F discussion we had regarding https://github.com/immersive-web/webxr/issues/243#issuecomment-454282109.


cwilso:
Does this PR fully address the tracking loss handling, or is there substantially more to do?

thetuvix:
Apologies for the delay!  Just emerging from a bunch of non-WebXR craziness here on my end...

This PR up to now had covered the tracking loss half of the explainer and I've just pushed a commit to cover tracking recovery as well.  This should bring the spatial tracking explainer up to date with where we landed after the F2F discussion regarding #243, including some of the changes that differed from the original proposal. (most notably that an unbounded reference space continues tracking its previous physical origin after tracking loss unless tracking recovers into a disjoint map fragment)

Pinging @NellWaliczek, @toji, @lincolnfrog, @Manishearth, @klausw, @darktears and @blairmacintyre to take a look at these explainer changes.  If we agree on these explainer-level definitions, I can leave more precise feedback on #580, #589 and #597.

spatial-tracking-explainer.md
```diff
@@ -228,18 +233,20 @@ For viewer poses, it's common in VR experiences to let the world drag along with
...
 
-#### Rays
-An `XRRay` object includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. The `XRRay` also defines a `matrix` which represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
+#### Tracking recovery
+As discussed above, during tracking loss, the viewer's pose will remain at their last-known position relative to each reference space, while continuing to incorporate orientation updates and any slight adjustments due to neck modeling. When the viewer then recovers positional tracking, developers may observe an instantaneous jump of the viewer pose. If the application continues to render during this time, the world will appear to drag along as the viewer moves around.
```
Manishearth: should this also mention input sources? It currently only covers tracking being lost for viewer spaces.

Manishearth: Should also redundantly mention that emulatedPosition is false, ideally

thetuvix: Thanks!  Added coverage of tracking recovery for input source poses.

thetuvix:
Resolved outstanding comments and merge conflicts. This PR should now address the full set of explainer changes resulting from #243.

What will remain after this PR are the core spec language changes that encode this behavior on top of the native origin spec language @Manishearth added in #621. That core spec approach of defining each space in terms of the behavior of its native origin should make it far easier to specify this tracking loss and recovery behavior in a consistent way.

Once this is merged, I can open a follow-on PR to nail down that core spec language.

spatial-tracking-explainer.md
```diff
@@ -364,19 +366,19 @@ xrReferenceSpace.addEventListener('reset', xrReferenceSpaceEvent => {
...
-* For an `XRUnboundedReferenceSpace`, when the user has travelled far enough from the origin of the reference space that floating point error would become problematic
+* For an `eye-level` or `floor-level` reference space, some XR systems have a mechanism for allowing the user to reset which direction is "forward" or re-center the scene's origin at their current location.
+* For a `bounded` reference space, a user steps outside the bounds of a "known" playspace and enters a different "known" playspace
+* For a `bounded` reference space, an inside-out based tracking system is temporarily unable to locate the user (e.g. due to poor lighting conditions in a dark hallway) and then recovers tracking in a new map fragment that cannot be related to the previous map fragment
+* For an `unbounded` reference space, when the user has travelled far enough from the origin of the reference space that floating point error would become problematic
```
Manishearth: Can we use `{{XRReferenceSpaceType/foo}}` for all of these?

thetuvix: This is explainer text rather than spec text, so I don't believe we can use those references here.

thetuvix:
@NellWaliczek, @toji, @klausw: Are you good to merge these tracking loss/recovery explainer changes so I can follow on with spec changes?

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
+* For a `bounded` reference space, the user may walk far enough to transition from the bounds of one user-defined playspace to another. In this case, the origin of the `bounded` reference space will snap to the defined origin of the new playspace bounds.
+* For an `unbounded` reference space, the user may walk through a dark hallway and regain tracking in a new room, with the system not knowing the spatial relationship between the two rooms. In this case, the origin of the `unbounded` reference space will snap to the position in the new room that results in the viewer's pose continuing seamlessly at the same coordinates in that reference space.
+
+In both cases above, the `onreset` event will fire for each affected reference space, indicating that the physical location of its origin has experienced a sudden discontinuity. Note that the `onreset` event fires only when the origin of the reference space itself jumps in the physical world, not when the viewer pose jumps within a stable reference space.
```
NellWaliczek: >Note that the `onreset` event fires only when the origin of the reference space itself jumps in the physical world, not when the viewer pose jumps within a stable reference space.

@Manishearth This looks like the spec text you may have been looking for.  Though, to be honest, it might make more sense for @thetuvix to fix #561 alongside the spec text changes that this PR will cause.

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
+
+When an input source recovers positional tracking, its pose will instantly jump back into position. If the device tracks the input source relative to the viewer (e.g. a motion controller tracked by headset cameras) and the viewer itself still does not have positional tracking, the input source pose will continue to have an `emulatedPosition` value of true in any space other than viewer space. Once the input source's position in the requested space is known again, `emulatedPosition` will become false.
```
NellWaliczek: >the input source pose will continue to have an `emulatedPosition` value of true in any space other than viewer space.

Can we make this language clearer now that there's actually a "viewer" reference space type?

spatial-tracking-explainer.md
```diff
@@ -257,6 +278,16 @@ It is expected that developers will often choose to preview `immersive` experien
 #### Unbounded to Bounded 
 When building an experience that is predominantly based on an `unbounded` reference space, developers may occasionally choose to switch to a `bounded` reference space. For example, a whole-home renovation experience might choose to switch to a bounded reference space for reviewing a furniture selection library.  If necessary to continue displaying content belonging to the previous reference space, developers may call the `XRFrame`'s `getPose()` method to re-parent nearby virtual content to the new reference space.
 
+#### Viewer space
```
NellWaliczek: This whole section is incorrect now :(

spatial-tracking-explainer.md
```diff
@@ -323,7 +354,7 @@ Some XR hardware with inside-out tracking has users establish "known spaces" tha
...
+The `XRReferenceSpace` type has an event, `onreset`, that is fired when there is a discontinuity of the location of the reference space's origin in the physical world.  This discontinuity may be caused for different reasons for each type, but the result is essentially the same, the perception of the user's location will have changed.  In response, pages may wish to reposition virtual elements in the scene or clear any additional transforms, such as teleportation transforms, that may no longer be needed.  The `onreset` event will fire prior to any poses being delivered with the new origin/direction, and all poses queried following the event must be relative to the reset origin/direction. The `transform` value in the `onreset` event data indicates the size of the jump of the reference space origin in the physical world, if known.
```
NellWaliczek: Are you explicitly trying to say that an orientation discontinuity will not cause it?

NellWaliczek: >The `transform` value in the `onreset` event data indicates the size of the jump of the reference space origin in the physical world, if known.

Why are you calling out the "size" specifically?  It's the size and direction, no?

NellWaliczek:
@thetuvix , I think we're aiming to get #640 and #642 merged by the end of day tomorrow.  If so, do you have a rough idea when you'll be able to rebase this PR on top of them for another round of reviews?

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
-  let pose = frame.getPose(xrSpace, xrReferenceSpace);
+  let pose = xrFrame.getPose(xrSpace, xrReferenceSpace);
   if (pose) {
     // Do a thing
   }
 ```
 
-The `emulatedPosition` attribute of `XRPose` indicates that the translation components of retrieved pose matrices may not be accurate. There are a number of reasons this might be the case. For example, a headset with orientation-only tracking capability may include position data to represent neck modeling. Another reason might be the underlying platform's tracking-loss behavior causes orientation data to be updated while it is unable to update position data. In these situations, the `emulatedPosition` attribute will be set to `true`.
+#### Tracking loss
+Developers should check initially that the result from `getPose()` is not null, as the pose of `space` within `relativeTo` may not have been established yet. For example, a viewer may not yet have been tracked within the application's `XRReferenceSpace`, or a motion controller may not yet have been observed after the user turned it on.
```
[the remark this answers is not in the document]
thetuvix: Just fixed this as part of this PR :smile:

spatial-tracking-explainer.md
thetuvix: The concept of passing the `viewer` reference space to `getPose` is now covered by the "Viewer reference space" section, so I just removed this section.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: Yes, I expect we'd decide that a never-observed anchor starts out with a null pose, as any other pose would be completely arbitrary.

It's an interesting idea for `requestReferenceSpace()` to wait until the viewer pose won't be null - however, I wonder if that won't cause other problems in apps that were tested on systems where this continues immediately.  I'll create a new issue to noodle on that separately.

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
 
-#### Rays
-An `XRRay` object includes both an `origin` and `direction`, both given as `DOMPointReadOnly`s. The `origin` represents a 3D coordinate in space with a `w` component that must be 1, and the `direction` represents a normalized 3D directional vector with a `w` component that must be 0. The `XRRay` also defines a `matrix` which represents the transform from a ray originating at `[0, 0, 0]` and extending down the negative Z axis to the ray described by the `XRRay`'s `origin` and `direction`. This is useful for positioning graphical representations of the ray.
+However, once a pose is initially established for a viewer or input source, pose matrices should continue to be provided even during tracking loss. The `emulatedPosition` attribute of `XRPose` indicates that the position component of the retrieved pose matrix does not represent an actively tracked position. There are a number of reasons this might be the case. For example:
```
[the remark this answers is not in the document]
thetuvix: So far, all we have are reference spaces (including `viewer` spaces) and input source spaces.  As we later define anchor spaces and other spaces, we'd need to augment the spec if we decide they behave differently than today's spaces.

For now, I've just changed this to:
> However, once a pose is initially established for a space (e.g. a `viewer` reference space or an input source's `gripSpace`), pose matrices should continue to be provided even during tracking loss.

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
+* A viewer with orientation-only tracking, whose position within an `eye-level` reference space represents neck modeling.
```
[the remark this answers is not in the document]
thetuvix: It seems compatible to me - the concept of [viewer](https://immersive-web.github.io/webxr/#viewer) remains a first-class concept in the spec.

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
+For viewer poses, it's common in VR experiences to let the world drag along with you during positional tracking loss, as this can be preferable to halting the experience. By continuing to render the experience with viewer poses even when `emulatedPosition` is true, this behavior can be obtained easily. In contrast, when building an AR experience where anchoring of rendered objects to the real world is more critical, developers can detect that the viewer pose's `emulatedPosition` has become false and stop rendering the main scene, falling back to an orientation-only warning.
+
+For input source poses, motion controllers may have their own inertial tracking that can continue updating orientation while positional tracking is lost. If using a controller primarily to target distant objects, developers may simply ignore `emulatedPosition` and continue to point using each frame's updated target ray as the user rotates the controller. If using a controller to do fine operations, such as painting at the controller's tip, developers may instead choose to stop painting when `emulatedPosition` becomes false, ensuring that only high-quality paint strokes are drawn.
+
+#### Tracking recovery
```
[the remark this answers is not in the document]
thetuvix: Like much of the explainer, this section both describes the API behavior that a developer "must" or "should" be able to count on from a WebXR UA, along with developer-focused guidance on how they can then best build an app on top of that behavior.

I do see that this section maybe has more "best practice" content than most, but it's woven in with the key paragraph that establishes the common situations where a UA "should" handle tracking loss in a specific way:

> However, once a pose is initially established for a space (e.g. a `viewer` reference space or an input source's `gripSpace`), pose matrices should continue to be provided even during tracking loss. The `emulatedPosition` attribute of `XRPose` indicates that the position component of the retrieved pose matrix does not represent an actively tracked position. There are a number of reasons this might be the case. For example:
> * A viewer with orientation-only tracking, whose position within a `local` reference space represents neck modeling.
> * A viewer that has temporarily lost positional tracking, whose position within a reference space represents the viewer's last-known position in that space, plus inertial dead reckoning and/or neck modeling to continue providing a position.
> * A motion controller with orientation-only tracking, which is positioned at an assumed position, e.g. by the user's hip.
> * A motion controller that has temporarily lost positional tracking but is still held, whose orientation continues to update at the last-known position relative to the viewer.

Do you believe we should more explicitly separate that paragraph from the guidance that follows?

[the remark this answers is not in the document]
thetuvix: Also, I tweaked the first sentence to say "primary reference space" rather than "`XRReferenceSpace`" to hopefully make this example feel less like a firm tie to that exact type.

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
+As discussed above, during tracking loss, the viewer's pose will remain at their last-known position within each reference space, while continuing to incorporate orientation updates and any slight adjustments due to neck modeling. When the viewer then recovers positional tracking, developers may observe an instantaneous jump of the viewer pose. If the application continues to render during this time, the world will appear to drag along as the viewer moves around.
+
+When the viewer then recovers positional tracking, the origin of each `eye-level`, `floor-level`, `bounded` and `unbounded` reference space will continue to track the same physical location as it did before tracking was lost. Therefore, in the first frame where the viewer pose's `emulatedPosition` is false again, the developer will observe an abrupt jump in the viewer pose back to its tracked position.
```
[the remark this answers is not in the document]
thetuvix: Does the phrasing cause misunderstandings in practice?

The `viewer` space precisely maps to the pose of the [viewer](https://immersive-web.github.io/webxr/#viewer).  Whichever one your brain binds "viewer" to, you are correct :smile:

spatial-tracking-explainer.md
thetuvix: No, I was just referring more explicitly to the "location" "in the physical world", to make clearer what is meant by a discontinuity of the origin.  In general, we've been using "location" when we want to more generically refer to the position/orientation of something.  Some readers may assume that "location" just means "position", but even then, 99% of the time, a reference space discontinuity will involve a discontinuity of "position", and so that is still the right mental model.

spatial-tracking-explainer.md
```diff
@@ -197,37 +197,58 @@ function onSessionStarted(session) {
...
+
+For input source poses, motion controllers may have their own inertial tracking that can continue updating orientation while positional tracking is lost. If using a controller primarily to target distant objects, developers may simply ignore `emulatedPosition` and continue to point using each frame's updated target ray as the user rotates the controller. If using a controller to do fine operations, such as painting at the controller's tip, developers may instead choose to stop painting when `emulatedPosition` becomes false, ensuring that only high-quality paint strokes are drawn.
+
+#### Tracking recovery
+As discussed above, during tracking loss, the viewer's pose will remain at their last-known position within each reference space, while continuing to incorporate orientation updates and any slight adjustments due to neck modeling. When the viewer then recovers positional tracking, developers may observe an instantaneous jump of the viewer pose. If the application continues to render during this time, the world will appear to drag along as the viewer moves around.
```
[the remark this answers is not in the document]
thetuvix: I think it was the second sentence in this paragraph actually that was out of place (it skips ahead to what's covered in the second paragraph) and made these first 2-3 paragraphs awkward.

I've removed the second sentence and left the rest as is - does this section flow better now?

thetuvix:
I've rebased this PR on the latest explainer changes, updated it to account for the recent API changes, and addressed or commented on all of @NellWaliczek's feedback above.

I'm taking a look now at the spec language changes needed to encode these explainer changes formally.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: My hesitation with this new phrasing is that pose data is always queried the same way (through `getPose()`) regardless of the type of `XRSpace` being located.  If future types inherited from `XRSpace` behave differently under tracking loss, we could end up in a situation where developers aren't prepared to handle those differences.  In other words, you're making the statement that the only time this should return `null` is during initialization and that once a pose is returned it will always be non-null, but I'm not sure we'd take that stance for anchors in the future.  Am I making sense?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Why `viewer` and not `reference space`?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Thinking about this more... maybe all it needs is subheadings to separate out the different sections of best practices?  WDYT?

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Maybe I'm looking for it to say "all viewer reference spaces' poses will remain"?  It's kind of a nit, though...

thetuvix:
@toji will be producing a first take on the core spec changes here, to save the time it will take me to get oriented from scratch in index.bs :smile:

Our goal is to then get all this merged by the F2F next week.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
blairmacintyre: I'm not sure I can see why we'd ever return non-null;  assuming there is some way to know that the tracking has stopped, continuing to return the last know pose is the only option that makes sense to me.  If it's optional, we'll get a pattern where developers (and libraries) will just mirror that last known pose anyway, and use "null" as a "use last know pose" flag.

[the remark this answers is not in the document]
thetuvix: There are two paths we may ultimately take with `XRSpace` tracking recovery for future spaces:
* We do as @blairmacintyre says and just decide that for all future spaces, we might as well just keep returning a last-known pose once any pose is delivered, and so all future spaces just behave as today's spaces do.
* We decide that it's simply too wrong to claim that we have any notion at all of where an `XRAnchor` is relative to the `viewer` reference space if you've recovered tracking in a disjoint new room, and so we want to be hardcore and return `null` for just that new flavor of `XRSpace`.  However, this decision won't surprise any existing apps - only apps that explicitly start using `XRAnchor` objects and their resulting spaces will ever observe the new behavior.  It seems OK for apps that explicitly opt in to a new scenario through new types to start getting more nuanced behavior from existing types, which we'd document at that point.

While I'm tempted toward the latter behavior to help stop apps from doing silly things when rendering anchored content, I believe the former is more consistent and easy to specify, and apps have the bit they need to not render at anchors when they're not actively tracked. (and as Blair points out, a naive abstraction layer above WebXR is likely to just implement "null" as "don't change the scene object's pose" anyway, and so there's a limit to how valuable being draconian is anyway)  However, whatever decision we do want to make later, I think we can do so compatibly, regardless of what we say now.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: In general, a single space "having tracking" is a squishy shorthand for the well-defined notion of whether _two_ particular spaces can be tracked relative to one another.  I'm using that here in the explainer to get folks aligned with some digestible examples before we get specific in the following paragraphs.

When using that shorthand, though, I tend to find myself thinking of a viewer (e.g. my headset) as "having tracking", rather than thinking of my primary reference space itself (e.g. my floor) as "having tracking".  While a viewer is a kind of reference space, being that generic here in this list of concrete examples felt like it would take away from readers grokking the overall concept.

spatial-tracking-explainer.md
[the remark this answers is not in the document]
thetuvix: Looking at the previous paragraph as well, I think one catch here is that I'm using the terms "viewer pose", "viewer recovers tracking", etc. to refer to one logical thing, where the viewer is and if it's tracked, that's then expressed in WebXR in two ways:
* The viewer pose (with emulatedPosition bool) returned by `getViewerPose(baseSpace)`
* The pose (with emulatedPosition bool) of the `viewer` reference space returned by `getPose(viewerSpace, baseSpace)`

I could be more specific throughout the explainer and say "`viewer` reference space" explicitly, but I'm not sure that excluding `getViewerPose` makes things more correct, since that's equally affected by all this tracking loss/recovery behavior too.

thetuvix:
Thank you so much to @toji for pulling together spec text here!  I think we're finally ready to merge.

Pinging @Manishearth, @klausw and @NellWaliczek to take a final look.

index.bs
```diff
@@ -841,16 +841,35 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
+    <dl class="switch">
+      <dt> If the tracking system provides a [=6DoF=] pose:
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |pose|'s {{XRPose/emulatedPosition}} to <code>false</code>.
+
+      <dt> Else if the tracking system provides a [=3DoF=] pose or indicates that the pose position is emulated:
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the tracking system's best estimate of the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=]. This MAY use a computed offset such as a neck or arm model. If a position estimate is not available, the last known position MUST be used.
```
NellWaliczek: nit: "This MAY use a computed offset such as a neck or arm model." -> "This MAY include a computed offset such as a neck or arm model.

index.bs
```diff
@@ -841,16 +841,35 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
-  1. Set |transform|'s {{XRRigidTransform/position}} to the location of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
-  1. Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+  1. Query the [=/XR device=]'s tracking system for |space|'s pose relative to |baseSpace| at the time represented by |frame|, then perform the following steps:
+    <dl class="switch">
+      <dt> If the tracking system provides a [=6DoF=] pose:
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |transform|'s {{XRRigidTransform/position}} to the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |pose|'s {{XRPose/emulatedPosition}} to <code>false</code>.
+
+      <dt> Else if the tracking system provides a [=3DoF=] pose or indicates that the pose position is emulated:
```
NellWaliczek: Nit: From what I remember with the Windows MR APIs, the function call that gets you a 6DOF pose will fail if it cannot be retrieved, and a different function call can then be used to get the 3DOF pose. Would it make sense instead to say "Else if the tracking system can provide"?  Or am I overthinking this?

index.bs
```diff
@@ -841,16 +841,35 @@ The [=effective origin=] of an {{XRSpace}} can only be observed in the coordinat
...
+        <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the last known orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+        <dd> Set |pose|'s {{XRPose/emulatedPosition}} boolean to <code>true</code>.
+      
+      <dt> Else if |space|'s pose relative to |baseSpace| has never been determined:
+        <dd> Set |pose| to <code>null</code>.
+    </dl>
+
+NOTE: The {{XRPose}}'s {{XRPose/emulatedPosition}} boolean does not indicate whether |baseSpace|'s position is emulated or not, only whether evaluating |space|'s position relative to |baseSpace| relies on emulation. For example, a controller with [=3DoF=] tracking would report poses with an {{XRPose/emulatedPosition}} of <code>true</code> when its {{targetRaySpace}} or {{gripSpace}} are queried against an {{XRReferenceSpace}}, but would likely report an {{XRPose/emulatedPosition}} of <code>false</code> if the pose of the {{targetRaySpace}} was queried in {{gripSpace}}, because the relationship between those two spaces should be known exactly.
```
NellWaliczek: Nit: Again, not sure if this adds clarify, but what about if this:
>but would likely report an {{XRPose/emulatedPosition}} of <code>false</code> if the pose of the {{targetRaySpace}} was queried in {{gripSpace}}, because the relationship between those two spaces should be known exactly.

was instead...
> but would report an {{XRPose/emulatedPosition}} of <code>false</code> if the pose of the {{targetRaySpace}} was queried in {{gripSpace}} should the relationship between those two spaces be known exactly.

index.bs
```diff
@@ -1290,9 +1309,7 @@ An {{XRPose}} describes a position and orientation in space relative to an {{XRS
 
 The <dfn attribute for="XRPose">transform</dfn> attribute describes the position and orientation relative to the base {{XRSpace}}.
 
-<section class="unstable">
 The <dfn attribute for="XRPose">emulatedPosition</dfn> attribute MUST be <code>false</code> if the {{XRPose/transform}}.{{XRRigidTransform/position}} values are based on sensor readings, or <code>true</code> if the position values are software estimations, such as those provided by a neck or arm model.
```
NellWaliczek: nit: "based exclusively on sensor readings"?

index.bs
```diff
@@ -1959,7 +1976,9 @@ A user agent MUST dispatch a <dfn event for="XRSession">selectend</dfn> event on
...
+A user agent MUST dispatch a <dfn event for="XRReferenceSpace">reset</dfn> event on an {{XRReferenceSpace}} when discontinuities of the [=native origin=] or [=effective origin=] occur, i.e. there are significant changes in the origin’s position or orientation relative to the user’s environment. (For example: After user recalibration of their XR device or if the XR device automatically shifts its origin after losing and regaining tracking.) A {{reset}} event MUST also be dispatched when the {{boundsGeometry}} changes for a {{XRBoundedReferenceSpace}}. A {{reset}} event SHOULD NOT be dispatched if the [=viewer=]'s pose experiences discontinuities but the {{XRReferenceSpace}}'s origin physical mapping remains stable, such as when the [=viewer=] momentarily loses and regains tracking within the same tracking area. The event MUST be of type {{XRReferenceSpaceEvent}}, and MUST be dispatched prior to the execution of any [=XR animation frame=]s that make use of the new origin.
```
NellWaliczek: > A {{reset}} event SHOULD NOT be dispatched if the [=viewer=]'s pose experiences discontinuities but the {{XRReferenceSpace}}'s origin physical mapping remains stable, such as when the [=viewer=] momentarily loses and regains tracking within the same tracking area.

Is this a "SHOULD NOT" or a "MUST NOT"?

NellWaliczek: Did we want to call out anything about the normal "unbounded" reference space origin adjustments not triggering this event like we do in the explainer text?

spatial-tracking-explainer.md
```diff
@@ -119,13 +127,14 @@ function onSessionStarted(session) {
...
 
-It is important to note that `XRViewerPose` objects retrieved using `local` reference spaces may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that `local` experiences must be resilient to position changes despite not being dependent on receiving them. Experiences are encouraged to fully support head movement in a typical seated range.
+It is important to note that `XRViewerPose` objects retrieved using a `local` reference space may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that local experiences must be resilient to position changes despite not being dependent on receiving them. Experiences are encouraged to fully support head movement in a typical seated range.
```
NellWaliczek: Should this paragraph be moved above to the parent section?  Then the similar paragraph in the section below can also be removed.

spatial-tracking-explainer.md
```diff
@@ -183,34 +192,65 @@ The `viewer` reference space is also useful when the developer wants to compare
...
+
+For example, it is common in VR experiences to let the world drag along with the user during positional tracking loss, as this can be preferable to halting the experience. By continuing to render the experience when `XRViewerPose.emulatedPosition` is reporting true, this behavior can be obtained easily. In contrast, when building an AR experience where anchoring of rendered objects to the real world is more critical, developers can detect that `XRViewerPose.emulatedPosition` is reporting false and stop rendering the main scene, falling back to an orientation-only warning.
+
+In contrast, many motion controllers have their own inertial tracking that can continue updating orientation while positional tracking is lost. If using a controller primarily to target distant objects, developers may choose to simply ignore `XRPose.emulatedPosition` and continue to point using each frame's updated target ray as the user rotates the controller. If using a controller to do fine operations, such as painting at the controller's tip, developers may instead choose to stop painting when `XRPose.emulatedPosition` becomes false, ensuring that only high-quality paint strokes are drawn.
+
```
NellWaliczek: Maybe add a paragraph about how developers can use the viewer reference space to display an "Establishing tracking" while other reference spaces may still be attempting to retrieve their first poses?  I'm also ok with this happening in a different PR ;)

index.bs
```diff
@@ -1959,7 +1976,9 @@ A user agent MUST dispatch a <dfn event for="XRSession">selectend</dfn> event on
...
+
+NOTE: Jumps in [=viewer=] position can be handled by the application by observing the {{XRPose/emulatedPosition}} boolean. If a jump in [=viewer=] position coincides with {{XRPose/emulatedPosition}} switching from <code>true</code> to <code>false</code>, it indicates that the [=viewer=] has regained tracking and their new position represents a correction from the previously emulated values.
```
NellWaliczek: I almost want to add one or two mores sentence about why developers should care to handle this at all.  I know this is the spec not the explainer, but even with understanding the purpose of this paragraph I had to read it three times to make sure I was getting it right.

Maybe pull from this explainer text?

>Therefore, in the first frame where the viewer pose's `emulatedPosition` is false again, the developer will observe an abrupt jump in the viewer pose back to its tracked position.

>  For experiences without a "teleportation" mechanic, this is generally the desired behavior. 

> However, if an experience does provide a "teleportation" mechanic, where the user can move through the virtual world without moving physically, it may be needlessly jarring to jump the user's position back after tracking recovery

toji:
@thetuvix: We can proceed in one of two ways at this point. Since this is your branch it's going to be easiest for you to address any nits before we merge. If you're in a position to do so and feel like you understand the requests well enough to act on them, then feel free to go ahead and make any needed updates, after which I'll merge it. (Feel free to make the spec changes too! I'm happy to review them.)

If you aren't able to address Nell's feedback soon for whatever reason, though, I'm happy to merge this now and I'll immediately follow it up with a nit-fix PR, since that would be more effective than trying to juggle branches any further.

Which would you prefer?

(Oh, and FYI: There's a warning about the branch not being able to rebase due to conflicts, but if I set the merge mode to "squash and merge" then it's not an issue, which is what we've traditionally been doing anyway to keep the master commit log relatively clean, so don't worry about trying to fix the rebase.)

klausw:
> Thank you so much to @toji for pulling together spec text here! I think we're finally ready to merge.
> 
> Pinging @Manishearth, @klausw and @NellWaliczek to take a final look.

I haven't been following this discussion closely, but I think the current version of the change looks good. Thanks for working on this!

index.bs
Manishearth: This seems to be at odds with the 3DOF pose thing? i.e. imagine the device has a 3DOF input controller `getPose(grip, targetRay)` should not return an emulated pose as explained here, but the spec text indicates that all 3DOF poses are emulated=true

index.bs
thetuvix: This is about whether a pose is available at this very moment, not whether the device is capable of such poses.  For example, an entity that is only visually tracked may update neither its position nor its orientation when unobserved, even if "the tracking system can provide" better poses generally.

To your specific recollection here, we do make your life difficult in the Windows MR APIs by having one reference space that provides a full 6DoF pose or nothing ([`SpatialStationaryFrameOfReference`](https://docs.microsoft.com/en-us/uwp/api/Windows.Perception.Spatial.SpatialStationaryFrameOfReference)), and then a second reference space that reliably provides a 3DoF pose all the time ([`SpatialLocatorAttachedFrameOfReference`](https://docs.microsoft.com/en-us/uwp/api/Windows.Perception.Spatial.SpatialLocatorAttachedFrameOfReference)), leaving it to the app to correlate the two during tracking loss.  This was a design mistake and I'm glad both WebXR and OpenXR take care of this for the app now :smile:

I think I'd just handwave all this under "tracking system provides a 3DoF pose" - it so happens that a UA may sometimes get that provided 3DoF pose from one mechanism or another.

index.bs
thetuvix: Agreed - this did seem too tentative.  I actually can't think of a case where the relationship _wouldn't_ be known statically, so I'll make it even stronger :smile:

thetuvix: I removed "likely", but left the "because" clause as is, since that really should be true all the time. (unless some badass future controller articulates in the middle!)

index.bs
thetuvix: Yea, this is a tricky thing to phrase to both meet developer expectations and not presume too much about how an implementation will work, since tracking systems today generally involve both "sensor readings" _and_ IMU inference, even during nominal 6DoF tracking.

Thanks to @toji, we now have far more precise spec language about exactly when `emulatedPosition` MUST be true or false in the switch at [line 852](https://github.com/immersive-web/webxr/pull/559/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR852).

Perhaps we just let that language control instead, and remove the normative clauses here:
> The `emulatedPosition` attribute is `false` when the `XRPose/transform` represents an actively tracked [=6DoF=] pose based on sensor readings, or `true` if its `XRRigidTransform/position` value includes a computed offset, such as that provided by a neck or arm model.

index.bs
thetuvix: Agreed - seems to be as much of a MUST as the rest of these assertions - a UA shouldn't just arbitrarily differ on whether it fires the `reset` event if it can't reasonably claim that the reference space's origin has actually shifted physically.

thetuvix: I added this text:

> A {{reset}} event also MUST NOT be dispatched as an {{unbounded}} reference space makes small adjustments to its [=native origin=] over time to maintain space stability near the user, if a significant discontinuity has not occurred.

index.bs
thetuvix: I clarified the text for the "6DoF pose" branch and the "3DoF pose" branch to make clear that these are reasoning about whether that exact relation of `space` to `baseSpace` is 6DoF or 3DoF at the moment, and not whether `space` itself is in some 6DoF or 3DoF state on its own:

> * If the tracking system provides an actively tracked or statically known [=6DoF=] pose for `space`'s pose relative to `baseSpace`:

> * Else if the tracking system provides a [=3DoF=] pose or a [=6DoF=] pose whose position is estimated for `space`'s pose relative to `baseSpace`:

In the case of a 3DoF controller that always estimates its position, `getPose(gripSpace, targetRaySpace)` would now more clearly fall into the first bullet, and so `emulatedPosition` would be `false`.

spatial-tracking-explainer.md
thetuvix: That could let us consolidate, but it seems like jumping ahead to dig into this detail before even establishing the basics of how the two spaces behave.  The `local-floor` version of this paragraph also has a more tailored teleport example, which seems good to keep.

spatial-tracking-explainer.md
thetuvix: I think it would be good to continue discussing that in a separate issue.  It does seem unfortunate after all this if apps have to jump through those hoops, especially if the behavior will differ across platforms and UAs, though it may be unavoidable.

index.bs
thetuvix: I fleshed out the note paragraph a bit - how does this look?

> NOTE: Jumps in [=viewer=] position can be handled by the application by observing the {{XRPose/emulatedPosition}} boolean. If a jump in [=viewer=] position coincides with {{XRPose/emulatedPosition}} switching from <code>true</code> to <code>false</code>, it indicates that the [=viewer=] has regained tracking and their new position represents a correction from the previously emulated values. For experiences without a "teleportation" mechanic, where the [=viewer=] can move through the virtual world without moving physically, this is generally the application's desired behavior. However, if an experience does provide a "teleportation" mechanic, it may be needlessly jarring to jump the [=viewer=]'s position back after tracking recovery. Instead, when such an application recovers tracking and receives a {{reset}} event, it can simply resume the experience from the [=viewer=]'s current position in the virtual world by absorbing that sudden jump in position since the previous frame into its teleportation offset.

thetuvix:
Just updated the spec in response to the feedback from @NellWaliczek's and @Manishearth.

Please take a final look at my spec updates here, especially to the core switch logic on [line 852](https://github.com/immersive-web/webxr/pull/559/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR852).

Thanks!

index.bs
Manishearth: Looks good, thanks!

index.bs
NellWaliczek: Wait now i'm confused... I thought we just said that this wasn't going to cause a 'reset' event?
> Instead, when such an application recovers tracking and receives a {{reset}} event, it can simply resume the experience from the [=viewer=]'s current position in the virtual world by absorbing that sudden jump in position into its teleportation offset.

index.bs
```diff
@@ -850,14 +850,14 @@ To <dfn>populate the pose</dfn> of an {{XRSpace}} |space| in an {{XRSpace}} |bas
   1. Let |transform| be |pose|'s {{XRPose/transform}}.
   1. Query the [=/XR device=]'s tracking system for |space|'s pose relative to |baseSpace| at the time represented by |frame|, then perform the following steps:
     <dl class="switch">
-      <dt> If the tracking system provides a [=6DoF=] pose:
+      <dt> If the tracking system provides an actively tracked or statically known [=6DoF=] pose for |space|'s pose relative to |baseSpace|:
```
NellWaliczek: nit: maybe flip this around to make sure that the "actively tracked" part is clear to be 6DOF
>      <dt> If the tracking system provides a statically known or actively tracked [=6DoF=] pose for |space|'s pose relative to |baseSpace|:

index.bs
```diff
@@ -850,14 +850,14 @@ To <dfn>populate the pose</dfn> of an {{XRSpace}} |space| in an {{XRSpace}} |bas
...
         <dd> Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
         <dd> Set |transform|'s {{XRRigidTransform/position}} to the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
         <dd> Set |pose|'s {{XRPose/emulatedPosition}} to <code>false</code>.
 
-      <dt> Else if the tracking system provides a [=3DoF=] pose or indicates that the pose position is emulated:
+      <dt> Else if the tracking system provides a [=3DoF=] pose or a [=6DoF=] pose whose position is estimated for |space|'s pose relative to |baseSpace|:
```
NellWaliczek: nit: What about...
><dt> Else if the tracking system provides a [=3DoF=] pose or a [=6DoF=] pose whose position is estimated and/or not actively tracked for |space|'s pose relative to |baseSpace|:

index.bs
toji: Read through this a couple of times as well. Sounds like this wasn't actually what was intended, because the entire rest of the note is talking about tracking recovery without a reset event. Seems like it would be fine, in context, to simply say:

> Instead, when such an application recovers tracking, it can simply resume the experience from the [=viewer=]'s current position...

thetuvix: Oops 😅  Good catch!  Fixed.

index.bs
thetuvix: One catch with that phrasing is a "statically known" transform (e.g. grip space to target ray space) might be interpreted as "not actively tracked".

I went with this:

> Else if the tracking system provides a [=3DoF=] pose or a [=6DoF=] pose whose position is neither actively tracked nor statically known for |space|'s pose relative to |baseSpace|:

This lost the word "estimated", but maybe being more specific here is better.  We can mop up any remaining concerns here when we resolve #534.

index.bs
thetuvix: I went with the following, to match the flow of the next bullet:

> If the tracking system provides a [=6DoF=] pose whose position is actively tracked or statically known for |space|'s pose relative to |baseSpace|:

index.bs
thetuvix: I'd gone with the less normative text above.

spatial-tracking-explainer.md
thetuvix: Let's continue this discussion in #674.

thetuvix:
@toji: I resolved the last `reset` goof and cleaned up the remaining feedback nits.  I believe we're ready to merge now!

Thanks everyone!

PR #560, opened 2019-03-13
Change XRRigidTransform inverse from a method to an attribute

toji:
Suggesting this change based on a conversation with @bfgeek about API ergonomics. His assertion was that if the XRRigidTransform object was (and always would be) immutable (which it is) then it would make sense to have the inverse be exposed as an attribute (lazily evaluated, of course) rather than a method. Using a method for a case like this is appropriate in scenarios where the desired behavior is to return something based on the current snapshot of a mutable object's state. But since the XRRigidTransform is immutable (and therefore will only ever have one inverse) we could opt for the arguably simpler syntax.

Thoughts?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -847,7 +847,7 @@ The <dfn attribute for="XRView">projectionMatrix</dfn> attribute provides a [=ma
 
 The <dfn attribute for="XRView">transform</dfn> attribute is the {{XRRigidTransform}} of the viewpoint.
 
-NOTE: The {{XRView/transform}} can be used to position camera objects in many rendering libraries. If a more traditional view matrix is needed by the application one can be retrieved by calling `view.transform.inverse().matrix`.
+NOTE: The {{XRView/transform}} can be used to position camera objects in many rendering libraries. If a more traditional view matrix is needed by the application one can be retrieved by calling `view.transform.inverse.matrix`.
 </section>
 
 XRViewport {#xrviewport-interface}
@@ -925,8 +925,7 @@ interface XRRigidTransform {
   readonly attribute DOMPointReadOnly position;
   readonly attribute DOMPointReadOnly orientation;
   readonly attribute Float32Array matrix;
-
-  XRRigidTransform inverse();
+  readonly attribute XRRigidTransform inverse;
 };
 </pre>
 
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -948,11 +947,11 @@ The <dfn attribute for="XRRigidTransform">position</dfn> attribute is a 3-dimens
...
 
-The <dfn attribute for="XRRigidTransform">matrix</dfn> attribute returns the transform described by the {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}} attributes as a [=matrix=].
+The <dfn attribute for="XRRigidTransform">matrix</dfn> attribute returns the transform described by the {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}} attributes as a [=matrix=]. This attribute SHOULD be lazily evaluated.
 
-An {{XRRigidTransform}} with a {{XRRigidTransform/position}} of <code>{ x: 0, y: 0, z: 0 w: 1 }</code> and an {{XRRigidTransform/orientation}} of <code>{ x: 0, y: 0, z: 0, w: 1 }</code> is known as an <dfn>identity transform</dfn>.
+The <dfn attribute for="XRRigidTransform">inverse</dfn> attribute returns a {{XRRigidTransform}} which, if applied to an object that had previously been transformed by the original {{XRRigidTransform}}, would undo the transform and return the object to it's initial pose. This attribute SHOULD be lazily evaluated.
```
NellWaliczek: Should we say that the inverse.inverse is the original object?

klausw: How about saying that `T.inverse.inverse` is equivalent to the original transform, but it may not be exactly equal due to rounding errors? I assume that's what you meant, but "is the original object" could be read as requiring that `T.inverse.inverse === T` in the sense of a true object identity.

I think the implementation should be allowed to handle each inverse as an individual calculation that returns a new object, and that wouldn't generally result in exact equality unless this variant is special-cased.

toji: I could see it go either way. It wouldn't be a huge burden to retain the "parent" transform. There's possibly some minor memory concerns that would come along with that, but it's probably not a big deal. It may improve performance in some rather weird edge cases, but I don't think we should be optimizing for that.

I would be tempted to leave it up to the UA, but there IS a minor compat issue that would come out of it. The biggest functionality fallout of a same-object policy here would be that if someone patches an attribute onto the transform, and then calls `T.inverse.inverse.inverse.inverse.inverse.inverse.inverse.inverse` that patched value would still be expected to be there. I don't know if that's behavior we necessarily want people to be relying on?

toji: After some further thought, I view this question as worthy of it's own discussion but not something that needs to block this particular change. Filed #576 to track the question.

Issue #542, opened 2019-02-25
Consider primary type plus options bag pattern in request methods


ddorwin:
`XR.requestSession()` and `XRSession.requestReferenceSpace()` both take a single options parameter. In many/most cases, only a single primary value from the options needs to be specified. It may be better to instead specify that primary value separately followed by an optional options parameter. This may both simplify the current API and allow for more flexibility for future extension.

Consider:
* The pair of methods in the `XR` interface are intended to be used together but do not share parameter types:
  `Promise<void> supportsSessionMode(XRSessionMode mode);`
  `Promise<XRSession> requestSession(optional XRSessionCreationOptions parameters);`
* `XRReferenceSpaceOptions` only contains a `type` and `subtype`.
  * These could be combined (https://github.com/immersive-web/webxr/issues/514), which would make the options parameter unnecessary while allowing one to be added in the future if necessary.
* IDL methods can accept different types in the same parameter slot (i.e., https://heycam.github.io/webidl/#idl-union). If the primary value is a separate parameter, subsequent parameters can be different types rather than needing to have a union of all possible values used for any primary mode/type.

With that in mind, the two methods might be changed as follows:
* `Promise<XRSession> requestSession(XRSessionMode mode, optional XRSessionCreationOptions options);`
  * Note that one argument is now required. This is a minor part of the change, but it seems reasonable to expect the caller to specify the mode they want just as they do for `supportsSessionMode()`.
* `Promise<XRReferenceSpace> requestReferenceSpace(XRReferenceSpaceType type);`
  * This assumes the suggestion in #514 is made. If not, some sort of options parameter would be added.

blairmacintyre:
How do you feel about `requestSession` succeeding but returning a session with a different mode?

I'm thinking about the case where a device supports both `ar` and `vr`, so `supportsSessionMode` returns true for both.  Then the UI presents the user with a choice, and they say they only way `ar` when the app asks for `vr` ... 

If `requestSession` fails, the app may simply give up and just assume the user refused (or couldn't) to XR.  

If `requestSession` succeeds, the app will need to look and notice that it's in a different mode.

NellWaliczek:
A device should support VR if it supports AR, so that case shouldn't happen.  Perhaps that's underspecified, though.

ddorwin:
I think that would be unexpected unless the definition and shape of the method are changed. That seems like a separate discussion, though.

cwilso:
[...]

PR #566, opened 2019-03-21
Remove XRSessionCreationOptions

toji:
As described in #542, this switches `requestSession` to a "primary type
plus options bag" pattern. However, since the dictionary only had a single
key after outputContext was moved this change ends up removing the
`XRSessionCreationOptions` dictionary entirely in favor of just passing in
the type enum directly.

In the future if additional options *are* needed for `requestSession`,
the expected pattern would be:

```
navigator.xr.requestSession('immersive-vr', { option: value });
```

### DIFF
--- a/spatial-tracking-explainer.md
+++ b/spatial-tracking-explainer.md
@@ -318,18 +318,6 @@ function onSessionStarted(session) {
 }
 ```
 
-While many sites will be able to provide this fallback, for some sites this will not be possible.  Under these circumstances, it is instead preferable for session creation to reject rather than spin up immersive display/tracking systems only to immediately exit the session.
-
-```js
-function beginImmersiveSession() {
-  xrDevice.requestSession({ immersive: true,  requiredReferenceSpaceType:'unbounded' })
-      .then(onSessionStarted)
-      .catch(err => {
-        // Error will indicate required reference space type unavailable
-      });
-}
-```
-
 ### Floor Alignment
 Some XR hardware with inside-out tracking has users establish "known spaces" that can be used to easily provide `XRBoundedReferenceSpace` and the `floor-level` subtype of `XRStationaryReferenceSpace`.  On inside-out XR hardware which does not intrinsically provide these known spaces, the User Agent must still provide `XRStationaryReferenceSpace` of subtype `floor-level`.  It may do so by estimating a floor level, but may not present any UI at the time the reference space is requested.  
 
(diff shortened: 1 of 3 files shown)


blairmacintyre:
I would like to suggest that we consider moving forward with https://github.com/immersive-web/geo-alignment and use this as an example of possible options.

spatial-tracking-explainer.md
```diff
@@ -318,18 +318,6 @@ function onSessionStarted(session) {
 }
 ```
 
-While many sites will be able to provide this fallback, for some sites this will not be possible.  Under these circumstances, it is instead preferable for session creation to reject rather than spin up immersive display/tracking systems only to immediately exit the session.
-
-```js
-function beginImmersiveSession() {
-  xrDevice.requestSession({ immersive: true,  requiredReferenceSpaceType:'unbounded' })
```
NellWaliczek: So this sample code made it into the explainer, but never made its way into the spec because there was confusion about the overall permissions/features model.  While we are working through the final shape of that design (oh, John..... *poke*poke*poke*) can we at least file an issue to cover adding this functionality back in once that's complete?

NellWaliczek: Alternatively, we can leave the options dictionary and put this in it ;)

blairmacintyre:
I like the idea of having the 1st parameter (session type) be explicit, but I'd still like to propose that we keep at least one "option" (I suggested `geoaligned` for requesting that the base coordinate system be geoaligned) and keep the optional `XRSessionCreationOptions` around.  It seems to me that we will want this eventually, so why not keep it?

NellWaliczek:
@toji, please correct me if I'm misinterpreting... but you're not saying you think we don't need it.  Just simply that we'll leave it off until a specific feature is actually merged into the spec that depends on it.

blairmacintyre:
Part of my motivation is that I suspect (we’ll, know) it will end up getting used, as people experiment with “next version” or “proposed” features.  

I expect I’ll put a bunch of things behind prefs (rather than just expose them) in the WebXR Viewer after moving it over to the “real” spec, for example.  So I’ll being using this immediately.  This, I’d prefer to have at least the name and format of the object defined if we’ll be adding it.

NellWaliczek:
@blairmacintyre The format of the object will almost certainly be a dictionary since that's the general pattern on the web and as such seems like a safe assumption.  (And as far as I'm aware, the idl type name shouldn't matter).  That said, I do understand your concern.  And my suspicion is that we're going to have something that proves the pattern before we hit CR, but I'm hesitant to put it in now so that we don't create a breaking change if we're wrong.  Hence, my suggestion to file an issue against CR to enable the extensibility without holding up this PR.  WDYT?

blairmacintyre:
Sounds reasonable to me, I'm not sure what to put in an issue.  

I've been pushing for geo-orientation, for example, for a while;  we have had it sitting in a repo (github.com/immersive-web/geo-orientation) for 7 months with no movement.   The current proposal would need a little bit of work, may as well wait till after CR so we have a defined API to work against.

toji:
Filed #591. Nell confirmed offline that she's comfortable merging this once that issue was filed, so doing so now.

PR #568, opened 2019-03-21
Validate position DOMPointInit

Manishearth:
Given that it's a float, comparison with 1 may not be the best idea, but I feel like erroring here is marginally better than ignoring invalid `w` values.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -937,6 +937,7 @@ The <dfn constructor for="XRRigidTransform">XRRigidTransform(|position|, |orient
 
   1. Let |transform| be a new {{XRRigidTransform}}.
   1. If |position| is not a {{DOMPointInit}} initialize |transform|'s {{XRRigidTransform/position}} to <code>{ x: 0.0, y: 0.0, z: 0.0, w: 1.0 }</code>.
+  1. If |position|'s {{DOMPointReadOnly/w}} value is not 1.0, throw a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


klausw:
I think people have a somewhat irrational (no pun intended) distrust of floating point arithmetic. If you start out with w=1, and apply matrix transforms involving add/multiply of values that are also exactly 0 or 1 (as is the case for the bottom row of a rigid transform matrix), the result will still exactly equal 1.0. Floating point arithmetic is exact for integers as long as you don't exceed the mantissa range (up to 2^52 or so for doubles), and the typical things you'd be doing with w values wouldn't lead to an approximately-but-not-exactly-1.0 value.

Manishearth:
Yeah, this was the assumption I was making too, floating point should be fine for ints here. And the fix for this error is straightforward, you can always set the last value to 1 if you need.

toji:
Hm... so this would explicitly prevent me from doing something along the lines of the following, right?

```js
let transform = new XRRigidTransform({x: 1, y: 2, z: 3}, someOrientation);
```

Because that actually seems pretty convenient.

Manishearth:
It would allow it as long as someOrientation has a unit value 1 -- if it doesn't you have a bug anyway since it's going to get overwritten

Manishearth:
Oh, sorry, position. That is a problem, hmm. Maybe provide our own init dicts?

Manishearth:
It seems like w=1 by default, so that looks like it would still work?

toji:
Oh really? Could you link to the relevant spec text? (I'm gonna try and hunt it down too.)

toji:
I guess this is the right link to cite: https://www.w3.org/TR/geometry-1/#DOMPoint

And in light of that, I'm good with this being a strict validation. Thanks!

Issue #567, opened 2019-03-21
Clarify what originOffset is applied to


Manishearth:
The spec says

>  The originOffset attribute is a XRRigidTransform that describes an additional translation and rotation to be applied to any poses queried using the XRReferenceSpace.

However, it's not exactly clear how exactly the rotation is applied: is it applied after the pose (like a periscope)? Or before the pose (as if the user is standing on a platform or something)? Or is it an offset being applied to the space?

I would have assumed it's a premultiplication with the pose, however [the chromium impl](https://cs.chromium.org/chromium/src/third_party/blink/renderer/modules/xr/xr_space.cc?q=xrspace&sq=package:chromium&g=0&l=92-93) seems to indicate that its _inverse_ is premultiplied, which makes me think it's logically a transformation applied to the entire space.

Could this be clarified? (May be worth doing as part of #565)

klausw:
I contributed some pictures and math here while working on the Chrome implementation: https://github.com/immersive-web/webxr/issues/477#issuecomment-471803936

The test page https://cs.chromium.org/chromium/src/third_party/webxr_test_pages/webxr-samples/input-selection.html exercises teleportation via originOffset manipulation, this was working as expected for me in Chrome both in 3DoF and 6DoF headsets:

My understanding is that the originOffset matrix transforms coordinates from world space (the virtual world's coordinate system) to tracking space (the roomscale area in the physical world), first rotating then applying translation. In the usual GL convention, these successive transforms get accumulated by multiplying them from the left:

```
originoffset_translation = (
     1 0 0 xform.position.x
     0 1 0 xform.position.y
     0 0 1 xform.position.z
     0 0 0 1
)

originoffset_rotation = (
    . . . 0
    . . . 0
    . . . 0
    0 0 0 1
)

point_worldspace = (px py pz 1)

point_trackingspace = originoffset_translation * originoffset_rotation * point_worldspace
```
 
The world space origin `point_worldspace = (0 0 0 1)` gets transformed to tracking space coordinates `(xform.position.x, xform.position.y, xform.position.z, 1)`, using the exact x/y/z values from the originOffset XRRigidTransform, and the resulting tracking space location stays the same when a rotation is added.

If you change the originOffset rotation component while keeping the originOffset position component unchanged, that corresponds to sticking a pin through the virtual world's origin that nails it to the originOffset position in tracking space (i.e. your play space), and then spinning the virtual world around that pin.

[...]

Issue #565, opened 2019-03-18
getViewerPose()s use of XRSpace doesn't quite specify how views work


Manishearth:
Fundamentally, the XRSpace is a position and an orientation, one that may change every frame.

`getViewerPose()` gets pose and view information for a viewer given an XR space.

However, the XR space has no inherent concept of a "view". When in immersive mode, aside from eye-level spaces, it doesn't quite make sense to request view data for the space since there's nothing telling you where the eyes _are_.

For example, what should `getViewerPose(identity)` return in immersive mode? Should it return just a single view? Should it return two views from the identity matrix, with the eyes offset based on device data? Which direction are the eyes offset in? Does this direction change as the viewer moves their head?

Manishearth:
Similarly, when there are multiple views, what exactly is originOffset affecting? Can I apply it by simply premultiplying the offset to the view matrices, or is there some deconstruction of pose data that I must first do?

It may be worth explicitly modeling this as a mathematical thing based off of pose information obtained from the device.

Manishearth:
Okay, so from reading the chromium source (which is in part written by @toji so I assume it follows the intent of the spec :smile: ) it seems like:

 - Fundamentally, a view is a set of offsets from the space. This may or may not involve orientation changes, but does not currently do so in the chromium impl, presumably because XR devices don't yet have/report offset orientation for eyes, which isn't too important anyway since the devices are for humans, not deer.
 - getViewerPose returns views obtained by applying the offset directly to the xrspace's pose (with originOffset applied). Essentially, it is "what if the current XR device were mounted at this XRSpace's pose".
 - eye-level is basically the default pose that most devices report. floor-level is the same pose with a "floor level transform" applied. "position-disabled" is the same pose with the position information zeroed out. `getViewerPose()` on these returns a set of view matrices that incorporate the pose information suitable for displaying.
 - `getViewerPose(identity)` basically will return a fixed view matrix, "what if the device were mounted at the origin"
 
I think the missing piece was understanding that `getViewerPose()` is "mounting" the XR device on whatever space you give to it, so different spaces don't have different poses.


I think we should explicitly mention this in the spec, by noting:
 - the session keeps track of the offset of the different views
 - each XRSpace has an inherent rigid transform (separate from originOffset), individual space types define how to obtain this for a given frame. It may be worth making input XR spaces their own type, as well as perhaps the identity reference XR space.
[...]

Issue #477, opened 2019-01-15
Does originOffset behave differently for identity reference space?


thetuvix:
> `XRViewerPose` objects retrieved with this reference space will have a transform that is equal to the reference space's `originOffset`.

Isn't this the reverse of what we defined `originOffset` to do for all other reference spaces?

**For the `"bounded"` space:**
Imagine that the viewer is located at position (2, 2, 2) within the default `"bounded"` reference space.  If I then "offset" the bounded reference space's "origin" by (2, 2, 2), I would expect the viewer to now be at position (0, 0, 0), since I shifted that space's origin over by (2, 2, 2) to now coincide with the viewer.

This means that **applying an `originOffset` of (2, 2, 2)** to the `"bounded"` reference space causes the viewer's pose to **translate by (-2, -2, -2)**.

**For the `"identity"` space:**
However, per this PR, if I apply an `originOffset` of (2, 2, 2) to the `"identity"` reference space, the viewer's position changes from (0, 0, 0) to (2, 2, 2).

This means that **applying an `originOffset` of (2, 2, 2)** to the `"identity"` reference space causes the viewer's pose to **translate by (2, 2, 2)**, the opposite behavior as with the `"bounded"` reference space.

_Originally posted by @thetuvix in https://github.com/immersive-web/webxr/pull/459/review_comment/create_

toji:
Sorry for the delay in commenting on this. This issue highlights the fact that there's actually two valid ways to think about this, and we need to:

  1. Make sure everyone agrees on which interpretation is desired.
  2. State which way the spec intends much more clearly.

And in order sort that out I'm going to subject you all to my scribbles. 😁 

![project - drawing 160404524081023811](https://user-images.githubusercontent.com/805273/51562472-bed5d080-1e3e-11e9-8d19-93816a397c99.png)

This is, I hope we can all agree, the default state of things. No `originOffset` has been applied, so the virtual scene's origin is aligned with the tracking system's origin, which in this case considers the center of the room at the floor to be `(0,0,0)`. Very typical `bounded` scenario.

My assumption when writing the `originOffset` text, and subsequently the `identity` reference space text, was that if you set it's `position` to, say, `(1, 0, -3)` you'd get an effect like this, which we'll call **Interpretation A**:

![project - drawing 13051976579337510665](https://user-images.githubusercontent.com/805273/51565661-6c4ce200-1e47-11e9-859e-3bad28df1ca9.png)

[...]

PR #569, opened 2019-03-23
Update spec and explainer to clarify pose and transform semantics

klausw:
This is intended to address issues #477, #565, and #567 which all involve areas where the spec isn't entirely clear about intended behavior.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -681,7 +681,7 @@ Spaces {#spaces}
 XRSpace {#xrspace-interface}
 ------------------
 
-An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s tracking systems. It conceptually corresponds to a specific origin and orientation, but does not directly expose any numerical coordinate values. {{XRSpace}}s MAY NOT have a fixed spatial relationship to one another or to any given {{XRReferenceSpace}}. The current transform between two {{XRSpace}}s can be evaluated by calling an {{XRFrame}}'s {{XRFrame/getPose()}} method. The interface is intentionally opaque.
+An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s tracking systems. It conceptually corresponds to a specific origin and orientation, but does not directly expose any numerical coordinate values. The spacial relationships between {{XRSpace}}s or {{XRReferenceSpace}} change over time in response to movement of tracked entities, but can also change without real-world movement as a result of ongoing tracking system recalibration. The transform between two {{XRSpace}}s at the time represented by an {{XRFrame}} can be evaluated by calling the frame's {{XRFrame/getPose()}} method. The interface is intentionally opaque.
 
 <pre class="idl">
 [SecureContext, Exposed=Window] interface XRSpace : EventTarget {
--- a/index.bs
+++ b/index.bs
@@ -664,7 +664,7 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method 
 
 <div class="algorithm unstable" data-algorithm="get-pose">
 
-When the <dfn method for="XRFrame">getPose(|destinationSpace|, |sourceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
+When the <dfn method for="XRFrame">getPose(|sourceSpace|, |destinationSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw a {{InvalidStateError}} and abort these steps.
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
--- a/index.bs
+++ b/index.bs
@@ -641,7 +641,7 @@ An {{XRFrame}} represents a snapshot of the state of all of the tracked objects 
   readonly attribute XRSession session;
 
   XRViewerPose? getViewerPose(XRReferenceSpace referenceSpace);
-  XRPose? getPose(XRSpace space, XRSpace relativeTo);
+  XRPose? getPose(XRSpace sourceSpace, XRSpace destinationSpace);
 };
 </pre>
 
@@ -658,20 +658,20 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method 
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
   1. If |referenceSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If the [=viewer=]'s pose cannot be determined relative to |referenceSpace|, return <code>null</code>
-  1. Return a new {{XRViewerPose}} describing the [=viewer=]'s pose relative to the origin of |referenceSpace| at the timestamp of the {{XRFrame}}.
+  1. Return a new {{XRViewerPose}} with an {{XRRigidTransform}} from the [=viewer=] space to |referenceSpace| at the timestamp of the {{XRFrame}}. The transform's {{XRRigidTransform/position}} is the location of the [=viewer=] in reference space coordinates.
 
 </div>
 
 <div class="algorithm unstable" data-algorithm="get-pose">
 
-When the <dfn method for="XRFrame">getPose(|space|, |relativeTo|)</dfn> method is invoked, the user agent MUST run the following steps:
+When the <dfn method for="XRFrame">getPose(|destinationSpace|, |sourceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. If the {{XRFrame}}'s [=active=] boolean is <code>false</code>, throw a {{InvalidStateError}} and abort these steps.
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
-  1. If |space|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |relativeTo|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |space|'s pose cannot be determined relative to |relativeTo|, return <code>null</code>
-  1. Return a new {{XRPose}} describing |space|'s pose relative to the origin of |relativeTo|.
+  1. If |destinationSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
+  1. If |sourceSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
+  1. If |destinationSpace|'s pose cannot be determined relative to |sourceSpace|, return <code>null</code>
+  1. Return a new {{XRPose}} containing the {{XRRigidTransform}} from |sourceSpace| to |destinationSpace|. The transform's {{XRRigidTransform/position}} is the location of the |sourceSpace| origin in |destinationSpace| coordinates.
 
 </div>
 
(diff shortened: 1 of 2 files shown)


klausw:
FWIW, I think originOffset makes it difficult to explain reference spaces things clearly. An XRReferenceSpace basically consists of two XRSpaces that are transformed into each other by the originOffset, and it's hard to explain how that transform works when the two spaces don't have names. I've used the name "effective reference space" for the offset-modified externally visible space and "unmodified reference space" for the internal space corresponding to an identity originOffset, but that's clunky.

index.bs
```diff
@@ -917,7 +917,7 @@ To <dfn>normalize</dfn> a list of components the UA MUST perform the following s
 XRRigidTransform {#xrrigidtransform-interface}
 ----------------
 
-An {{XRRigidTransform}} is a transform described by a {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}}. When interpreting an {{XRRigidTransform}} the {{XRRigidTransform/orientation}} is always applied prior to the {{XRRigidTransform/position}}.
+An {{XRRigidTransform}} is a transform from a source {{XRSpace}} to a destination {{XRSpace}} described by a {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}}. When applying an {{XRRigidTransform}} the {{XRRigidTransform/orientation}} is always applied prior to the {{XRRigidTransform/position}}. Effectively, the source space is rotated around its origin using the {{XRRigidTransform/orientation}}, and then the sourc e space's origin is translated to the {{XRRigidTransform/position}} in destination space.
```
Manishearth: nit: spelling

klausw: Oops, fixed. Thanks!

klausw:
Rebased to pick up the `inverse()`=>`inverse` attribute change from 9ec29f5760f048053dbae3b4f598705573a21a3e, and combined into a single commit.

index.bs
```diff
@@ -681,7 +681,7 @@ Spaces {#spaces}
 XRSpace {#xrspace-interface}
 ------------------
 
-An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s tracking systems. {{XRSpace}}s MAY NOT have a fixed spatial relationship to one another or to any given {{XRReferenceSpace}}. The transform between two {{XRSpace}} can be evaluated by calling an {{XRFrame}}'s {{XRFrame/getPose()}} method. The interface is intentionally opaque.
+An {{XRSpace}} describes an entity that is tracked by the [=/XR device=]'s tracking systems. It conceptually corresponds to a specific origin and orientation, but does not directly expose any numerical coordinate values. {{XRSpace}}s MAY NOT have a fixed spatial relationship to one another or to any given {{XRReferenceSpace}}. The current transform between two {{XRSpace}}s can be evaluated by calling an {{XRFrame}}'s {{XRFrame/getPose()}} method. The interface is intentionally opaque.
```
toji: "current" is a hard concept to pin down in WebXR. This whole sentence could probably stand to be rephrased to clarify that. Something along the lines of:

"The transform between two {{XRSpace}}s at the time represented by an {{XRFrame}} can be evaluated by calling the frame's {{XRFrame/getPose()}} method."

Manishearth:
> FWIW, I think originOffset makes it difficult to explain reference spaces things clearly

Filed https://github.com/immersive-web/webxr/issues/580

klausw:
If I'm understanding it right, the consensus from the meeting was that we'll be revisiting originOffset as per @Manishearth's #580, but that shouldn't block us from proceeding with the bulk of the changes in this proposal.  The originOffset part of it is localized to just one "unstable" paragraph

Reviewers, please let me know if you want me to make further changes, or if it's OK as-is.

Manishearth:
Yeah, I would still like this pull request to land; it contains changes that are independent of originOffset, and the changes that depend on originOffset are still an unambiguous improvement, which at worst will be overwritten by #580.

index.bs
klausw: Changed in 30ca9d2 . I've also rephrased the previous sentence to make it explicit that the spatial relationship can change even if there's no real-world movement. New text:

"The spacial relationships between {{XRSpace}}s or {{XRReferenceSpace}} change over time in response to movement of tracked entities, but can also change without real-world movement as a result of ongoing tracking system recalibration. The transform between two {{XRSpace}}s at the time represented by an {{XRFrame}} can be evaluated by calling the frame's's {{XRFrame/getPose()}} method."

klausw: Another reason for rephrasing that sentence was that I wanted to avoid the MAY NOT construction - that's not one of the [rfc 2119](https://www.ietf.org/rfc/rfc2119.txt) key terms, and it could be misinterpreted as disallowing fixed spacial relationships as opposed to warning not to assume them.

index.bs
```diff
@@ -658,20 +658,20 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
   1. Let |session| be the {{XRFrame}}'s {{XRFrame/session}} object.
-  1. If |space|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |relativeTo|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
-  1. If |space|'s pose cannot be determined relative to |relativeTo|, return <code>null</code>
-  1. Return a new {{XRPose}} describing |space|'s pose relative to the origin of |relativeTo|.
+  1. If |destinationSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
+  1. If |sourceSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
+  1. If |destinationSpace|'s pose cannot be determined relative to |sourceSpace|, return <code>null</code>
+  1. Return a new {{XRPose}} containing the {{XRRigidTransform}} from |sourceSpace| to |destinationSpace|. The transform's {{XRRigidTransform/position}} is the location of the |sourceSpace| origin in |destinationSpace| coordinates.
```
klausw: Does anyone have good naming suggestions here? "sourceSpace" and "destinationSpace" seemed like the best fit for describing the transform, but it's a bit clunky when talking about getPose.

Do any of these variants look better? I've added snippets how the naming would affect this paragraph and the XRRigidTransform definition below.

`getPose(XRSpace sourceSpace, XRSpace destinationSpace);` =>
> Return a new {{XRPose}} containing the {{XRRigidTransform}} from |sourceSpace| to |destinationSpace|. The transform's {{XRRigidTransform/position}} is the location of the |sourceSpace| origin in |destinationSpace| coordinates.
>
> An {{XRRigidTransform}} is a transform from a source {{XRSpace}} to a destination {{XRSpace}} described by ...

`getPose(XRSpace space, XRSpace relativeTo);` =>
> Return a new {{XRPose}} containing the {{XRRigidTransform}} from |space| to |relativeTo|. The transform's {{XRRigidTransform/position}} is the location of the |space| origin in |relativeTo| coordinates.
>
> An {{XRRigidTransform}} is a transform from an {{XRSpace}} to a relativeTo {{XRSpace}} described by ...

`getPose(XRSpace targetSpace, XRSpace baseSpace);` =>
> Return a new {{XRPose}} containing the {{XRRigidTransform}} from |targetSpace| to |baseSpace|. The transform's {{XRRigidTransform/position}} is the location of the |targetSpace| origin in |baseSpace| coordinates.
>
> An {{XRRigidTransform}} is a transform from a target {{XRSpace}} to a base {{XRSpace}} described by ...

`getPose(XRSpace transformedSpace, XRSpace baseSpace);` =>
> Return a new {{XRPose}} containing the {{XRRigidTransform}} from |transformedSpace| to |baseSpace|. The transform's {{XRRigidTransform/position}} is the location of the |transformedSpace| origin in |baseSpace| coordinates.
>
> An {{XRRigidTransform}} is a transform from a transformed {{XRSpace}} to a base {{XRSpace}} described by ... _(This one sounds backwards)_

klausw: @NellWaliczek please see [^^^](https://github.com/immersive-web/webxr/pull/569/files#r277060085), trying to get the naming sorted out in relation to your request from https://github.com/immersive-web/webxr/pull/589#discussion_r277049962

Manishearth: I hope to somewhat redundantly clarify `getPose` with some matrix math so that the confusion doesn't get too bad. I kinda like `source` `destination`.

thetuvix: Of the naming options above, I prefer `targetSpace`/`baseSpace`.  Specifically, I worry that `sourceSpace`/`destinationSpace` would present a reversed mental model to the way in which `getPose` will be increasingly used in practice.

As we start to establish plane anchors, mesh anchors, face anchors, etc., there will be an increasing number of static and dynamic entities in WebXR that are tracked over time by `XRSpace` instances.  To place these entities within their scene each frame, developers will use `xrFrame.getPose(interestingEntitySpace, myReferenceSpace)` and expect to get back the position and orientation of that entity within their engine's world coordinates (which are underwritten by `myReferenceSpace`), so they can update that entity's scene object transform in their engine.  The mental model for `getPose` in particular will start to become "what's the position and orientation of this single entity in my reference space", moreso than "get me a model transform I can use to generally transform coordinates from this entity space into my reference space".

However, the above is ultimately just me arguing that we should preference naming that's intuitive for the `position`/`orientation` attributes on `XRRigidTransform` rather than the `matrix` attribute.  In the end, @klausw had a key observation in https://github.com/immersive-web/webxr/issues/580#issuecomment-479595386 that applies here:

> I'm starting to think that a core part of the confusion is that a tracked object's XRPose has-a XRRigidTransform, and the transform's position component corresponds to the tracked object's coordinates in reference space as expected, but a XRRigidTransform and its matrix are a transform from object space to reference space. That's internally consistent, but it's confusing when you supply a transform by itself as an argument (as in the constructor arg here) because it's easily interpreted as saying "please apply this transform" which would be backwards. In a way, OpenXR's approach of directly putting position + orientation on the pose objects helps avoid this ambiguity.

Ultimately, **any naming pattern and/or conceptual framework we come up with that makes `position`/`orientation` make more sense will make `matrix` make less sense, and vice versa**.

In OpenXR, we've managed to sidestep entire classes of confusion around the order or direction in which matrices are applied, whether they're pre-multiplied or post-multiplied, whether they are row-major/column-major, etc. by **banishing matrices from the API altogether**.  Rather than expressing transform matrices "from" or "to" various spaces, all poses are expressed only as a `position` vector and `orientation` quaternion "in" a given space.

Even for view matrices, as discussed in #447, we found in OpenXR that any non-trivial engine ends up positioning a camera object within its scene in world coordinates anyway, and so even there it ends up being more convenient for developers to reason about the position and orientation of their view rather than getting a view matrix directly.  You can see this pattern in OpenXR in [xrLocateViews](https://www.khronos.org/registry/OpenXR/specs/0.90/man/html/#_xrlocateviews_3), which returns a set of [XrView](https://www.khronos.org/registry/OpenXR/specs/0.90/man/html/#_xrview_3) structs, each containing the position and orientation of a given view.

Specifically in the case of OpenXR reference spaces with a non-identity pose applied:
* Each reference space type is specified in terms of how its "natural" origin behaves.
* If you apply a non-identity pose when creating a reference space, that's specified as placing the origin of your reference space at that position and orientation "in" the "natural" reference space.

By defining the origin of any space as either a well-specified location in the physical world or a position and orientation within some other such space, all space origins have a clear binding to a place in the physical world.  It then falls out without ambiguity how you "get" the "pose" of any one space "in" another.

If we ourselves continue to struggle with these naming and direction issues each week, despite being embedded in the details of the spec, perhaps it's the lesser evil to just remove `matrix` from `XRRigidTransform`?  This will forcibly snap the developer's mental model to an unambiguous position and orientation approach throughout the API, and then we can design our naming to be consistent with that model.

Manishearth: I think the problem here isn't `matrix`, the problem is the use of the word `transform` itself.

I point this out here: https://github.com/immersive-web/editor-collab/pull/36#discussion_r278361579 , there are two ways one can look at a transform. _Just_ talking about position/orientation as a "pose" is clear, however when you talk of the "transform" of a position/orientation that can mean opposite things.

If we can carefully clarify our use of the word "transform", talking about matrices isn't hard.

I feel like we should definitely try and keep `matrix` around for developer convenience.

klausw:
Retiring this PR, please see the replacement #609 "Overhaul XRSpace, get(Viewer)Pose definitions". We can check separately to see if there's pieces in here we'd like to keep, i.e. for the explainer, but that would be a separate new PR if applicable.

toji:
The primary thing that my PR doesn't capture, upon re-reading this, is the explainer changes. While it would be good to ensure that they're consistent, I'm inclined to view updating the explainer on topics like this as less critical, simply because the spec is authoritative and should be rigorous while the explainers are intended to be higher level.

Issue #489, opened 2019-01-17
Enable developers to distinguish between screen-based and head-worn sessions


NellWaliczek:
Developers will often wish to provide different interaction models for screen-based (aka hand-held) sessions and head-worn sessions.  A screen-based session can comfortably accommodate UI on top of the 3D rendering's near plane, in other words, drawn in screen-space.  Head-worn sessions cannot use this approach and must place interactive elements in world-space.  This may end up having the same solution as #400, but this is issue is focused on ensuring developers know when do provide which type of UI.

klausw:
Careful, I think it's not safe to treat "near plane" and "screen space" as equivalent.

Mathematically, the projection matrix is derived based on intersecting rays with the near plane, but this doesn't mean that this corresponds to the distance to an actual screen. For example, a zSpace-style 3D monitor is able to display objects that appear in front of the monitor surface, and this requires the near plane to be closer to the eye than the physical screen. If an app tries to draw UI at the near plane distance, this may look odd on such a system. As far as I know, the UI quality would be highest if it's drawn at the actual screen distance with no extra parallax.

Also, even if an app primarily targets a hand-held use case, there seems to be a risk of unfortunate breakage on HMDs if the API makes bad assumptions due to this - drawing UI around the near plane on an HMD is likely to be uncomfortably close.

How about a per-frame property where the UA provides a suggested UI rectangle based on app-provided hints?

For example, the app could request a "largest readable area" quad which would be fullscreen on a hand-held session, a generously-sized floating quad for an HMD (while not extending into blurry peripheral vision), or the actual screen position for a zSpace-style 3D monitor or diorama mode.

Separately, it could express how important it is to keep this visible, with the UA choosing appropriate placement. The same hints could be used for a separate DOM layer as per issue #400, but I think it would also be useful without layer support.

kearwood:
> How about a per-frame property where the UA provides a suggested UI rectangle based on app-provided hints?

I like this option -- In addition to providing a safer abstraction of the hardware differences, it could further enable accessibility use cases by letting the UA take into consider various constraints set by the user.

NellWaliczek:
Potentially related to the work now happening in https://github.com/immersive-web/layers

PR #573, opened 2019-03-26
Allow flexible depth for screen-based ray input

klausw:
For screen-based input, it seems overly specific to require that the ray origin must be at the near plane. If the system knows the actual position of the screen in 3D space, i.e. for a zSpace-style display or for smartphone AR, it would be preferable to use the actual mapping of the touch point into virtual world space for this. For example, if you use touch controls to add bubbles or other particle effects to the scene, it would look most natural to have those actually originate at the touched point.

As an additional complication, https://github.com/immersive-web/webxr/pull/548 changed the core spec to explicitly allow reversed-Z where `depthFar` is less than `depthNear`. If the actual screen plane isn't available, it would be more consistent to use the closer of the two instead of always using `depthNear`.

As far as I can tell the spec itself doesn't currently require any particular depth for `XRTargetRayMode` `screen`.

See also https://github.com/immersive-web/webxr/issues/489#issuecomment-455291966 - I think it's problematic to assume that the screen plane corresponds to the near plane.


toji:
This LGTM. I had originally thought that the screen position and the near plane would be equivalent, but after reading through this I realize that was an incorrect assumption. I appreciate that this version of the text still allows for mapping to the near plane if the screen's relationship to the space is unknown.

I generally see this as uncontroversial, since it wouldn't even require an implementation to change that had been following the previous text, so I'm going to merge it.

PR #584, opened 2019-04-12
`boundsGeometry` must account for `originOffset`

toji:
Added a paragraph indicating that the `boundsGeometry` of an
 `XRBoundedReferenceSpace` MUST be transformed by `originOffset.inverse`
 in order to ensure it stays well-aligned to the physical environment.


NellWaliczek:
This seems pretty intertwined with PR #589.  Perhaps we should just consolidate them?

PR #585, opened 2019-04-12
Reject outstanding promises on session shut down

### LINKED ISSUE #556: Clarify/Add requirement to explicitly reject promises in error cases
The spec should call out that in error cases promises should be rejected rather than allowed to time out.  Specifically in cases where there may be an outstanding promise and the device is unplugged/disconnected during that promise.
Perhaps InvalidStateError? Perhaps as a separate "Error Handling" section?

toji:
Fixes #556. Adds a catch-all statement that promises shouldn't outlive
the session they were returned from and indicates what error they should
reject with when the session shuts down.

Pinging @alcooper91 to take a look.


alcooper91:
Looks good to  me.

PR #586, opened 2019-04-12
`inverse` attribute always returns the same object

### LINKED ISSUE #576: Should XRRigidTransform.inverse.inverse return the original object?
Question asked by @NellWaliczek on https://github.com/immersive-web/webxr/pull/560#discussion_r270994813 Effectively, should to following code be guaranteed to evaluate to true?

```
let A = new XRRigidTransform(somePosition, someOrientation);
let B = A.inverse;
let C = B.inverse;
assert(A === C);
```

It wouldn't be a huge burden to retain the "parent" transform. There's possibly some minor memory concerns that would come along with that, but it's probably not a big deal. It may improve performance in some rather weird edge cases, but I don't think we should be optimizing for that.

I would be tempted to leave it up to the UA, but there IS a minor compat issue that would come out of it. The biggest functionality fallout of a same-object policy here would be that if someone patches an attribute onto the transform, and then calls `T.inverse.inverse` that patched value would still be expected to be there. I don't know if that's behavior we necessarily want people to be relying on, and as a result I would suggest not implementing this, but I do think it's worthy of discussion.

A related topic: Should `XRRigidTransform.inverse` return the same object every time it's called? That actually seems more logical to me, and probably yields real performance benefits in cases where the attribute is referenced multiple times, so I'd advocate for enforcing that.

toji:
Fixes #576. Specifies that the object returned by `XRRigidTransform`'s
`inverse` attribute must always be the same JS object and, furthermore,
the `inverse` of a transform returned by `inverse` must be the
originating transform.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -963,7 +963,7 @@ interface XRRigidTransform {
   readonly attribute DOMPointReadOnly position;
   readonly attribute DOMPointReadOnly orientation;
   readonly attribute Float32Array matrix;
-  readonly attribute XRRigidTransform inverse;
+  [SameObject] readonly attribute XRRigidTransform inverse;
 };
 </pre>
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -987,7 +987,7 @@ The <dfn attribute for="XRRigidTransform">orientation</dfn> attribute is a quate
 
 The <dfn attribute for="XRRigidTransform">matrix</dfn> attribute returns the transform described by the {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}} attributes as a [=matrix=]. This attribute SHOULD be lazily evaluated.
 
-The <dfn attribute for="XRRigidTransform">inverse</dfn> attribute returns a {{XRRigidTransform}} which, if applied to an object that had previously been transformed by the original {{XRRigidTransform}}, would undo the transform and return the object to it's initial pose. This attribute SHOULD be lazily evaluated.
+The <dfn attribute for="XRRigidTransform">inverse</dfn> attribute returns a {{XRRigidTransform}} which, if applied to an object that had previously been transformed by the original {{XRRigidTransform}}, would undo the transform and return the object to it's initial pose. This attribute SHOULD be lazily evaluated. The {{XRRigidTransform}} returned by {{inverse}} MUST return the originating {{XRRigidTransform}} as it's {{inverse}}.
```
Manishearth: "its"

toji: Thanks, fixed.

Issue #514, opened 2019-02-07
Combine XRStationaryReferenceSpaceSubtype into XRReferenceSpaceType


ddorwin:
https://github.com/immersive-web/webxr/pull/502 merged `subtype` into XRReferenceSpaceOptions, and have `type` and (optional) `subtype` are now in the same dictionary.

In addition to having a member that only applies when the type is set to one value, it's not clear what value there is in having subtypes for `"stationary"` rather than just making three types for stationary. It seems that it would be simpler for developers to only need to think about choosing the type(s) they support. This would also simplify the spec a little bit as described below.
<br/>

`XRStationaryReferenceSpace` exists only to provide the `subtype` member. We probably don't need this (see https://github.com/immersive-web/webxr/issues/513). Even if we did, it would probably make sense to provide the type, which would now include the existing subtypes, in the base XRReferenceSpace.

I don't think there is a need to have a 1:1 mapping of `type` to `XRReferenceSpace` sub-interfaces. Thus, we can probably remove `XRStationaryReferenceSpace` and `XRUnboundedReferenceSpace`, further simplifying the API and spec. Sub-interfaces can always be added later if necessary. The spec and applications will need to be updated to take advantage of new attributes anyway.

toji:
I know that when you reached out to me internally about this I let the conversation drop prior to fully resolving this, so sorry for that.

First off, I agree that in the spirit of #513 (and a desire to do a general clean-up pass over the whole API) we could probably drop the `subtype` member and then reduce the reference space types other than `XRBoundedReferenceSpace` to simply return an opaque `XRReferenceSpace` type. It shouldn't be too disruptive later to re-introduce those types either. The only thing it should affect in real applications is people doing `typeof` tests and that wouldn't be a meaningful way to differentiate the types anyway, so the risk of breaking content is low.

As for the type/subtype division at creation time, I'm not sure I see as much benefit in bikeshedding that particular decision? I think we want to keep the arguments as a dictionary for future expansion, so we're not reducing the number of types there. Merging names like "stationary-floor-level", while verbose, could certainly work. But I don't think it's any different in terms of cognitive load for developers, so I'd lean towards not changing the already established API shape in favor of concentrating on other known issues.

(Actually, I'll make an argument for going the other way and saying that it might be kind of nice if we determined a default `subtype` for "stationary" reference spaces. Probably 'eye-level'? At that point the division between `type` and `subtype` is more justified from an ergonomics standpoint.)

[...]

Issue #579, opened 2019-04-02
Clarify assumptions about stationary reference space movement


klausw:
Currently, the spec says: 

> An {{XRStationaryReferenceSpace}} represents a tracking space that the user is not expected to move around within. Tracking in a {{stationary}} reference space is optimized for the assumption that the user will not move much beyond their starting point, if at all.

I think this can be problematic since a 6DoF system will still report substantial movement even though the application has requested a stationary reference space.

This reference space should be interpreted as not *requiring* the user to move significantly from their starting position, but the application still needs to be able to handle such movement when it happens, and it must not assume that the user's position will always be essentially equivalent to their starting point position.

For example, the user may well take a couple of steps in the real world to take a closer look at a virtual object, and then remain standing in that position while using teleport or other locomotion to move around the game world. If a teleport mechanic is intended to put the user's feet in a new location, the application must adjust for the current displacement when calculating the new originOffset. (See https://github.com/immersive-web/webxr/issues/477 for more discussion of this.) If the application just places the new origin at a clicked-on floor location, the user's feet will be in the wrong place in the virtual world.

Similarly, if a game features NPC enemies that attack the player, it's essential that they aim at the displaced position and not at the origin. (IIRC Skyrim VR had that issue in early builds, where the player thought they were in cover, but were actually getting hit by enemies striking at the empty space around the origin.)

A separate issue is that the user agent doesn't enforce any restrictions on the user's pose displacement. If an application needs to prevent unexpected view positions from movement to avoid immersion-breaking behavior such as looking through walls or into objects, it needs to take its own measures such as fading or blurring the rendered view when the user is too far from the origin.

Overall, I think the distinction should be that an application should request a bounded reference space only when it has a way to make use of the bounds, for example by placing virtual objects within matching bounds in the virtual world so that they are reachable without artificial locomotion. An application that doesn't make use of that should feel free to use a stationary reference frame, but in any case needs to be able to deal with user position displacements in the range of a few meters as the user moves around.

cwilso:
Should be fixed by the fix for #597

klausw:
The clarification in #597 is useful, but I think it doesn't really resolve this issue. I think there are several things going on here:

* how are stationary reference types and their origins defined? (#597 does address this)
* what assumptions can applications make about user movement? For example, should a "stationary/floor-level" mode app work correctly for a user moving around a few steps in a roomscale tracked volume?
[...]

PR #587, opened 2019-04-12
Consolidate reference space types and interfaces

toji:
Since #574 leaves XRStationaryReferenceSpace with nothing else in it, taking the opportunity to consolidate it and XRUnboundedReferenceSpace into XRReferenceSpace, as was discussed in #514. 

Originally was part of #574 but separated the consolidation out at Nell's request.


index.bs
```diff
@@ -757,23 +755,24 @@ dictionary XRReferenceSpaceOptions {
 };
 </pre>
 
-An {{XRReferenceSpace}} is obtained by calling {{XRSession/requestReferenceSpace()}}, which creates an instance of an interface extending {{XRReferenceSpace}}, determined by the {{XRReferenceSpaceOptions/type}} value of the {{XRReferenceSpaceOptions}} dictionary passed into the call:
+An {{XRReferenceSpace}} is obtained by calling {{XRSession/requestReferenceSpace()}}, which creates an instance of an {{XRReferenceSpace}} or an interface extending it, determined by the {{XRReferenceSpaceOptions/type}} value of the {{XRReferenceSpaceOptions}} dictionary passed into the call. The {{XRReferenceSpaceOptions/type}} indicates the tracking behavior that the reference space will exhibit:
+
+  - Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRReferenceSpaceType">identity</dfn> creates an {{XRReferenceSpace}} instance. It represents a reference space without any tracking behavior. The {{XRPose/transform}} of any {{XRPose}} queried from an {{identity}} reference space MUST be an [=identity transform=] offset by the {{originOffset}}.
```
Manishearth: Given the discussion in https://github.com/immersive-web/webxr/issues/565 this characterization seems inaccurate? The pose queried changes based on the base space used, a more accurate characterization is that "getViewerPose" should always return the identity transform (potentially offset with originOffset).

toji:
Combined this PR with another one that I had in progress to consolidate the reference space type and subtype enums because @NellWaliczek and I both think these changes make the most sense when taken together. As a result this PR now fully fixes #514 and fixes #542.

index.bs
toji: Reworked this text, though I think it's going to need to be updated again when we land some of the planed changes re: describing these spaces in terms of the native origin as you've started looking at.

index.bs
```diff
@@ -752,27 +741,25 @@ dictionary XRReferenceSpaceOptions {
...
 
-  - Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRReferenceSpaceType">identity</dfn> creates an {{XRReferenceSpace}} instance.
+An {{XRReferenceSpace}} is obtained by calling {{XRSession/requestReferenceSpace()}}, which creates an instance of an {{XRReferenceSpace}} or an interface extending it, determined by the {{XRReferenceSpaceType}} enum value passed into the call. The type indicates the tracking behavior that the reference space will exhibit:
 
-  - Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRReferenceSpaceType">stationary</dfn> creates an {{XRStationaryReferenceSpace}} instance.
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">identity</dfn> creates an {{XRReferenceSpace}} instance. It represents a reference space where the [=viewer=] is always at the origin (prior to applying the {{originOffset}}). Only the {{XRSession/viewerSpace}} and {{XRInputSource/targetRaySpace}}'s of {{XRInputSource}}'s with an {{XRInputSource/targetRayMode}} of {{XRTargetRayMode/screen}} can be spatially related to an {{identity}} reference space.
```
NellWaliczek: > It represents a reference space where the [=viewer=] is always at the origin (prior to applying the {{originOffset}}).

Instead of this, can we describe the behavior in terms of the native origin now that we have that variable?

NellWaliczek: This is really a comment for all of the reference space types

index.bs
```diff
@@ -752,27 +741,25 @@ dictionary XRReferenceSpaceOptions {
...
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">position-disabled</dfn> creates an {{XRReferenceSpace}} instance. It represents a tracking space where orientation is tracked but the [=viewer=]s position is always reported as being at the origin.
 
-  - Passing a {{XRReferenceSpaceOptions/type}} of <dfn enum-value for="XRReferenceSpaceType">unbounded</dfn> creates an {{XRUnboundedReferenceSpace}} instance if supported by the [=/XR device=] and the {{XRSession}}.
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">eye-level</dfn> creates an {{XRReferenceSpace}} instance. It represents a tracking space with it's origin near the user's head at the time of creation. The exact position and orientation will be initialized based on the conventions of the underlying platform. When using this reference space the user is not expected to move beyond their initial position much, if at all, and tracking is optimized for that purpose. For devices with [=6DoF=] tracking, {{eye-level}} reference spaces should emphasize keeping the origin stable relative to the user's environment.
```
NellWaliczek: > The exact position and orientation will be initialized based on the conventions of the underlying platform.

Should we change this to say that the originOffset should be set such that the behavior is always the same?

NellWaliczek: > When using this reference space the user is not expected to move beyond their initial position much,

Didn't @klausw have concerns about this phrasing not being clear enough?

index.bs
```diff
@@ -752,27 +741,25 @@ dictionary XRReferenceSpaceOptions {
...
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">unbounded</dfn> creates an {{XRReferenceSpace}} instance if supported by the [=/XR device=] and the {{XRSession}}. It represents a tracking space where the user is expected to move freely around their environment, potentially even long distances from their starting point. Tracking in an {{unbounded}} reference space is optimized for stability around the user's current position, and as such the tracking origin may drift over time.
+
+Note: The {{position-disabled}} type is primarily intended for use with pre-rendered media such as panoramic photos or videos. It should not be used for most other media types due to user discomfort associated with the lack of a neck model or full positional tracking.
   
-Devices that support {{XRReferenceSpaceType/stationary}} reference spaces MUST support all {{XRStationaryReferenceSpaceSubtype}}s.
+Devices that support any of the following {{XRReferenceSpaceType}}s MUST support all three: {{XRReferenceSpaceType/position-disabled}}, {{XRReferenceSpaceType/eye-level}}, {{XRReferenceSpaceType/floor-level}}.
```
NellWaliczek: It would be nice to add one sentence explaining why.

spatial-tracking-explainer.md
```diff
@@ -7,7 +7,7 @@ A big differentiating aspect of XR, as opposed to standard 3D rendering, is that
...
+There are several types of reference spaces: `inline`, `position-disabled`, `eye-level`, `floor-level`, `bounded`, and `unbounded`.  A `bounded` experience is one in which the user will move around their physical environment to fully interact, but will not need to travel beyond a fixed boundary defined by the XR hardware.  An `unbounded` experience is one in which a user is able to freely move around their physical environment and travel significant distances.  `position-disabled`, `eye-level`, and `floor-level` experiences are ones which do not require the user to move around in space, and includes "seated" or "standing" experiences.  Examples of each of these types of experiences can be found in the detailed sections below.
```
NellWaliczek: >and includes "seated" or "standing" experiences

"include"

index.bs
toji: This actually raises a good question about our understanding of the native origin: I guess I had been viewing it as the "base" origin for a reference space, regardless of if the UA had to apply some internal tweaks to get it there. This comment suggests you see the native origin as a function of the underlying XR backend and completely out of the UAs control, with origin offsets being applied in any case where the desired reference space does not already happen to match the platform-provided one. I'm fine with that interpretation, though if that's the case we should not land this PR until after #612 lands and removes the `originOffset` attribute.

toji:
Made a few tweaks in response to Nell's comments, but had an open question about the native origin/origin offset thing.

index.bs
Manishearth: I've been viewing it the same way (and in fact Nell's comment didn't initially make sense to me since I wasn't sure what the behavior was supposed to be the same to). 

Her interpretation makes sense too, though it surprises me a bit.

I guess I'm unclear on when such a situation can occur.

NellWaliczek: I'm not sure it really strongly matters either way, but the term "native" origin implies to me that it's the origin as defined by the underlying tracking system...

NellWaliczek: Tracked in #579

index.bs
```diff
@@ -766,21 +766,21 @@ Each {{XRReferenceSpace}} has a <dfn for="XRReferenceSpace">type</dfn>, which is
...
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">unbounded</dfn> creates an {{XRReferenceSpace}} instance if supported by the [=/XR device=] and the {{XRSession}}. It represents a tracking space where the user is expected to move freely around their environment, potentially even long distances from their starting point. Tracking in an {{unbounded}} reference space is optimized for stability around the user's current position, and as such the [=native origin=] may drift over time.
+
+Devices that support any of the following {{XRReferenceSpaceType}}s MUST support all three: {{XRReferenceSpaceType/position-disabled}}, {{XRReferenceSpaceType/eye-level}}, {{XRReferenceSpaceType/floor-level}}. The minimum sensor data necessary for supporting all three reference spaces is the same.
```
NellWaliczek: > The minimum sensor data necessary for supporting all three reference spaces is the same.
	
Not necessarily the case for the floor-level type, yes?

Manishearth:
Btw, I've rebased this over  https://github.com/immersive-web/webxr/pull/612 in https://github.com/Manishearth/webxr/tree/consolidate


(I need this to start working on https://github.com/immersive-web/editor-collab/pull/38 )

index.bs
toji: The word "minimum" is doing a lot of lifting in that sentence. You can get an emulated floor-level with nothing but orientation data and a static offset.

index.bs
```diff
@@ -766,21 +766,21 @@ Each {{XRReferenceSpace}} has a <dfn for="XRReferenceSpace">type</dfn>, which is
...
 
-  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">identity</dfn> creates an {{XRReferenceSpace}} instance. It represents a reference space where the [=viewer=] is always at the origin (prior to applying the {{originOffset}}). Only the {{XRSession/viewerSpace}} and {{XRInputSource/targetRaySpace}}'s of {{XRInputSource}}'s with an {{XRInputSource/targetRayMode}} of {{XRTargetRayMode/screen}} can be spatially related to an {{identity}} reference space.
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">identity</dfn> creates an {{XRReferenceSpace}} instance. It represents a reference space where the [=viewer=] is always at the [=native origin=]. Only the {{XRSession/viewerSpace}} and {{XRInputSource/targetRaySpace}}'s of {{XRInputSource}}'s with an {{XRInputSource/targetRayMode}} of {{XRTargetRayMode/screen}} can be spatially related to an {{identity}} reference space.
```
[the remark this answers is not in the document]
toji: Does that matter, in this case, if it can't relate to anything else?

toji:
Apologies, @Manishearth. I actually merged this in the opposite order (resolving the conflicts along the way.) Hopefully you can rebase your change onto the new master without much trouble?

Manishearth:
Oh, that's fine, I hadn't made any major changes yet :)

Issue #588, opened 2019-04-15
Mark up readonly attributes with `[SameObject]` wherever applicable


toji:
#586 makes use of the [`[SameObject]`](https://heycam.github.io/webidl/#SameObject) extended attribute for the first time in the spec, but it should be applied to many more attributes. Many of the readonly attributes in the spec are not expected to change their value for the lifetime of the parent object and thus should be marked as returning the same object each time.

Examples include (but are certainly not limited to):

 - `XRSession.viewerSpace`
 - `XRRigidTransform.position`
 - `XRRigidTransform.orientation`
 - `XRPose.transform`
 - `XRWebGLLayer.context`

And so on.

PR #589, opened 2019-04-15
Change originOffset w/ Reference Space constructor

### LINKED ISSUE #580: Should origin offset be a new reference space instead?
See: https://github.com/immersive-web/webxr/pull/569#issuecomment-475823936

> FWIW, I think originOffset makes it difficult to explain reference spaces things clearly. An XRReferenceSpace basically consists of two XRSpaces that are transformed into each other by the originOffset, and it's hard to explain how that transform works when the two spaces don't have names. I've used the name "effective reference space" for the offset-modified externally visible space and "unmodified reference space" for the internal space corresponding to an identity originOffset, but that's clunky.

In today's call we discussed this a bit, an alternate model suggested by @klausw was to have might be to instead have a constructor for origin offsets, probably something like:

```webidl
[SecureContext, Exposed=Window,
 Constructor(XRReferenceSpace base, XRRigidTransform offset)]
interface XROffsetReferenceSpace : XRReferenceSpace {
  attribute XRRigidTransform offset;
  readonly attribute XRReferenceSpace base;
}
```

where if you want to apply an origin offset, you effectively construct a _new_ space wrapping over the space you have, and use it where you need the offset. This avoids having to maintain the clunky distinction in the spec, especially for things like input spaces.

toji:
Fixes #580. Changes originOffset to be immutable and establish that the
way that the originOffset changes is by constructing a new reference
space with the base space and an additional offset to apply on top of
it. This should hopefully solve several timing issues related to
changing the originOffset mid-frame.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -771,12 +771,10 @@ dictionary XRReferenceSpaceOptions {
   XRStationaryReferenceSpaceSubtype subtype;
 };
 
-
-[SecureContext, Exposed=Window,
- Constructor(XRReferenceSpace base, XRRigidTransform originOffset)]
+[SecureContext, Exposed=Window]
 interface XRReferenceSpace : XRSpace {
-  readonly attribute XRReferenceSpace? base;
-  readonly attribute XRRigidTransform originOffset;
+  XRReferenceSpace getOffsetReferenceSpace(XRRigidTransform originOffset);
+
   attribute EventHandler onreset;
 };
 </pre>
@@ -803,22 +801,6 @@ Note: The {{position-disabled}} subtype is primarily intended for use with pre-r
   
 Devices that support {{XRReferenceSpaceType/stationary}} reference spaces MUST support all {{XRStationaryReferenceSpaceSubtype}}s.
 
-<!--
-The <dfn constructor for="XRReferenceSpace">XRReferenceSpace(|base|, |originOffset|)</dfn> constructor MUST perform the following steps when invoked:
-
-  1. Let |refSpace| be a new {{XRReferenceSpace}}.
-  1. If |base|'s {{XRReferenceSpace/base}} is <code>null</code>, set |refSpace|'s {{XRReferenceSpace/base}} to |base|.
-  1. Else set |refSpace|'s {{XRReferenceSpace/base}} to |base|'s {{XRReferenceSpace/base}}.
-  1. Set |refSpace|'s {{XRReferenceSpace/originOffset}} to |originOffset| multiplied by |base|'s {{XRReferenceSpace/originOffset}}.
-  1. Return |refSpace|.
-
-</div>
--->
-
-<section class="unstable">
-The <dfn attribute for="XRReferenceSpace">originOffset</dfn> attribute gets and sets the {{XRReferenceSpace}}'s [=XRSpace/origin offset=]. Changes to the {{originOffset}} take effect immediately, and subsequent poses queried with the {{XRReferenceSpace}} will take into account the [=effective origin=].
-</section>
-
 The <dfn attribute for="XRReferenceSpace">onreset</dfn> attribute is an [=Event handler IDL attribute=] for the {{reset}} event type.
 
 <div class="algorithm" data-algorithm="create-reference-space">
@@ -836,8 +818,21 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
   1. If |type| is {{stationary}}, let |referenceSpace| be a new {{XRStationaryReferenceSpace}} with a [=XRStationaryReferenceSpace/subtype=] of |options|'s {{XRReferenceSpaceOptions/subtype}}.
   1. Else if |type| is {{bounded}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
   1. Else if |type| is {{unbounded}}, let |referenceSpace| be a new {{XRUnboundedReferenceSpace}}.
-  1. Set |referenceSpace|'s {{XRReferenceSpace/originOffset}} to an [=identity transform=].
   1. Return |referenceSpace|.
+
+</div>
+
+<div class="algorithm" data-algorithm="get-offset-space">
+The <dfn method for="XRReferenceSpace">getOffsetReferenceSpace(|originOffset|)</dfn> method MUST perform the following steps when invoked:
+
+  1. Let |base| be the {{XRReferenceSpace}} the method was called on.
+  1. If |base| is an instance of {{XRStationaryReferenceSpace}}, let |offsetSpace| be a new {{XRStationaryReferenceSpace}}.
+  1. Else if |base| is an instance of {{XRBoundedReferenceSpace}}, let |offsetSpace| be a new {{XRBoundedReferenceSpace}} and set |offsetSpace|'s {{boundsGeometry}} to |base|'s {{boundsGeometry}}, with each point transformed by the {{XRRigidTransform/inverse}} of |originOffset|.
+  1. Else if |base| is an instance of {{XRUnboundedReferenceSpace}}, let |offsetSpace| be a new {{XRUnboundedReferenceSpace}}.
+  1. Else let |offsetSpace| be a new {{XRReferenceSpace}}.
+  1. Set |offsetSpace|'s [=native origin=] to |base|'s [=native origin=].
+  1. Set |offsetSpace|'s [=origin offset=] to the result of [=multiply transforms|multiplying=] |originOffset| by |base|'s [=origin offset=].
+  1. Return |offsetSpace|.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -750,15 +750,19 @@ dictionary XRReferenceSpaceOptions {
   XRStationaryReferenceSpaceSubtype subtype;
 };
 
-[SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {
-  attribute XRRigidTransform originOffset;
+
+[SecureContext, Exposed=Window,
+ Constructor(XRReferenceSpace base, XRRigidTransform originOffset)]
+interface XRReferenceSpace : XRSpace {
+  readonly attribute XRReferenceSpace? base;
```
Manishearth: `[SameObject]` on both in accordance with https://github.com/immersive-web/webxr/issues/588 ?

index.bs
```diff
@@ -779,10 +783,18 @@ Note: The {{position-disabled}} subtype is primarily intended for use with pre-r
...
 Devices that support {{XRReferenceSpaceType/stationary}} reference spaces MUST support all {{XRStationaryReferenceSpaceSubtype}}s.
 
-<section class="unstable">
-The <dfn attribute for="XRReferenceSpace">originOffset</dfn> attribute is a {{XRRigidTransform}} that describes an additional translation and rotation to be applied to any poses queried using the {{XRReferenceSpace}}. It is initially set to an [=identity transform=]. Changes to the {{originOffset}} take effect immediately, and subsequent poses queried with the {{XRReferenceSpace}} will take into account the new transform.
+The <dfn constructor for="XRReferenceSpace">XRReferenceSpace(|base|, |originOffset|)</dfn> constructor MUST perform the following steps when invoked:
+
+  1. Let |refSpace| be a new {{XRReferenceSpace}}.
+  1. If |base|'s {{XRReferenceSpace/base}} is <code>null</code>, set |refSpace|'s {{XRReferenceSpace/base}} to |base|.
+  1. Else set |refSpace|'s {{XRReferenceSpace/base}} to |base|'s {{XRReferenceSpace/base}}.
+  1. Set |refSpace|'s {{XRReferenceSpace/originOffset}} to |originOffset| multiplied by |base|'s {{XRReferenceSpace/originOffset}}.
```
Manishearth: Need to clarify pre vs post multiply here.

Ideally avoid talking about multiplication at all, instead saying "Set refSpace's originOffset to a transform where base's originOffset is applied first, and originOffset is applied second".

index.bs
```diff
@@ -750,15 +750,19 @@ dictionary XRReferenceSpaceOptions {
   XRStationaryReferenceSpaceSubtype subtype;
 };
 
-[SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {
-  attribute XRRigidTransform originOffset;
+
+[SecureContext, Exposed=Window,
+ Constructor(XRReferenceSpace base, XRRigidTransform originOffset)]
```
NellWaliczek: Can we change this pattern to not be a constructor, but rather a function on the XRReferenceSpace object being based on? Maybe `XRReferenceSpace.getOffsetReferenceSpace()`

index.bs
NellWaliczek: In accordance with our move to remove attributes that reflect the object's creation parameters, this should probably be removed altogether.

index.bs
Manishearth: Ah, yeah, I guess we should figure out how to consistently talk about this first instead of attempting to fix it in each use site in an ad hoc manner.

index.bs
```diff
@@ -822,9 +836,22 @@ The origin of a {{XRBoundedReferenceSpace}} MUST be positioned at the floor, suc
...
+  1. Return |refSpace|.
+
+</div>
+
 The <dfn attribute for="XRBoundedReferenceSpace">boundsGeometry</dfn> attribute describes the border around the {{XRBoundedReferenceSpace}}, which the user can expect to safely move within. 
 
-The polygonal boundary is given as an array of {{DOMPointReadOnly}}s, which represents a loop of points at the edges of the safe space. The points describe offsets from the {{XRReferenceSpace}} origin in meters. Points MUST be given in a clockwise order as viewed from above, looking towards the negative end of the Y axis. The {{DOMPointReadOnly/y}} value of each point MUST be <code>0</code> and the {{DOMPointReadOnly/w}} value of each point MUST be <code>1</code>. The bounds can be considered to originate at the floor and extend infinitely high. The shape it describes MAY be convex or concave.
+The polygonal boundary is given as an array of {{DOMPointReadOnly}}s, which represents a loop of points at the edges of the safe space. The points describe offsets from the {{XRReferenceSpace}} origin in meters.
```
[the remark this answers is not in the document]
NellWaliczek: Seems like you're trying to address the same issue in this PR that you're addressing in a different way in #584?  Perhaps let's consolidate these changes together so they can be viewed in one place?

index.bs
```diff
@@ -779,10 +783,18 @@ Note: The {{position-disabled}} subtype is primarily intended for use with pre-r
...
+  1. If |base|'s {{XRReferenceSpace/base}} is <code>null</code>, set |refSpace|'s {{XRReferenceSpace/base}} to |base|.
+  1. Else set |refSpace|'s {{XRReferenceSpace/base}} to |base|'s {{XRReferenceSpace/base}}.
+  1. Set |refSpace|'s {{XRReferenceSpace/originOffset}} to |originOffset| multiplied by |base|'s {{XRReferenceSpace/originOffset}}.
+  1. Return |refSpace|.
 
-Note: Changing the {{originOffset}} between pose queries in a single [=XR animation frame=] is not advised, since it will cause inconsistencies in the tracking data and rendered output.
+</div>
+
+<section class="unstable">
+The <dfn attribute for="XRReferenceSpace">originOffset</dfn> attribute is a {{XRRigidTransform}} that describes an additional translation and rotation to be applied to any poses queried using the {{XRReferenceSpace}}.
```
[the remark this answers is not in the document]
klausw: @toji, are you intentionally reversing the meaning of the `originOffset` here? If the `originOffset` provides the position and orientation of the offset space, where the position member is exactly the origin position in base space coordinates, queries using the offset space need to apply the **inverse** of the `originOffset` transform when returning poses, assuming you want to match the voted resolution for #477. (cf. #569 and the rough consensus from #580.)

index.bs
```diff
@@ -822,9 +836,22 @@ The origin of a {{XRBoundedReferenceSpace}} MUST be positioned at the floor, suc
 
 Note: Other XR platforms sometimes refer to the type of tracking offered by a {{bounded}} reference space as "room scale" tracking. An {{XRBoundedReferenceSpace}} is not intended to describe multi-room spaces, areas with uneven floor levels, or very large open areas. Content that needs to handle those scenarios should use an {{unbounded}} reference space.
 
+The <dfn constructor for="XRBoundedReferenceSpace">XRBoundedReferenceSpace(|base|, |originOffset|)</dfn> constructor MUST perform the following steps when invoked:
+
+  1. Let |refSpace| be a new {{XRBoundedReferenceSpace}}.
+  1. If |base|'s {{XRReferenceSpace/base}} is <code>null</code>, set |refSpace|'s {{XRReferenceSpace/base}} to |base|.
+  1. Else set |refSpace|'s {{XRReferenceSpace/base}} to |base|'s {{XRReferenceSpace/base}}.
+  1. Set |refSpace|'s {{XRReferenceSpace/originOffset}} to |originOffset| multiplied by |base|'s {{XRReferenceSpace/originOffset}}.
```
klausw: Is there an existing definition what it means to multiply XRRigidTransforms? This needs to be unambiguous. I think it's backwards as written.

Let's say we set up a base space and two chained offsets like this:

```js
let baseSpaceA = ...; starting reference space A
let offsetSpaceA = new XRBoundedReferenceSpace(baseSpaceA, originOffsetA);
let offsetSpaceB = new XRBoundedReferenceSpace(offsetSpaceA, originOffsetB);
```

Each originOffset's matrix is a transform from offset space to base space. Space B's base space corresponds to space A's offset space.

You can use the matrices to transform coordinates in `offsetSpaceB` to coordinates in `baseSpaceA`:

```
pointInBaseA = baseAFromOffsetA * offsetAFromOffsetB * pointInOffsetB;
             = offsetSpaceA.originOffset.matrix * offsetSpaceB.originOffset.matrix * pointInOffsetB
```


and therefore, space B's originOffset.matrix is space A's originOffset.matrix times space B's originOffset.matrix, or for spec purposes:

```
1. Set |refSpace|'s {{XRReferenceSpace/originOffset}} to |base|'s {{XRReferenceSpace/originOffset}} multiplied by |originOffset|.
```

NellWaliczek: ➕ 1 to having this defined as an "algorithm" in the spec so we can easily reference it in other spec "algorithms".  @klausw , do you want to post a PR with this clarification text and Brandon and I can review it?  I think getting that nailed down first will go a long way towards our goal of algorithmically defining the behavior of `getPose()` and `getViewerPose()`.

klausw: @NellWaliczek That's what I was trying to do in my existing PR, see especially https://github.com/immersive-web/webxr/pull/569/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR955 which tries to be quite specific about this.

Can we discuss that in that PR? I'm happy to extend this further, i.e. adding more info about chaining transforms, but I'd like to get consensus on the basics first.

toji: Sorry, I think there's a misunderstanding here between Nell and Klaus. I believe the algorithm which Nell is talking about is one that describes how two XRRigidTransforms are multiplied. (At least that's how it reads given the context.) The line in Klaus' PR that was linked details (very well) how a matrix is derived from the rigid transform.

I do think that having an in-spec algorithm that describes how transforms are multiplied would be extremely useful. That's something that Nell or I could sketch out if needed, but either way I'd want to have Klaus and/or Manish review it to ensure we're all on the same page. Then that should serve as a fundamental building block though which we can help describe a great many of the rest of these concepts.

klausw: @toji The text I have now in #569 specifies the content of a XRRigidTransform matrix and how it transforms numeric coordinates, and this should constrain implementations to one specific interpretation of how the transforms work. As a result, there should only be one correct way to chain transforms.

I don't think it makes sense to have an independent PR or discussion about chaining transforms unless we can get agreement on these basic properties of transforms. There's no separate design freedom for the underlying matrix math of XRRigidTransform multiplication.  I agree it would be helpful to add an explanation to the text, but I think this would be appropriate as a followup since it's derived from the basic properties and definitions of the transforms and matrices.

I'm fairly confident that the chaining I described in [my comment above](https://github.com/immersive-web/webxr/pull/589#discussion_r275979306) matches #569, assuming usual column-major GL matrices applied to column vectors, but I'd appreciate an independent sanity check.

Would it help to schedule some time with interested parties to talk through this? I feel as if we're still a bit stuck on this issue even after discussing it in two of the weekly conf calls, and I'm unsure if there's actual disagreement or not.

ddorwin:
For reference, this PR was moved to https://github.com/immersive-web/webxr/pull/612#issuecomment-487255447.

PR #594, opened 2019-04-19
[Editiorial] Status metadata does not need "w3c/"

himorin:
to fix regression at @2f84c43

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2,7 +2,7 @@
 Shortname: webxr
 Title: WebXR Device API
 Group: immersivewebwg
-Status: w3c/ED
+Status: ED
 TR: https://www.w3.org/TR/webxr/
 ED: https://immersive-web.github.io/webxr/
 Repository: immersive-web/webxr


toji:
Looks good, thanks!

toji:
Sorry, just pushed through a revert of this because I noticed it was breaking local builds of the spec. When I run `make` locally, which calls out to the Bikeshed web service via curl, I get the following message:

**FATAL ERROR:** You used Status: ED, but that's limited to the 'w3c', 'khronos', or 'fido' orgs, and your group 'immersivewebwg' isn't recognized as being in any of those orgs. If this is wrong, please file a Bikeshed issue to categorize your group properly, and/or try:
Status: w3c/ED
Status: khronos/ED
Status: fido/ED

himorin:
I'm quite sorry to make things messy... I've put two PRs to bikeshed to fix a 'FATAL ERROR' you noted (I'm not sure why I have not noticed when I've tested locally...), and to fix one of our template in bikeshed even when we use w3c/XX as "Status".

himorin:
Two PRs are merged, status with both "w3c/ED" and "ED" should work now (at least I've tested with current HEAD of bikeshed) and output correct html without "FATAL ERROR". So, even with "w3c/ED", Editors' Draft logo at top left will be shown (which are already taken into gh-pages at 0bfd2a5).
@toji if you think this change is fine, please put at some point..

toji:
Thank you for that. Rolled this fix into another minor cleanup in https://github.com/immersive-web/webxr/commit/3cdebce178a89daceae5e1ac3f3a6e622a644a13

PR #595, opened 2019-04-19
Rename supportsSessionMode to supportsSession

### LINKED ISSUE #592: supportsSessionMode -> supportsSession?
Now that #566 has merged, I noticed that it looks a teensy bit weird to have `supportsSessionMode(mode)` and `requestSession(mode)` side by side on the `navigator.xr` interface. The difference in names made more sense when `requestSession` took a full dictionary, but if all it takes to request a session is the `XRSessionMode` then it seems like the name `supportsSessionMode` is unnecessarily specific. Should we change it back to `supportsSession` to give the calls back some symmetry?

toji:
Fixes #592. Brings name inline with requestSession, which now takes the
same arguments.

This is a simple enough change that I figured I'd just have a PR ready to go if we agree with #592.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -172,7 +172,7 @@ Application flow {#applicationflow}
 
 Most applications using the WebXR Device API will follow a similar usage pattern:
 
-  * Query {{XR/supportsSessionMode()|navigator.xr.supportsSessionMode()}} to determine if the desired type of XR content is supported by the hardware and UA.
+  * Query {{XR/supportsSession()|navigator.xr.supportsSession()}} to determine if the desired type of XR content is supported by the hardware and UA.
   * If so, advertise the XR content to the user.
   * Wait for the user to [=triggered by user activation|trigger a user activation event=] indicating they want to begin viewing XR content.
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -270,13 +270,13 @@ When the <dfn method for="XR">supportsSessionMode(|mode|)</dfn> method is invoke
 
 </div>
 
-Calling {{XR/supportsSessionMode()}} MUST NOT trigger device-selection UI as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
+Calling {{XR/supportsSession()}} MUST NOT trigger device-selection UI as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation. Additionally, calling {{XR/supportsSession()}} MUST NOT interfere with any running XR applications on the system.
```
NellWaliczek: Good catch that this didn't make it over from the explainer.  We should probably also make it clear that this should not cause any other applications to launch on the system such as SteamVR.

toji: Done!

PR #599, opened 2019-04-22
Add [SameObject] to all applicable readonly attributes

### LINKED ISSUE #588: Mark up readonly attributes with `[SameObject]` wherever applicable

Manishearth:
Fixes #588

There are only two non-primitive `readonly attribute`s that this does not affect:

 - `XRSession.renderState`: This can be updated with `updateRenderState()`
 - `XRBoundedReferenceSpace.boundsGeometry`: I'm unsure if this array can be updated if settings change

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -363,7 +363,7 @@ enum XREnvironmentBlendMode {
   // Attributes
   readonly attribute XREnvironmentBlendMode environmentBlendMode;
   readonly attribute XRRenderState renderState;
-  readonly attribute XRSpace viewerSpace;
+  [SameObject] readonly attribute XRSpace viewerSpace;
 
   // Methods
   void updateRenderState(optional XRRenderStateInit state);
@@ -670,7 +670,7 @@ An {{XRFrame}} represents a snapshot of the state of all of the tracked objects 
 
 <pre class="idl">
 [SecureContext, Exposed=Window] interface XRFrame {
-  readonly attribute XRSession session;
+  [SameObject] readonly attribute XRSession session;
 
   XRViewerPose? getViewerPose(XRReferenceSpace referenceSpace);
   XRPose? getPose(XRSpace space, XRSpace relativeTo);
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -544,8 +544,8 @@ dictionary XRRenderStateInit {
   readonly attribute double depthNear;
   readonly attribute double depthFar;
   readonly attribute double? inlineVerticalFieldOfView;
-  readonly attribute XRLayer? baseLayer;
-  readonly attribute XRPresentationContext? outputContext;
+  [SameObject] readonly attribute XRLayer? baseLayer;
+  [SameObject] readonly attribute XRPresentationContext? outputContext;
```
toji: These two attributes are the only ones I question in this change, because it implies that each time the render state changes the session attribute will be a new object. That's actually not a bad assumption to make, and it's probably preferable to having `XRRenderState` be a live object, but I'd like to get that clearly defined one way or the other in the spec prior to adding a dependency on that behavior here.

Manishearth: Oh, hm, that's fair. I'll remove that and file an issue

toji:
Looks great, thank you! I have one change request but otherwise I'm happy to merge this.

You are correct about the `boundsGeometry` being a live attribute as we have it defined now, by the way.

Manishearth:
Done.

Manishearth:
Should I leave a note on XRBoundedReferenceSpace that `boundsGeometry` may change?

Also I added in two uncommitted SameObject changes that I'd forgotten to push: `context` and `framebuffer` on `XRWebGLLayer`. Does that sound correct?

toji:
Hm......... `context` should be safe, but I'm actually not sure about `framebuffer`. Let me look into that a bit.

Manishearth:
Yeah, I was a bit iffy about that one too

toji:
Nope, actually `framebuffer` should be fine because while the attachments need to swap out on a per-frame basis the framebuffer itself can be a single object throughout! And with that in mind I think I'll merge this now. Thank you!

PR #604, opened 2019-04-23
Remove ability to dynamically modify originOffset

Manishearth:
Addresses point 3 of https://github.com/immersive-web/editor-collab/pull/36#pullrequestreview-229743358

This change actually makes it impossible to use originOffset -- maybe I misinterpreted that point?

r? @toji


Manishearth:
wrong repo, https://hub.github.com doesn't like forks that aren't named the same

PR #606, opened 2019-04-24
[minor] Views don't provide view matrices anymore

Manishearth:
We removed the `viewMatrix` attribute quite a while ago.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1064,7 +1064,7 @@ The <dfn attribute for="XRPose">emulatedPosition</dfn> attribute MUST be <code>f
 XRViewerPose {#xrviewerpose-interface}
 -------------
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1064,7 +1064,7 @@ The <dfn attribute for="XRPose">emulatedPosition</dfn> attribute MUST be <code>f
...
+An {{XRViewerPose}} is an {{XRPose}} describing the state of a <dfn>viewer</dfn> of the XR scene as tracked by the [=/XR device=]. A [=viewer=] may represent a tracked piece of hardware, the observed position of a users head relative to the hardware, or some other means of computing a series of viewpoints into the XR scene. {{XRViewerPose}}s can only be queried relative to an {{XRReferenceSpace}}. It provides, in addition to the {{XRPose}} values, an array of [=view=]s which include rigid transforms and projection matrices. These matrices should be used by the application when render a frame of an XR scene.
```
toji: > ...include rigid transforms and projection matrices. These matrices should be used by the application when render a frame of an XR scene.

Good catch and a nice change. One minor suggestion clarify things even further:

"...include rigid transforms **to indicate the viewpoint** and projection matrices. These **values** should be used by the application when render a frame of an XR scene."

Manishearth: fixed

toji:
Thanks!

PR #612, opened 2019-04-27
Make origin offset immutable

### LINKED ISSUE #580: Should origin offset be a new reference space instead?

toji:
Fixes #580. Changes originOffset to be immutable and establish that the
way that the originOffset changes is by creating a new reference
space with the `getOffsetReferenceSpace()` method of the base space.
This should hopefully solve several timing issues related to changing the
originOffset mid-frame.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -830,9 +830,7 @@ Note: Other XR platforms sometimes refer to the type of tracking offered by a {{
 
 The <dfn attribute for="XRBoundedReferenceSpace">boundsGeometry</dfn> attribute describes the border around the {{XRBoundedReferenceSpace}}, which the user can expect to safely move within. 
 
-The polygonal boundary is given as an array of {{DOMPointReadOnly}}s, which represents a loop of points at the edges of the safe space. The points describe offsets from the {{XRReferenceSpace}} origin in meters.
-
-Relative to the {{XRBoundedReferenceSpace}}'s [=native origin=], points MUST be given in a clockwise order as viewed from above, looking towards the negative end of the Y axis. The {{DOMPointReadOnly/y}} value of each point MUST be <code>0</code> and the {{DOMPointReadOnly/w}} value of each point MUST be <code>1</code>. The bounds can be considered to originate at the floor and extend infinitely high. The shape it describes MAY be convex or concave.
+The polygonal boundary is given as an array of {{DOMPointReadOnly}}s, which represents a loop of points at the edges of the safe space. The points describe offsets from the {{XRReferenceSpace}} origin in meters. Points MUST be given in a clockwise order as viewed from above, looking towards the negative end of the Y axis. The {{DOMPointReadOnly/y}} value of each point MUST be <code>0</code> and the {{DOMPointReadOnly/w}} value of each point MUST be <code>1</code>. The bounds can be considered to originate at the floor and extend infinitely high. The shape it describes MAY be convex or concave.
 
 Note: Content should not require the user to move beyond the {{boundsGeometry}}. It is possible for the user to move beyond the bounds if their physical surroundings allow for it, resulting in position values outside of the polygon they describe. This is not an error condition and should be handled gracefully by page content.
 
--- a/index.bs
+++ b/index.bs
@@ -756,15 +756,19 @@ enum XRReferenceSpaceType {
   "unbounded"
 };
 
-[SecureContext, Exposed=Window] interface XRReferenceSpace : XRSpace {
-  attribute XRRigidTransform originOffset;
+
+[SecureContext, Exposed=Window,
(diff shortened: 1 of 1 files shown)


toji:
This is the same PR as #589, but due to a dumb branch deletion on my part and a dumber restriction on GitHub's part, that PR is now un-re-openable forevermore. So y'all get this one now instead. 😝 

FWIW this addresses the feedback given in the previous PR, primarily moving the offset reference space creation mechanism from a constructor to a method, which definitely feels cleaner. PTAL!

index.bs
```diff
@@ -822,6 +818,20 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
 
 </div>
 
+<div class="algorithm" data-algorithm="get-offset-space">
+The <dfn method for="XRReferenceSpace">getOffsetReferenceSpace(|originOffset|)</dfn> method MUST perform the following steps when invoked:
+
+  1. Let |base| be the {{XRReferenceSpace}} the method was called on.
+  1. If |base| is an instance of {{XRStationaryReferenceSpace}}, let |offsetSpace| be a new {{XRStationaryReferenceSpace}}.
```
NellWaliczek: Are you planning to rebase on top of #587?

index.bs
```diff
@@ -822,6 +818,20 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
...
+  1. Else if |base| is an instance of {{XRBoundedReferenceSpace}}, let |offsetSpace| be a new {{XRBoundedReferenceSpace}} and set |offsetSpace|'s {{boundsGeometry}} to |base|'s {{boundsGeometry}}, with each point transformed by the {{XRRigidTransform/inverse}} of |originOffset|.
+  1. Else if |base| is an instance of {{XRUnboundedReferenceSpace}}, let |offsetSpace| be a new {{XRUnboundedReferenceSpace}}.
+  1. Else let |offsetSpace| be a new {{XRReferenceSpace}}.
+  1. Set |offsetSpace|'s [=native origin=] to |base|'s [=native origin=].
+  1. Set |offsetSpace|'s [=origin offset=] to the result of [=multiply transforms|multiplying=] |originOffset| by |base|'s [=origin offset=].
```
NellWaliczek: Someone not me should check that this math is in the right order?

index.bs
```diff
@@ -853,7 +863,9 @@ Note: Other XR platforms sometimes refer to the type of tracking offered by a {{
...
+The polygonal boundary is given as an array of {{DOMPointReadOnly}}s, which represents a loop of points at the edges of the safe space. The points describe offsets from the {{XRReferenceSpace}} origin in meters.
+
+Relative to the {{XRBoundedReferenceSpace}}'s [=native origin=], points MUST be given in a clockwise order as viewed from above, looking towards the negative end of the Y axis. The {{DOMPointReadOnly/y}} value of each point MUST be <code>0</code> and the {{DOMPointReadOnly/w}} value of each point MUST be <code>1</code>. The bounds can be considered to originate at the floor and extend infinitely high. The shape it describes MAY be convex or concave.
```
NellWaliczek: These changes don't seem like they belong in this PR?

index.bs
Manishearth: It's correct, here `new * old` will provide a transformation where `old` is applied first and `new` second.

index.bs
toji: Comments in https://github.com/immersive-web/webxr/pull/587#discussion_r279622789 may indicate that it's better to land this one first?

index.bs
toji: Fair enough. Removed.

toji:
Addressed the feedback about the unnecessary new text, and the other two comments appear to be potential non-issues? Please take another look.

PR #610, opened 2019-04-26
Explicitly specify how the views array is populated

Manishearth:
Moved from https://github.com/immersive-web/editor-collab/pull/42

fixes https://github.com/immersive-web/webxr/issues/565

r? @NellWaliczek

PR #614, opened 2019-04-27
Explicitly specify how the views array is populated

### LINKED ISSUE #565: getViewerPose()s use of XRSpace doesn't quite specify how views work

Manishearth:
Moved from https://github.com/immersive-web/webxr/pull/610 which got accidentally closed due to a branch being deleted

fixes #565

r? @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -389,6 +389,8 @@ enum XREnvironmentBlendMode {
 
 Each {{XRSession}} has a <dfn for="XRSession">mode</dfn>, which is one of the values of {{XRSessionMode}}.
 
+The {{XRSession/viewerSpace}} on an {{XRSession}} has a <dfn for="XRSession/viewerSpace">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the underlying device.
+
 <div class="algorithm" data-algorithm="initialize-session">
 
 To <dfn>initialize the session</dfn>, given |session| and |mode|, the user agent MUST run the following steps:
@@ -687,6 +689,16 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method 
   1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Let |pose| be a new {{XRViewerPose}} object.
   1. [=Populate the pose=] of |session|'s {{XRSession/viewerSpace}} in |referenceSpace| at the time represented by |frame| into |pose|.
+  1. If |pose| is <code>null</code> return <code>null</code>.
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/viewerSpace/list of views=] on the {{XRSession/viewerSpace}} of {{XRFrame/session}}, perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|'s {{XRView/eye}} to |view|'s [=view/eye=]
+      1. Initialize |xrview|'s {{XRView/projectionMatrix}} to |view|'s [=view/projection matrix=]
+      1. Let |offset| be an {{XRRigidTransform}} equal to the [=view offset=] of |view|
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply transforms|multiplying=] the |offset| transform by the {{XRViewerPose}}'s {{XRPose/transform}}
+      1. [=list/Append=] |xrview| to |xrviews|
+  1. Set |pose|'s {{XRViewerPose/views}} to |xrviews|
   1. Return |pose|.
 
 </div>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -687,6 +689,16 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
+  1. If |pose| is <code>null</code> return <code>null</code>.
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/viewerSpace/list of views=] on the {{XRSession/viewerSpace}} of {{XRFrame/session}}, perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|'s {{XRView/eye}} to the appropriate eye the |view| is expected to render to. For views that have no such associated eye, this defaults to {{XREye/left}}.
+      1. Initialize |xrview|'s {{XRView/projectionMatrix}} to a [=matrix=] describing the projection to be used when rendering the [=view=], provided by the underlying XR system.
+      1. Let |offset| be an {{XRRigidTransform}} equal to the [=view offset=] of |view|
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply transforms|multiplying=] the |offset| transform by the {{XRViewerPose}}'s {{XRPose/transform}}
+      1. [=list/Append=] |xrview| to |xrviews|
+  1. Set |pose|'ss {{XRViewerPose/views}} to |xrviews|
```
toji: |pose|'ss -> |pose|'s

index.bs
```diff
@@ -687,6 +689,16 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
   1. If |frame|'s [=animationFrame=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Let |pose| be a new {{XRViewerPose}} object.
   1. [=Populate the pose=] of |session|'s {{XRSession/viewerSpace}} in |referenceSpace| at the time represented by |frame| into |pose|.
+  1. If |pose| is <code>null</code> return <code>null</code>.
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/viewerSpace/list of views=] on the {{XRSession/viewerSpace}} of {{XRFrame/session}}, perform the following steps:
```
toji: To prevent this algorithm from getting too long and hard to follow, I think it would be good to move this for loop out into it's own algorithm which is referenced here. (ie: "Let |xrviews| be the [=list of XRViews=] for |frame|." or similar)

index.bs
```diff
@@ -687,6 +689,16 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|'s {{XRView/eye}} to the appropriate eye the |view| is expected to render to. For views that have no such associated eye, this defaults to {{XREye/left}}.
+      1. Initialize |xrview|'s {{XRView/projectionMatrix}} to a [=matrix=] describing the projection to be used when rendering the [=view=], provided by the underlying XR system.
```
toji: Can we move the description of the eye and projection matrix down to the section defining a view and then link them here? That way the algorithm stays terse and the definition can be more expressive if needed. (The way |offset| is handled is perfect, basically.)

index.bs
Manishearth: This was initially the case, Nell asked me to inline it :smile:

NellWaliczek: Slight misunderstanding.  I'm happy to have the view population part be a separate algorithm.  I wanted the looping to be done here, though.

NellWaliczek: Oh wait, I mixed up two comments.  Yeah, I'm not super opinionated on this one.  It just didn't seem worth having a separate algorithm that wasn't ever invoked from more than one place.

toji: LOL. Sorry, don't want to ping-pong you between editors feedback. Just leave it as-is.

Manishearth:
Updated!

(I'm okay with re-splitting out the initialize algorithm, or splitting out just the loop part if y'all want it)

index.bs
```diff
@@ -389,6 +389,8 @@ enum XREnvironmentBlendMode {
 
 Each {{XRSession}} has a <dfn for="XRSession">mode</dfn>, which is one of the values of {{XRSessionMode}}.
 
+The {{XRSession/viewerSpace}} on an {{XRSession}} has a <dfn for="XRSession/viewerSpace">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the underlying device.
```
NellWaliczek: > underlying device

Don't we have a variable that defines this?

index.bs
```diff
@@ -687,6 +689,16 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
...
   1. Let |pose| be a new {{XRViewerPose}} object.
   1. [=Populate the pose=] of |session|'s {{XRSession/viewerSpace}} in |referenceSpace| at the time represented by |frame| into |pose|.
+  1. If |pose| is <code>null</code> return <code>null</code>.
+  1. Let |xrviews| be an empty [=/list=].
+  1. For each [=view=] |view| in the [=XRSession/viewerSpace/list of views=] on the {{XRSession/viewerSpace}} of {{XRFrame/session}}, perform the following steps:
+      1. Let |xrview| be a new {{XRView}} object.
+      1. Initialize |xrview|'s {{XRView/eye}} to |view|'s [=view/eye=]
+      1. Initialize |xrview|'s {{XRView/projectionMatrix}} to |view|'s [=view/projection matrix=]
+      1. Let |offset| be an {{XRRigidTransform}} equal to the [=view offset=] of |view|
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply transforms|multiplying=] the |offset| transform by the {{XRViewerPose}}'s {{XRPose/transform}}
```
NellWaliczek: Someone not me should confirm this shouldn't be an inverse...

index.bs
```diff
@@ -877,7 +889,15 @@ Views {#views}
...
+
+A [=view=] has an associated internal <dfn>view offset</dfn>, which is an {{XRRigidTransform}} describing the position and orientation of the [=view=] in the {{XRSession/viewerSpace}}'s [=coordinate system=].
+
+A [=view=] has an associated <dfn for="view">projection matrix</dfn>which is a [=matrix=] describing the projection to be used when rendering the [=view=], provided by the underlying XR device. The [=view/projection matrix=] MAY include transformations such as shearing that prevent the projection from being accurately described by a simple frustum.
+
+A [=view=] has an associated <dfn for="view">eye</dfn>which is an {{XREye}} describing which eye this view is expected to be shown to. If the view does not have an intrinsically associated eye (the display is monoscopic, for example) this value MUST be set to {{XREye/"left"}}.
```
NellWaliczek: > If the view does not have an intrinsically associated eye (the display is monoscopic, for example) this value MUST be set to {{XREye/"left"}}.

We probably don't want CAVE systems reporting "left" for all n views.  Can you please file a separate issue to address this?

index.bs

index.bs
Manishearth: https://github.com/immersive-web/webxr/issues/620

index.bs
Manishearth: @toji or @klausw ?

index.bs
NellWaliczek: Thanks!

index.bs
toji: I'm gonna go out on a limb here and say that I'm *pretty sure* that this shouldn't be an inverse (unlike the handling of the bounds geometry, which did need to be.)

Gonna merge this now, and accept public shaming from @klausw if I'm wrong.

Manishearth: I'm more concerned about the multiplication order than whether it should be an inverse, but I'm reasonably sure about both.

PR #616, opened 2019-04-29
Remove gravity mention

Manishearth:
Brought up by @blairmacintyre in https://github.com/immersive-web/webxr/pull/609#discussion_r279159519

To repeat the concerns there: This makes less sense when in a vehicle, and even less sense in space. It's adequate to define the Y axis as being a deliberately vague notion of "up", since that's more tied to user perception rather than physical concepts. That should be enough to get people to match implementations

(The orientation of the Y axis of natural origins isn't observable if it's consistent across origins anyway, but it's good to avoid overspeccing where unnecessary)

r? @toji


toji:
Gonna merge this given that Nell's time has been at a premium recently and I doubt that the scope of this change is one that will concern her much.

PR #618, opened 2019-04-30
Remove non-object [SameObject]s

Manishearth:
These are bools and enums which aren't objects, `[SameObject]` will not work on them.

r? @toji


toji:
Ooh! Good catch. Thanks!

PR #621, opened 2019-05-01
Explicitly spec out native origins

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/597 and #581

Moved from https://github.com/immersive-web/editor-collab/pull/38, totally rewritten

r? @toji @NellWaliczek

PR #624, opened 2019-05-02
Change inputSources getter from method to attrib.

### LINKED ISSUE #608: Specify live-ness of XRInputSources
Just got done consulting with @bfgeek on the subject of whether or not `XRInputSource`s and their child attributes (such as `gamepad`) should be "live" on not. That is, should an `XRInputSource` that you get from the `getInputSources()` call on one frame be updates with new values on the next frame (ie: live), or should you have call `getInputSources()` again which will return a new object representing the same input source but with new data. (ie: a snapshot). This is an issue that came up during Chrome implementation work and one that ended up biting the Gamepad API pretty hard a few years back, so it's in our interest to specify it now.

The general sentiment is that it's a bit weird to have live objects returned from a getter function, which is the way we currently have it. On the web platform getters typically should return snapshots, as that helps alleviate lifetime issues for the objects. If we want to represent the objects as live ones that update in-place, it's recommended that we have the accessor be an attribute rather than a method, since that helps imply that you'll be getting back the same objects with each access.

Given this, I think we need to clarify the behavior of `XRInputSource`s in one of three ways:

 1. Change the `getInputSources()` method to a `inputSources` attribute and specify that the objects in the array are updated in place.
 2. Leave `getInputSources()` as a method and specify that it returns a snapshot of the input state which does not subsequently update.
 3. Leave `getInputSources()` as a method but specify that it returns a live array with big flashing red text in the spec so that nobody misses it, and also go to great pains to describe exactly how those returned values lifetime corresponds to the lifetime of the session object.

I have a preference for 1, but would be fine with 2 if the rest of the working group wanted it. I will accept 3 only if you have sufficient bribe money available for immediate deposit.

toji:
/Fixes #608

Changes the `getInputSources()` method on `XRSesssion` to a readonly
attribute `inputSources`. This better reflects the expected behavior
of the returned array being a live object that is updated in-place as
input sources are added and removed.

### DIFF
--- a/input-explainer.md
+++ b/input-explainer.md
@@ -31,10 +31,10 @@ In addition to a targeting ray, all input sources provide a mechanism for the us
 ## Basic usage
 
 ### Enumerating input sources
-Calling the `getInputSources()` function on an `XRSession` will return a list of all `XRInputSource`s that the user agent considers active. The properties of an `XRInputSource` object are immutable. If a device can be manipulated in such a way that these properties can change, the `XRInputSource` will be removed from the array and a new entry created.
+The `inputSources` attribute on an `XRSession` will return a list of all `XRInputSource`s that the user agent considers active. The properties of an `XRInputSource` object are immutable. If a device can be manipulated in such a way that these properties can change, the `XRInputSource` will be removed from the array and a new entry created.
 
 ```js
-let inputSources = xrSession.getInputSources();
+let inputSources = xrSession.inputSources;
 ```
 
 When input sources are added to or removed from the list of available input sources the `inputsourceschange` event must be fired on the `XRSession` object to indicate that any cached copies of the list should be refreshed. In addition, the `inputsourceschange` event will also fire once after the session creation callback completes. This event is of the type `XRInputSourceChangeEvent` and will contain three attributes: `session` is associated session being changed, `added` is the new input sources, and `removed` is the input sources that will no longer be reported.
@@ -50,7 +50,7 @@ function onSessionStarted(session) {
 
 let xrInputSources = null;
 function onInputSourcesChange(event) {
-  xrInputSources = event.session.getInputSources();
+  xrInputSources = event.session.inputSources;
 }
 ```
 
@@ -92,7 +92,7 @@ function onSessionStarted(session) {
 All three events are `XRInputSourceEvent` events. When fired the event's `inputSource` attribute will contain the `XRInputSource` that produced the event. The event's `frame` attribute will contain a valid `XRFrame` that can be used to call `getPose()` at the time the selection event occurred. Calling the frame's `getViewerPose()` method will throw an `InvalidState` error. (The viewer's pose can still be queried by passing an `XRReferenceSpace` of type `viewer` to `XRFrame.getPose()`.)
 
 ### Transient input sources
-Some input sources are only be added to the list of input sources while an action is occurring. For example, those with an `targetRayMode` of 'screen' or those with `targetRayMode` of 'gaze' which are triggered by a voice command. In these cases, `XRInputSource` is only present in the array returned by `getInputSources()` during the lifetime of the action. In this circumstance, the order of events is as follows:
+Some input sources are only be added to the list of input sources while an action is occurring. For example, those with an `targetRayMode` of 'screen' or those with `targetRayMode` of 'gaze' which are triggered by a voice command. In these cases, `XRInputSource` is only present in the `inputSources` array during the lifetime of the action. In this circumstance, the order of events is as follows:
 1. Optionally `pointerdown`
 1. `inputsourceschange` for add
 1. `selectstart`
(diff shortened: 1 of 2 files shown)


toji:
@AdaRoseCannon: Is the magical /fixes syntax wired up for comments yet, and if so did I fail to invoke it properly above?

Manishearth:
> live object that is updated in-place as input sources are added and removed.

I don't think this can be a FrozenArray in such a case. May need a new interface.

toji:
My understanding was that FrozenArrays could be updated by the UA, and that the "frozen" aspect simply reflected that the developer couldn't alter it as they could other arrays, but that would be good to verify.

toji:
At the very least The WebHID API thinks it's possible. :)

https://wicg.github.io/webhid/index.html#hidcollectioninfo

Manishearth:
> A frozen array type is a parameterized type whose values are references to objects that hold a fixed length array of unmodifiable values. The values in the array are of type T.

I'm reading this as the array length being fixed from all sides, but the values can be internally modified (but not by users)

toji:
Ah, that might be right. That would be a meaningful limitation in this case.

Manishearth:
@heycam can you confirm my reading of how FrozenArray works?

ddorwin:
I'm not sure what /Fixes is supposed to do, but if you put "Fix #..." in the summary, GitHub will automatically close the issue when the PR lands. I think this also causes a silent message in the issue that it is referenced by this PR.

There are other syntaxes too. While not required, "Fix #abc: description" is a nice format.

heycam:
That's right, `FrozenArray<T>` means a JS `Array` object that has been frozen in the same way that calling `Object.freeze()` does.  So the values cannot cannot be changed, even by the engine.

https://heycam.github.io/webidl/#es-frozen-array

Manishearth:
I guess we need a custom indexable/iterable interface here.

(We can explicitly leave it undefined as to what the behavior is if the list of sources gets too large)

toji:
Bleh. Thanks @heycam for the confirmation. I'd rather not get into defining new, custom sequence types in this spec. Perhaps simply leaving it as a getter method is the right move here, with some added clarity about the liveness of the `XRInputSource`s themselves. I'll try to consult with @bfgeek about this as well to further evaluate the API ergonomics.

Manishearth:
Yeah, that feels best. It also fixes the iteration problem: when you call getInputSources you get a view into what the input sources were at the time, and you call it whenever you want an up to date view (maybe with an event handler for letting you know the list has changed)

This does potentially mean that we need to add a "disconnected" state/event to XRInputSource and specify what it does when the source is disconnected.

toji:
Possibly. I'm hesitant to add a connected/disconnected state simply because I've seen that it's been a fairly confusing thing for implementers and developers with the Gamepad API, but I'm not against it if has a clear benefit for developers and we can be really crisp about it's definition.

Manishearth:
I care about this more from the point of view that we should specify how an `XRInputSource` behaves when disconnected, without necessarily exposing this state to users.

e.g. does getPose() start failing on input spaces? is it frozen? can it be either?

toji:
Put up a pull request that covers this and #625 to better facilitate discussion. (Assuming that it will need rebasing on top of #621, but that shouldn't change it's substance much.)

toji:
Consulted with @bfgeek about this and he clarified the "custom indexable/iterable interface" option that @Manishearth mentioned earlier and which I had misunderstood. (Sorry!) For those of you who, like me, were not familiar with the pattern it's something that apparently shows up semi-commonly in web specs and has well understood IDL semantics. Changing to a custom iterable type in our IDL would look something like this: 

```webidl
interface XRInputSourceArray {
  iterable<XRInputSource>;
  readonly attribute unsigned long length;
  getter XRInputSource(unsigned long index);
};

partial interface XRSession {
  [SameObject] readonly attribute XRInputSourceArray inputSources;
}
```

Which isn't quite as bad as what I had imagined. It's still not ideal to have to define the array type ourselves, but I can see that being a reasonable path here and would like feedback from other members on their preference.

Manishearth:
We might need to define how such an iterator gets invalidated or updated on input state changes. Presumably it's a frozen view over the input sources at the time it's fetched (call `getInputSources()` again if you want it updated)

But yeah, this is more or less what I was suggesting. Kinda like HTMLCollection, but more up-to-date (HTMLCollection is legacy IIRC and may not use `iterable`)

toji:
Here's an example of a more recent API that does use `iterable`: https://drafts.css-houdini.org/css-typed-om/#cssnumericarray

toji:
Updated this PR to make use of `iterable`. I'm pretty happy with this approach, but would be interested to hear if there were any concerns about going this route.

input-explainer.md
```diff
@@ -31,10 +31,10 @@ In addition to a targeting ray, all input sources provide a mechanism for the us
 ## Basic usage
 
 ### Enumerating input sources
-Calling the `getInputSources()` function on an `XRSession` will return a list of all `XRInputSource`s that the user agent considers active. The properties of an `XRInputSource` object are immutable. If a device can be manipulated in such a way that these properties can change, the `XRInputSource` will be removed from the array and a new entry created.
+The `inputSources` attribute on an `XRSession` will return a list of all `XRInputSource`s that the user agent considers active. The properties of an `XRInputSource` object are immutable. If a device can be manipulated in such a way that these properties can change, the `XRInputSource` will be removed from the array and a new entry created.
```
NellWaliczek: >will return

Slightly strange phrasing.

Issue #478, opened 2019-01-15
How does the identity reference space relate to any other space?


thetuvix:
@RafaelCintron asked me a good question here that is underspecified.  What happens if an app attempts to relate the `"identity"` reference space to any other space?  I'd expect `getPose` to return null for that request, since a disconnected identity space is by definition not positioned with any stable relation to any real-world objects.  However, we should probably specify that explicitly.

_Originally posted by @thetuvix in https://github.com/immersive-web/webxr/pull/459/files_

NellWaliczek:
Fix was removed from #491 because it was inadequate

toji:
Coming back around and giving this some further thought: We probably need this to have more complex behavior than simply returning null for everything other than `getViewerPose`. The reason being that an `identity` reference space will be used primarily for `inline` sessions and in that context we still want to be able to properly process `screen` input sources so that click and tap interactions can continue using a unified model. That means the `targetRaySpace` needs to give a sensible result.

So that's a bit harder to explain cleanly than "It's null for pretty much everything." We could carve out screen input sources as a singular exception, I guess, but I'm not sure if that's sufficient. Alternately we could say that it should return all other poses relative to the viewer, but that actually means you could reasonably produce tracking values for pretty much everything, which grinds against the permissions model we want to have for `identity`, which is that it doesn't require permissions.

Manishearth:
> Alternately we could say that it should return all other poses relative to the viewer, but that actually means you could reasonably produce tracking values for pretty much everything

Not necessarily, to query any space that can be related to identity in a way that requires tracking (e.g. `eye-level`) you'd need tracking permissions _anyway_, yes?

See also: https://github.com/immersive-web/webxr/issues/565 , where I was initially confused as to what identity actually _was_ since I assumed you could relate it to other spaces.

cwilso:
Should be fixed by the fix for #597

Manishearth:
See also: https://github.com/immersive-web/webxr/issues/623

PR #626, opened 2019-05-03
Simplify available spaces

### LINKED ISSUE #623: Should identity and viewerSpace be merged?
Brought up by @NellWaliczek in https://github.com/immersive-web/webxr/pull/621#discussion_r280544401

See also: https://github.com/immersive-web/webxr/issues/478

`identity` reference spaces and `session.viewerSpace` are both reference spaces with a native origin that matches the viewer/headset pose.

The differences are:

 - `viewerSpace` is not a reference space and cannot be fed to `getViewerPose()`. It also does not have an origin offset
 - `identity` is (currently) restricted as to what spaces it can be related against in `getPose()`, currently you can only relate it against `viewerSpace` and `screen-space` targetRaySpace`s. This is for privacy/permissions reasons.

It should be possible to merge them. [I don't think it is necessary to restrict `identity`'s ability to relate to other spaces for _permissions_ reasons:](https://github.com/immersive-web/webxr/issues/478#issuecomment-483840400) the only way relating an identity space with other spaces can leak pose information is if you have another space that is not tied to the user's pose, and we can permissions-gate those anyway. Any information someone wants to leak from `identity` can be leaked through `viewerSpace` anyway, so whatever we design for permissions needs to be able to handle both.

One non-permissions concern of merging the two is that the concept of an identity space makes less sense in non-screen-space experiences, it's only an identity when related to the viewer pose (ignoring origin offsets).

A potential way of fixing that is to replace "identity" with a "viewer" reference space. This space can be related to everything, and it now has an origin offset. `XRSession.viewerSpace` returns a "viewer" reference space with no offset (for convenience). This seems conceptually clearer to me, and it keeps all the capabilities both spaces have right now.

### LINKED ISSUE #625: Does the "position-disabled" reference space still make physical sense?
Now that we're settling in on clean physically-based reference space definitions (which is great, especially for ensuring the associativity of space relationships!), I wonder if "position-disabled" reference spaces will operate as we intend for their 360-degree video scenarios when on stereo headsets.

Under this definition, `getViewerPose` would still return non-identity eye positions for each `XRView` relative to this reference space, since the eyes are indeed in physically different places.  That would cause the app to render each eye from some pose other than (0, 0, 0), leading to weird parallax artifacts relative to its finite-sized video sphere, which is the primary problem that "position-disabled" was trying to avoid.  I'd really hate to see us backpedal on the physically-based definitions and introduce some special case to `getViewerPose` for "position-disabled" spaces to magically zero out the `XRView` poses, since we should otherwise be fully clean on reference spaces behaving identically, just with differently-behaving origins.

Alternatively, **what if we just replaced the "position-disabled" reference space with a `disablePosition` option on `getViewerPose`**?  360-video apps could then just use the "eye-level" reference space like other seated experiences, with their one difference being a parameter they pass to one function whose behavior they actually wish to impact.  By adding the option to the affected method, all other use of reference spaces stays clean and easy to understand.

The simplest option here would just be to require that 360-video apps use a big enough video sphere (e.g. 1000m) that these parallax issues from neck modeling and stereo separation in "eye-level" spaces would fall away.  If we do that, **we could just cut the "position-disabled" space entirely**.  Perhaps in the interest of getting quickly to WebXR 1.0, that's the best answer?

_Originally posted by @thetuvix in https://github.com/immersive-web/webxr/pull/621/review_comment/create_

toji:
/Fixes #625
/Fixes #623 

(Happy to split these two out, but it felt like it would be clearer to view
both changes at once.)

Removes the position-disabled reference space from the spec and
explainers. Developers that wish to display the types of content this
reference space supported (primarily 360 media) should instead manually
zero-out the position of the XRView transforms when rendering.

Also consolidates the `viewerSpace` attribute and the `identity` reference
space into a single new reference space called `viewer`. This new
space is functionally equivalent to the previous `viewerSpace` but
fetched via `requestReferenceSpace()` with the same guarantee of
availability as `identity` had previously. All other spaces relate to it
like any other space.


toji:
I think this also fixes #478.

index.bs
```diff
@@ -792,9 +784,7 @@ An {{XRReferenceSpace}} is most frequently obtained by calling {{XRSession/reque
...
 
-Devices that support any of the following {{XRReferenceSpaceType}}s MUST support all three: {{XRReferenceSpaceType/position-disabled}}, {{XRReferenceSpaceType/eye-level}}, {{XRReferenceSpaceType/floor-level}}. The minimum sensor data necessary for supporting all three reference spaces is the same.
-
-Note: The {{position-disabled}} type is primarily intended for use with pre-rendered media such as panoramic photos or videos. It should not be used for most other media types due to user discomfort associated with the lack of a neck model or full positional tracking.
+Devices that support {{XRReferenceSpaceType/eye-level}} reference spaces MUST support {{XRReferenceSpaceType/floor-level}} reference spaces, through emulation if necessary, and visa-versa. The minimum sensor data necessary for supporting both reference spaces is the same.
```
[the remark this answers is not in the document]

spatial-tracking-explainer.md
```diff
@@ -147,20 +147,19 @@ function onSessionStarted(session) {
 }
 ```
 
-#### Position-disabled reference spaces
+### Viewer reference spaces
+A _viewer_ reference space tracks the origin of the viewer. This type of reference space is primarily used for creating inline experiences with no tracking information. Instead, developers may use `XRReferenceSpace.originOffset` which is described in the [Application supplied transforms section](#application-supplied-transforms). An example usage of an _viewer_ reference space is a furniture viewer that will use [click-and-drag controls](#click-and-drag-view-controls) to rotate the furniture. It also supports cases where the developer wishes to avoid displaying any type of tracking consent prompt to the user prior while displaying inline content.
```
NellWaliczek: >A _viewer_ reference space tracks the origin of the viewer.

This phrasing sounds a bit strange now... I know this is the explainer, but maybe something about the origin of the viewer reference space always being at the viewer's position and orientation?

NellWaliczek: Especially when the next sentence immediately says "no tracking information".

NellWaliczek: > Instead, developers may use `XRReferenceSpace.originOffset`

Also we now have a problem.  Some of the spec is the official design and some of the explainer is the official design.  Where as before we could always say the explainer was the most up to date... how do we want to deal with this? File issues against the explainer?

spatial-tracking-explainer.md
```diff
@@ -147,20 +147,19 @@ function onSessionStarted(session) {
...
-
-Some example use cases: 
-* 360 photo/video viewer
+This type of reference space is requested with a type of `viewer` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have a `transform` that is equal to the reference space's `originOffset` and the `XRView` matrices will be offset accordingly.
 
 ```js
 let xrSession = null;
 let xrReferenceSpace = null;
 
+// Create an 'viewer' reference space
```
NellWaliczek: typo "an"

spatial-tracking-explainer.md
```diff
@@ -171,27 +170,13 @@ function onSessionStarted(session) {
...
-  xrSession.requestReferenceSpace('identity')
-  .then((referenceSpace) => {
-    xrReferenceSpace = referenceSpace;
-  })
-  .then(setupWebGLLayer)
-  .then(() => {
-    xrSession.requestAnimationFrame(onDrawFrame);
-  });
-}
+  let pose = xrFrame.getPose(preferredInputSource.gripSpace, xrSession.viewerSpace);
```
NellWaliczek: xrSession.viewerSpace doesn't exist with this PR anymore, right?

spatial-tracking-explainer.md
```diff
@@ -289,10 +264,10 @@ inlineCanvas.addEventListener('pointermove', onPointerMove);
...
+Inline sessions, by definition, do not require a user gesture or user permission to create, and as a result there must be strong limitations on the pose data that can be reported for privacy and security reasons. Requests for `identity` reference spaces will always succeed. Requests for a `bounded` or an `unbounded` reference space will always be rejected on inline sessions. Requests for an `eye-level` or `floor-level` reference space may succeed but may also be rejected if the UA is unable provide any tracking information such as for an inline session on a desktop PC or a 2D browser window in a headset. The UA is also allowed to request the user's consent prior to returning an `eye-level` or `floor-level` reference space.
```
NellWaliczek: > Requests for `identity` reference spaces will always succeed.

'viewer'

spatial-tracking-explainer.md
```diff
@@ -289,10 +264,10 @@ inlineCanvas.addEventListener('pointermove', onPointerMove);
...
 
 ### Ensuring hardware compatibility
-Immersive sessions will always be able to provide `eye-level`, `floor-level`, and `position-disabled` reference spaces, but may not support other `XRReferenceSpace` types due to hardware limitations.  Developers are strongly encouraged to follow the spirit of [progressive enhancement](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) and provide a reasonable fallback behavior if their desired `bounded` or `unbounded` reference space is unavailable.  In many cases it will be adequate for this fallback to behave similarly to an inline preview experience.
+Immersive sessions will always be able to provide `eye-level` and `floor-level` reference spaces, but may not support other `XRReferenceSpace` types due to hardware limitations.  Developers are strongly encouraged to follow the spirit of [progressive enhancement](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) and provide a reasonable fallback behavior if their desired `bounded` or `unbounded` reference space is unavailable.  In many cases it will be adequate for this fallback to behave similarly to an inline preview experience.
```
NellWaliczek: >Immersive sessions will always be able to provide `eye-level` and `floor-level` 

Should be

> Immersive sessions will always be able to provide `viewer`,`eye-level`, and `floor-level`

spatial-tracking-explainer.md
```diff
@@ -370,7 +345,6 @@ How to pick a reference space:
 | Type                | Examples                                      |
 | ------------        | --------------------------------------------- |
 | `identity`          | - In-page content preview<br>- Click/Drag viewing |
```
NellWaliczek: `viewer`

spatial-tracking-explainer.md
```diff
@@ -461,7 +432,6 @@ interface XRPose {
 
 enum XRReferenceSpaceType {
   "identity",
```
NellWaliczek: `viewer`

spatial-tracking-explainer.md
```diff
@@ -387,7 +361,6 @@ How to pick a reference space:
 | Type                | Inline             | Immersive  |
 | ------------        | ------------------ | ---------- |
 | `identity`          | Guaranteed         | Guaranteed |
```
NellWaliczek: viewer

index.bs
```diff
@@ -780,9 +774,7 @@ Each {{XRReferenceSpace}} has a <dfn for="XRReferenceSpace">type</dfn>, which is
...
-
-  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">position-disabled</dfn> creates an {{XRReferenceSpace}} instance.  It represents a tracking space with a [=native origin=] that tracks the position of the [=viewer=], with a constant orientation initialized based on the conventions of the underlying [=/XR device=]. It can be used to create experiences where the movement of the user is disregarded, as it will not report changes in position when used with {{XRFrame/getViewerPose()}}.
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">viewer</dfn> creates an {{XRReferenceSpace}} instance representing the {{XRSession}}'s [=XRSession/viewer space=].
```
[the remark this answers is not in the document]
Manishearth: Yeah I figured it would go the other way too. Perhaps talk about it explicitly as having an native origin matching that of the [=viewer=], and then define viewerSpace as returning a {{XRReferenceSpaceType/viewer}} reference space with zero offset.

spatial-tracking-explainer.md
toji: Boy, I'm just failing at this search and replace thing left and right...

spatial-tracking-explainer.md
```diff
@@ -92,9 +92,9 @@ function onSessionStarted(session) {
...
+It is important to note that `XRViewerPose` objects retrieved using `floor-level` and `eye-level` reference spaces may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that `floor-level` and `eye-level` experiences should be resilient to position changes despite not being dependent on receiving them.
```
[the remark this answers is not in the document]
toji: Given that this is referring to the code using the API and not the API itself, we have no authority to enforce a MUST here. I'm open, however, to petitioning the W3C to change that! Would make my life a hell of a lot easier. ;)

spatial-tracking-explainer.md
toji: Correct, thanks.

index.bs
```diff
@@ -792,9 +784,7 @@ An {{XRReferenceSpace}} is most frequently obtained by calling {{XRSession/reque
...
 
-Devices that support any of the following {{XRReferenceSpaceType}}s MUST support all three: {{XRReferenceSpaceType/position-disabled}}, {{XRReferenceSpaceType/eye-level}}, {{XRReferenceSpaceType/floor-level}}. The minimum sensor data necessary for supporting all three reference spaces is the same.
-
-Note: The {{position-disabled}} type is primarily intended for use with pre-rendered media such as panoramic photos or videos. It should not be used for most other media types due to user discomfort associated with the lack of a neck model or full positional tracking.
+Devices that support {{XRReferenceSpaceType/eye-level}} reference spaces MUST support {{XRReferenceSpaceType/floor-level}} reference spaces, through emulation if necessary, and vice versa. The minimum sensor data necessary for supporting both reference spaces is the same.
```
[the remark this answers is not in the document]
toji: Dropped the second sentence. Feels overly repetitive when we're only talking about the two types anyway.

toji:
Updated to address all applicable feedback. Please take another look!

spatial-tracking-explainer.md
```diff
@@ -92,7 +92,7 @@ function onSessionStarted(session) {
...
 
 ### Stationary experiences
-A _stationary_ experience is one which does not require the user to move around in space.  This includes several categories of experiences that developers are commonly building today, and multiple reference space types are available to address each class of experience.  "Standing" experiences can be created with a `floor-level` reference space.  "Seated" experiences can be created with an `eye-level` reference space.
+A _stationary_ experience is one which does not require the user to move around in space.  This includes several categories of experiences that developers are commonly building today, and multiple reference space types are available to address each class of experience.  "Standing" experiences can be created with a `floor-level` reference space.  "Seated" experiences can be created with an `eye-level` reference space. Orientation-only experiences such as 360 photo/video viewers can also be created with an `eye-level` reference space, though in that case positional data should be ignored in order to display the media properly.
```
NellWaliczek: >though in that case positional data should be ignored in order to display the media properly.

Is this actually the guidance?  Thought it was to render the content on sufficiently far away surrounding geometry....

spatial-tracking-explainer.md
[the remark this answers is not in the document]
NellWaliczek: Ok, fine. ;). How about "must" then?

spatial-tracking-explainer.md
```diff
@@ -148,20 +148,20 @@ function onSessionStarted(session) {
...
+A _viewer_ reference space's origin is always at the position and orientation of the viewer device. This type of reference space is primarily used for creating inline experiences with no tracking of the viewer relative to it's physical environment. Instead, developers may use `XRReferenceSpace.originOffset` which is described in the [Application supplied transforms section](#application-supplied-transforms). An example usage of an _viewer_ reference space is a furniture viewer that will use [click-and-drag controls](#click-and-drag-view-controls) to rotate the furniture. It also supports cases where the developer wishes to avoid displaying any type of tracking consent prompt to the user prior while displaying inline content.
```
NellWaliczek: > A _viewer_ reference space's origin is always at the position and orientation of the viewer device.

Eh... now this sounds in conflict with the behavior of the originOffset.

index.bs
```diff
@@ -492,7 +492,7 @@ The <dfn attribute for="XRSession">environmentBlendMode</dfn> attribute returns
 
 NOTE: Most Virtual Reality devices exhibit {{XREnvironmentBlendMode/opaque}} blending behavior. Augmented Reality devices that use transparent optical elements frequently exhibit {{XREnvironmentBlendMode/additive}} blending behavior, and Augmented Reality devices that use passthrough cameras frequently exhibit {{XREnvironmentBlendMode/alpha-blend}} blending behavior.
 
-Each {{XRSession}} has a <dfn for="XRSession">viewer space</dfn>, which is an {{XRReferenceSpace}} with a [=native origin=] which tracks the position and orientation of the [=viewer=]. The [=XRSession/viewer space=] has a <dfn for="XRSession/viewer space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=].
+Each {{XRSession}} has a <dfn for="XRSession">viewer space</dfn>, which is an {{XRReferenceSpace}} of type {{XRReferenceSpaceType/viewer}} with an [=identity transform=] [=XRSpace/origin offset=]. The [=XRSession/viewer space=] has a <dfn for="XRSession/viewer space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=].
```
NellWaliczek: Should we be calling it a "viewer reference space"? It'd be so long that way....

spatial-tracking-explainer.md
NellWaliczek: >`XRReferenceSpace.originOffset`

Are we planning to fixing the explainer text about originOffset in a separate PR?

Manishearth:
```suggestion
A _viewer_ reference space's native origin is always at the position and orientation of the viewer device. This type of reference space is primarily used for creating inline experiences with no tracking of the viewer relative to it's physical environment. Instead, developers may use `XRReferenceSpace.originOffset` which is described in the [Application supplied transforms section](#application-supplied-transforms). An example usage of an _viewer_ reference space is a furniture viewer that will use [click-and-drag controls](#click-and-drag-view-controls) to rotate the furniture. It also supports cases where the developer wishes to avoid displaying any type of tracking consent prompt to the user prior while displaying inline content.
```

spatial-tracking-explainer.md
toji: Either approach works, and there's minor pros/cons to each but no so much that it really makes a difference in terms of guidance here.

spatial-tracking-explainer.md
toji: I guess the way to fix that would be to lift the concept of "native origin" into the explainer, but I'm not convinced that's necessary.

toji: Yeah, that's a large enough change that I'd feel more comfortable handling it separately from this issue.

index.bs
toji: Meh, it's not all that bad. Made the change for clarity's sake.

toji:
Okay, remainder of Nell's comments are addressed or responded to.

spatial-tracking-explainer.md
Manishearth: I guess the explainer doesn't talk of originOffset at all. Hm. I'd say we should fix this separately, if at all, and for now just say "origin" to be consistent with the rest?

NellWaliczek: Fine by me.  Can you file an issue to track that?

NellWaliczek: That's fine, though we should call out that this will need to be clarified when we fix the issue @toji will file to address originOffset in the explainer. :)

spatial-tracking-explainer.md
NellWaliczek: Then let's say something like this?

> Orientation-only experiences such as 360 photo/video viewers can also be created with an `eye-level` reference space, and in that case one method would be to ignore the positional data in order to display the media properly.

spatial-tracking-explainer.md
toji: #634

spatial-tracking-explainer.md
toji: Did one better and expanded this into it's own paragraph to give both options and explain why.

Issue #583, opened 2019-04-05
Decide what to do with Gamepad.index


toji:
One issue that we currently have with input sources is that if we have two controllers of the same type without different handedness (think Vive controllers prior to handedness estimation) and then one of them disconnects we have no way of indicating which one it was outside of tracking the pose frame to frame and trying to identify which one the remaining controller is closest to, which... 🤢

The Gamepad API addresses this by leaving holes in the gamepads array, so that a single controller remains in the same index for the lifetime of it's connection. [Developers are opinionated about this approach.](https://twitter.com/stuk/status/1107754849062412288)

I would propose instead that we should use the `Gamepad.index` attribute to uniquely identify devices within a session. The value should be something that monotonically increases with each newly connected device and doesn't actually correspond to the element index in the array at all, which allows us to avoid array gaps. Thoughts?

toji:
Edited the title of this to be a bit more generic, because while I think we could use the index as an arbitrary counter the bigger issue is simply deciding what it should be used for, period. (As pointed out by Nell in https://github.com/immersive-web/webxr/pull/628#discussion_r283589964)

cwilso:
This should just be -1.  Gamepad says index is a signed long.  (Chrome needs an impl. fix.)

ddorwin:
There seems to be a gap between @toji's https://github.com/immersive-web/webxr/issues/583#issuecomment-492079244 29 days ago and a constant value being the fix for this issue. Was the decision to not pursue those earlier ideas made somewhere else or is #690 just a related fix?

toji:
Yes, we should have made that clearer: It may be beneficial to do something along the lines of what I stated in the first comment, but we don't anticipate working that out in time for the VR complete milestone. Since it's possible that `0` will be a valid value in a future system like that, though, and almost certain that `-1` will not be, AND since the Gamepad API spec allows `-1`, we decided that it would be best to make the initial value the clearly-not-a-valid-index one.

ddorwin:
Thanks for the update. SGTM.

NellWaliczek:
A clarification about the original reason this issue was filed... the inputsourceschange event was modified during the hit-testing-explainer design and now has "added" and "removed" attributes that list the XRInputSources that were added and removed respectively.  This, combined with the XRInputSource objects now being tagged as [SameObject], should mean #690 effectively addresses this issue completely.

PR #628, opened 2019-05-06
Describe how the input source list is maintained

### LINKED ISSUE #465: Document how the list of active input sources is maintained
https://immersive-web.github.io/webxr/#list-of-active-input-sources

An XRSession has the capability of returning the list of inputs see: https://immersive-web.github.io/webxr/#dom-xrsession-getinputsources

This production and maintenance of this frozen array needs to be documented.

toji:
Fixes #465.

Defines what should happen as input sources for a session are added,
removed, and changed. Also clarifies the definitions for some of the
related concepts.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -471,7 +471,43 @@ When the <dfn method for="XRSession">requestReferenceSpace(|options|)</dfn> meth
 
 </div>
 
-The <dfn attribute for="XRSession">inputSources</dfn> attribute returns the {{XRSession}}'s [=list of active input sources=].
+Each {{XRSession}} has a <dfn>list of active XR input sources</dfn> (a [=/list=] of {{XRInputSource}}) which MUST be initially an empty [=/list=].
+
+The <dfn attribute for="XRSession">inputSources</dfn> attribute returns the {{XRSession}}'s [=list of active XR input sources=].
+
+The user agent MUST monitor any [=XR input source=]s associated with the [=/XR Device=], including detecting when [=XR input source=]s are added, removed, or changed.
+
+<div class="algorithm" data-algorithm="on-input-source-added">
+
+When a new [=XR input source=] becomes available for {{XRSession}} |session|, the user agent MUST run the following steps:
+
+  1. Let |inputSource| be a new {{XRInputSource}}.
+  1. Add |inputSource| to |session|'s [=list of active XR input sources=] associated with the new [=XR input source=].
+  1. [=Queue a task=] that fires a {{XRSessionEvent}} named {{inputsourceschange!!event}} on |session|.
+
+</div>
+
+<div class="algorithm" data-algorithm="on-input-source-removed">
+
+When an [=XR input source=] is no longer available for {{XRSession}} |session|, the user agent MUST run the following steps:
+
+  1. Let |inputSource| be the {{XRInputSource}} in |session|'s [=list of active XR input sources=] associated with the [=XR input source=] that is not longer available.
+  1. Remove |inputSource| from |session|'s [=list of active XR input sources=].
+  1. [=Queue a task=] that fires a {{XRSessionEvent}} named {{inputsourceschange!!event}} on |session|.
+
+</div>
+
+<div class="algorithm" data-algorithm="on-input-source-change">
+
+When the intrinsic properties of an [=XR input source=] change for {{XRSession}} |session|, such as the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or presence of a {{XRInputSource/gripSpace}} or {{XRInputSource/gamepad}}, the user agent MUST run the following steps:
+
+  1. Let |oldInputSource| be the {{XRInputSource}} in |session|'s [=list of active XR input sources=] previously associated with the [=XR input source=] that changed.
+  1. Let |newInputSource| be a new {{XRInputSource}}.
+  1. Remove |oldInputSource| from |session|'s [=list of active XR input sources=].
+  1. Add |newInputSource| to |session|'s [=list of active XR input sources=] associated with the [=XR input source=] that changed.
+  1. [=Queue a task=] that fires a {{XRSessionEvent}} named {{inputsourceschange!!event}} on |session|.
+
+</div>
 
 Each {{XRSession}} has a <dfn>environment blending mode</dfn> value, which is a enum which MUST be set to whichever of the following values best matches the behavior of imagery rendered by the session in relation to the user's surrounding environment.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -475,11 +475,47 @@ When the <dfn method for="XRSession">requestReferenceSpace(|options|)</dfn> meth
...
+
+</div>
+
+<div class="algorithm" data-algorithm="on-input-source-change">
+
+When the intrinsic properties of an [=XR input source=] change for {{XRSession}} |session|, such as the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or presence of a {{XRInputSource/gripSpace}} or {{XRInputSource/gamepad}}, the user agent MUST run the following steps:
+
+  1. Let |oldInputSource| be the {{XRInputSource}} in |session|'s [=list of active XR input sources=] previously associated with the [=XR input source=] that changed.
+  1. Let |newInputSource| be a new {{XRInputSource}}.
+  1. Remove |oldInputSource| from |session|'s [=list of active XR input sources=].
```
Manishearth: Should we also unset an `active` flag on the source? Make `getPose()` on inactive input source spaces returns null (or throw an error for better diagnostics?).

In other words, force people to reload input source lists whenever they change. Otherwise they'll continue using a nonexistant input space.

toji: Good suggestion! I'll incorporate that.

toji:
Pushed some changes to this PR, thanks for your patience!

> Is there a reason this PR isn't reflecting the behavior of the inputsourceschange event...

Huh. Turns out that the spec was missing the XRInputSourceChangeEvent entirely! Fixed that now, as well as making sure the algorithms populate the `added`/`removed` arrays as needed.

One minor concern: The event name is  "inputsource**s**change" while the interface, as defined in the explainer, is "XRInputSourceChangeEvent" (note the singular "Source"). We should probably normalize that, and it could easily be done in this PR. I suggest making the interface use the plural form?

> Is the expectation that left/right motion controllers will each fire their own event even if they're detected on the same frame?

Nope, that was an oversight on my part. Fixed in the latest commit.

> Are there any developer expectations we should be setting about when input sources may change?

We certainly can, though what exactly those expectations should be is not readily apparent to me at the moment and is probably best discussed in a separate issue.

index.bs
toji: Following up: This is ever-so-slightly more complex than I had originally anticipated, primarily because I think the "active" flag should actually reside on the space and not the input source. That implies that we probably want to make more consistent use of it elsewhere in the spec, so I wanted to give it some more thought before jumping into a change like that.

Manishearth: Totally okay with it being a followup, just something to keep in mind.

NellWaliczek:
Looking at the other open PRs made me realize that one really important thing is missing.  We haven't defined the expected behavior of the array when an item is added/removed.  Do gaps fill in?  What about for transient input sources?

Manishearth:
> Looking at the other open PRs made me realize that one really important thing is missing. We haven't defined the expected behavior of the array when an item is added/removed. Do gaps fill in? What about for transient input sources?

@NellWaliczek There's some discussion on that in #624. It's not touching array ordering concerns, though.

cwilso:
Also fixes #632

index.bs
```diff
@@ -486,33 +486,44 @@ The user agent MUST monitor any [=XR input source=]s associated with the [=/XR D
...
+When new [=XR input source=]s become available for {{XRSession}} |session|, the user agent MUST run the following steps:
 
-  1. Let |inputSource| be a new {{XRInputSource}}.
-  1. Add |inputSource| to |session|'s [=list of active XR input sources=] associated with the new [=XR input source=].
-  1. [=Queue a task=] that fires a {{XRSessionEvent}} named {{inputsourceschange!!event}} on |session|.
+  1. Let |added| be a new [=/list=].
+  1. For each new [=XR input source=]:
+    1. Let |inputSource| be a new {{XRInputSource}}.
+    1. Add |inputSource| to |added|.
+    1. Add |inputSource| to |session|'s [=list of active XR input sources=] associated with the [=XR input source=].
```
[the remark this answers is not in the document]
toji: Ooh! Good catch! Thanks!

index.bs
```diff
@@ -486,33 +486,44 @@ The user agent MUST monitor any [=XR input source=]s associated with the [=/XR D
...
+    1. Add |inputSource| to |removed|.
+    1. Remove |inputSource| from |session|'s [=list of active XR input sources=].
+  1. [=Queue a task=] that fires a {{XRInputSourceChangeEvent}} named {{inputsourceschange!!event}} on |session| with {{XRInputSourceChangeEvent/removed}} set to |removed|.
 
 </div>
 
 <div class="algorithm" data-algorithm="on-input-source-change">
 
-When the intrinsic properties of an [=XR input source=] change for {{XRSession}} |session|, such as the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or presence of a {{XRInputSource/gripSpace}} or {{XRInputSource/gamepad}}, the user agent MUST run the following steps:
+When the intrinsic properties of any [=XR input source=]s change for {{XRSession}} |session|, such as the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or presence of a {{XRInputSource/gripSpace}} or {{XRInputSource/gamepad}}, the user agent MUST run the following steps:
```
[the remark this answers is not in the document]
toji: I think I started that list thinking it was going to be an example then listed them all anyway and forgot to fix the start of the sentence. 😅

toji:
Update done with a sick, sleeping 2yo on my lap. Apologies from any oversights/typos as a result.

> I support the plural form of the event :)

Done!

> We haven't defined the expected behavior of the array when an item is added/removed.

I've updated the algorithms to use some [pre-defined list operations](https://infra.spec.whatwg.org/#list-extend), which should remove some of the ambiguity around how it should behave.

index.bs
```diff
@@ -1229,7 +1232,7 @@ Gamepad API Integration {#gamepad-api-integration}
 
   - {{XRInputSource/gamepad}} MUST NOT be included in the array returned by {{navigator.getGamepads()}}.
   - {{XRInputSource/gamepad}}'s {{Gamepad/index}} attribute MUST be <code>0</code>.
```
NellWaliczek: >{{XRInputSource/gamepad}}'s {{Gamepad/index}} attribute MUST be <code>0</code>.

Gah this makes me so nervous.. people are going to use this and index into inputSources.... Can we file a separate issue (and label it to be an external dependency) to figure out how we want to handle this with the Gamepad spec folks? (yes, I'm looking at you @toji)

toji: Re-cast #583 as a broader issue to address this topic.

PR #629, opened 2019-05-06
Define event order of input sources

### LINKED ISSUE #540: Update spec text to clarify the order in which input events should fire
Based on a clarification to the explainer that was merged in #492, the order of event firing should also be updated in the actual spec text

toji:
Fixes #540

Defines the order that events fire in during the handling of transient events.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -493,11 +493,7 @@ When <dfn for="XRSession" lt="add input source">new [=XR input source=]s become 
 
 <div class="algorithm" data-algorithm="on-input-source-removed">
 
-<<<<<<< HEAD
 When any previously added <dfn for="XRSession" lt="remove input source">[=XR input source=]s are no longer available</dfn> for {{XRSession}} |session|, the user agent MUST run the following steps:
-=======
-When an <dfn for="XRSession" lt="remove input source">[=XR input source=] is no longer available</dfn> for {{XRSession}} |session|, the user agent MUST run the following steps:
->>>>>>> Define the behavior of transient inputs
 
   1. Let |removed| be a new [=/list=].
   1. For each [=XR input source=] that is no longer available:
@@ -1180,8 +1176,6 @@ interface XRInputSource {
 };
 </pre>
 
-Each {{XRInputSource}} SHOULD define a <dfn>primary action</dfn>. The [=primary action=] is a platform-specific action that, when engaged, produces {{selectstart}}, {{selectend}}, and {{XRSession/select}} events. Examples of possible [=primary action=]s are pressing a trigger, touchpad, or button, speaking a command, or making a hand gesture. If the platform guidelines define a recommended primary input then it should be used as the [=primary action=], otherwise the user agent is free to select one.
-
 The <dfn attribute for="XRInputSource">handedness</dfn> attribute describes which hand the [=XR input source=] is associated with, if any. Input sources with no natural handedness (such as headset-mounted controls or standard gamepads) or for which the handedness is not currently known MUST set this attribute {{XRHandedness/none}}.
 
 The <dfn attribute for="XRInputSource">targetRayMode</dfn> attribute describes the method used to produce the target ray, and indicates how the application should present the target ray to the user if desired.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1206,6 +1204,32 @@ The <dfn attribute for="XRInputSource">gamepad</dfn> attribute is a {{Gamepad}}
...
+
+  1. Fire any <code>"pointerdown"</code> events produced by the [=XR input source=]'s action, if necessary.
+  1. [=add input source|Add the XR input source=] to the [=list of active XR input sources=].
+  1. [=Queue a task=] that fires a {{XRInputSourceEvent}} named {{selectstart!!event}} on |session|.
+
+</div>
+
+<div class="algorithm" data-algorithm="on-transient-input-end">
+
+When a [=transient input source=] for {{XRSession}} |session| ends it's [=primary action=] the UA MUST run the following steps:
```
NellWaliczek: nit: its

index.bs
```diff
@@ -1206,6 +1204,32 @@ The <dfn attribute for="XRInputSource">gamepad</dfn> attribute is a {{Gamepad}}
 NOTE: {{XRInputSource}}s with a <code>null</code> {{gamepad}} can still fire {{select}}, {{selectstart}}, and {{selectend}} events to report binary inputs.
 </section>
 
+Transient input {#transient-input}
+-----------------------
+Some [=/XR Device=]s may support <dfn>transient input sources</dfn>, where the [=XR input source=] is only meaningful while performing it's [=primary action=]. An example would be mouse, touch, or stylus input against an {{XRSessionMode/inline}} {{XRSession}}, producing an {{XRInputSource}} with a {{targetRayMode}} set to {{screen}}. This type of [=XR input source=] would only be present in the session's [=list of active XR input sources=] for the duration of the the {{selectstart}}, {{select}}, and {{selectend}} event sequence.
```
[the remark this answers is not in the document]
NellWaliczek: Also, are we explicitly stating anywhere in the spec that mouse, touch, stylus MUST create a transient input source?  I didn't see it...?

index.bs
NellWaliczek: What about when the primary action is cancelled?

index.bs
```diff
@@ -1206,6 +1204,32 @@ The <dfn attribute for="XRInputSource">gamepad</dfn> attribute is a {{Gamepad}}
...
+
+</div>
+
+<div class="algorithm" data-algorithm="on-transient-input-end">
+
+When a [=transient input source=] for {{XRSession}} |session| ends it's [=primary action=] the UA MUST run the following steps:
+
+  1. [=Queue a task=] that fires a {{XRInputSourceEvent}} named {{select!!event}} on |session|.
+  1. Fire any <code>"click"</code> events produced by the [=XR input source=]'s action, if necessary.
+  1. [=Queue a task=] that fires a {{XRInputSourceEvent}} named {{selectend!!event}} on |session|.
```
NellWaliczek: Are we explicitly stating this event ordering for non-transient input sources anywhere in the spec?

index.bs
```diff
@@ -1206,6 +1204,32 @@ The <dfn attribute for="XRInputSource">gamepad</dfn> attribute is a {{Gamepad}}
...
 
+Transient input {#transient-input}
+-----------------------
+Some [=/XR Device=]s may support <dfn>transient input sources</dfn>, where the [=XR input source=] is only meaningful while performing it's [=primary action=]. An example would be mouse, touch, or stylus input against an {{XRSessionMode/inline}} {{XRSession}}, producing an {{XRInputSource}} with a {{targetRayMode}} set to {{screen}}. This type of [=XR input source=] would only be present in the session's [=list of active XR input sources=] for the duration of the the {{selectstart}}, {{select}}, and {{selectend}} event sequence.
+
+<div class="algorithm" data-algorithm="on-transient-input-start">
+
+When a [=transient input source=] for {{XRSession}} |session| begins it's [=primary action=] the UA MUST run the following steps:
+
+  1. Fire any <code>"pointerdown"</code> events produced by the [=XR input source=]'s action, if necessary.
```
NellWaliczek: I'm curious about why the first two steps are in this order?  I don't think I have strong opinion either way I suppose?

toji:
Regarding why this is for transient sources only: Any of the non-transient input sources I can think of wouldn't be firing interleaved pointer/click events? Also, this is how it was presented in the explainer and the issue in questions (#540) specifically called out that the new explainer text needed to be pulled into the spec.

And yes, that was probably too narrow of a reading of the issue/explainer text on my part. I'll address the feedback soon.

index.bs
toji: That was the order that you had in the explainer. I don't recall us having a conversation about it at any point, but I do recall looking at your PR and wondering if it mattered before deciding that it probably did not and thus wasn't worth discussing.

toji:
Updated to address feedback. Big change is section talking about order of events for non-transient input sources. PTAL!

index.bs
NellWaliczek: Haha, well I don't think I have a strong opinion either?

index.bs
```diff
@@ -1206,16 +1200,46 @@ The <dfn attribute for="XRInputSource">gamepad</dfn> attribute is a {{Gamepad}}
  - A button which reports analog values
  - An axis
 
-NOTE: {{XRInputSource}}s with a <code>null</code> {{gamepad}} can still fire {{select}}, {{selectstart}}, and {{selectend}} events to report binary inputs.
+Each [=XR input source=] SHOULD define a <dfn>primary action</dfn>. The [=primary action=] is a platform-specific action that, when engaged, produces {{selectstart}}, {{selectend}}, and {{XRSession/select}} events. Examples of possible [=primary action=]s are pressing a trigger, touchpad, or button, speaking a command, or making a hand gesture. If the platform guidelines define a recommended primary input then it should be used as the [=primary action=], otherwise the user agent is free to select one.
```
NellWaliczek: >{{selectstart}}, {{selectend}}, and {{XRSession/select}}

Why the different formatting?

NellWaliczek: >recommended primary input

This is a bit confusing?  Perhaps say "button"?

index.bs
```diff
@@ -1206,16 +1200,46 @@ The <dfn attribute for="XRInputSource">gamepad</dfn> attribute is a {{Gamepad}}
...
+<div class="algorithm" data-algorithm="on-input-end">
+
+When an [=XR input source=] for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
+
+  1. [=Queue a task=] that fires a {{XRInputSourceEvent}} named {{select!!event}} on |session|.
+  1. [=Queue a task=] that fires a {{XRInputSourceEvent}} named {{selectend!!event}} on |session|.
+
+</div>
+
+Sometimes platform-specific behavior can result in a [=primary action=] being interrupted or cancelled. For example, a [=/XR input source=] may be removed from the [=/XR device=] after the [=primary action=] is started but before it ends.
```
NellWaliczek: Are there other reasons for this to occur?  If not, perhaps it would make the spec text clearer if we landed a fix for the issue described in this comment https://github.com/immersive-web/webxr/pull/628#discussion_r282987983 first?

index.bs
toji: Agreed that it sounds a bit awkward, but I wanted to be inclusive to inputs such as air taps and voice commands.

index.bs
toji: There are other reasons, I'm just not sure how many we really want to get into in the spec text? For example: If you started a mouse click on an inline canvas and then moved the cursor off before releasing?

Issue #633, opened 2019-05-08
Do "eye-level" and "floor-level" fit with "bounded" and "unbounded"?


thetuvix:
When seeing the simpler list of 5 reference spaces proposed in #626, it jumped out at me that `eye-level` and `floor-level` don't really fit as names in the same list with `bounded` and `unbounded`:

> There are several types of reference spaces: `viewer`, `eye-level`, `floor-level`, `bounded`, and `unbounded`.

For example, the `bounded` reference space is also a "floor level" space and the `unbounded` reference space is also an "eye level" space.  As a room-scale developer approaching this API fresh, how do I know if the reference space I want is `floor-level` or `bounded`, when what I really want is both?

It may make more sense in our recently-flattened reference space list to keep the term "stationary", so that it's clear how those two spaces differ from the other spaces.

_A strawman space list that keeps "stationary" and "floor" orthogonal, with "eye" as the implicit default:_
* `viewer`
* `stationary`
* `stationary-floor`
* `bounded-floor`
* `unbounded`

Independently, I'd shared the [spatial tracking explainer](https://github.com/immersive-web/webxr/blob/master/spatial-tracking-explainer.md) with @mattdatwork from Unity who expressed that the term "local" from OpenXR made more sense to him alongside "bounded" and "unbounded" than the term "stationary".

_If we went the "local" route, we'd end up with this strawman instead:_
* `viewer`
* `local`
* `local-floor`
* `bounded-floor`
* `unbounded`

Thoughts?

thetuvix:
Pinging @NellWaliczek, @toji, @Manishearth, @klausw and @blairmacintyre to take a look here.

thetuvix:
One potential advantage of the `local`/`local-floor` naming here would be if we had the spec specifically suggest that the local reference spaces follow platform conventions for experiences that should operate nearby the user.

For example, the spec currently states the following:

>   - Passing a type of <dfn enum-value for="XRReferenceSpaceType">eye-level</dfn> creates an {{XRReferenceSpace}} instance. It represents a tracking space with a [=native origin=] near the user's head at the time of creation. The exact position and orientation will be initialized based on the conventions of the underlying platform. ~~When using this reference space the user is not expected to move beyond their initial position much, if at all, and tracking is optimized for that purpose.~~ For devices with [=6DoF=] tracking, {{eye-level}} reference spaces should emphasize keeping the origin stable relative to the user's environment.

We could say the following instead:

[...]

PR #640, opened 2019-05-14
Rename reference space types

toji:
Renames several reference space types for clarity, as recommended in
issue #633. Intent is to better communicate which reference spaces are
floor-aligned and avoid potential developer confusion that may arise
from use of the term "stationary".

Renamed reference spaces:
 - `eye-level` -> `local`
 - `floor-level` -> `local-floor`
 - `bounded` -> `bounded-floor`

### DIFF
--- a/spatial-tracking-explainer.md
+++ b/spatial-tracking-explainer.md
@@ -91,16 +91,9 @@ function onSessionStarted(session) {
 
 There is no mechanism for getting a floor-relative _unbounded_ reference space. This is because the user may move through a variety of elevations (via stairs, hills, etc), making identification of a single floor plane impossible.
 
-### Local experiences
-A _local_ experience is one which does not require the user to move around in space.  This includes several categories of experiences that developers are commonly building today, and multiple reference space types are available to address each class of experience. "Seated" experiences can be created with an `local` reference space. "Standing" experiences can be created with a `local-floor` reference space.
-
-Orientation-only experiences such as 360 photo/video viewers can also be created with an `local` reference space by either explicitly ignoring the pose's positional data or displaying the media "infinitely" far away from the viewer. If such position mitigation steps are not taken the user may perceive the geometry the media is displayed on, leading to discomfort.
-
-It is important to note that `XRViewerPose` objects retrieved using `local` and `local-floor` reference spaces may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that `local` and `local-floor` experiences must be resilient to position changes despite not being dependent on receiving them.
-
 #### Local reference spaces
 
-Sometimes referred to as "seated", this reference space's origin will be initialized at a position near the viewer's position at the time of creation. The exact `x`, `y`, `z`, and orientation values will be initialized based on the conventions of the underlying platform for `local` experiences. Some platforms may initialize these values to the viewer's exact position/orientation at the time of creation. Other platforms that allow users to reset a common local origin shared across multiple apps may use that origin instead.
+A _local_ experience is one which does not require the user to move around in space. Sometimes referred to as "seated", this reference space's origin will be initialized at a position near the viewer's position at the time of creation. The exact `x`, `y`, `z`, and orientation values will be initialized based on the conventions of the underlying platform for `local` experiences. Some platforms may initialize these values to the viewer's exact position/orientation at the time of creation. Other platforms that allow users to reset a common local origin shared across multiple apps may use that origin instead.
 
 Some example use cases: 
 * Immersive 2D video viewer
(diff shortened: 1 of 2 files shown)


spatial-tracking-explainer.md
```diff
@@ -92,15 +92,15 @@ function onSessionStarted(session) {
 There is no mechanism for getting a floor-relative _unbounded_ reference space. This is because the user may move through a variety of elevations (via stairs, hills, etc), making identification of a single floor plane impossible.
 
 ### Stationary experiences
```
thetuvix: Should we refer to these now as "local experiences" in the explainer, to avoid the same "stationary" ambiguity that we're avoiding in the API?

spatial-tracking-explainer.md
```diff
@@ -92,15 +92,15 @@ function onSessionStarted(session) {
...
 
-It is important to note that `XRViewerPose` objects retrieved using `floor-level` and `eye-level` reference spaces may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that `floor-level` and `eye-level` experiences must be resilient to position changes despite not being dependent on receiving them.  
+It is important to note that `XRViewerPose` objects retrieved using `local-floor` and `local` reference spaces may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that `local-floor` and `local` experiences must be resilient to position changes despite not being dependent on receiving them.  
 
-#### Floor-level reference spaces
+#### Local-floor reference spaces
```
thetuvix: Is it now more sensible for the explainer to introduce "local" before "local-floor"?

spatial-tracking-explainer.md
```diff
@@ -281,9 +281,9 @@ function onSessionStarted(session) {
   xrSession.requestReferenceSpace('unbounded').then((referenceSpace) => {
     xrReferenceSpace = referenceSpace;
   }).catch(() => {
-    // If an unbounded reference space is not available, request an eye-level
```
NellWaliczek: It's totally minor, but I feel like the "local-floor" and "bounded-floor" would make a better example here.

toji:
Thanks for the feedback! Resolved it and pushed again.

spatial-tracking-explainer.md
```diff
@@ -91,16 +91,9 @@ function onSessionStarted(session) {
...
-
 #### Local reference spaces
 
-Sometimes referred to as "seated", this reference space's origin will be initialized at a position near the viewer's position at the time of creation. The exact `x`, `y`, `z`, and orientation values will be initialized based on the conventions of the underlying platform for `local` experiences. Some platforms may initialize these values to the viewer's exact position/orientation at the time of creation. Other platforms that allow users to reset a common local origin shared across multiple apps may use that origin instead.
+A _local_ experience is one which does not require the user to move around in space. Sometimes referred to as "seated", this reference space's origin will be initialized at a position near the viewer's position at the time of creation. The exact `x`, `y`, `z`, and orientation values will be initialized based on the conventions of the underlying platform for `local` experiences. Some platforms may initialize these values to the viewer's exact position/orientation at the time of creation. Other platforms that allow users to reset a common local origin shared across multiple apps may use that origin instead.
```
NellWaliczek: > Some platforms may initialize these values to the viewer's exact position/orientation at the time of creation. Other platforms that allow users to reset a common local origin shared across multiple apps may use that origin instead.

This is something we may choose to require standardization around for privacy/security reasons....

spatial-tracking-explainer.md
```diff
@@ -124,9 +117,11 @@ function onSessionStarted(session) {
 }
 ```
 
+Orientation-only experiences such as 360 photo/video viewers can also be created with an `local` reference space by either explicitly ignoring the pose's positional data or displaying the media "infinitely" far away from the viewer. If such position mitigation steps are not taken the user may perceive the geometry the media is displayed on, leading to discomfort.
+
 #### Local-floor reference spaces
```
NellWaliczek: Should this be "###"?

spatial-tracking-explainer.md
```diff
@@ -149,6 +144,8 @@ function onSessionStarted(session) {
 }
 ```
 
+It is important to note that `XRViewerPose` objects retrieved using `local` and `local-floor` reference spaces may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that `local` and `local-floor` experiences must be resilient to position changes despite not being dependent on receiving them.
```
NellWaliczek: Maybe this should go into the section above? I'm not sure how to make sure this info doesn't get lost in the sea of words ^_^;

spatial-tracking-explainer.md
toji: Agreed. But that's a matter for a separate issue.

spatial-tracking-explainer.md
toji: Moved it up a section and partially repeated that paragraph again down here. It does feel less likely to get overlooked that way.

spatial-tracking-explainer.md
thetuvix: This seems similar in privacy/security impact to bounded experiences to me - the user's first head pose can be relative to a user-calibrated origin, either nearby for `local`/`local-floor` or by the room center for `bounded`.

Do you see specific privacy/security impact around `local'?

PR #642, opened 2019-05-14
Updated the explainer with new origin offset mechanics.

### LINKED ISSUE #634: Update origin offset handling in the explainer.
When we made the recent change from the `originOffset` attribute to the `getOffsetReferenceSpace()` method (#580) the explainer missed getting the new definitions. Let's do a pass over it to make sure the origin offset language is consistent with the spec.

toji:
/fixes #634


spatial-tracking-explainer.md
```diff
@@ -209,35 +209,31 @@ An `XRRay` object includes both an `origin` and `direction`, both given as `DOMP
...
 
-Developers can specify application-specific transforms by setting the `originOffset` attribute of any `XRReferenceSpace`. The `originOffset` is initialized to an identity transform, and any values queried using the `XRReferenceSpace` will be offset by the `position` and `orientation` the `originOffset` describes. The `XRReferenceSpace`'s `originOffset` can be updated at any time and will immediately take effect, meaning that any subsequent poses queried with the `XRReferenceSpace` will take into account the new `originOffset`. Previously queried values will not be altered. Changing the `originOffset` between pose queries in a single frame is not advised, since it will cause inconsistencies in the tracking data and rendered output.
+Developers can specify application-specific transforms by calling the `getOffsetReferenceSpace()` method of any `XRReferenceSpace`. This returns a new `XRReferenceSpace` with an origin equal to the origin of the parent reference space offset by the by the `position` and `orientation` described by the `XRRigidTransform` passed to `getOffsetReferenceSpace()`.
```
NellWaliczek: Is there a reason not to introduce the concept of a native origin in the explainer?  I suspect it would make things quite a bit clearer...

spatial-tracking-explainer.md
```diff
@@ -209,35 +209,31 @@ An `XRRay` object includes both an `origin` and `direction`, both given as `DOMP
...
 
 #### Inline to Immersive
-It is expected that developers will often choose to preview `immersive` experiences with a similar experience `inline`. In this situation, users often expect to see the scene from the same perspective when they make the transition from `inline` to `immersive`. To accomplish this, developers should grab the `transform` of the last `XRViewerPose` retrieved using the `inline` session's `XRReferenceSpace` and set it as the `originOffset` of the `immersive` session's `XRReferenceSpace`. The same logic applies in the reverse when exiting `immersive`.
+It is expected that developers will often choose to preview `immersive` experiences with a similar experience `inline`. In this situation, users often expect to see the scene from the same perspective when they make the transition from `inline` to `immersive`. To accomplish this, developers should grab the `transform` of the last `XRViewerPose` retrieved using the `inline` session's `XRReferenceSpace` and pass it to `getOffsetReferenceSpace()` on the `immersive` session's `XRReferenceSpace` to produce an appropriately offset reference space. The same logic applies in the reverse when exiting `immersive`.
```
NellWaliczek: > To accomplish this, developers should grab the `transform` of the last `XRViewerPose` retrieved using the `inline` session's `XRReferenceSpace` and pass it to `getOffsetReferenceSpace()` on the `immersive` session's `XRReferenceSpace` to produce an appropriately offset reference space. The same logic applies in the reverse when exiting `immersive`.

Oh ick.  This means that developers will need to create a throw-away reference space so they can call `getOffsetReferenceSpace()` on it?  Should we be considering allowing an optional `XRRigidTransform` as a parameter to `requestReferenceSpace()`?

spatial-tracking-explainer.md
```diff
@@ -209,35 +209,31 @@ An `XRRay` object includes both an `origin` and `direction`, both given as `DOMP
...
 
 ### Click-and-drag view controls
 Frequently with inline sessions it's desirable to have the view rotate when the user interacts with the inline canvas. This is useful on devices without tracking capabilities to allow users to still view the full scene, but can also be desirable on devices with some tracking capabilities, such as a mobile phone or tablet, as a way to adjust the users view without requiring them to physically turn around.
 
-By updating the `originOffset` in response to pointer events, pages can provide basic click-and-drag style controls to allow the user to pan the view around the immersive scene.
+By calling `getOffsetReferenceSpace()` in response to pointer events, pages can provide basic click-and-drag style controls to allow the user to pan the view around the immersive scene.
```
NellWaliczek: Given that we're encouraging this behavior in developers, I wonder if we want to add some spec text to encourage UAs to minimized duplicated data each time a `getOffsetReferenceSpace()` is called?

spatial-tracking-explainer.md
toji: I doubt that the "throwaway" reference space + one offset one will be all that expensive relative to creating the reference space with the right offset to begin with, but it's a valid API ergonomics issue. It should definitely be handled in a separate issue from this one, however, and I'd argue that it sounds to me like the type of convenience feature that would work really well for a "vNext" edition of the spec.

spatial-tracking-explainer.md
toji: By "duplicated data" I assume you mean the pose data that gets polled from the native API? (To which I agree.) That seems like it's technically an implementation detail, but I don't see much issue with adding a non-normative section to indicate that we expect that some apps may create a LOT of these. I'll add something to that effect.

toji:
Addressed feedback. PTAL!

spatial-tracking-explainer.md
```diff
@@ -208,35 +208,31 @@ An `XRRay` object includes both an `origin` and `direction`, both given as `DOMP
...
+Developers can specify application-specific transforms by calling the `getOffsetReferenceSpace()` method of any `XRReferenceSpace`. This returns a new `XRReferenceSpace` with an "effective origin" equal to the native origin of the parent reference space offset by the by the `position` and `orientation` described by the `XRRigidTransform` passed to `getOffsetReferenceSpace()` plus any offset applied to the reference space previously if it was also created with `getOffsetReferenceSpace()`.
```
klausw: I think it's confusing to use "plus" to refer to chaining or multiplying transforms.

Duplicate words: "by the by the"

"origin equal to ... native origin offset by" seems ill-defined. This would need to be expressed in terms of equivalence of transforms, or equality of coordinates after being transformed.

How about something like this?

Developers can specify application-specific transforms by calling the `getOffsetReferenceSpace()` method of any `XRReferenceSpace`. This returns a new `XRReferenceSpace` where the `XRRigidTransform` passed to `getOffsetReferenceSpace()` describes the position and orientation of the offset space's origin in relation to the base reference space's origin. Specifically, `originOffset.position` contains the coordinates of the new origin relative to the base reference space's origin. If the base reference space was also created with `getOffsetReferenceSpace()`, the overall offset is the combination of both transforms. 

Optional example:

```js
let offsetASpace = baseSpace.getOffsetReferenceSpace(offsetAInBase);
// offsetAInBase.matrix = baseFromA
let offsetBSpace = offsetASpace.getOffsetReferenceSpace(offsetBInA);
// offsetBInA.matrix = AFromB

// equivalent to:
let offsetBSpace = baseSpace.getOffsetReferenceSpace(offsetBInBase);
// offsetBInBase.matrix = baseFromA * AFromB = offsetAInBase.matrix * offsetBInA.matrix
```

spatial-tracking-explainer.md
```diff
@@ -208,35 +208,31 @@ An `XRRay` object includes both an `origin` and `direction`, both given as `DOMP
...
 
 A common use case for this attribute would be for a "teleportation" mechanic, where the user "jumps" to a new point in the virtual scene, after which the selected point is treated as the new virtual origin which all tracked motion is relative to.
 
 ```js
 // Teleport the user a certain number of meters along the X, Y, and Z axes
```
klausw: This example is too vague to be useful, there's no explanation which sign of delta{X,Y,Z} corresponds to which of the movement direction interpretations.

Do you want to do something more along the lines of my more specific example in my old PR?

https://github.com/immersive-web/webxr/pull/569/files#diff-f02c24f867e61b8e58dc8f2c228f476cR242

spatial-tracking-explainer.md
```diff
@@ -208,35 +208,31 @@ An `XRRay` object includes both an `origin` and `direction`, both given as `DOMP
...
 function teleport(deltaX, deltaY, deltaZ) {
-  let currentOrigin = xrReferenceSpace.originOffset;
-  xrReferenceSpace.originOffset = new XRRigidTransform(
-      { x: currentOrigin.position.x + deltaX,
-        y: currentOrigin.position.y + deltaY,
-        z: currentOrigin.position.z + deltaZ },
-      currentOrigin.orientation);
+  xrReferenceSpace = xrReferenceSpace.getOffsetReferenceSpace(
```
billorr-google: I'd love to see a super-specific example of how a reference space and an offset reference space interact, not just how they can be used.  Would it make sense to add something like this?

For example:
offsetSpace = xrReferenceSpace.getOffsetReferenceSpace(new XRRigidTransform({x: 1, y:0, z:0});
// (0,0,0) in offsetSpace coordinates is (1,0,0) in xrReferenceSpace coordinates.

Similar for chained offsets to make it extra clear how they are chained.
offsetSpace1 = xrReferenceSpace.getOffsetReferenceSpace(new XRRigidTransform({x: 1,y: 0, z: 0});
offsetSpace2 = offsetSpace1.getOffsetReferenceSpace(<rotation around y axes by 180>);
(0,0,0) in offsetSpace2 is (0,0,0) in offsetspace1 is (1, 0, 0) in xrReferenceSpace, but
(1,0,0) in offsetSpace2 is (-1,0,0) in offsetSpace1 is (0,0,0) in xrReferenceSpace.

spatial-tracking-explainer.md
```diff
@@ -246,18 +242,10 @@ const RAD_PER_PIXEL = Math.PI / 180.0; // (1 degree)
...
-      {
-        x: o.x * c - o.z * s,
-        y: o.y * c + o.w * s,
-        z: o.z * c + o.x * s,
-        w: o.w * c - o.y * s
-      });
+
+  // Computes a quaternion to rotate around the Y axis.
+  xrReferenceSpace = xrReferenceSpace.getOffsetReferenceSpace(
+    new XRRigidTransform(null, { x: 0, y: s, z: 0, w: c }));
```
billorr-google: If the API allows chaining transforms, numerical error may accumulate.  Should the spec/explainer call this out as a warning to implementers?

klausw: I think this example may be a bit confusing, it implies that it rotates by a given angle, but the usual "quaternion from axis angle" formulas use sin/cos of half angles internally. I guess it doesn't matter since it's based on an arbitrary scale factor, but could lead to wrong results if people generalize it.

klausw:
@billorr-google wrote:

> I'd love to see a super-specific example of how a reference space and an
> offset reference space interact, not just how they can be used. Would it
> make sense to add something like this?


FWIW, I had a specific example in my old PR:
https://github.com/immersive-web/webxr/pull/569/files#diff-f02c24f867e61b8e58dc8f2c228f476cR259

*From: *billorr-google <notifications@github.com>
*Date: *Tue, May 14, 2019 at 5:10 PM
*To: *immersive-web/webxr
*Cc: *Klaus Weidner, Comment

*@billorr-google* commented on this pull request.

spatial-tracking-explainer.md
toji: The proposed text here is perfect! I'm going to use it word-for-word. Thank you!

spatial-tracking-explainer.md
toji: Updated the example to use comments and logic from your old PR (adapted to the new API shape.)

spatial-tracking-explainer.md
NellWaliczek: Sounds good.  Perhaps let's file the issue and put it in the "Future" milestone.

spatial-tracking-explainer.md
NellWaliczek: Thanks for fixing this!

spatial-tracking-explainer.md
toji: Fixed

toji: Added a comment to this effect.

spatial-tracking-explainer.md
```diff
@@ -151,9 +151,9 @@ function onSessionStarted(session) {
...
 
-This type of reference space is requested with a type of `viewer` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have a `transform` that is equal to the reference space's `originOffset` and the `XRView` matrices will be offset accordingly.
+This type of reference space is requested with a type of `viewer` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have an identity `transform` (plus an offset if the reference space was created by calling `getOffsetReferenceSpace()`). `XRView`s populated from a `viewer` reference space will still be offset from the `XRViewerPose`'s `transform` as usual.
```
NellWaliczek: This somewhat confused me... perhaps this instead?
```suggestion
This type of reference space is requested with a type of `viewer` and returns a basic `XRReferenceSpace`. `XRViewerPose` objects retrieved with this reference space will have an identity `transform` (plus an offset if the reference space was created by calling `getOffsetReferenceSpace()`). `XRView`s populated from a `viewer` reference space will still be offset from the `XRViewerPose`'s `transform` in the same manner as all other reference spaces.
```

spatial-tracking-explainer.md
```diff
@@ -210,60 +210,54 @@ An `XRRay` object includes both an `origin` and `direction`, both given as `DOMP
...
+Developers can specify application-specific transforms by calling the getOffsetReferenceSpace() method of any XRReferenceSpace. This returns a new XRReferenceSpace where the XRRigidTransform passed to getOffsetReferenceSpace() describes the position and orientation of the offset space's origin in relation to the base reference space's origin. Specifically, originOffset.position contains the coordinates of the new origin relative to the base reference space's origin. If the base reference space was also created with getOffsetReferenceSpace(), the overall offset is the combination of both transforms.
```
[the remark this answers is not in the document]
klausw: > Seems strange to be calling out position, but not orientation. Almost implies that orientation is ignored.

That wasn't the intent, a reason for having this here is that it's a fairly straightforward example that's easy to sanity check in terms of coordinates, and the previous sentence already mentioned "position and orientation". It's difficult to make a short-but-concrete example based on quaternions.

We could add something along these lines, but I think this may get clunky?

... and originOffset.orientation is a quaternion representing a rotation that transforms vectors in offset space coordinates to their coordinate values in the base reference space.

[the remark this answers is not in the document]
thetuvix: In general, if we could centrally establish once in the spec that a "pose" is a position and orientation applied in a specific order within a base space, we could easily and consistently describe all such use of poses throughout the spec in a straightforward manner:

> This returns a new XRReferenceSpace where the XRRigidTransform passed to getOffsetReferenceSpace() describes the pose of the offset space's origin within the base space.

PR #645, opened 2019-05-14
Remove redundant "abort these steps" in requestSession() algorithm

ddorwin:
"and abort these steps" is unnecessary as these steps are aborted two steps later. This would also prevent "pending immersive session" from being cleared.


toji:
Good catch, thanks!

PR #648, opened 2019-05-15
Clarify assumptions for local and local-floor tracking

klausw:


spatial-tracking-explainer.md
```diff
@@ -119,7 +119,7 @@ function onSessionStarted(session) {
...
+It is important to note that `XRViewerPose` objects retrieved using `local` reference spaces may include position information as well as rotation information.  For example, hardware which does not support 6DOF tracking (ex: GearVR) may still use neck-modeling to improve user comfort. Similarly, a user may lean side-to-side on a device with 6DOF tracking (ex: HTC Vive). The result is that `local` experiences must be resilient to position changes despite not being dependent on receiving them. Experiences are encouraged to fully support head movement in a typical seated range. For some hardware the poses may even include substantial position offsets, for example if the user stands up from their seated position and walks around. The experience can choose to react to this appropriately, for example by fading out the rendered view when intersecting geometry. Overriding the pose's reported head position, for example clamping to stay within an expected volume, can be very uncomfortable to users and should be avoided.
```
toji: This is a great addition, but I feel like it would read a bit better if we put most of your added text in it's own paragraph. Maybe something like so:

> ... The result is that `local` experiences must be resilient to position changes despite not being dependent on receiving them, **and are encouraged to fully support head movement in a typical seated range.**
>
> **For some hardware the poses may even include substantial position offsets, for example if the user stands up from their seated position and walks around.**...

klausw:
> This is a great addition, but I feel like it would read a bit better if
we put most of your added text in it's own paragraph. Maybe something like
so:

Done.

*From: *Brandon Jones <notifications@github.com>
*Date: *Wed, May 15, 2019 at 11:32 AM
*To: *immersive-web/webxr
*Cc: *Klaus Weidner, Author

*@toji* commented on this pull request.

toji:
Thanks!

PR #649, opened 2019-05-15
Fix multiplication order for transforms

klausw:
The multiplication order in the spec needs to match the matrix
semantics we've agreed on, and the usage appeared to be backwards.

Specific justifications following.

getViewerPose: if |offset| is the view offset of an eye view, the expected
semantics would be that it's a pose-like transform where the transform's
position is the eye view's coordinates in XRViewerPose space. The
corresponding matrix is viewerFromEye. XRViewerPose contains the viewer
pose in the base reference space, so its transform's position is the viewer
position in base space coordinates, and the transform is baseFromViewer.

Chaining these transforms works as follows:

```js
offset.matrix = viewerFromEye;
viewerPose.transform.matrix = baseFromViewer;

view.transform.matrix = baseFromEye
                      = baseFromViewer * viewerFromEye
                      = viewerPose.transform.matrix * offset.matrix
```

For originOffset, the logic is similar:

```js
let offsetASpace = baseSpace.getOffsetReferenceSpace(offsetAInBase);
// offsetAInBase.matrix = baseFromA
let offsetBSpace = offsetASpace.getOffsetReferenceSpace(offsetBInA);
// offsetBInA.matrix = AFromB

// equivalent to:
let offsetBSpace = baseSpace.getOffsetReferenceSpace(offsetBInBase);
// offsetBInBase.matrix = baseFromA * AFromB = offsetAInBase.matrix * offsetBInA.matrix
```

Lastly, for XRRay, the convention is to transform points or vectors by
premultiplying the transform matrix from the left, I've rephrased those usages
to make that explicit. The current text was basically saying "multiply
vector by matrix" which would be backwards.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -733,7 +733,7 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method 
       1. Initialize |xrview|'s {{XRView/eye}} to |view|'s [=view/eye=]
       1. Initialize |xrview|'s {{XRView/projectionMatrix}} to |view|'s [=view/projection matrix=]
       1. Let |offset| be an {{XRRigidTransform}} equal to the [=view offset=] of |view|
-      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply transforms|multiplying=] the |offset| transform by the {{XRViewerPose}}'s {{XRPose/transform}}
+      1. Set |xrview|'s {{XRView/transform}} property to the result of [=multiply transforms|multiplying=] the {{XRViewerPose}}'s {{XRPose/transform}} by the |offset| transform
       1. [=list/Append=] |xrview| to |xrviews|
   1. Set |pose|'s {{XRViewerPose/views}} to |xrviews|
   1. Return |pose|.
@@ -851,7 +851,7 @@ The <dfn method for="XRReferenceSpace">getOffsetReferenceSpace(|originOffset|)</
   1. If |base| is an instance of {{XRBoundedReferenceSpace}}, let |offsetSpace| be a new {{XRBoundedReferenceSpace}} and set |offsetSpace|'s {{boundsGeometry}} to |base|'s {{boundsGeometry}}, with each point multiplied by the {{XRRigidTransform/inverse}} of |originOffset|.
   1. Else let |offsetSpace| be a new {{XRReferenceSpace}}.
   1. Set |offsetSpace|'s [=native origin=] to |base|'s [=native origin=].
-  1. Set |offsetSpace|'s [=origin offset=] to the result of [=multiply transforms|multiplying=] |originOffset| by |base|'s [=origin offset=].
+  1. Set |offsetSpace|'s [=origin offset=] to the result of [=multiply transforms|multiplying=] |base|'s [=origin offset=] by |originOffset|.
   1. Return |offsetSpace|.
 
 </div>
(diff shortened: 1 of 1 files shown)


klausw:
@Manishearth @toji @NellWaliczek @billorr-boogle: can you please sanity check this? I'm fairly sure the proposed change is correct, but we should make sure we're all on the same page for this. Let me know if you want to discuss this more.

index.bs
```diff
@@ -851,7 +851,7 @@ The <dfn method for="XRReferenceSpace">getOffsetReferenceSpace(|originOffset|)</
   1. If |base| is an instance of {{XRBoundedReferenceSpace}}, let |offsetSpace| be a new {{XRBoundedReferenceSpace}} and set |offsetSpace|'s {{boundsGeometry}} to |base|'s {{boundsGeometry}}, with each point multiplied by the {{XRRigidTransform/inverse}} of |originOffset|.
   1. Else let |offsetSpace| be a new {{XRReferenceSpace}}.
   1. Set |offsetSpace|'s [=native origin=] to |base|'s [=native origin=].
-  1. Set |offsetSpace|'s [=origin offset=] to the result of [=multiply transforms|multiplying=] |originOffset| by |base|'s [=origin offset=].
+  1. Set |offsetSpace|'s [=origin offset=] to the result of [=multiply transforms|multiplying=]|base|'s [=origin offset=] by |originOffset|.
```
toji: Nit: We lost a space between `[=multiply transforms|multiplying=]` and `|base|`

klausw: Oops, fixed.

toji:
Thanks for this! The description and spec change *sound* correct to me, though I'd definitely want @Manishearth to also chime in since I believe at least the `XRView` logic was something he added recently.

Manishearth:
IMO this is incomplete: The choice can be made either way here until we clarify what XRRigidTransform's `matrix` is. As mentioned before, the following text:

```
The matrix attribute returns the transform described by the position and orientation attributes as a matrix. This attribute SHOULD be lazily evaluated.
```

can mean one of two things: either a transform that takes an object and offsets it by the position/orientation, or a transform that takes an object and transforms it into the coordinates of the space described by the offset. The latter is the inverse of the former.

I'm worried of bouncing back and forth on this (which is what we've been doing so far): as long as this isn't explicitly specced, there will always be some confusion

The definition we've implicitly settled (I _think_) on is "a transform that takes an object and offsets it by the position/orientation, orientation applied first", but we should be clear about that, and also include some math (IMO). I've been planning to make a PR for this (especially since this is bikeshed time), doing the following:

 - Spec out what `matrix` does, both conceptually and mathematically
 - Maybe remove matrix decomposition from the transform multiplication algorithms, operate on rotations and translations directly (with a note mentioning the matrix method too)
 - Go through the spec ensuring that all the multiplication orders make sense

Thoughts? This can still land as-is, though, the math is correct, if we continue with the implicit assumptions we've been making about `matrix`.

klausw:
@Manishearth I agree that the current XRRigidTransform description is underspecified, but can we address this in a followup if we agree that this PR matches the intended interpretation?

I've previously attempted to make the transforms unambiguous by adding additional language about source and destination of transforms in XRRigidTransform and getPose, along with adding coordinate-based clarifications, but there seemed to be strong resistance against doing this.

See for example [this part of the closed PR #569](https://github.com/immersive-web/webxr/pull/569/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR920):

> An _XRRigidTransform_ is a transform from a source _XRSpace_ to a destination _XRSpace_ described by a _position_ and _orientation_. When applying an _XRRigidTransform_ the _orientation_ is always applied prior to the _position_. Effectively, the source space is rotated around its origin using the _orientation_, and then the source space's origin is translated to the _XRRigidTransform/position_ in destination space.
> 
> The _position_ attribute is a 3-dimensional point, given in meters, for the source space's origin location in the destination space's coordinate system.
> 
> The _matrix_ attribute returns the transform described by the _position_ and _orientation_ attributes as a matrix. This attribute SHOULD be lazily evaluated. The top left 3x3 sub-matrix is the rotation matrix corresponding to the orientation, its column values are the source space's axis directions as unit vector coordinates in destination space. The fourth column contains the position. Premultiplying this matrix onto a column vector of source space coordinates produces a column vector of destination space coordinates.

klausw:
(To be clear, I'm not proposing using the old phrasing I've quoted above, I agree that it's clunky, and the source/destination space approach is still confusing. But I do think that it's helpful to explain positions and matrix values in terms of coordinates in a given space.)

Manishearth:
Yeah, like I said this can land as-is, I just wanted to point out that this only shifts the ambiguity to the matrix definition.

But that is okay, since we're reasonably clear as to what the semantics of the matrix are, we just haven't specced it out.

toji:
Thanks Klaus and Manish! Merging.

Manishearth:
Should I open a dedicated issue to discuss remaining matrix clarifications?

PR #651, opened 2019-05-16
Explicitly spec out when requestReferenceSpace() can reject queries

### LINKED ISSUE #564: requestReferenceSpace() should talk more about when you can reject the request
There's a list of situations where `requestReferenceSpace()` can fail listed in [the explainer](https://github.com/immersive-web/webxr/blob/master/spatial-tracking-explainer.md#practical-usage-guidelines), this should be part of the spec.

Manishearth:
fixes #564

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -837,11 +837,26 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
 
   1. Let |session| be the {{XRSession}} object that requested creation of a reference space.
   1. Let |type| be set to the {{XRReferenceSpaceType}} passed to {{requestReferenceSpace()}}.
-  1. If |type| is {{bounded-floor}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
-  1. Else let |referenceSpace| be a new {{XRReferenceSpace}}.
-  1. Initialize |referenceSpace|'s [=XRReferenceSpace/type=] to be |type|.
-  1. Initialize [=XRSpace/session=] to be |session|.
-  1. Return |referenceSpace|.
+  1. If the [=reference space is supported=] for |type| and |session|, run the following steps:
+    1. If |type| is {{bounded-floor}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
+    1. Else let |referenceSpace| be a new {{XRReferenceSpace}}.
+    1. Initialize |referenceSpace|'s [=XRReferenceSpace/type=] to be |type|.
+    1. Initialize [=XRSpace/session=] to be |session|.
+    1. Return |referenceSpace|.
+  1. Else, return <code>null</code>.
+
+</div>
+
+<div class="algorithm" data-algorithm="reference-space-supported">
+To check if a <dfn>reference space is supported</dfn> for a given reference space type |type| and {{XRSession}} |session|, run the following steps:
+
+
+  1. If |type| is {{inline}}, return <code>true</code>.
+  1. If |type| is {{local}} or {{local-floor}}, and |session| is an [=immersive session=], return <code>true</code>.
+  1. If |type| is {{local}} or {{local-floor}}, and the [=/XR device=] supports reporting orientation data, return <code>true</code>.
+  1. If |type| is {{bounded-floor}} and |session| is an [=immersive session=], return if [=bounded reference spaces are supported=] by the [=/XR device=].
+  1. If |type| is {{unbounded}}, |session| is an [=immersive session=], and the [=/XR device=] supports reporting position and orientation data, return <code>true</code>.
+  1. Return <code>false</code>.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


Manishearth:
r? @NellWaliczek 

not sure if this is what you were looking for. I can also change the language to be stronger/weaker on this.

(I suspect we'll have to tweak this a bit as our permissions model comes in)

index.bs
```diff
@@ -837,6 +837,12 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
 
   1. Let |session| be the {{XRSession}} object that requested creation of a reference space.
   1. Let |type| be set to the {{XRReferenceSpaceType}} passed to {{requestReferenceSpace()}}.
+  1. If |type| is {{bounded-floor}} or {{unbounded}}, and |session|'s [=XRSession/mode=] is {{XRSessionMode/inline}}, return <code>null</code>
+  1. If |type| is {{bounded-floor}} or {{unbounded}}, and |session|'s [=XRSession/mode=] is not {{XRSessionMode/inline}} the user agent may choose to return <code>null</code> if such experiences are not supported by the underlying [=/XR device=]
+  1. If |type| is {{local-floor}} or {{local}}, and |session|'s [=XRSession/mode=] is {{XRSessionMode/inline}} the user agent may choose to return <code>null</code> if such experiences are not supported by the underlying [=/XR device=] in non-immersive sessions
+
+      NOTE: All other combinations of {{XRReferenceSpaceType}}s and [=XRSession/mode=]s are guaranteed by this spec, provided the [=XRSession/mode=] itself is supported.
```
toji: I can't imagine placing a note in the middle of an algorithm like this formats nicely? It should definitely be moved outside the algorithm.

toji:
Happened to be able to catch @bfgeek and ask him about this. (Presented as "I want my spec algorithm to have a virtual function, is that a thing?") He wasn't aware of any other specs that did something similar, and recommended a "manual V-Table dispatch" (because we are oh so *painfully* geeky) like so:
**requestReferenceSpace algorithm:**
  1. If the [=reference space is supported=] of type |type| etc etc...

Then, near the reference space definitions, define the support tests if they're simple enough or link off to a new algorithm if they're complex:
**reference space is supported:**
  1. If |type| is {{inline}} return true.
  1. If |type| is {{local}} return true if |session| is an [=immersive session=] or etc. etc...
  1. If |type| is {{bounded}} return if [=bounded reference spaces are supported=]

And then in a contextually relevant spot:
**bounded reference spaces are supported:**
  1. If the [=XR device=] cannot report boundaries, return false.
  1. If the [=XR device=] cannot identify the height of the user's physical floor, return false.
  1. Return true.

(Obviously I'm being pretty hand-wavy with all of that.) Sorry that this is turning out to be a bigger issue than it first appeared.

index.bs
Manishearth: Notes like this seem to work fine, they get indented. A couple specs do this

Manishearth:
Thanks, I don't mind being more explicit in spec text :) I wasn't quite comfortable with the compact text in the current PR either but I didn't know enough about the various criteria for support to write more. This seems good!

index.bs
toji: Well color me surprised! Today I learned something new.

Manishearth:
Redid it. I'm not quite sure what the criteria for inline local and unbounded spaces should be, but I took a guess.

index.bs
```diff
@@ -837,11 +837,26 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
...
-  1. If |type| is {{bounded-floor}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
-  1. Else let |referenceSpace| be a new {{XRReferenceSpace}}.
-  1. Initialize |referenceSpace|'s [=XRReferenceSpace/type=] to be |type|.
-  1. Initialize [=XRSpace/session=] to be |session|.
-  1. Return |referenceSpace|.
+  1. If the [=reference space is supported=] for |type| and |session|, run the following steps:
+    1. If |type| is {{bounded-floor}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
+    1. Else let |referenceSpace| be a new {{XRReferenceSpace}}.
+    1. Initialize |referenceSpace|'s [=XRReferenceSpace/type=] to be |type|.
+    1. Initialize [=XRSpace/session=] to be |session|.
```
toji: Should say `Initialize |referenceSpace|'s [=XRSpace/session=] to be |session|.`

index.bs
```diff
@@ -837,11 +837,26 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
...
+
+<div class="algorithm" data-algorithm="reference-space-supported">
+To check if a <dfn>reference space is supported</dfn> for a given reference space type |type| and {{XRSession}} |session|, run the following steps:
+
+
+  1. If |type| is {{inline}}, return <code>true</code>.
+  1. If |type| is {{local}} or {{local-floor}}, and |session| is an [=immersive session=], return <code>true</code>.
+  1. If |type| is {{local}} or {{local-floor}}, and the [=/XR device=] supports reporting orientation data, return <code>true</code>.
+  1. If |type| is {{bounded-floor}} and |session| is an [=immersive session=], return if [=bounded reference spaces are supported=] by the [=/XR device=].
+  1. If |type| is {{unbounded}}, |session| is an [=immersive session=], and the [=/XR device=] supports reporting position and orientation data, return <code>true</code>.
```
toji: I'm don't think this is an adequate requirement for `unbounded`, but I'm not sure how to describe it better. @thetuvix, do you have any thoughts about an appropriate support test here?

index.bs
toji: This doesn't look like it was actually resolved in the latest commit?

Manishearth: no idea what happened there, but fixed

index.bs
toji: For the sake of moving this PR forward, I propose we change this to something along the lines of "supports stable tracking near the user over an unlimited distance, return <code>true</code>."

If there's a more accurate way to describe the requirements for an `unbounded` space then we can fix it up later, but that should be good enough for now and prevent implementers from saying "Oh, by this definition my Vive *totally* supports `unbounded`!"

Manishearth: Fixed.

toji:
Awesome, thanks!

index.bs
```diff
@@ -877,6 +892,15 @@ Each {{XRBoundedReferenceSpace}} has a <dfn for="XRBoundedReferenceSpace">native
 
 The <dfn attribute for="XRBoundedReferenceSpace">boundsGeometry</dfn> attribute is an array of {{DOMPointReadOnly}}s such that each entry is equal to the entry in the {{XRBoundedReferenceSpace}}'s [=XRBoundedReferenceSpace/native bounds geometry=] premultiplied by the {{XRRigidTransform/inverse}} of the [=XRSpace/origin offset=]. In other words, it provides the same border in {{XRBoundedReferenceSpace}} coordinates relative to the [=XRSpace/effective origin=].
 
+
+<div class="algorithm" data-algorithm="bounded-space-supported">
```
NellWaliczek: @thetuvix are you on board with this language?  If not, please feel free to propose an alternative when you update your tracking loss PR.

thetuvix: This seems like a reasonable approach.  If the device can produce a floor-based coordinate system, but doesn't know boundaries, it should only support `local-floor`.

If the device somehow knows boundaries without knowing the floor, I suppose the UA could estimate the floor as with `local-floor`, but I don't know of any such devices today, so that would be more theoretical.  We can always relax this later.

index.bs
```diff
@@ -837,11 +837,26 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
...
+
+</div>
+
+<div class="algorithm" data-algorithm="reference-space-supported">
+To check if a <dfn>reference space is supported</dfn> for a given reference space type |type| and {{XRSession}} |session|, run the following steps:
+
+
+  1. If |type| is {{inline}}, return <code>true</code>.
+  1. If |type| is {{local}} or {{local-floor}}, and |session| is an [=immersive session=], return <code>true</code>.
+  1. If |type| is {{local}} or {{local-floor}}, and the [=/XR device=] supports reporting orientation data, return <code>true</code>.
```
thetuvix: To confirm, apps should be able to get a `local` reference space even before the session becomes immersive?

toji: Sessions do not "become" immersive, they are a single mode for their lifetime. Inline sessions should be able to create `local` reference spaces, assuming the UA has properly handled the privacy implications (ongoing discussion).

PR #652, opened 2019-05-16
Document the formatting of Gamepad ids

### LINKED ISSUE #550: Format of Gamepad identifiers
I wanted to separate this discussion out from the rest of the gamepad-as-the-input-solution work, because it's a reasonably complex topic on it's own.

To recap a bit of the prior discussion here, first off there's some guidelines that we've already set that should be considered (taken from #499):

- The `id` MAY be `'unknown'` if the type of input source cannot be reliably identified or the UA determines that the input source type must be masked for any reason. Applications should render a generic input device in this case.
- Inline sessions MUST only expose `id`s of `'unknown'`.
- Otherwise the `id` must be a lower-case string that describes the physical input source.
- It must not include an indication of the handedness of the input source (such as `oculus-touch-left`), as that can be determined from the `handedness` attribute.

Beyond that, I had previously had text that said "For most devices this SHOULD be of the format `<vendor>-<product-id>`. For example: `oculus-touch`." But this is an unfortunately loose guideline and could result in manual work being done by the UA to expose every device. Ideally we would want a system that surfaces a name provided by the underlying platform with minimal modification.

To that end @thetuvix had previously suggested we use something like the hex representation of USB vendor and product IDs, resulting in strings such as `045e-065d`. That's not human readable, but at least it's automated and consistent. My primary concern with it, however, is that not all platforms will surface their input devices in a way that provides access to USB metadata like this (and depending on the OS/device it may not have said metadata at all.)

Some of the native APIs have built-in methods for exposing device identifiers of various forms. For example, in OpenVR you can query a "Prop_RenderModelName_String", "Prop_ModelNumber_String", and "Prop_ManufacturerName_String". Given that we expect to have multiple different devices exposing themselves through such and interface it would make sense to use those. However, that could run into some issues if, for example, the strings they use for an Oculus Touch are different than what's reported when using the Oculus API directly. Also, if the strings reported from the API have the device handedness baked in it would be difficult to scrub that out in a consistent and safe way.

Finally, some APIs may not report a device identifier at all, since there's assumed to only be a single type of controller that they'll be used with. In that case we don't have to report `unknown` since we would know what the device is, but it'll be up to the UA to make up a value.

I don't have any great suggestions for this aside from suggesting we just do `<manufacturer>-<model>`, pull those values from the API whenever possible, and if the best way to consume them is USB IDs then so be it. This is effectively the approach that Nell has been taking to the names in the [Gamepad Mappings repo](https://github.com/immersive-web/xr-gamepad-mappings/tree/master/mappings/WebXR) thus far, if you want an idea of what that would look like.

Any suggestions to allow this to be more automated/consistent are appreciated!

toji:
/fixes #550

Describes how the layout and model of a device should be encoded into
the `gamepad.id`. Does not add a registry identifier yet because we need
to establish the gamepad registry properly before that can be accurately
documented.

Also: I finally got to use Bikeshed's railroad diagram feature! *Toot toot!*

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1311,7 +1311,29 @@ The {{XRInputSource/gamepad}}'s {{Gamepad/id}} also enforces additional behavior
   - If the input device cannot be reliably identified the {{XRInputSource/gamepad}}'s {{Gamepad/id}} MUST be <code>"unknown"</code>.
   - If the UA masks the input device type for any reason the {{XRInputSource/gamepad}}'s {{Gamepad/id}} MUST be <code>"unknown"</code>
 
-ISSUE: <a href="https://github.com/immersive-web/webxr/issues/550">GitHub #550</a> - Additional restrictions, still under discussion, will be applied to the formatting of the {{XRInputSource/gamepad}}'s {{Gamepad/id}} attribute. Until those restrictions have been identified, all {{XRInputSource/gamepad}} {{Gamepad/id}}'s should default to <code>"unknown"</code>.
+When a {{XRInputSource/gamepad}}'s {{Gamepad/id}} is not "unknown", it MUST conform to the following format:
+
+<pre class='railroad'>
+N: layout
+Opt:
+  Seq:
+    T: :
+    N: model
+</pre>
+
+The <dfn for="gamepad id">layout</dfn> indicates the capabilities and mapping of inputs to the {{Gamepad}} {{Gamepad/buttons}} and {{Gamepad/axes}}. Multiple different types of physical device MAY report the same layout as long as they are functionally equivalent and use identical {{Gamepad}} mappings. The [=gamepad id/layout=] string MUST be lowercase and contain no spaces or colons, with separate words concatenated with a hyphen (<code>-</code>) character. If multiple UAs expose the same device, they SHOULD make an effort to report the same [=gamepad id/layout=] string.
+
+<p class="note">
+To illustrate the above principles, all Windows Mixed Reality controllers should share a single [=gamepad id/layout=] string, such as <code>"windows-mixed-reality"</code>, because they all share the same inputs and are exposed the same way, even though there are several different models of the controller.
+</p>
+
+The <dfn for="gamepad id">model</dfn>, if present, indicates the exact vendor and model of the physical device. It is intended to allow applications to determine the correct mesh to display to users in order to match the device they are holding. It MUST be separated from the [=gamepad id/layout=] string by a colon (<code>:</code>) character, and MUST be a lowercase string containing no spaces or colons, with separate words concatenated with a hyphen (<code>-</code>) character. If the platform provides an appropriate identifier, such as a USB vendor and product ID, it SHOULD be used. Otherwise a descriptive name should be chosen, using the vendors preferred verbiage when possible. If multiple UAs expose the same device, they SHOULD make an effort to report the same [=gamepad id/model=] string.
+
+If a [=gamepad id/model=] cannot be determined, it MUST be omitted from the {{Gamepad/id}}, along with the preceding colon.
+
+<p class="note">
+If a UA is able to identify that an input device conforms to the capabilities and layout of an Oculus Touch controller, but is unable to identify the model of device, it may report an {{Gamepad/id}} of <code>"oculus-touch"</code>. If the UA was, however, able to determine the model of the device, it should report an {{Gamepad/id}} along the lines of <code>"oculus-touch:2nd-gen"</code>.
+</p>
 </section>
 
 <section class="unstable">


index.bs
```diff
@@ -1311,7 +1311,29 @@ The {{XRInputSource/gamepad}}'s {{Gamepad/id}} also enforces additional behavior
...
+
+If a [=gamepad id/model=] cannot be determined, it MUST be omitted from the {{Gamepad/id}}, along with the preceding colon.
+
+<p class="note">
+If a UA is able to identify that an input device conforms to the capabilities and layout of an Oculus Touch controller, but is unable to identify the model of device, it may report an {{Gamepad/id}} of <code>"oculus-touch"</code>. If the UA was, however, able to determine the model of the device, it should report an {{Gamepad/id}} along the lines of <code>"oculus-touch:2nd-gen"</code>.
```
jacobcdewitt: Is the intention of this section to specifically describe how to set the ID for an Oculus Touch controller, or is it meant to be an example of how to set ID strings depending on what is/isn't known about the controller? In the latter case, starting the note with "example:" would make it more clear.

toji: It is indeed intended to only be an example, and I struggled with how to communicate that appropriately. (Obviously I didn't do a great job at it if you're asking. 😉) I'll take another stab at this. Maybe it's time to make up imaginary VR devices/controllers?

index.bs
```diff
@@ -1311,7 +1311,29 @@ The {{XRInputSource/gamepad}}'s {{Gamepad/id}} also enforces additional behavior
   - If the input device cannot be reliably identified the {{XRInputSource/gamepad}}'s {{Gamepad/id}} MUST be <code>"unknown"</code>.
   - If the UA masks the input device type for any reason the {{XRInputSource/gamepad}}'s {{Gamepad/id}} MUST be <code>"unknown"</code>
 
-ISSUE: <a href="https://github.com/immersive-web/webxr/issues/550">GitHub #550</a> - Additional restrictions, still under discussion, will be applied to the formatting of the {{XRInputSource/gamepad}}'s {{Gamepad/id}} attribute. Until those restrictions have been identified, all {{XRInputSource/gamepad}} {{Gamepad/id}}'s should default to <code>"unknown"</code>.
+When a {{XRInputSource/gamepad}}'s {{Gamepad/id}} is not "unknown", it MUST conform to the following format:
```
NellWaliczek: For whatever reason I thought the "it MUST conform" was applying to the gamepad not the id and I was super confused for a moment...

index.bs
```diff
@@ -1311,7 +1311,29 @@ The {{XRInputSource/gamepad}}'s {{Gamepad/id}} also enforces additional behavior
...
+
+<pre class='railroad'>
+N: layout
+Opt:
+  Seq:
+    T: :
+    N: model
+</pre>
+
+The <dfn for="gamepad id">layout</dfn> indicates the capabilities and mapping of inputs to the {{Gamepad}} {{Gamepad/buttons}} and {{Gamepad/axes}}. Multiple different types of physical device MAY report the same layout as long as they are functionally equivalent and use identical {{Gamepad}} mappings. The [=gamepad id/layout=] string MUST be lowercase and contain no spaces or colons, with separate words concatenated with a hyphen (<code>-</code>) character. If multiple UAs expose the same device, they SHOULD make an effort to report the same [=gamepad id/layout=] string.
```
NellWaliczek: >If multiple UAs expose the same device, they SHOULD make an effort to report the same [=gamepad id/layout=] string.

Is this part intended to be replaced by the registry?

index.bs
```diff
@@ -1311,7 +1311,29 @@ The {{XRInputSource/gamepad}}'s {{Gamepad/id}} also enforces additional behavior
...
+
+<p class="note">
+To illustrate the above principles, all Windows Mixed Reality controllers should share a single [=gamepad id/layout=] string, such as <code>"windows-mixed-reality"</code>, because they all share the same inputs and are exposed the same way, even though there are several different models of the controller.
+</p>
+
+The <dfn for="gamepad id">model</dfn>, if present, indicates the exact vendor and model of the physical device. It is intended to allow applications to determine the correct mesh to display to users in order to match the device they are holding. It MUST be separated from the [=gamepad id/layout=] string by a colon (<code>:</code>) character, and MUST be a lowercase string containing no spaces or colons, with separate words concatenated with a hyphen (<code>-</code>) character. If the platform provides an appropriate identifier, such as a USB vendor and product ID, it SHOULD be used. Otherwise a descriptive name should be chosen, using the vendors preferred verbiage when possible. If multiple UAs expose the same device, they SHOULD make an effort to report the same [=gamepad id/model=] string.
```
NellWaliczek: nit: "mesh" is a bit too specific.  It's really the model because it includes the materials, etc.

index.bs
NellWaliczek: Maybe just start the paragraph with "For example"?

toji:
This direction was decided against at the June F2F. Will be creating a new PR soon based on the latest comments in #550.

Issue #582, opened 2019-04-05
Describe in detail the way used to obtain XRRay.matrix


bialpio:
The way it's currently specified, the matrix that's supposed to be returned from XRRay.matrix is not unique - there are infinitely many ways of transforming one ray into another as the roll is not constrained in this problem. The specification should describe exactly how the XRRay.matrix should be calculated to prevent interoperability issues between different implementations.

NellWaliczek:
@thetuvix IIRC, you shared a similar concern?  Would you mind elaborating on the explanation you gave me the other day?

thetuvix:
My concern (https://github.com/immersive-web/webxr/pull/569#discussion_r277617621) has been more general, as to whether the WebXR `matrix` attributes are likely to create more convenience or confusion for applications.

Even among those of us who live and breathe the WebXR spec every day, I've seen us struggle on a regular basis to recall whether `matrix` attributes should be a **"transform from"** or a **"transform to"** among the two spaces in question.  This just came up again today in https://github.com/immersive-web/webxr/pull/649#issuecomment-492772969.  Meanwhile, I haven't seen such ambiguity around the `position`/`orientation` attributes on `XRRigidTransform` (in either WebXR or OpenXR) or the `origin`/`direction` attributes on `XRRay` - those are all just interpreted as **"coordinates in"** the base space.  Given this, I expect that any `matrix` convention that we choose and rigorously specify will create confusion as compared to `position`/`orientation` or `origin`/`direction`, unless we are following a strong existing convention used elsewhere on the web.

That said, it could still be worth introducing and strictly defining some convention for these `matrix` attributes if that then enabled apps to directly operate on the `Float32Array` more efficiently than creating a matrix from the `position` vector and `orientation` quaternion.  However, if we do define some WebXR matrix convention here, the representation we choose must then exactly match the internal matrix representation used by the app's 3D math library (left-handed vs. right-handed? column-major vs. row-major? Y-up vs. Z-up?) - otherwise, the app will need to transform the matrix data WebXR gives them anyway.

@toji mentioned in https://github.com/immersive-web/webxr/pull/609#discussion_r279578804 that WebGL does not specify any particular matrix representation.  However, the [following claim in the spec](https://github.com/immersive-web/webxr/blob/a3dd6831f383e8a8d1f6f070b31de0bca6ecc0e5/index.bs#L970) does assert that the format WebXR uses has significant alignment with WebGL, `DOMMatrix` and third-party math libraries:

[...]

PR #654, opened 2019-05-17
Some matrix clarifications

Manishearth:
We've been circling around whether transforms are "transforms from" or "transforms to" a bunch, see https://github.com/immersive-web/webxr/issues/582#issuecomment-492879145 , https://github.com/immersive-web/webxr/pull/649#issuecomment-492772969 for the latest examples.

These are some clarifications that may help.

I have a potential fix for #582 ready as well, but I want to land that separately since that actually adds new semantics as opposed to clarifying existing semantics.

r? @toji @NellWaliczek

cc @klausw @thetuvix

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -743,11 +743,11 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method 
 
 <div class="algorithm unstable" data-algorithm="get-pose">
 
-When the <dfn method for="XRFrame">getPose(|sourceSpace|, |destinationSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
+When the <dfn method for="XRFrame">getPose(|space|, |referenceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
 
   1. Let |frame| be the target {{XRFrame}}
   1. Let |pose| be a new {{XRPose}} object.
-  1. [=Populate the pose=] of |sourceSpace| in |destinationSpace| at the time represented by |frame| into |pose|.
+  1. [=Populate the pose=] of |space| in |referenceSpace| at the time represented by |frame| into |pose|.
   1. Return |pose|.
 
 </div>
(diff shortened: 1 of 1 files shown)


Manishearth:
While I was working on this I realized the fundamental issue was that we're thinking about these transforms as transforms "between spaces", which is what's ambiguous (since you can transform the origins of the spaces, or you can transform the coordinates of objects in the spaces).

This uniformly makes us talk of these as positions and orientations floating in space, among other things.

Manishearth:
When I first wanted to do this PR it had a lot more things that needed doing, fortunately all the recent spec work has pinned down _most_ of the ambiguous bits around this already.

index.bs
```diff
@@ -1039,6 +1039,8 @@ The <dfn attribute for="XRRigidTransform">orientation</dfn> attribute is a quate
 
 The <dfn attribute for="XRRigidTransform">matrix</dfn> attribute returns the transform described by the {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}} attributes as a [=matrix=]. This attribute SHOULD be lazily evaluated.
 
+NOTE: This matrix when premultiplied with a column vector will rotate the vector by the 3D rotation described by {{XRRigidTransform/orientation}}, and then translate it by {{XRRigidTransform/position}}. Mathematically in column-vector notation, this is <code>M = T * R</code>, where <code>T</code> is a translation matrix corresponding to {{XRRigidTransform/position}} and  <code>R</code> is a rotation matrix corresponding to {{XRRigidTransform/orientation}}.
```
klausw: How about "This matrix premultiplied onto a column vector"? "premultiplied with a vector" sounds backwards since the vector is on the right.

index.bs
```diff
@@ -1126,10 +1129,10 @@ An {{XRPose}} describes a position and orientation in space relative to an {{XRS
 };
 </pre>
 
-The <dfn attribute for="XRPose">transform</dfn> attribute describes the {{XRRigidTransform}} between two {{XRSpace}}.
+The <dfn attribute for="XRPose">transform</dfn> attribute describes the position and orientation relative to the underlying {{XRSpace}}.
```
klausw: There's some potential confusion here since the getPose definition uses sourceSpace and destinationSpace in its arguments, and it's not obvious which of these is the "underlying" space. 

How about using `relative to the reference {{XRSpace}}` here, and changing the getPose definition to match? Something like this?

```
XRPose? getPose(XRSpace space, XRSpace referenceSpace);
...
When the <dfn method for="XRFrame">getPose(|space|, |referenceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
...
  1. [=Populate the pose=] of |space| in |referenceSpace| at the time represented by |frame| into |pose|.
 ```

Not sure what the best name is for the first argument. The proposal here is roughly following the explainer which uses `getPose(xrSpace, xrReferenceSpace)`. Other options would be `posedSpace`, `objectSpace`, `trackedSpace`?

File: index.bs [hunk 3]
```diff
@@ -1126,10 +1129,10 @@ An {{XRPose}} describes a position and orientation in space relative to an {{XRS
...
 
 <section class="unstable">
-The <dfn attribute for="XRPose">emulatedPosition</dfn> attribute MUST be <code>false</code> if the {{XRPose/transform}}.{{XRRigidTransform/position}} values are based on sensor readings, or <code>true</code> if the position values are software estimations, such as those provided by a neck or arm model.
+The <dfn attribute for="XRPose">emulatedPosition</dfn> attribute MUST be <code>false</code> if the {{XRPose/transform}}. {{XRRigidTransform/position}} values are based on sensor readings, or <code>true</code> if the position values are software estimations, such as those provided by a neck or arm model.
```
klausw: Accidentally added space?

index.bs
Manishearth: Oh, I like this. This better clarifies the source/destination thing too.

Manishearth:
Addressed

index.bs
```diff
@@ -743,11 +743,11 @@ When the <dfn method for="XRFrame">getViewerPose(|referenceSpace|)</dfn> method
 
 <div class="algorithm unstable" data-algorithm="get-pose">
 
-When the <dfn method for="XRFrame">getPose(|sourceSpace|, |destinationSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
+When the <dfn method for="XRFrame">getPose(|space|, |referenceSpace|)</dfn> method is invoked, the user agent MUST run the following steps:
```
toji: I'm slightly concerned that using "referenceSpace" here may erroneously imply that the input can ONLY be an `XRReferenceSpace`. That said, naming is hard and there's not a lot a great options that I can think of here that address what you and Klaus were talking about. So maybe it's a non-issue?

Manishearth: Yeah, that's fair. I think ultimately it's clear from the types, though.

_shrug_

toji: Hey, I jumped the gun on this and merged it before Nell had a chance to sign off, and she shares my concern about using `referenceSpace` as a name here. As a result I'm going to do a follow up to adjust the name again. I'll post a PR when I have something so we can make sure the name is agreeable all around.

Manishearth: Sounds good. You can just revert https://github.com/immersive-web/webxr/pull/654/commits/a8563000d2857d9912cc574cafbb79d4b1e0cc3b

klausw:
[bikeshed]
How about baseSpace, coordinateSpace, or underlyingSpace?
[/bkeshed]

toji:
Great suggestions, thanks!

thetuvix:
This is a great change!  Once we start to include offset spaces, anchor spaces, etc., I've found it far more straightforward to reason about the pose of one space "in" another.  Aligning to that model within our API naming should help the developers using WebXR as well!

At the risk of Brandon calling me out for aligning WebXR to OpenXR :smile:, I'll put in my vote for `baseSpace`. (I share the concerns around developer confusion if we call the parameter `referenceSpace`)

We can then describe `getPose` to developers simply:
> `getPose(space, baseSpace)` returns the pose of `space` within `baseSpace` at the time represented by this `XRFrame`.

toji:
Thanks all! PR to change the name is now up at #659

PR #655, opened 2019-05-17
Make XRRay.matrix unique, add steps for obtaining it

### LINKED ISSUE #582: Describe in detail the way used to obtain XRRay.matrix

Manishearth:
I'm not 100% sure if I like the existence of this attribute, but I took a stab at making it conceptually unique. While there are many 3D rotations between two vectors, there is only one 2D rotation (and also, it is the easiest to compute -- most algorithms for "rotation between two vectors" will give you this). Such a rotation will basically keep the perpendicular to `direction` and `-z` unchanged.

I've added text specifying that the matrix must be of this type, as well as a partial algorithm for computation (I don't elaborate on how the axis-angle rotation matrix is to be obtained)

r? @NellWaliczek

fixes #582

cc @thetuvix @bialpio

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1106,9 +1106,25 @@ The <dfn attribute for="XRRay">origin</dfn> attribute defines the 3-dimensional 
 The <dfn attribute for="XRRay">direction</dfn> attribute defines the ray's 3-dimensional directional vector. The {{XRRay/direction}}'s {{DOMPointReadOnly/w}} attribute MUST be <code>0.0</code> and the vector MUST be normalized to have a length of <code>1.0</code>.
 
 <section class="unstable">
-The <dfn attribute for="XRRay">matrix</dfn> attribute is a [=matrix=] which represents the transform from a ray originating at <code>[0, 0, 0]</code> and extending down the negative Z axis to the ray described by the {{XRRay}}'s {{XRRay/origin}} and {{XRRay/direction}}.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1106,9 +1106,25 @@ The <dfn attribute for="XRRay">origin</dfn> attribute defines the 3-dimensional
 The <dfn attribute for="XRRay">direction</dfn> attribute defines the ray's 3-dimensional directional vector. The {{XRRay/direction}}'s {{DOMPointReadOnly/w}} attribute MUST be <code>0.0</code> and the vector MUST be normalized to have a length of <code>1.0</code>.
 
 <section class="unstable">
-The <dfn attribute for="XRRay">matrix</dfn> attribute is a [=matrix=] which represents the transform from a ray originating at <code>[0, 0, 0]</code> and extending down the negative Z axis to the ray described by the {{XRRay}}'s {{XRRay/origin}} and {{XRRay/direction}}.
+The <dfn attribute for="XRRay">matrix</dfn> attribute is a [=matrix=] which represents the transform from a ray originating at <code>[0, 0, 0]</code> and extending down the negative Z axis to the ray described by the {{XRRay}}'s {{XRRay/origin}} and {{XRRay/direction}}. Such a matrix MUST be one that has a rotation component which leaves any vector perpendicular to {{XRRay/direction}} and the <code>Z</code> axis unchanged. This attribute MUST be computed by [=XRRay/obtain the matrix|obtaining the matrix=] for the {{XRRay}}. This attribute SHOULD be lazily evaluated.
```
NellWaliczek: >Such a matrix MUST be one that has a rotation component which leaves any vector perpendicular to {{XRRay/direction}} and the <code>Z</code> axis unchanged.

I'm pretty sure I get what you're trying to say here, but this was a bit confusing to me.  Perhaps just because 3D math isn't my bread-and-butter.  Would it be simpler to refer to the cross product?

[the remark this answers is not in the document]
Manishearth: Will clarify!

Manishearth: The cross product has no direction when the vectors are parallel, so I'd have to have an extra bit of text saying that the rotation is zero in such a case. This phrasing conveniently avoids this problem since when dir and Z are parallel, there are infinite perpendicular vectors, and keeping them all unchanged implies rotation=0.

toji:
Thanks Manish! LGTM too.

Issue #341, opened 2018-04-11
Trying to avoid additional canvas when mirroring


TrevorDev:
I'm on the BabylonJS team and we are working on adding WebXR support. When prototyping I ran into an issue when trying to avoid the use of multiple canvases.

I managed to modify Babylon to get prototypes working 🙏 in chrome canary by following these samples:
https://github.com/immersive-web/webxr-samples/blob/master/magic-window.html 
https://github.com/immersive-web/webxr-samples/blob/master/mirroring.html
but am wondering if there is a way to reuse an existing canvas for mirroring/magic window scenarios instead of overlaying canvases like in the samples.

Currently most Babylon projects start with a single canvas and a webGL context to render a scene. When starting to displaying on the hmd in XR I can set the xr device on the existing webGL context to render to the headset with. However, based on the samples, to mirror the display another canvas is needed to be created with a context of xrpresent. Since the original canvas is no longer used for displaying is there any method to use it as the mirroring canvas to avoid the additional complexity of managing multiple canvas/contexts or is there something I'm missing?

brianchirls:
I'm with @TrevorDev on this one. I'm in the process of porting my business/enterprise data viz applications over to XR and am finding this aspect of it extremely cumbersome, particularly when it comes to "spectator" mode. There are *very* few examples out there that do mirroring, much less a third-person spectator view. Three.js (and by extension, A-Frame) doesn't support it in core, and @toji's [samples](https://github.com/immersive-web/webxr-samples) are good but artificial. So it doesn't seem like the use case has been well tested in practice.

For non-entertainment, non-gaming applications, assume that XR is not the primary concern. Most users will want a screen view first and only optionally enter XR later. They will almost always want a spectator view if the device supports it. Further, almost all WebVR/WebXR work in the wild has a full-window canvas. For business applications, the canvas is likely to be part of a greater page layout, so it's not as simple as just setting everything to `window.innerWidth` by `window.innerHeight`. Layout becomes difficult to reason about, and keeping track of multiple canvases only makes it worse.

I'm sorry I haven't had time to be involved in these discussions the past several months. I should be more available now, if someone wants to talk through these use cases in more detail.

toji:
FWIW this is an issue that may be made significantly easier with an upcoming proposal for allowing a single session to transition between inline and immersive modes. (Wouldn't solve the issue entirely but I do think it would make it more tractable.)

PR #656, opened 2019-05-17
Single-canvas inline, drop XRPresentationContext

### LINKED ISSUE #636: Simplification: Enable single-canvas inline, defer desktop mirroring
I'd like to propose that we seriously evaluate allowing a single-canvas method of rendering inline content, deferring the mirroring feature on desktop, (and as a result also deferring the `XRPresentationContext` and `XRRenderState.outputContext` attribute) till after the core spec is complete.

**Reason:** I think I've personally heard more developers express confusion about how `XRPresentationContext` works and why it's required for `inline` sessions than just about any other feature in the API. Devs seem to understand it slightly better when talking about mirroring, but for `inline` there's a lot of confusion as to why the output doesn't simply go to the original WebGL canvas by default. I frequently end up explaining it in terms of more advanced use cases with galleries of sessions on a single page, but the fact of the matter is that most developers are not going to be pursuing that type of use case, and thus it just sounds like additional work for no benefit to them.

Additionally, the requirement to use an `XRPresentationContext` has been very problematic for developing an efficient polyfill of `inline` features. It forces an awkward series of copies to try and get the image from the WebGL context to the fake 'xrpresent' context that can absolutely tank performance. (It's not as much of an issue for `immersive` sessions, since the polyfill doesn't need to support mirroring on mobile and has to do a copy anyway to get the lens distortion correct.) If we could drop the two-canvas requirement it would make polyfilling the API core much more performant and practical.

For mirroring, we find that pretty much every native API has a built-in mirroring capability. For desktop APIs there's usually a way to open a mirroring window for development, debugging, or streaming. For mobile devices there's often a way to stream to an external device like Chromecasts, which our API won't be able to provide anyway, and as such the mirroring feature is semi-redundant. (And removing it does not prevent spectator rendering or demo-driver use cases.)

Finally, some native implementations (such as Windows Mixed Reality and OpenXR) don't provide a mirror texture, so those platforms would either have to do a second composite in browser in order to show a proper mirror or just show nothing, in which case the mirroring isn't a dependable feature anyway (and we should make sure that we communicate when it's available an when it's not.) 

**Proposal:** We could add a method of creating a WebGL layer that specifically indicates that the WebGL default backbuffer should be used rather than an XR-specific framebuffer. For example, let's assume that it's an option for the `XRWebGLLayer` constructor:

```js
// This is a terrible option name and I will die of shame if we actually use it.
let xrLayer = new XRWebGLLayer(session, gl, { useWebGLDefaultBackbuffer: true });
```

This layer now has several unique properties: All of the framebuffer creation options (`framebufferScaleFactor`, `alpha`, `depth`, etc.) are ignored, because the target framebuffer already exists. The layer's `framebuffer` attribute will be `null` and the `framebufferWidth` and `framebufferHeight` just reflect the `gl.drawingBufferWidth` and `gl.drawingBufferHeight`. This means that you can bind to it and use it like you would a normal WebGL layer in your drawing loop, but you'll just be rendering straight into the WebGL canvas instead. Also, because of that, when we start supporting multiple layers we'd restrict it so that layers of this type cannot be mixed with any other layers since we wouldn't actually be doing any compositing here. It's just a "passthrough" layer of sorts. (And **no.** That can't be the option name. Because AR.) Finally, for obvious reasons, you couldn't create a layer with this option for `immersive` sessions.

The restrictions above sound annoying, but the overall reduction in complexity of both the spec and implementation should be significant. The WebGL behavior is basically free, because it's largely just letting WebGL do what it already does and simply leveraging the layer as means for unifying the rendering of simple, first-gen WebXR content.

**Forward Compatibility:** I want to preserve the capability to add mirroring/advanced inline features back in shortly after the core API is locked down because there ARE cases where it's useful/necessary. Developers who want to show multi-layer content inline would have to use an `XRPresentationContext`, as would developers that want to use a single context with multiple sessions. But at least then we would be saying "You have to do this *because you need these features*" Rather than "You're forced to jump through this hoop because someone else might need to do something fancier some day."

We need to ensure that it can properly be feature detected and could be added back in without interfering with the core. Given that `canvas.getContext()` is specced to return `null` for unknown context names, that gives devs a really clean feature detect point to say "hey, is more advanced inline/mirroring available?" Additionally there will be several detectable attributes (such as `XRRenderState.outputContext`) that developers would be able to check for at that point.

**Conclusion:** This is a large change and probably feels kind of hacky upon initial reading, but I strongly encourage us to look for a way to achieve the above goals, even if my specific proposal isn't taken up. Again, this is probably the single most complained-about aspect of the API in my experience, and I think that we have an opportunity to make the lives of ourselves and developers simpler here.

toji:
/fixes #636

This change adds a path for using a single WebGL canvas to drive inline
sessions, rather than needing both a WebGL context and an
XRPresentationContext before the inline session can function. The way to
do this is to create an `XRWebGLLayer` with the `useDefaultFramebuffer`
option set to `true` and set it as the session's `baseLayer`.

In addition to the above, this change removes the XRPresentationContext
entirely, and along with it the ability to do desktop mirroring (and
more advanced inline scenarios.) This is to simplify the core API, and
equivalent functionality will be reconsidered in the future because the
functionality that they represent is valuable.


toji:
@NellWaliczek noted that I forgot to make the associated explainer changes, so I'll make sure to get those done ASAP.

toji:
Updated the explainer! PTAL!

index.bs
```diff
@@ -646,7 +668,7 @@ When an {{XRSession}} |session| receives updated [=viewer=] state from the [=/XR
 
   1. If |session|'s [=list of pending render states=] is not empty, [=apply pending render states=].
   1. If |session|'s {{XRSession/renderState}}'s {{XRRenderState/baseLayer}} is <code>null</code>, abort these steps.
-  1. If |session|'s  [=XRSession/mode=] is {{XRSessionMode/"inline"}} and |session|'s {{XRSession/renderState}}'s {{XRRenderState/outputContext}} is <code>null</code>, abort these steps.
+  1. If |session|'s  [=XRSession/mode=] is {{XRSessionMode/"inline"}} and |session|'s [=XRSession/output canvas=] is <code>null</code>, abort these steps.
```
NellWaliczek: Still feels worth keeping the output canvas variable on the render state given the steps needed to change its value.

index.bs
```diff
@@ -1489,35 +1512,48 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
...
+    1. Initialize |layer|'s {{XRWebGLLayer/antialias}} to |layerInit|'s {{XRWebGLLayerInit/antialias}} value.
+    1. If |layerInit|'s {{XRWebGLLayerInit/ignoreDepthValues}} value is <code>false</code> and the [=XR Compositor=] will make use of depth values, Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} to <code>false</code>.
+    1. Else Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} to <code>true</code>
+    1. Initialize |layer|'s {{XRWebGLLayer/framebuffer}} to a new [=opaque framebuffer=] created with |context|.
+    1. Initialize the |layer|'s [=swap chain=].
+    1. If |layer|'s [=swap chain=] was unable to be created for any reason, throw an {{OperationError}} and abort these steps.
+  1. Else if |layer|'s [=XRWebGLLayer/use default framebuffer=] boolean is <code>true</code>:
+    1. If |session|'s [=mode=] is not {{XRSessionMode/inline}}, throw an {{InvalidStateError}} and abort these steps.
+    1. Initialize |layer|'s {{XRWebGLLayer/antialias}} to |layer|'s {{XRWebGLLayer/context}}'s [=actual context parameters=] {{WebGLContextAttributes|antialias}} value.
+    1. Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} to <code>true</code>.
```
NellWaliczek: Why require this?

index.bs
```diff
@@ -1489,35 +1512,48 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
...
+  1. If |layer|'s [=XRWebGLLayer/use default framebuffer=] boolean is <code>false</code>:
+    1. Initialize |layer|'s {{XRWebGLLayer/antialias}} to |layerInit|'s {{XRWebGLLayerInit/antialias}} value.
+    1. If |layerInit|'s {{XRWebGLLayerInit/ignoreDepthValues}} value is <code>false</code> and the [=XR Compositor=] will make use of depth values, Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} to <code>false</code>.
+    1. Else Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} to <code>true</code>
+    1. Initialize |layer|'s {{XRWebGLLayer/framebuffer}} to a new [=opaque framebuffer=] created with |context|.
+    1. Initialize the |layer|'s [=swap chain=].
+    1. If |layer|'s [=swap chain=] was unable to be created for any reason, throw an {{OperationError}} and abort these steps.
+  1. Else if |layer|'s [=XRWebGLLayer/use default framebuffer=] boolean is <code>true</code>:
+    1. If |session|'s [=mode=] is not {{XRSessionMode/inline}}, throw an {{InvalidStateError}} and abort these steps.
+    1. Initialize |layer|'s {{XRWebGLLayer/antialias}} to |layer|'s {{XRWebGLLayer/context}}'s [=actual context parameters=] {{WebGLContextAttributes|antialias}} value.
```
NellWaliczek: Where's the equivalent for the other properties such as depth?

index.bs
```diff
@@ -1544,7 +1580,7 @@ The [=native WebGL framebuffer resolution=] is determined by running the followi
 
   1. Let |session| be the target {{XRSession}}.
   1. If |session|'s [=XRSession/mode=] value is not <code>"inline"</code>, set the [=native WebGL framebuffer resolution=] to the resolution required to have a 1:1 ratio between the pixels of a framebuffer large enough to contain all of the session's {{XRView}}s and the physical screen pixels in the area of the display under the highest magnification and abort these steps. If no method exists to determine the native resolution as described, the [=recommended WebGL framebuffer resolution=] MAY be used.
-  1. If |session|'s [=XRSession/mode=] value is <code>"inline"</code>, set the [=native WebGL framebuffer resolution=] to the size of the |session|'s {{XRSession/renderState}}'s {{XRRenderState/outputContext}}'s {{XRPresentationContext/canvas}} in physical display pixels and reevaluate these steps every time the size of the canvas changes or the {{XRRenderState/outputContext}} is changed.
+  1. Else the set the [=native WebGL framebuffer resolution=] to <code>1 x 1</code>.
```
NellWaliczek: Why 1x1?  Should this not be the framebuffer size?

index.bs
```diff
@@ -1450,6 +1472,7 @@ typedef (WebGLRenderingContext or
          WebGL2RenderingContext) XRWebGLRenderingContext;
 
 dictionary XRWebGLLayerInit {
```
NellWaliczek: Should we throw an exception if useDefaultFramebuffer is true and framebuffferScaleFactor isn't 1?

index.bs
```diff
@@ -537,7 +537,32 @@ The <dfn attribute for="XRSession">environmentBlendMode</dfn> attribute returns
 
 NOTE: Most Virtual Reality devices exhibit {{XREnvironmentBlendMode/opaque}} blending behavior. Augmented Reality devices that use transparent optical elements frequently exhibit {{XREnvironmentBlendMode/additive}} blending behavior, and Augmented Reality devices that use passthrough cameras frequently exhibit {{XREnvironmentBlendMode/alpha-blend}} blending behavior.
 
-Each {{XRSession}} has a <dfn for="XRSession">viewer reference space</dfn>, which is an {{XRReferenceSpace}} of type {{XRReferenceSpaceType/viewer}} with an [=identity transform=] [=XRSpace/origin offset=]. The [=XRSession/viewer reference space=] has a <dfn for="XRSession/viewer reference space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=].
+Each {{XRSession}} has a <dfn for="XRSession">output canvas</dfn>, which is an {{HTMLCanvasElement}}. The [=XRSession/output canvas=] is the DOM element that will display any content rendered for an {{XRSessionMode/inline}} {{XRSession}}.
```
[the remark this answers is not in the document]
toji: I'm not sure if I grok what you're referring to re: "a strong statement than I'd been expecting" here. If it's the requirement that an output canvas be present in order for rAFs to process, today we already require that the inline context has an `outputContext`.

Also, To clarify: When `XRPresentationContext` (or similar) is re-introduced in the future it would function as one possible source of an output canvas. So the message here is ultimately that the output canvas is the thing that output will eventually show up on, whether via this "direct output" mechanism or by an `XRPresentationContext`-style interface.

index.bs
```diff
@@ -537,7 +537,32 @@ The <dfn attribute for="XRSession">environmentBlendMode</dfn> attribute returns
 
 NOTE: Most Virtual Reality devices exhibit {{XREnvironmentBlendMode/opaque}} blending behavior. Augmented Reality devices that use transparent optical elements frequently exhibit {{XREnvironmentBlendMode/additive}} blending behavior, and Augmented Reality devices that use passthrough cameras frequently exhibit {{XREnvironmentBlendMode/alpha-blend}} blending behavior.
 
-Each {{XRSession}} has a <dfn for="XRSession">viewer reference space</dfn>, which is an {{XRReferenceSpace}} of type {{XRReferenceSpaceType/viewer}} with an [=identity transform=] [=XRSpace/origin offset=]. The [=XRSession/viewer reference space=] has a <dfn for="XRSession/viewer reference space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=].
+Each {{XRSession}} has a <dfn for="XRSession">output canvas</dfn>, which is an {{HTMLCanvasElement}}. The [=XRSession/output canvas=] is the DOM element that will display any content rendered for an {{XRSessionMode/inline}} {{XRSession}}. 
+
+<div class="algorithm" data-algorithm="determine-output-canvas">
```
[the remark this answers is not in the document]
toji: Derp. Yeah. That would probably be a better place for it, huh?

index.bs
toji: If we do that we'd basically want to state that when this flag is present nothing else can be specified, because it should all defer to the GL context attributes. I'm hesitant to say that it should throw, though, because if we add new options in the future they would then start throwing on newer implementations and being ignored on older implementations. It seems more consistent to just say that everything else is ignored when this flag is present.

index.bs
toji: depth, alpha, and stencil don't have corresponding attributes on the layer, because they are guaranteed to be initialized to the options that were passed in and we don't have attributes that only reflect inputs anymore.

We don't need to do any handling at all of those same options in this specific scenario, because the context will already have it's own context creation attributes set in that regard and we won't be overriding them.

index.bs
toji: I guess we don't have to, but I have a hard time imagining a scenario where 

A) An inline session is making use of the depth buffer.
B) The default depth buffer is allocated in such a way that it could be used in that scenario.

Still, that doesn't mean that CAN'T happen, so it's probably not worth explicitly requiring here.

index.bs
```diff
@@ -537,7 +537,32 @@ The <dfn attribute for="XRSession">environmentBlendMode</dfn> attribute returns
...
+
+An {{XRSessionMode/inline}} {{XRSession}} is considered a <dfn for="XRSession">direct output session</dfn> if rendering commands performed for the {{XRSession}} are executed in such a way that they are directly displayed on the page, rather than being processed by the [=XR Compositor=].
+
+<div class="algorithm" data-algorithm="determine-direct-output-session">
+
+To determine if an {{XRSession}} |session| is a [=direct output session=], the user agent MUST run the following steps:
+
+  1. If |session|'s [=XRSession/mode=] is not {{XRSessionMode/inline}}, return <code>false</code>.
+  1. Let |baseLayer| be |session|'s {{XRSession/renderState}}'s {{XRRenderState/baseLayer}}.
+  1. If |baseLayer| is an instance of an {{XRWebGLLayer}} with a [=XRWebGLLayer/use default framebuffer=] boolean set to <code>true</code>, return <code>true</code>.
```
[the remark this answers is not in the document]
NellWaliczek: Is this still pending?

index.bs
```diff
@@ -560,7 +568,20 @@ The <dfn attribute for="XRSession">environmentBlendMode</dfn> attribute returns
 
 NOTE: Most Virtual Reality devices exhibit {{XREnvironmentBlendMode/opaque}} blending behavior. Augmented Reality devices that use transparent optical elements frequently exhibit {{XREnvironmentBlendMode/additive}} blending behavior, and Augmented Reality devices that use passthrough cameras frequently exhibit {{XREnvironmentBlendMode/alpha-blend}} blending behavior.
 
-Each {{XRSession}} has a <dfn for="XRSession">viewer reference space</dfn>, which is an {{XRReferenceSpace}} of type {{XRReferenceSpaceType/viewer}} with an [=identity transform=] [=XRSpace/origin offset=]. The [=XRSession/viewer reference space=] has a <dfn for="XRSession/viewer reference space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=].
+An {{XRSessionMode/inline}} {{XRSession}} is considered a <dfn for="XRSession">direct output session</dfn> if rendering commands performed for the {{XRSession}} are executed in such a way that they are directly displayed on the page, rather than being processed by the [=XR Compositor=].
```
NellWaliczek: Do we really want to be making a statement that a "session" is a direct output session given that this is really a property of the base layer and the base layer can be changed?  If we're treating this PR as a two-way door decision then it seems like we might want to keep these changes encapsulated within the layer spec text.

index.bs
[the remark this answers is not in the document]
NellWaliczek: Ah, sorry I wasn't clear.  I was surprised that we're stating that a session always has an output canvas, even if it's null for immersive sessions.  I wasn't clear that this was strictly necessary or if instead we could just refer to the properties it was computed off of.  Not a big deal, tbh.

index.bs
NellWaliczek: Ah, actually the bigger issue is that we have those values defined on the initialization dictionary, but we don't actually seem to do anything with them in any of our algorithms even for the opaque framebuffer O_O

index.bs
NellWaliczek: You're totally right about framebufferScaleFactor.  We specified it to be ignorable by the UA which is why it's an attribute on the layer (so developers can know if it was actually applied).

Ok, but then what does that mean about the other flags that actually do defer to the actual GL context attributes?  Are they reported on the gl context itself?  If so, then yeah we can probably gracefully ignore the invalid data.  Probably worth explicitly stating that in the spec though so no one tries to implement a system that attempts to change those properties.

toji:
Thanks, Nell, for pointing out this effectively fixes #341 as well.

index.bs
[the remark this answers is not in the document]
toji: Ah, sorry, bad git commit caused me to miss this. Actually a good thing, though, because I realized afterwards I hadn't fully moved it. That's fixed now.

index.bs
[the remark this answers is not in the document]
toji: It's primarily a convenience reference for the rest of the spec text to point back to and say "I'm computing (width/height/position/etc) off of this thing, which will eventually have a couple of different ways to set it.

index.bs
toji: They are reported on the GL context. Via `gl.getContextAttributes()` specifically. I'll add some text about it.

index.bs
toji: Moved to the `XRRenderState` now, which I agree makes more sense. I also changed it to talk about the render state being in "direct output mode" which sounds more mutable.

index.bs
```diff
@@ -491,6 +491,13 @@ When requested, the {{XRSession}} MUST <dfn>apply pending render states</dfn> by
       <dt> Otherwise
       <dd> Set |activeState|'s [=XRRenderState/output canvas=] to <code>null</code>.
     </dl>
+  1. Set |activeState|'s [=XRRenderState/direct output=] boolean based on the following:
```
NellWaliczek: Can this section be combined with the one above?  Or at the very least the switch be based on whether the "output canvas" property is null?

index.bs
```diff
@@ -568,20 +575,7 @@ The <dfn attribute for="XRSession">environmentBlendMode</dfn> attribute returns
...
-  1. return <code>false</code>.
-
-</div>
-
-Each {{XRSession}} has a <dfn for="XRSession">viewer reference space</dfn>, which is an {{XRReferenceSpace}} of type {{XRReferenceSpaceType/viewer}} with an [=identity transform=] [=XRSpace/origin offset=]. The [=XRSession/viewer reference space=] has a <dfn for="XRSession/viewer reference space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=]. If the {{XRSession}} is a [=direct output session=] the [=list of views=] MUST contain a single [=view=].
+Each {{XRSession}} has a <dfn for="XRSession">viewer reference space</dfn>, which is an {{XRReferenceSpace}} of type {{XRReferenceSpaceType/viewer}} with an [=identity transform=] [=XRSpace/origin offset=]. The [=XRSession/viewer reference space=] has a <dfn for="XRSession/viewer reference space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=]. If the {{XRSession}}'s {{XRSession/renderState}} is in [=XRRenderState/direct output=] mode the [=list of views=] MUST contain a single [=view=].
```
NellWaliczek: nit: not your PR, but perhaps we want to call [=XRSession/viewer reference space=] something like [=XRSession/default viewer reference space=] or [=XRSession/primary viewer reference space=] to distinguish it from other viewer reference spaces that can be created?

NellWaliczek: How about....
```suggestion
Each {{XRSession}} has a <dfn for="XRSession">viewer reference space</dfn>, which is an {{XRReferenceSpace}} of type {{XRReferenceSpaceType/viewer}} with an [=identity transform=] [=XRSpace/origin offset=]. The [=XRSession/viewer reference space=] has a <dfn for="XRSession/viewer reference space">list of views</dfn>, which is a [=/list=] of [=view=]s corresponding to the views provided by the [=/XR device=]. If the {{XRSession}}'s {{XRSession/renderState}}'s [=XRRenderState/direct output=] is set to <code>true</code> the [=list of views=] MUST contain a single [=view=].
```

index.bs
```diff
@@ -628,7 +622,11 @@ dictionary XRRenderStateInit {
 };
 </pre>
 
-Each {{XRRenderState}} has a <dfn for="XRRenderState">output canvas</dfn>, which is an {{HTMLCanvasElement}}. The [=XRRenderState/output canvas=] is the DOM element that will display any content rendered for an {{XRSessionMode/inline}} {{XRSession}}.
+Each {{XRRenderState}} has a <dfn for="XRRenderState">output canvas</dfn>, which is an {{HTMLCanvasElement}} initially set to <code>null</code>. The [=XRRenderState/output canvas=] is the DOM element that will display any content rendered for an {{XRSessionMode/inline}} {{XRSession}}.
+
+Each {{XRRenderState}} also has <dfn for="XRRenderState">direct output</dfn> boolean, which is initially <code>false</code>. The {{XRRenderState}} is considered to be in [=XRRenderState/direct output=] mode if rendering commands performed for an {{XRSessionMode/inline}} {{XRSession}} are executed in such a way that they are directly displayed on the page, rather than being processed by the [=XR Compositor=].
```
NellWaliczek: What about 

>are executed in such a way that they are directly displayed into [=XRRenderState/output canvas=], rather than first being processed by the [=XR Compositor=].

index.bs
NellWaliczek: Can we file an issue to make sure we don't lose track of this needing to get done?

index.bs
NellWaliczek: Did this get added and I missed it?

index.bs
```diff
@@ -1548,40 +1567,58 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
...
     <dl class="switch">
       <dt> If |layerInit|'s {{XRWebGLLayerInit/ignoreDepthValues}} value is <code>false</code> and the [=XR Compositor=] will make use of depth values
       <dd> Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} to <code>false</code>
       <dt> Otherwise
       <dd> Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} to <code>true</code>
     </dl>
-  1. Initialize |layer|'s {{XRWebGLLayer/framebuffer}} to a new [=opaque framebuffer=] created with |context|.
-  1. Initialize the |layer|'s [=swap chain=].
-  1. If |layer|'s [=swap chain=] was unable to be created for any reason, throw an {{OperationError}} and abort these steps.
+  1. Initialize |layer|'s [=XRWebGLLayer/use default framebuffer=] to |layerInit|'s {{XRWebGLLayerInit/useDefaultFramebuffer}}.
```
NellWaliczek: Let's modify this so that inline sessions cannot accept the creation of XRWebGLLayers with useDefaultFramebuffer set to false.

NellWaliczek:
If "useDefaultFramebuffer" is confusing, what about "useDirectOutput" or "disableComposition"?

toji:
Ooh! "disableComposition" or similar is tempting. I do worry that it might be a bit too abstract to be intuitive for developers (Wait! Composition sounds important! Isn't that what makes the canvas show up on the page? I heard a tech talk about that once...) But within the context of WebXR it's sensible.

index.bs
toji: It did get addressed but I apparently didn't push it? Sorry! Change in question was https://github.com/immersive-web/webxr/pull/656/commits/f363b89c6ed504295667df8d3f1be6df86597749

index.bs
toji: Yeah, we can do that. This will need to be torn apart when we get back to more advanced canvas ouput, but that's a problem for future me. :)

index.bs
toji: SGTM!

index.bs
toji: Filed this as #671

toji:
Okay, updated this to use the option name 'compositionDisabled' (putting "disabled" at the end seems like the more common pattern). Thanks for the suggestion! Considered `xrCompositionDisabled` to clarify the concern in my previous comment, but ultimately it just felt redundant since we were setting this on an XRLayer and not an HTML element. Also changed any instances of "direct output" in the document to read "composition disabled" as well, because now that the option is not WebGL-specific it made more sense to use it everywhere.

explainer.md
```diff
@@ -411,31 +377,25 @@ These scenarios can make use of inline sessions to render tracked content to the
...
 
-Similar to mirroring, to make use of this mode the  `XRRenderState`'s `outputContext` must be set. At that point content rendered to the `XRRenderState`'s `baseLayer` will be rendered to the canvas associated with the `outputContext`. The UA is also allowed to composite in additional content if desired. (In the future, if multiple `XRLayers` are used their composited result will be what is displayed in the `outputContext`.)
+To make use of this mode a `XRWebGLLayer` must be created with the `useDefaultFramebuffer` option set to `true`. This instructs the layer to not allocate a new WebGL framebuffer but instead set the `framebuffer` attribute to `null`. That way when `framebuffer` is bound all WebGL commands will naturally execute against the WebGL context's default framebuffer and display on the page like any other WebGL content. When that layer is set as the `XRRenderState`'s `baseLayer` the inline session is able to render it's output to the page.
```
RafaelCintron: In the future, I presume we're going to want inline sessions to participate in new XRLayers types that require composition.  How is that going to work in light of the fact that inline sessions are required to use `useDefaultFramebuffer=true`?  Are any layers besides `XRWebGLLayer` going be ignored in that case? 

I worry about the compatibility aspects of these types of flags. The goal of WebXR is to be a write once, run anywhere deal. When it comes to drawing stuff, the spec should ideally say "You render your stuff to this framebuffer,  we'll take care of composing your layers together with things the browser wants to draw and everything will just work." The more special canvas contexts, useXXXOnlyWorksInInline flags, etc, the more complexity we add for web developers and the more time we'll spend debugging sites that are broken on some hardware but not others.

toji: > In the future, I presume we're going to want inline sessions to participate in new XRLayers types that require composition.

Yes, absolutely!

> How is that going to work in light of the fact that inline sessions are required to use useDefaultFramebuffer=true? Are any layers besides XRWebGLLayer going be ignored in that case?

Once we have new layer types and/or multi-layer the restriction that inline session layers must use that flag would be lifted (and I expect most layers wouldn't have the flag at all.) At that point if a layer IS constructed with the flag it would be an error to use it with any other layers.

> I worry about the compatibility aspects of these types of flags... The more special canvas contexts, useXXXOnlyWorksInInline flags, etc, the more complexity we add for web developers and the more time we'll spend debugging sites that are broken on some hardware but not others.

Completely agree, but in that regard I feel like this is a lateral move rather than a regression. Previously we were asking developers to use a special secondary canvas for inline sessions that was necessary in order to support features that didn't actually exist yet, so we were already making users jump through hoops without any clear benefits to doing so. With this approach we are scaling back to hoop jumping to setting a single boolean option at which point everything works the way they already expect WebGL to work for inline content. Then when those eventual advanced features ARE ready we can introduce the more complex route with a much more tractable explanation of why they're necessary.

To address a comment that I remember from the call, it looks like we could say "turn this flag on automatically for inline" and be done with it. But then we're in a weird position down the road where when layer compositing IS a thing and we expect most content will eventually want to use it, now inline has to flip the same damn flag in reverse for the majority of content just to get the same default behavior as immersive. I'd rather have the default be compositing is used everywhere and for this first release we have this weird little required opt-out wart.

index.bs
```diff
@@ -1592,6 +1592,8 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
 
 </div>
 
+Note: When a {{XRWebGLLayerInit}}'s {{XRWebGLLayerInit/useDefaultFramebuffer}} is set to <code>true</code> all other values on the {{XRWebGLLayerInit}} object are ignored, since the {{WebGLRenderingContext}}'s default framebuffer was already allocated using the context's [=actual context parameters=] and cannot be overridden.
```
NellWaliczek: > [=actual context parameters=]

Is this defined somewhere?

index.bs
NellWaliczek: On this week's call we discussed the option to throw an error if an XRLayer is created without composition disabled on an inline session.. I know at the time you were on the fence... What conclusion did you come to?

index.bs
toji: Yes. When the .bs is built it links here: https://www.khronos.org/registry/webgl/specs/latest/1.0/#actual-context-parameters

explainer.md
NellWaliczek: @rafaelcintron Does that address your concern?  If so, I think we'd like to try to get this merged today or tomorrow...

index.bs
toji: I made it an error. See https://github.com/immersive-web/webxr/pull/656/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR1576

NellWaliczek: Apparently I'm incapable of reading ^_^

explainer.md
```diff
@@ -365,41 +365,7 @@ function checkARSupport() {
...
-        onSessionStarted(session);
-      })
-      .catch((reason) => { console.log("requestSession failed: " + reason); });
-}
-```
-
-### Inline sessions
+## Inline sessions
 
 There are several scenarios where it's beneficial to render a scene whose view is controlled by device tracking within a 2D page. For example:
```
thetuvix: The Tango link on line 373 is no longer live.

explainer.md
RafaelCintron: Since we're removing XRPresentationContext (which I'm fully supportive of), now is an opportune time to eliminate any lingering immersive vs. inline differences as much as we can. 

The way the spec is written today, setting the flag to true means content will not work in immersive sessions. This creates compatibility problems for developers who only have access to inline session hardware. I understand that the API cannot hide all differences between hardware types but having this flag, in my view, does not meet the bar for introducing compatibility problems.

Why can't we eliminate the flag and have the returned framebuffer always be non-null?

explainer.md
toji: Thanks for pointing that out! Replaced with ARCore/ARKit links

explainer.md
toji: I see what you're saying now, but I'm not sure how to achieve that in a way that actually makes developers lives simpler (and doesn't turn into a really sticky spec issue).

If the returned framebufer is always non-null then that means that content rendered to it is going somewhere other than the default backbuffer for the WebGL canvas, and as such needs an explicit mechanism to display it on the page. There's a couple of ways that could happen:

 - We have it bound to some sort of explicit output surface. This is what `xrpresent` did for us, and what we're trying to move away from.
 - We declare that we are taking ownership of the WebGL canvas when a layer is created like this and start displaying our content in favor of the default framebuffer. This seems pretty messy to me, would probably require coordination with the WebGL WG to achive, and at the very least Ken has expressed that he's less than thrilled with the idea when I've floated it with him in the past. There's also ownership lifetime issues to worry about in this model, and it would likely force something like spectator mode to use an OffscreenCanvas or second context altogether.
 - We can require the developer to manually blit the rendered content from our framebuffer (probably surfaced as a texture in that case) into the WebGL default framebuffer. This now requires WAY more work on the developers part to support inline content, and increases the divergence of the two rendering paths, which I think is what we're both trying to reduce.
 - A variant of the previous option, we could provide a way to surface the framebuffer content as an ImageBitmap, which could then be shown directly via a `bitmaprenderer` context or easily converted into a texture for WebGL use. Again, though, this is significantly *more* work in order to get inline to display properly. And we wouldn't want to expose the ImageBitmap producer to immersive sessions because it would either hold references to swap chain surfaces in an environment where we want to control their flow *or* necessitate a copy. And we still run into the core issue that the code paths between immersive and inline and now more divergent than ever.

In light of the above, it seemed to me that flipping a single boolean flag and then having everything else "just work" was an attractive option. I've coded up a couple of tests that use this route already and the code ends up looking like this:

```js
let layer = new WebGLLayer(session, gl, { compositionDisabled: mode == 'inline'; });
```

RafaelCintron: > We have it bound to some sort of explicit output surface. This is what xrpresent did for us, and what we're trying to move away from.

This is along the lines of what I had in mind, but without the extra IDL of XRPresentationContext.

What is your plan for introducing composition to inline sessions in the future?  Won't that require "taking ownership of the WebGL canvas" like you described?  If so, seems like we should have a discussion of the implications sooner rather than later, especially if the flag is going to end up in the final, ratified spec.  Or will inline session never be able to take advantage of composition?  

My high order bit is minimizing the number of differences between inline and immersive for core scenarios.  Without this PR, developers who create xrpresent contexts will at least have the 3D portion of their content work correctly in immersive. From that standpoint, this change is more of a regression than a lateral move.

toji: > This is along the lines of what I had in mind, but without the extra IDL of XRPresentationContext.

I'm really curious what you have in mind for the output surface in that case, without introducing additional IDL?

> What is your plan for introducing composition to inline sessions in the future?

Essentially to re-introduce `XRPresentationContext` or something like it, though I think there's still reasonable design discussions to be had around the shape of that API (hence my curiosity about what you had in mind above). I do strongly feel there's value to having composited output from inline sessions, but it's very hard to justify to developers the need to juggle additional canvases and new interfaces to do so when we don't yet support any scenarios that compositing would actually benefit from (multilayer, etc.)

Also, I missed emphasizing previously that one of the explicit goals of this change is to make the polyfill more performant, and I don't see a way to do that if a secondary output surface is involved, whereas the proposal in this PR makes it trivial. (To be clear: I'm also OK with the polyfill staying limited to the core API)

RafaelCintron: > Essentially to re-introduce XRPresentationContext or something like it, though I think there's still reasonable design discussions to be had around the shape of that API (hence my curiosity about what you had in mind above). 

I see. I didn't realize that you were already counting on inline developers having to change their code in other ways in order to get composition in the future.

When we add `XRPresentationContext` in the future, could we detect "use composition" intent for inline sessions by virtue of developers using `XRPresentationContext` in the WebGL layer's init parameters?  This way, we can remove the flag now instead of having awkward "we may remove this flag in the future" verbiage or get stuck having to support `XRPresentationContext` use without composition down the road.

toji: We don't want the `XRPresentationContext` to be added to the WebGL layer directly, because the entire point is that it accepts the composited results, not the results from a single layer. That's why it was on the session's `XRRenderState` previously, which is where I would be inclined to put it again in the future.

RafaelCintron: OK.  Why can't we use the presence of `XRPresentationContext` on `XRRenderState` to determine whether the developer has opted into composition for inline sessions?

toji:
Nell and I just talked about this change and can see a couple of paths forward for it.

The most conservative path is what's already in this PR. Having an explicit flag that opts _out_ of composition keeps the internal logic the most consistent and introduces no changes to layer logic based on the type of session the layer is created with. It has the downside of introducing a developer-side logic branch based on whether or not the session is inline with the implied promise that that bit of weirdness may go away in the future as we move to more complex layer behavior.

A more developer-friendly but internally "magical" approach would be to omit the flag and have composition disabled for any WebGL layer constructed against an inline session. This means the app code branches less (worth noting that inline vs. immersive still requires some minor logical differences at other points, so this doesn't completely unify the code paths, just gets us closer.) From there, as we expand layer capabilities there's a few viable routes to allow composition of inline layers again:

1. We add a flag to explicitly turn inline layering on, which now means that to get the same behavior as immersive sessions we've introduced a branch, which was the thing that got everyone all riled up in the first place, but at least the initial use case is cleaner.
1. We declare that XRWebGLLayer is forevermore the "weird one", is the only layer type that doesn't support compositing, and if you want more advanced features like inline layering you just have to use a newer style of layer that we've defined in a hypothetical layers module.
1. We allow the layer's framebuffer to flit in and out of existence when a compositing output surface is set on the XRSession. This gives us the more unified behavior in terms of application code, but also introduces "spooky action at a distance" behavior to the API where setting a property on the session magically makes a completely separate object start behaving differently (not to mention start allocating new memory).

The upside is that we don't really have to choose any of the three above paths right now, they're all viable in the future depending on what direction we decide to take with the rest of the API. What really needs to be decided right now is if we want to keep the flag that this PR describes right now (and Chrome 76 happens to implement) or if we want to drop the flag and have the behavior based on session type instead with the understanding that we'll have to make one of the three followup decisions in the future.

(Worth noting that no matter what we chose we're definitely dropping XRPresentationContext from the VR complete milestone, because nobody likes it, its mother was a hamster, and its father smelt of elderberries.)

Nell mentioned that she would appreciate @ddorwin's input on this matter, and I agree. I'd also appreciate a followup from @RafaelCintron. The editors are eager to get this issue wrapped up well before the end of the current milestone, however, and will be arbitrarily choosing a direction and running with it if we haven't reached consensus by the end of the week.

thetuvix:
> We declare that XRWebGLLayer is forevermore the "weird one", is the only layer type that doesn't support compositing, and if you want more advanced features like inline layering you just have to use a newer style of layer that we've defined in a hypothetical layers module.

In a world where we don't generalize the WebXR 1.0 Core module to take an array of layers or to extract WebGL 1 to a `XRLayerSource` abstraction, we already know that `XRWebGLLayer` will be "the weird one" in other ways.  Perhaps we just go all in there?

ddorwin:
@toji and I discussed. He may have more detailed takeaways, but here's a summary of my thoughts:
- It seems odd to have a value that effectively duplicates the mode.
  - The `XRWebGLLayer` constructor takes an `XRSession`, so it already has this information.
- It's not clear that `compositionDisabled` is the right concept and/or that there won't be another signal.
  - For example, using more than one layer will require use of an additional API mechanism, and this could be used as the indication that composition should always be enabled.
  - Supporting multiple layers inline may require doing something different with respect to the <canvas>. See also the next item
- When adding more layers, we should ensure they work in all modes, including inline.
  - It sounds like this will be complex, and we'll need to work with the canvas and/or WebGL folks.
  - It would be a bit unfortunate if we forced developers to put `compositionDisabled` in their apps when we didn't really need it.
  - There shouldn't be a risk that we would want to break backwards compatibility, so relying on the combination of the mode and use of `baseLayer` seems fine.
- It seems reasonable to use _composition disabled_ as an internal state in the algorithms if it helps. This value could be set appropriately by the algorithms related to a future layers mechanism. (Or replaced if that makes more sense and is backwards-compatible.)

toji:
Since it seems increasingly obvious that we're leaning towards an implicit flag based on session mode, I've gone ahead and made that change. The internal "composition disabled" flag is still present (as David mentioned) and all of the logic depending on it is unchanged, I just updated how it gets set in the constructor and removed the explicit flag from the init dictionary.

index.bs
```diff
@@ -1614,7 +1619,7 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
 
 </div>
 
-Note: When a {{XRWebGLLayerInit}}'s {{XRWebGLLayerInit/compositionDisabled}} is set to <code>true</code> all other values on the {{XRWebGLLayerInit}} object are ignored, since the {{WebGLRenderingContext}}'s default framebuffer was already allocated using the context's [=actual context parameters=] and cannot be overridden.
+Note: If a {{XRWebGLLayer}}'s {{XRWebGLLayerInit/composition disabled}} boolean is set to <code>true</code> all values on the {{XRWebGLLayerInit}} object are ignored, since the {{WebGLRenderingContext}}'s default framebuffer was already allocated using the context's [=actual context parameters=] and cannot be overridden.
```
NellWaliczek: Should both of these have been changed to XRWebGLLayer?

toji: Oops! Yes. One sec.

toji: Fixed

RafaelCintron:
I am confused. @toji states: 

> I've gone ahead and made that change. The internal "composition disabled" flag is still present (as David mentioned) and all of the logic depending on it is unchanged, I just updated how it gets set in the constructor and removed the explicit flag from the init dictionary.

In the explainer, I still see `compositionDisabled` listed in the `XRWebGLLayerInit` dictionary and used in sample code.  `compositionDisabled` is not, however, listed in the `XRWebGLLayerInit` dictionary in the official spec.  I presume the spec text is correct and the explainer is wrong?

I agree with the choice of having `compositionDisabled` be an internal-only variable that is not exposed to web developers.  Now that we don't have to obsess with initial dictionary values, consider renaming it 'compositionEnabled' to avoid double negatives.

toji:
That was an oversight. Sorry. Fixed in 2e825a618a3be714d6349e93887d0d861a611662

PR #657, opened 2019-05-17
Properly format conditionals

### LINKED ISSUE #647: Do not use successive steps for handling conditionals in algorithms
The spec currently uses the following pattern for conditionals where there are steps to be run in multiple branches:
>1. Do some stuff.
>1. If foo, do...
>1. Else do...
>1. Do more stuff.

IMO, the steps would be easier to read if the branches were more differentiated from the other steps. See, for example, see steps 10 and 12 of https://w3c.github.io/encrypted-media/#get-supported-configuration-and-consent.

Manishearth:
Fixes #647

Contains no behavior changes, only formatting. In some cases we had "If foo do ... and abort these steps. Else do ...", I just simplified these cases by removing the else.

[Rendered](https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/Manishearth/webxr/conditionals/index.bs)

cc @ddorwin

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -237,8 +237,13 @@ Each time the [=list of XR devices=] changes the user agent should <dfn>select a
   1. Let |oldDevice| be the [=XR/XR device=].
   1. If the [=list of XR devices=] is an empty [=/list=], set the [=XR/XR device=] to null.
   1. If the [=list of XR devices=]'s [=list/size=] is one, set the [=XR/XR device=] to the [=list of XR devices=][0].
-  1. If there are any active {{XRSession}}s and the [=list of XR devices=] [=list/contains=] |oldDevice|, set the [=XR/XR device=] to |oldDevice|.
-  1. Else, set the [=XR/XR device=] to a device of the user agent's choosing.
+  1. Set the [=XR/XR device=] based on the following:
+    <dl class="switch">
+      <dt> If there are any active {{XRSession}}s and the [=list of XR devices=] [=list/contains=] |oldDevice|
+      <dd> Set the [=XR/XR device=] to |oldDevice|
+      <dt> Otherwise
+      <dd> Set the [=XR/XR device=] to a device of the user agent's choosing
+    </dl>
   1. If this is the first time devices have been enumerated or |oldDevice| equals the [=XR/XR device=], abort these steps.
   1. [=Shut down the session|Shut down=] any active {{XRSession}}s.
   1. Set the [=XR compatible=] boolean of all {{WebGLRenderingContextBase}} instances to `false`.
@@ -267,7 +272,7 @@ When the <dfn method for="XR">supportsSession(|mode|)</dfn> method is invoked, i
   1. [=ensures an XR device is selected|Ensure an XR device is selected=].
   1. If the [=XR/XR device=] is null, [=reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}} and abort these steps.
   1. If the [=XR/XR device=]'s [=list of supported modes=] does not [=list/contain=] |mode|, [=reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}} and abort these steps.
-  1. Else [=/resolve=] |promise|.
+  1. [=/Resolve=] |promise|.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -293,18 +298,29 @@ When the <dfn method for="XR">requestSession(|mode|)</dfn> method is invoked, th
...
+    1. Set [=pending immersive session=] to be <code>true</code>.
   1. [=ensures an XR device is selected|Ensure an XR device is selected=].
-  1. If the [=XR/XR device=] is null, [=reject=] |promise| with null.
-  1. Else if the [=XR/XR device=]'s [=list of supported modes=] does not [=list/contain=] |mode|, [=reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}}.
-  1. Else if |immersive| is <code>true</code> and the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a "{{SecurityError}}" {{DOMException}}.
+  1. Follow the steps for the first matching conditions:
+      <dl class="switch">
+      <dt> If the [=XR/XR device=] is <code>null</code>
+      <dd> [=Reject=] |promise| with <code>null</code>.
+      <dt> Else if the [=XR/XR device=]'s [=list of supported modes=] does not [=list/contain=] |mode|
```
ddorwin: Given the intro, I think this can just be "If..."

index.bs
```diff
@@ -293,18 +298,29 @@ When the <dfn method for="XR">requestSession(|mode|)</dfn> method is invoked, th
...
+      <dd> [=Reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}}.
+      <dt> Else if |immersive| is <code>true</code> and the algorithm is not [=triggered by user activation=]
+      <dd> [=Reject=] |promise| with a "{{SecurityError}}" {{DOMException}}.
+    </dl>
```
ddorwin: There should probably be an Otherwise given the intro. It can just say "Continue to the next step." or something like that.

index.bs
```diff
@@ -293,18 +298,29 @@ When the <dfn method for="XR">requestSession(|mode|)</dfn> method is invoked, th
...
     1. Abort these steps.
   1. Let |session| be a new {{XRSession}} object.
   1. [=Initialize the session=] with |session| and |mode|.
-  1. If |immersive| is <code>true</code>, set the [=active immersive session=] to |session|, and set [=pending immersive session=] to <code>false</code>.
-  1. Else append |session| to the [=list of inline sessions=].
+  1. Potentially set the [=active immersive session=] based on the following:
+    <dl class="switch">
+      <dt> If |immersive| is <code>true</code>
+      <dd> Set the [=active immersive session=] to |session|, and set [=pending immersive session=] to <code>false</code>.
+      <dt> Else
```
ddorwin: Line 244 uses Otherwise, which is also used in at least some other specs. Most importantly, this spec should be consistent.

Manishearth:
Addressed

index.bs
```diff
@@ -838,12 +861,19 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
...
   1. If the [=reference space is supported=] for |type| and |session|, run the following steps:
+    1. Initialize |referenceSpace| based on the following:
+      <dl class="switch">
+        <dt> If |type| is {{bounded-floor}}
+        <dd> Let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
+        <dt> Else
+        <dd> Let |referenceSpace| be a new {{XRReferenceSpace}}.
+      </dl>
     1. If |type| is {{bounded-floor}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
     1. Else let |referenceSpace| be a new {{XRReferenceSpace}}.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -838,12 +861,17 @@ When an {{XRReferenceSpace}} is requested, the user agent MUST <dfn>create a ref
   1. Let |session| be the {{XRSession}} object that requested creation of a reference space.
   1. Let |type| be set to the {{XRReferenceSpaceType}} passed to {{requestReferenceSpace()}}.
   1. If the [=reference space is supported=] for |type| and |session|, run the following steps:
-    1. If |type| is {{bounded-floor}}, let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
-    1. Else let |referenceSpace| be a new {{XRReferenceSpace}}.
+    1. Initialize |referenceSpace| based on the following:
+      <dl class="switch">
+        <dt> If |type| is {{bounded-floor}}
+        <dd> Let |referenceSpace| be a new {{XRBoundedReferenceSpace}}.
+        <dt> Else
```
ddorwin: There are six instances of `Else`. Should those be `Otherwise` for consistency?

toji: Honsetly I think `Else` is fine in these cases, especially if it's just a simple `if/else` statement. Where I would expect `Otherwise` to apply is mostly at the end of a switch with several branches. (Functioning as `default` in a C++ switch, essentially)

Manishearth: It seems like other specs use `Otherwise` with this most of the time (as @ddorwin said above)

https://drafts.csswg.org/css-transforms-2/#serialization-of-the-computed-value , https://drafts.csswg.org/web-animations-2/#the-intrinsic-iteration-duration-of-a-group-effect 
https://html.spec.whatwg.org/multipage/dom.html#directionality-of-the-attribute

PR #659, opened 2019-05-20
Rename getPose arg referenceSpace->baseSpace

toji:
Addresses concerns raised after landing #654

The use of `referenceSpace` here may inadvertently suggest that it only
takes an `XRReferenceSpace` object, when in reality it can take any
`XRSpace`. Using `baseSpace` as a replacement at the suggestion of
@klausw and @thetuvix.

Plus it's fun to say: `getPose` finds the place of your space in the
baseSpace's space! 😉

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1201,7 +1201,7 @@ An {{XRPose}} describes a position and orientation in space relative to an {{XRS
 };
 </pre>
 
-The <dfn attribute for="XRPose">transform</dfn> attribute describes the position and orientation relative to the reference {{XRSpace}}.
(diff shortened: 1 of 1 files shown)


toji:
Thanks for catching that!

toji:
@NellWaliczek is the one that initially asked for this change, but is unavailable for reviews for the next few days. In the meantime I'd prefer to get this landed now under the assumption that it's less confusing than the previous verbiage. I'll happily take the blame if Nell takes issue with `baseSpace` as well when she gets back. 😅

PR #660, opened 2019-05-20
Clarify what value a touchpad should report when not being touched.

### LINKED ISSUE #622: What values to use for untouched Gamepad touchpad?
For WebXR Gamepads, there are two possible ways of reporting the input axis values for a touchpad that is not currently touched:
A) last position at which the touchpad was touched
B) origin (0, 0)

After some discussion, @toji and I agree that option B would be the better choice and that the restriction should be documented in the spec. However, we're not sure if this restriction on touchpad behavior should be in the Gamepad spec and apply to all Gamepads or if it should instead be in the WebXR spec as an additional constraint on how WebXR Gamepads must behave.

toji:
/fixes #622

Specifies that touchpads that aren't actively being touched MUST report axes of 0. If/when this policy gets adopted by the Gamepad API we can remove this additional restriction.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1379,6 +1379,7 @@ Gamepad API Integration {#gamepad-api-integration}
   - {{XRInputSource/gamepad}} MUST NOT be included in the array returned by {{navigator.getGamepads()}}.
   - {{XRInputSource/gamepad}}'s {{Gamepad/index}} attribute MUST be <code>0</code>.
   - {{XRInputSource/gamepad}}'s {{Gamepad/connected}} attribute MUST be <code>true</code> until the {{XRInputSource}} is removed from the [=list of active XR input sources=] or the {{XRSession}} is ended.
+  - If an axis reported by the {{XRInputSource/gamepad}}'s {{Gamepad/axes}} array represents an axis of a touchpad, the value MUST be <code>0</code> when the touchpad is not being touched.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1379,6 +1379,7 @@ Gamepad API Integration {#gamepad-api-integration}
   - {{XRInputSource/gamepad}} MUST NOT be included in the array returned by {{navigator.getGamepads()}}.
   - {{XRInputSource/gamepad}}'s {{Gamepad/index}} attribute MUST be <code>0</code>.
   - {{XRInputSource/gamepad}}'s {{Gamepad/connected}} attribute MUST be <code>true</code> until the {{XRInputSource}} is removed from the [=list of active XR input sources=] or the {{XRSession}} is ended.
+  - If an axis reported by the {{XRInputSource/gamepad}}'s {{Gamepad/axes}} array represents an axis of a touchpad, the value MUST be <code>0</code> when the touchpad is not being touched.
```
NellWaliczek: What about
```suggestion
  - If an axis reported by the {{XRInputSource/gamepad}}'s {{Gamepad/axes}} array represents an axis of a touchpad, the value MUST be <code>0</code> when the associated {{GamepadButton/touched}} is <code>false</code>.
```

toji: Good suggestion, thanks!

PR #661, opened 2019-05-20
Better define gamepad placeholder buttons and axes

### LINKED ISSUE #570: Clarify behavior of buttons/axes at reserved indices
The "xr-standard" Gamepad Mapping section (added by #553) says, "Additional buttons or axes may be exposed after these reserved indices, and SHOULD appear in order of decreasing importance." However, the values and/or behavior of the buttons or axes at the reserved indices is not defined.

@toji says there is established behavior for Gamepad API, but it doesn't appear to be documented there either. It would be helpful to have some sort of text in the WebXR spec giving a hint of the intended behavior. It might be ideal if this pointed to general guidance in the Gamepad API spec.

toji:
/fixes #570


index.bs
```diff
@@ -1471,7 +1475,9 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
+Devices that lack one of the inputs listed in the tables above MUST preserve their place in the {{Gamepad/buttons}} or {{Gamepad/axes}} array, reporting a <dfn>placeholder button</dfn> or <dfn>placeholder axis</dfn>, respectively. A [=placeholder button=] MUST report <code>0</code> for {{GamepadButton/value}}, <code>false</code> for {{GamepadButton/pressed}}, and <code>false</code> for {{GamepadButton/touched}}. A [=placeholder axis=] MUST report <code>0</code>. [=Placeholder buttons=] and [=placeholder axis|axes=] MUST be omitted if they are the last element in the array or all following elements are also [=placeholder buttons=] or [=placeholder axis|axes=].
```
NellWaliczek: >Devices that lack one of the secondary inputs

Just a nit to indicate that this applies only to the optional entries.

toji: Done

PR #662, opened 2019-05-21
Update Explainer Text for new requestReferenceSpace modes names

bricetebbs:


dontcallmedom:

toji:
Good catch, thank you!

PR #684, opened 2019-06-05
Handle detached arrays

### LINKED ISSUE #630: Determine how to handle detached Typed Arrays
Matrices are currently returned as a `Float32Array`, which is a view over an `ArrayBuffer`. As it turns out, `ArrayBuffers` can be detached and tossed over a worker boundary, which then makes any views into that buffer invalid. In WebXR terms this means that a matrix that was previously returned as a `Float32Array` would no longer provide the matrix data it needs, nor could it be updated in-place if the spec called for it.

We should specify exactly how the spec reacts to detached typed arrays, or alter the spec design to avoid the problem entirely.

Manishearth:
Implements Option 2 from https://github.com/immersive-web/webxr/issues/630#issuecomment-495781562

fixes #630

r? @toji @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -80,6 +80,8 @@ spec: Gamepad; urlPrefix: https://w3c.github.io/gamepad/#
     type: attribute; text: pressed; for: GamepadButton; url: dom-gamepadbutton-pressed
     type: attribute; text: value; for: GamepadButton; url: dom-gamepadbutton-value
 spec:html; type:method; for:HTMLCanvasElement; text:getContext(contextId); url: https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext
+spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
+    type: method; text: IsDetachedBuffer; url: sec-isdetachedbuffer
 </pre>
 
 <link rel="icon" type="image/png" sizes="32x32" href="favicon-32x32.png">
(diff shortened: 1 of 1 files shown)


Manishearth:
Given that there's now an explicit algorithm for obtaining XRRigidTransform.matrix it may also be worth fleshing it out by saying exactly how this matrix is to be obtained, the same way we do for XRRay already.

Manishearth:
filed https://github.com/immersive-web/webxr/issues/686

index.bs
```diff
@@ -1106,10 +1111,21 @@ The <dfn attribute for="XRRigidTransform">position</dfn> attribute is a 3-dimens
 
 The <dfn attribute for="XRRigidTransform">orientation</dfn> attribute is a quaternion describing the rotational component of the transform. The {{XRRigidTransform/orientation}} MUST be normalized to have a length of <code>1.0</code>.
 
-The <dfn attribute for="XRRigidTransform">matrix</dfn> attribute returns the transform described by the {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}} attributes as a [=matrix=]. This attribute SHOULD be lazily evaluated.
+The <dfn attribute for="XRRigidTransform">matrix</dfn> attribute returns the transform described by the {{XRRigidTransform/position}} and {{XRRigidTransform/orientation}} attributes as a [=matrix=]. This attribute MUST be computed by [=XRRigidTransform/obtain the matrix|obtaining the matrix=] for the {{XRRigidTransform}}.
```
Squareys: @Manishearth @toji Sorry for the post-merge comment here:

The "lazy evaluated" part originally introduced in https://github.com/immersive-web/webxr/pull/560 disappeared here. As it is viable to use a DualQuaternion as view transform directly, it would make sense to keep the matrix computation optional. It's just a nit-optimization for a currently less common use case, of course.

Or does "MUST be computed" imply it's computed on access?

Manishearth: Yep.

Squareys: Alright, awesome, thanks for the clarification!

PR #687, opened 2019-06-06
Change blur/focus to visibilitychange

### LINKED ISSUE #468: Document effects when session is blurred
BS Issue: https://immersive-web.github.io/webxr/#blur-all-sessions

This is to do with an immersive session being put into the background rather than visual blurring. See: https://immersive-web.github.io/webxr/#blurred

* We should indicate in what way inputs are not updated, are the in accessible, do the position/rotation values just remain stationary, do they get set to 0 or null?
* We should indicate expected behaviour when framerate throttling, do we inform developers of the new frame rate, to avoid giving the illusion of performance issues in the metrics?

### LINKED ISSUE #488: "focus"/"blur" naming on XRSession may be confusing
The naming of the `"focus"` and `"blur"` events on `XRSession` are somewhat confusing to me.

I hadn't thought about those events for a bit and when I saw them today in the spec, I presumed they were telling the app when it could expect or not expect to receive XR input updates such as pose updates, `select` events and other input source button presses.  Basically, I presumed these events to be the equivalent of [`"focus"`](https://developer.mozilla.org/en-US/docs/Web/Events/focus)/[`"blur"`](https://developer.mozilla.org/en-US/docs/Web/Events/blur) in the core DOM.  However, these events actually tell the app when its pixels are invisible or visible, allowing it to save effort by pausing its rendering.

While it is likely that XR input will _also_ be removed from the app during the example situation of a sensitive overlay being displayed, that seems like a secondary effect.  Specifically, if XR input is removed from the page temporarily due to a non-sensitive _partial_ overlay, the UA _would_ expect the page to continue rendering.  However, it's unlikely that a UA could raise the `"blur"` event to communicate the loss of input, since we'd have already suggested that pages stop rendering when they receive that event.

This seems more similar to the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) and its `visibilitychange` event, which serves an equivalent purpose for 2D content, letting the page know when the user can't see it, allowing the page to stop expensive rendering or pause content.  Perhaps a single `visibilitychange` event better fits our intended developer mental model here?

toji:
A proposal, based on discussion at the F2F.

Fixes #488
Fixes #468

Changes the `blur` and `focus` events to `visibilitychange`, along with
a `visibilityState` attribute on the `XRSession`. The `visibilityState`
may be `visible`, `hidden`, or `background`. This is pretty close to the
`document.visibilityState` attribute and associated event, with the
exception of the `background` state.


index.bs
```diff
@@ -1927,11 +1941,7 @@ The user agent MUST provide the following new events. Registration for and firin
...
-A user agent MAY dispatch a <dfn event for="XRSession">blur</dfn> event on an {{XRSession}} to indicate that presentation to the {{XRSession}} by the page has been suspended by the user agent, OS, or XR hardware. While an {{XRSession}} is <dfn>blurred</dfn> it remains active but it may have its frame production throttled. This is to prevent tracking while the user interacts with potentially sensitive UI. For example: The user agent SHOULD blur the presenting application when the user is typing a URL into the browser with a virtual keyboard, otherwise the presenting page may be able to guess the URL the user is entering by tracking their head motions. The event MUST be of type {{XRSessionEvent}}.
-
-A user agent MAY dispatch a <dfn event for="XRSession">focus</dfn> event on an {{XRSession}} to indicate that presentation to the {{XRSession}} by the page has resumed after being suspended. The event MUST be of type {{XRSessionEvent}}.
-</section>
+A user agent MAY dispatch a <dfn event for="XRSession">visibilitychange</dfn> event on an {{XRSession}} to indicate that the [=XRSession/visibility state=] of the {{XRSession}} has changed. The event MUST be of type {{XRSessionEvent}}.
```
Manishearth: Should we require it be dispatched whenever the value of `{{XRSession/visibilityState}}` changes?

We probably need this MAY so that UAs aren't required to support noticing changes in visibility state, but if they can report these changes through the attribute they should also trigger the event.

toji: You're absolutely right. I've update the PR so that changes to the attribute MUST trigger the event, but that the attribute only SHOULD be affected by platform-level changes, because I don't think we can require that in all cases.

Manishearth: Yeah, we shouldn't require UAs to notice everything.

index.bs
```diff
@@ -363,7 +363,7 @@ enum XRSessionMode {
...
 
-An <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or an {{immersive-ar}} session. [=Immersive sessions=] MUST provide some level of [=viewer=] tracking, and content MUST be shown at the proper scale relative to the user and/or the surrounding environment. Additionally, [=Immersive sessions=] MUST be given <dfn>exclusive access</dfn> to the [=/XR device=], meaning that while the [=immersive session=] is not [=blurred=] the HTML document is not shown on the [=/XR device=]'s display, nor is content from other applications shown on the [=/XR device=]'s display.
+An <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or an {{immersive-ar}} session. [=Immersive sessions=] MUST provide some level of [=viewer=] tracking, and content MUST be shown at the proper scale relative to the user and/or the surrounding environment. Additionally, [=Immersive sessions=] MUST be given <dfn>exclusive access</dfn> to the [=/XR device=], meaning that while the [=immersive session=] is {{XRVisibilityState/visible}} the HTML document is not shown on the [=/XR device=]'s display, nor is content from other applications shown on the [=/XR device=]'s display.
```
thetuvix: May be worth noting that the HTML document may still be visible on another display during an immersive session, such as a desktop monitor.

Also may be worth discussing what the HTML document being hidden by an immersive session causes for the 2D [Document.hidden](https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden) and [Document.visibilityState](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) attributes.

index.bs
```diff
@@ -549,26 +555,36 @@ When the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or prese
...
 
-<section class="unstable">
-The <dfn attribute for="XRSession">onblur</dfn> attribute is an [=Event handler IDL attribute=] for the {{blur}} event type.
+  - A state of <dfn enum-value for="XRVisibilityState">visible</dfn> indicates that imagery rendered by the {{XRSession}} can be seen by the user and {{XRSession/requestAnimationFrame()}} callbacks are processed at the [=/XR device=]'s native refresh rate. Input is processed by the {{XRSession}} normally.
+
+  - A state of <dfn enum-value for="XRVisibilityState">hidden</dfn> indicates that imagery rendered by the {{XRSession}} cannot be seen by the user. {{XRSession/requestAnimationFrame()}} callbacks will not be processed until the [=visibility state=] changes. Input is not processed by the {{XRSession}}.
+
+  - A state of <dfn enum-value for="XRVisibilityState">background</dfn> indicates that imagery rendered by the {{XRSession}} may be seen by the user, but is not the primary focus. {{XRSession/requestAnimationFrame()}} callbacks MAY be throttled. Input is not processed by the {{XRSession}}.
```
thetuvix: I would call this `visible-background` or `visible-blurred` or such, to indicate that this is also a visible state, and I'd mention it between `visible` and `hidden`.  My first thought when I saw `background` at the bottom of the enum was that the session was even more hidden than `hidden`, being fully "backgrounded" or such.

index.bs
thetuvix: May be meaningful on headsets (i.e. when there is no desktop monitor to still show the HTML document) to mirror what signals the Fullscreen API sends regarding 2D document visibility.

index.bs
thetuvix: I have a moderate preference for `visible-blurred`, since `background` seems to mean other things throughout the web. which is partially what led me astray:
* https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background
* https://developer.mozilla.org/en-US/docs/Web/CSS/background

index.bs
```diff
@@ -1958,7 +1968,7 @@ Gaze Tracking {#gazetracking-security}
...
 
-To prevent this risk the user agent MUST [=blur all sessions=] when the users is interacting with sensitive, trusted UI such as URL bars or system dialogs. Additionally, to prevent a malicious page from being able to monitor input on a other pages the user agent MUST [=blur all sessions=] on non-focused pages.
+To prevent this risk the user agent MUST set the [=visibility state=] of all {{XRSession}}s to {{XRVisibilityState/hidden}} or {{XRVisibilityState/background}} with throttled frame production when the users is interacting with sensitive, trusted UI such as URL bars or system dialogs. Additionally, to prevent a malicious page from being able to monitor input on a other pages the user agent MUST set the [=visibility state=] of all {{XRSession}}s on non-focused pages to {{XRVisibilityState/hidden}}.
```
thetuvix: Interesting - is the idea then that during `background` an app's frames are not being displayed, but it should keep its render loop going?  Or the idea that during `background` the app is throttled to 5 fps, counting on reprojection to keep the view reasonable?  Do we have a belief that UAs can throttle enough to eliminate the privacy risk without causing users to be uncomfortable?

thetuvix: Or sometimes `background` is during a sensitive overlay (e.g. 5 fps) and sometimes `background` is during a non-sensitive overlay (e.g. 60/90 fps)?

index.bs
darktears: Per https://codepen.io/anon/pen/byXVXX on Chrome, entering and leaving Fullscreen doesn't trigger any visibility change event which is to be expected tbh. You're watching one of the element (e.g. video) in fullscreen so the page is technically visible.

darktears: > May be worth noting that the HTML document may still be visible on another display during an immersive session, such as a desktop monitor.
> 
> Also may be worth discussing what the HTML document being hidden by an immersive session causes for the 2D [Document.hidden](https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden) and [Document.visibilityState](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) attributes.

Interesting question. I think if the document is still visible on another monitor you don't want to touch the visibilityState attribute on the 2D document because as you said it's "visible" and a secondary user (not the one in the HMD) would expect the page to continue rendering and working (it could be a control panel for e.g. beside the mirroring) therefore we shouldn't send such signal to developers (because they mentally map it to, my page is hidden, I need to stop doing work)

index.bs
toji: Yes. This is to allow for UAs to have some flexibility in how they display sensitive content AND to let UAs take advantage of some platforms features where the content is still visible even when the input focus is elsewhere. Primary example of that I can think of is Steam VR, which continues to display the app (lower res and greyed out, but still live) when you pull up the Steam overlay.
2019-06-07-PM_04_11_46 This is a situation in which the app would be put in that state but still pump frames at ~full speed.

You could also picture a scenario where the app is placed in that state but rendering is severely throttled to the point where it may only render intermittently to patch up around reprojection or even just process one frame to make a cubemap. This could be either for privacy reasons OR simply because the device lacks the horsepower to render two realtime stereo scenes at once. I don't believe the API should differentiate between those scenarios to the developer. Whether or not the UA is comfortable with that being sufficient to preserve privacy is up to them, though the `hidden` state is always available as a last resort.

index.bs
toji: Sounds reasonable, and I agree with "visible-" prefix communicating the right thin here. Went with `visible-blurred`, but I'm not overtly attached to it either should a better state name come up.

index.bs
toji: Added a note specifying that this attribute has no bearing on the document's `visibilityState` and specified the tethered desktop headset scenario as an example. In general I would say that even on devices without an external display I'd expect showing immersive content to indicate that the document visibility state is still 'visible' because the page _is_ still showing you content, just not HTML content. (And that may change in the future when we talk about DOM layers.) This apparently would keep us inline with fullscreen, as Alexis pointed out.

toji:
Updated to take into account feedback from Alex.

darktears:
LGTM

index.bs
```diff
@@ -549,26 +555,38 @@ When the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or prese
...
+
+  - A state of <dfn enum-value for="XRVisibilityState">hidden</dfn> indicates that imagery rendered by the {{XRSession}} cannot be seen by the user. {{XRSession/requestAnimationFrame()}} callbacks will not be processed until the [=visibility state=] changes. Input is not processed by the {{XRSession}}.
+
+The <dfn attribute for="XRSession">visibilityState</dfn> attribute returns the {{XRSession}}'s [=visibility state=]. The <dfn attribute for="XRSession">onvisibilitychange</dfn> attribute is an [=Event handler IDL attribute=] for the {{visibilitychange}} event type.
+
+The [=visibility state=] MAY be changed by the user agent at any time, and the user agent SHOULD monitor the XR platform when possible to observe when session visibility has been affected external to the user agent and update the [=visibility state=] accordingly. Additionally, the {{XRSession}} MUST have it's [=visibility state=] set to {{XRVisibilityState/hidden}} or {{XRVisibilityState/visible-blurred}} with frame production throttled when the user is interacting with potentially sensitive UI.
```
NellWaliczek: >when the user is interacting with potentially sensitive UI.

Can it change within an XRFrame?

NellWaliczek: >when the user is interacting with potentially sensitive UI.

It's a bit pedantic, but we should probably should explicitly state this means UI provided by the UA not the page.

index.bs
```diff
@@ -549,26 +555,38 @@ When the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or prese
...
 
-The <dfn attribute for="XRSession">onfocus</dfn> attribute is an [=Event handler IDL attribute=] for the {{focus}} event type.
+<section class="note">
+EXAMPLE: The user agent must change the [=visibility state=] of all {{XRSession}}s that can track the [=viewer=]'s pose when the user is typing a URL into the browser with a virtual keyboard, otherwise the page may be able to guess the URL the user is entering by tracking their head motions.
```
NellWaliczek: Should this really be a non-normative note?  It seems like we should explicitly define the concept of trusted UI (though whether it's required to implement is a separate question) and have it be a MUST that the page is visibility-blurred if the trusted UI is showing.

index.bs
```diff
@@ -549,26 +555,38 @@ When the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or prese
...
 </section>
 
+NOTE: The {{XRSession}}'s [=visibility state=] has no bearing on the visibility of the HTML document. Depending on the system configuration the page may continue to be visible while an [=immersive session=] is active. (For example, a headset connected to a PC may continue to display the page on the monitor while the headset is viewing content from an [=immersive session=].) Developers should continue to rely on the [Page Visibility API](https://w3c.github.io/page-visibility/) to determine page visibility.
```
NellWaliczek: Why is this a "NOTE:" when the previous paragraph is in a class="note" section?

NellWaliczek: >no bearing 

Is this true?  Based on the behavior of fullscreen you're discussing in the comments above, I agree that the document shouldn't be hidden when an immersive session is happing.  However, personally, I'd expect the session to be "hidden" if the page was "hidden.

index.bs
```diff
@@ -583,8 +601,6 @@ The <dfn attribute for="XRSession">onselect</dfn> attribute is an [=Event handle
 
 We still need to document what happens when we <dfn>end the session</dfn>. (This is <a href="https://github.com/immersive-web/webxr/issues/467">filed</a>.)
 
-We still need to document what happens when we <dfn lt="blur all sessions">blur all sessions</dfn>. (This is <a href="https://github.com/immersive-web/webxr/issues/468">filed</a>.)
-
 We still need to document what happens when we <dfn>poll the device pose</dfn> (This is <a href="https://github.com/immersive-web/webxr/issues/480">filed</a>.)
```
NellWaliczek: Remove this line now that 480 is closed?

index.bs
NellWaliczek: Actually we say something similar to my suggestion in the section on trusted UI, and this should probably just move there (since we're going to have to make that properly normative shortly anyway)

index.bs
```diff
@@ -1958,7 +1970,7 @@ Gaze Tracking {#gazetracking-security}
...
 
-To prevent this risk the user agent MUST [=blur all sessions=] when the users is interacting with sensitive, trusted UI such as URL bars or system dialogs. Additionally, to prevent a malicious page from being able to monitor input on a other pages the user agent MUST [=blur all sessions=] on non-focused pages.
+To prevent this risk the user agent MUST set the [=visibility state=] of all {{XRSession}}s to {{XRVisibilityState/hidden}} or {{XRVisibilityState/visible-blurred}} with throttled frame production when the users is interacting with sensitive, trusted UI such as URL bars or system dialogs. Additionally, to prevent a malicious page from being able to monitor input on a other pages the user agent MUST set the [=visibility state=] of all {{XRSession}}s on non-focused pages to {{XRVisibilityState/hidden}}.
```
NellWaliczek: > with throttled frame production

This is a bit overly specific, I think?  Quantized may be sufficient?  And if not, we should enforce that the state should be 'hidden'.

index.bs
```diff
@@ -385,9 +385,16 @@ enum XREnvironmentBlendMode {
...
+enum XRVisibilityState {
+  "visible",
+  "visible-blurred",
+  "hidden",
+};
+
 [SecureContext, Exposed=Window] interface XRSession : EventTarget {
   // Attributes
   readonly attribute XREnvironmentBlendMode environmentBlendMode;
+  readonly attribute XRVisibilityState visibilityState;
```
NellWaliczek: [SameObject]?

index.bs
toji: Because I didn't want the previous section to start with "NOTE: EXAMPLE: ..." and NOTE is the only text that auto-generates that styling, but EXAMPLE felt more appropriate.

toji: Hm. This is maybe an issue of language precision, then. In my mind "X has no bearing on Y" means that "Y is not affected by X" but does not necessarily imply that "X is not affected by Y". In any case, the fact that there is any confusion here likely means that this warrants rephrasing. I'll take another stab at it.

index.bs
toji: I don't believe that's applicable here, because the state changes over time. (Also not sure if `[SameObject]` is applicable to enums at all.)

index.bs
toji: Pedantry approved. 👍

explainer.md
```diff
@@ -277,27 +277,33 @@ function drawScene(view) {
 
 Because the `XRViewerPose` inherits from `XRPose` it also contains a `transform` describing the position and orientation of the viewer as a whole relative to the `XRReferenceSpace` origin. This is primarily useful for rendering a visual representation of the viewer for spectator views or multi-user environments.
 
-### Handling suspended sessions
+### Handling sessions visibility
```
NellWaliczek: nit "session"

explainer.md
```diff
@@ -277,27 +277,33 @@ function drawScene(view) {
...
 
-The UA may temporarily "suspend" a session at any time. While suspended a session has restricted or throttled access to the XR device state and may process frames slowly or not at all. Suspended sessions can be reasonably be expected to be resumed at some point, usually when the user has finished performing whatever action triggered the suspension in the first place.
+The UA may temporarily hide a session at any time. While hidden a session has restricted access to the XR device state and frames will not be processed. Hidden sessions can be reasonably be expected to be made visible again at some point, usually when the user has finished performing whatever action triggered the session to hide in the first place.
```
NellWaliczek: Add a sentence at the end that clarifies that they may not be made visible, so don't count on it. (i.e. if the page is navigated away from)

explainer.md
```diff
@@ -277,27 +277,33 @@ function drawScene(view) {
...
 
-The UA may suspend a session if allowing the page to continue reading the headset position represents a security or privacy risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output.
+The UA may hide a session if allowing the page to continue reading the headset position represents a security or privacy risk (like when the user is entering a password or URL with a virtual keyboard, in which case the head motion may infer the user's input), or if other content is obscuring the page's output.
```
NellWaliczek: >other content is obscuring the page's output

May need to clarify that the other content isn't drawn by the page.

index.bs
NellWaliczek: Oh, oops... i copied the wrong bit of text... updated the comment above...

toji:
Done, done, and done. Thanks!

index.bs
toji: Okay, made it explicit that the value can't change during an XR frame.

index.bs
```diff
@@ -575,7 +575,7 @@ Each {{XRSession}} has a <dfn for="XRSession">visibility state</dfn> value, whic
...
 
-The [=visibility state=] MAY be changed by the user agent at any time, and the user agent SHOULD monitor the XR platform when possible to observe when session visibility has been affected external to the user agent and update the [=visibility state=] accordingly. Additionally, the {{XRSession}} MUST have it's [=visibility state=] set to {{XRVisibilityState/hidden}} or {{XRVisibilityState/visible-blurred}} with frame production throttled when the user is interacting with potentially sensitive UI from the user agent (such as entering a URL).
+The [=visibility state=] MAY be changed by the user agent at any time other than during the processing of an[=XR animation frame=], and the user agent SHOULD monitor the XR platform when possible to observe when session visibility has been affected external to the user agent and update the [=visibility state=] accordingly. Additionally, the {{XRSession}} MUST have it's [=visibility state=] set to {{XRVisibilityState/hidden}} or {{XRVisibilityState/visible-blurred}} with frame production throttled when the user is interacting with potentially sensitive UI from the user agent (such as entering a URL).
```
NellWaliczek: > an[=XR animation frame=]

missing a space

index.bs
```diff
@@ -362,7 +362,7 @@ enum XRSessionMode {
...
 
-An <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or an {{immersive-ar}} session. [=Immersive sessions=] MUST provide some level of [=viewer=] tracking, and content MUST be shown at the proper scale relative to the user and/or the surrounding environment. Additionally, [=Immersive sessions=] MUST be given <dfn>exclusive access</dfn> to the [=/XR device=], meaning that while the [=immersive session=] is not [=blurred=] the HTML document is not shown on the [=/XR device=]'s display, nor is content from other applications shown on the [=/XR device=]'s display.
+An <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or an {{immersive-ar}} session. [=Immersive sessions=] MUST provide some level of [=viewer=] tracking, and content MUST be shown at the proper scale relative to the user and/or the surrounding environment. Additionally, [=Immersive sessions=] MUST be given <dfn>exclusive access</dfn> to the [=/XR device=], meaning that while the [=immersive session=] is {{XRVisibilityState/visible}} the HTML document is not shown on the [=/XR device=]'s display, nor is content from other applications shown on the [=/XR device=]'s display.
```
NellWaliczek: >nor is content from other applications shown on the [=/XR device=]'s display.

IIRC there's a way in OpenVR to overlay content from one app on top of another?  If so, I'm a bit concerned that this implies this would be prohibited, but I'm not sure the browser would actually be in a position to disable that?

index.bs
```diff
@@ -547,25 +553,33 @@ When the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or prese
...
+
+  - A state of <dfn enum-value for="XRVisibilityState">hidden</dfn> indicates that imagery rendered by the {{XRSession}} cannot be seen by the user. {{XRSession/requestAnimationFrame()}} callbacks will not be processed until the [=visibility state=] changes. Input is not processed by the {{XRSession}}.
+
+The <dfn attribute for="XRSession">visibilityState</dfn> attribute returns the {{XRSession}}'s [=visibility state=]. The <dfn attribute for="XRSession">onvisibilitychange</dfn> attribute is an [=Event handler IDL attribute=] for the {{visibilitychange}} event type.
+
+The [=visibility state=] MAY be changed by the user agent at any time other than during the processing of an[=XR animation frame=], and the user agent SHOULD monitor the XR platform when possible to observe when session visibility has been affected external to the user agent and update the [=visibility state=] accordingly. Additionally, the {{XRSession}} MUST have it's [=visibility state=] set to {{XRVisibilityState/hidden}} or {{XRVisibilityState/visible-blurred}} with frame production throttled when the user is interacting with potentially sensitive UI from the user agent (such as entering a URL).
```
NellWaliczek: > sensitive UI from the user agent (such as entering a URL).

I know we're about to revamp it, but can we make this link to the security section's "Trusted Environment" section?

index.bs
```diff
@@ -547,25 +553,33 @@ When the {{XRInputSource/handedness}}, {{XRInputSource/targetRayMode}}, or prese
...
+
+NOTE: The {{XRSession}}'s [=visibility state=] does not necessarily imply the visibility of the HTML document. Depending on the system configuration the page may continue to be visible while an [=immersive session=] is active. (For example, a headset connected to a PC may continue to display the page on the monitor while the headset is viewing content from an [=immersive session=].) Developers should continue to rely on the [Page Visibility API](https://w3c.github.io/page-visibility/) to determine page visibility.
```
NellWaliczek: > Developers should continue to rely on the [Page Visibility API](https://w3c.github.io/page-visibility/) to determine page visibility.

It feels like this is something we should be normative about so there's consistency across user agents. For example, if there is no secondary display showing the 2D page should it or shouldn't it report that the page is visible?  If the browser window on a secondary screen is minimized, the session should or shouldn't change it's visibility state?

index.bs
NellWaliczek: Alternatively, can we just move this whole chunk down to the Trusted Environment section?  It has a lot of overlap, but that section doesn't talk about frame production being throttled.

>Additionally, the {{XRSession}} MUST have it's [=visibility state=] set to {{XRVisibilityState/hidden}} or {{XRVisibilityState/visible-blurred}} with frame production throttled when the user is interacting with potentially sensitive UI from the user agent (such as entering a URL).

NellWaliczek: >frame production throttled

I'm still unclear if this is the only allowable mitigation?  Would quantization be adequate if the underlying platform had an easy way to support this in its reprojection pipeline?

toji: It may be. I'll remove that as a requirement for now.

index.bs
toji: As discussed in our call, now filed as #696

index.bs
toji: Updated with your suggested wording from our call.

PR #688, opened 2019-06-10
Remove XRLayer base type.

### LINKED ISSUE #268: Is XRLayer useful?
Right now it's an empty interface. Do we anticipate putting things there in the future, so that having a common base class is desirable? If so, that would be great material for the explainer.

toji:
/fixes #268

Also changed `baseLayer` to only accept an `XRWebGLLayer`. This can be
extended in the future in a few ways:

 - Re-introduce XRLayer and change `baseLayer` to take one.
 - Update `baseLayer` to accept a union of layer types.
 - Leave `baseLayer` as WebGL-only and enforce that new layer types must
   use whatever layer array mechanism is introduced.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -145,7 +145,7 @@ Only one immersive session per XR hardware device is allowed at a time across th
 
 Once the session has started, some setup must be done to prepare for rendering.
 - An `XRReferenceSpace` should be created to establish a space in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
-- An `XRLayer` must be created and set as the `XRSession`'s `renderState.baseLayer`. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
+- An `XRWebGLLayer` must be created and set as the `XRSession`'s `renderState.baseLayer`. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
 - Then `XRSession.requestAnimationFrame` must be called to start the render loop pumping.
 
 ```js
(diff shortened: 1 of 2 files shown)


explainer.md
```diff
@@ -145,7 +145,7 @@ Only one immersive session per XR hardware device is allowed at a time across th
 
 Once the session has started, some setup must be done to prepare for rendering.
 - An `XRReferenceSpace` should be created to establish a space in which `XRViewerPose` data will be defined. See the [Spatial Tracking Explainer](spatial-tracking-explainer.md) for more information.
-- An `XRLayer` must be created and set as the `XRSession`'s `renderState.baseLayer`. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
+- An `XRWebGLLayer` must be created and set as the `XRSession`'s `renderState.baseLayer`. (`baseLayer` because future versions of the spec will likely enable multiple layers, at which point this would act like the `firstChild` attribute of a DOM element.)
```
NellWaliczek: > at which point this would act like the `firstChild` attribute of a DOM element.

This statement is a bit specific regarding a design we haven't actually decided yet... Can we just remove it?

index.bs
```diff
@@ -1523,14 +1523,7 @@ NOTE: This diagram demonstrates how two potential controllers would be exposed w
...
 
-XRLayer {#xrlayer-interface}
--------
-
-An {{XRLayer}} defines a source of bitmap images and a description of how the image is to be rendered to the [=/XR device=]. Initially only one type of layer, the {{XRWebGLLayer}}, is defined but future revisions of the spec may extend the available layer types.
-
-<pre class="idl">
-[SecureContext, Exposed=Window] interface XRLayer {};
-</pre>
+NOTE: While this specification only defines the {{XRWebGLLayer}} layer, future modules are expected to add additional layer types and the image sources that they draw from.
```
NellWaliczek: > modules

Teeeeechnically we haven't officially decided this is the plan... Perhaps let's use a less specific phrase for the moment :)

Issue #578, opened 2019-04-02
Define a registry for defining how UAs expose input devices


ddorwin:
While the WebXR spec (after #553 is merged) provides some guidelines around exposing controllers as Gamepads, there is still room for interpretation that could lead to different behavior among implementations. In addition, future controller/input mechanisms (i.e., hands as exposed by a specific technology) may not map well to `"xr-standard"` or otherwise have obvious mappings.

To encourage consistency and interoperability, details about how specific controllers should be exposed by user agent implementations should be defined in a registry. The registry would be separate from the spec, allowing out-of-band updates (per an established process). It would both clearly define how implementations should expose specific controllers and thus how applications and libraries (e.g., https://github.com/immersive-web/xr-gamepad-mappings) can expect them to be exposed.

The registry would map some well understood definition of a physical controller to:
* `id` string
* `GamepadMappingType`
* The physical button that represents the [primary action](file:///src/scratch/webxr/index.html#primary-action)
* A clear definition of how the physical buttons are exposed via the `Gamepad` interface
  * This would include which buttons are not to be exposed, such as those reserved by the system and/or user agents.

Ideally, the entry would also include some concrete information, such as USB vendor and product IDs and/or identifier exposed by native runtime(s).

A single entry could cover multiple devices (i.e., revisions of a controller) as long as the physical appearance (due to the [intent](https://github.com/immersive-web/webxr/issues/336#issuecomment-478712300) for applications to use the `id` to choose a controller model) and button mapping are the same. If there are multiple versions of hardware with the same product name, they may be identified by version, year, or similar.

Such a registry would also solve a couple open issues:
* Format of Gamepad identifiers (#550): While the registry could have guidelines, there wouldn’t be a need to define a perfect rule covering all hardware because the registry would specify the identifier.
* As such, identically-/similarly-named controllers (e.g., “Oculus Touch” per https://github.com/immersive-web/webxr/issues/550#issuecomment-469818569) can be clearly and easily differentiated.

<br/>One additional possibility is that [XR-specific] GamepadMappingTypes also be specified in a registry. These would be added much more conservatively, but it might be preferable to be able to define new ones without going through the full spec process. This would also allow slimming the core WebXR spec a bit.

<br/>For reference, there is an open issue for the W3C Process Document (https://github.com/w3c/w3process/issues/168) to define a process for handling registries. This shouldn’t block us from proceeding, and other specs have successfully used registries in the past.

jacobcdewitt:
[...]

PR #695, opened 2019-06-13
Refactor input identification

### LINKED ISSUE #550: Format of Gamepad identifiers

toji:
Fixes #550.

Removed any logic regarding using the gamepad id for rendering or
layout identification. Introduced the `renderId` and `gamepadLayout`
attributes to `XRInputSource` in it's place.


input-explainer.md
```diff
@@ -228,21 +228,25 @@ For `tracked-pointer` input sources, it is often appropriate for the application
...
+The WebXR Device API currently does not offer any way to retrieve renderable resources that represent the input devices from the API itself, and as such the `XRInputSource`'s `renderId` must be used to identify and load an appropriate resources from the application's server or a CDN. The `renderId` provides a list of strings that identify the device, ordered from most accurate to least accurate. Applications should iterate through the list until a string is located that corresponds to a known model and display that. The example below presumes that the `getInputSourceRenderableModel` call would do the required lookup and caching.
 
 ```js
 function loadRenderableInputModels(xrInputSource) {
-  // Don't load renderable models if the input source doesn't have a gamepad. 
-  if (!inputSource.gamepad)
+  // Don't load renderable models if the input source doesn't identify what the
```
ddorwin: This logic isn't necessary as the for loop will get skipped.

You could move this comment to after the for loop to explain what happens when that point is reached.

input-explainer.md
```diff
@@ -285,19 +289,11 @@ Examples of input sources that may expose their state this way include Oculus To
...
-  - Inline sessions MUST only expose `id`s of `'unknown'`.
-  - Otherwise the `id` must be a lower-case string that describes the physical input source.
-    - The exact format is [still under discussion](https://github.com/immersive-web/webxr/issues/550), but will probably need to include at least an indication of the input device's manufacturer and model. (Such as `oculus-touch`)
-    - It must not include an indication of the handedness of the input source (such as `oculus-touch-left`), as that can be determined from the `handedness` attribute.
-    - UAs SHOULD make an effort to align on the strings that are returned for any given device.
-
-All other attributes behave as described in the [Gamepad](https://w3c.github.io/gamepad/) specification.
+The exact button and axes layout is given by the `XRInputSource`'s `gamepadLayout` attribute, which contains an array of strings that identify the button and axes layout or subsets of it, ordered from most specific to least specific.
```
ddorwin: Ditto on naming/plurality and the language around the order.

input-explainer.md
```diff
@@ -228,21 +228,25 @@ For `tracked-pointer` input sources, it is often appropriate for the application
...
 
-The WebXR Device API currently does not offer any way to retrieve renderable resources that represent the input devices from the API itself, and as such the `Gamepad`'s `id` must be used as a key to load an appropriate resources from the application's server or a CDN. The example below presumes that the `getInputSourceRenderableModel` call would do the required lookup and caching.
+The WebXR Device API currently does not offer any way to retrieve renderable resources that represent the input devices from the API itself, and as such the `XRInputSource`'s `renderId` must be used to identify and load an appropriate resources from the application's server or a CDN. The `renderId` provides a list of strings that identify the device, ordered from most accurate to least accurate. Applications should iterate through the list until a string is located that corresponds to a known model and display that. The example below presumes that the `getInputSourceRenderableModel` call would do the required lookup and caching.
```
ddorwin: nit: "application's server or a CDN" is just one possible mechanism. It could be generated or part of an offline PWA, for example.

ddorwin: Given that this is the explainer, does it make sense to provide examples of what the lists might contain?

NellWaliczek:
Thanks for putting this together!  I read it over and I think my main question/concern is how, as a developer, I'm expected to use the layout and render parts differently?  For example, right now the xr-gamepad-mappings library uses a single ID to pull a json file with the details of the controller.  This file includes how to map elements in the buttons/axes array into specific physical controller parts (i.e. thumbsticks).  But it also contains information about how to animate the virtual model to match the behavior of the physical gamepad.

Under this design I'm having a bit of a hard time picturing the correct way to update the library.  Do I start by walking through the fallbacks for the layouts until I find a match?  Once I have that match, am I guaranteed that all the renderIds will be compatible with all the layouts?  I'd guess not?  Or am I missing something here?

If I'm not misunderstanding, perhaps an alternative approach would be to make the retrieval of renderId a function call that takes a layoutId as a parameter?

toji:
Addressed Davids' feedback outside of the name issue. Also haven't addressed Nell's comments. Need to give them both some more thought.

ddorwin:
Replying to @NellWaliczek's question, perhaps applications (or libraries) would just not animate button presses, etc. if they don't recognize both values or do not recognize them both as a valid pair together. The unrecognized pairs shouldn't be much more common than not recognizing one of the values. The registry (#578) should help ensure this.

@NellWaliczek wrote:
> If I'm not misunderstanding, perhaps an alternative approach would be to make the retrieval of renderId a function call that takes a layoutId as a parameter?

`gamepadLayout` is specific to Gamepad, so it would be weird to make obtaining `renderId` dependent on it in the spec. The xr-gamepad-mappings library could maybe make such an association, though.

NellWaliczek:
Yeah I see what you mean about that being a bit weird.  In theory a renderId can exist for XRInputSources without a Gamepad associated with them. 

My concern is that we'll end up with a mismatch.  As in, you end up with a layout that doesn't match with the model that is being rendered.  Not sure if that's a real concern... @thetuvix do you have an opinion on this?

And for what its worth, not animating is a pretty bad experience since the user will have no idea if the input source is working if they can't see a visual response.

ddorwin:
I'm not sure if this solves the problem, but another way the layouts could be exposed would be to have the Gamepad layout list returned by a method. This might also be a bit more future-compatible since it wouldn't bake `gamepad` further into the API.

Some possible examples:
1. `sequence<DOMString> getLayout(DOMString renderId);`
1. `sequence<DOMString> getLayout(Gamepad gamepad);`
1. `sequence<DOMString> getLayout(DOMString renderId, Gamepad gamepad);`

Note on the parameters in these examples:
* `renderId` would be one of the values from the `renderIds` attribute.
  - Something to think about is whether this would add complexity to implementations.
* `gamepad` would be the `gamepad` attribute on the `XRInputSource` object.
  - This might not do much other than indicate that the layout is being requested for a Gamepad object vs. something else - see below.
* In the future, the `Gamepad` parameter could become optional or a union in order to support future mechanisms that replace Gamepad.

NellWaliczek:
After thinking on this a bit more and rereading... I'm still kinda lost.  I really don't understand the relationship between renderId and layoutId....Would it be possible to update this PR with a few additional bits of sample code?  

For example, should I check layoutId or renderId first?  If I only recognize the 2nd entry in layoutIds, can I be sure that all of the renderIds will work with it?  What about if I check renderId first?  Will all the layoutIds work with it?

Relatedly, does a layoutId need to be limited to things that have gamepads?  Or will it work for things with articulated hand tracking?

toji:
After talking with Nell about this on Friday and doing some additional research, I've updated this proposal to hopefully be a bit more straightforward. The latest change combines the two previous lists of IDs into a single list, now called `profiles`. This is intended to cover device appearance, gamepad layout, and any other needs that we might have in the future.

It might feel more limiting at first glance, but after giving some thought I feel like this approach is actually a bit more flexible and can handle a variety of situations. This is based on a few realizations:

- Frequently appearance and layout will be linked. This is especially true when animating models based on the current input state, and even if you try to separate them developers will still try to infer one from the other. So a string like "oculus-touch" should be fine serving as both.
- It prevents UAs from having multiple layouts for a single visual ID. Sounds like a bug, but I think it's a feature. This means that if you're going to use a particular profile string you have to ensure you're synced up with the rest of the UAs regarding how that string is used, which ultimately is better for devs. It also naturally encourages gamepad layouts to be more predictable, since we'll likely end up with a few distinct "tracks" of devices that all are working to maintain compatibility between devices.
- You could actually mix and match "visual" and "layout" profiles if you wanted, but the reality of it is that they'll end up being both no matter what you try. Take, for example, the `["samsung-odyssey", "windows-mixed-reality", "touchpad-thumbstick-wand"]` example in the PR: Samsung Odyssey really only describes the visual design of the device, while Windows mixed reality can be thought of only describing the layout, but the reality is if your app recognizes a Samsung Odyssey, it will know that it's using a Windows Mixed Reality layout (even if internally it's just a pointer to a common data set), and if you recognized Windows Mixed reality as a layout at all then you likely have an appropriate mesh for it, and even in the fallback case, which ostensibly only defines the layout, it's still extremely likely that developers will come up with a generic-looking "stick with a touchpad and a joystick" model to use in the case that this is the only thing present.
- By not explicitly stating that it's a gamepad layout identifier or visual identifier we also leave things open for this to eventually help identify other aspects of the input device, like (for example) what style of hand tracking is being used.

Finally, it seems as though OpenXR's "[Interaction profiles](https://www.khronos.org/registry/OpenXR/specs/0.90/html/xrspec.html#semantic-path-interaction-profiles)" concept could be exposed reasonably via this mechanic. We would need a simple mapping from their path names to an appropriately formatted profile name and gamepad layout, but fundamentally they both provide a description of how the device expects to be used. (Would love any comments @thetuvix has in that regard).

index.bs
```diff
@@ -1354,6 +1355,22 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
 
 The {{gripSpace}} MUST be <code>null</code> if the input source isn't inherently trackable such as for input sources with a {{targetRayMode}} of {{XRTargetRayMode/gaze}} or {{XRTargetRayMode/screen}}.
 
+The <dfn attribute for="XRInputSource">profiles</dfn> attribute is a [=/list=] of [=input profile name=]s indicating both the preffered visual representation and the mapping of the buttons and axes of the {{XRInputSource/gamepad}} attribute of the {{XRInputSource}}. Names are given in decending order of preference.
```
NellWaliczek: nit: preffered

NellWaliczek: > the mapping of the buttons and axes of the {{XRInputSource/gamepad}} attribute of the {{XRInputSource}}

This is very specific to motion controllers with buttons.  Can we adjust this text so that it can also reflect variances in hand tracking capabilities even if no motion controller is present?

index.bs
```diff
@@ -1354,6 +1355,22 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
...
+
+Any [=input profile name=]s given after the first should provide fallback values that represent alternative visual representations and/or {{XRInputSource/gamepad}} layouts of the device. This may include a more generic or prior version of the device, a more widely recognized device that is sufficently similar, or a broad description of the device type (such as "touchpad-wand"). If multiple profiles are given the {{XRInputSource/gamepad}} layouts they describe must all represent a superset or subset of every other profile in the list.
```
NellWaliczek: nit: "after the first entry in the list"

NellWaliczek: >{{XRInputSource/gamepad}} layouts 

Ok, now I'm second guessing my recollection of our conversation... is this intended to be only for gamepads?

NellWaliczek: Probably need to call out that the list should be in order of decreasing specificity where possible.  We probably also want to state that the last item in the list should not contain any vendor specific information.  It doesn't have to be this PR, but we probably want to enumerate a concrete list of allowable "last item in the list" profile names that matches what's supported by the "xr-mapping".  Though that could be defined in the registry if necessary.

NellWaliczek: nit: comma after "given"

>If multiple profiles are given the {{XRInputSource/gamepad}} layouts they describe must all represent a superset or subset of every other profile in the list.

index.bs
```diff
@@ -1354,6 +1355,22 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
...
+
+If the {{XRSession}}'s [=mode=] is {{XRSessionMode/inline}}, {{XRInputSource/profiles}} MUST be an empty list. If the input device cannot be reliably identified, or the user agent wishes to mask the input device being used, it MAY choose to only report generic [=input profile name=]s or an empty list.
```
NellWaliczek: Can we make these two separate lines? The two sentences aren't actually related and that confused me for a moment...

NellWaliczek: > If the {{XRSession}}'s [=mode=] is {{XRSessionMode/inline}}, {{XRInputSource/profiles}} MUST be an empty list.

Is that true?  We don't want a profile for "touch" or something?

index.bs
```diff
@@ -1354,6 +1355,22 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
...
+
+<p class="note">
+For example, the Samsung Odyssey controller is a design variant of the standard Windows Mixed Reality controller. Both controllers share the same inputs. As a result, the {{XRInputSource/profiles}} for a Samsung Odyssey controller could be: <code>["samsung-odyssey", "windows-mixed-reality", "touchpad-thumbstick-wand"]</code>. Here the first two profiles would indicate the same input layout for the {{XRInputSource/gamepad}} attribute, but the appearance of the controller is more precisely communicated by the first profile in the list, with the second profile describing an acceptable substitue. The last profile is a generic fallback that describes the device in the roughest sense. (It's a wand with a touchpad and thumbstick.)
```
NellWaliczek: >Both controllers share the same inputs.

could maybe be

>Both controllers share the same input layouts.

NellWaliczek: Heh, I think Twitter has informed us that the word "controller" is preferable over the word "wand"

index.bs
```diff
@@ -1354,6 +1355,22 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
...
+
+Similarly, the Valve Index controller is backwards compatible with the HTC Vive controller, but the Index controller has additional buttons and axes. As a result, the {{XRInputSource/profiles}} for the Valve Index controller could be: <code>["valve-index", "htc-vive", "touchpad-thumbstick-wand"]</code>. In this case the gamepad layout described by the <code>"valve-index"</code> profile must be a superset of the layout described by the <code>"htc-vive"</code> profile. Also, the <code>"valve-index"</code> profile indicates the precise appearance of the controller, while the <code>"htc-vive"</code> controller has a significantly different appearance. In this case the UA would have deemed that difference acceptable. And as in the first example, the last profile is a generic fallback.
+
+(Exact strings are examples only. This specification does not make any recommendation for profile strings.)
```
NellWaliczek: Is this the text we should be using if we're deferring the enumeration to a separate registry document?

input-explainer.md
```diff
@@ -228,21 +228,35 @@ For `tracked-pointer` input sources, it is often appropriate for the application
...
 The majority of `tracked-pointer` input sources will have a non-null `gamepad` attribute on the `XRInputSource` object. The `Gamepad`'s `id` is used to determine what should be rendered if the app intends to visualize the input source itself, rather than an alternative virtual object. (See the section on [Button and Axis State](#button-and-axis-state) for more details.)
 
-The WebXR Device API currently does not offer any way to retrieve renderable resources that represent the input devices from the API itself, and as such the `Gamepad`'s `id` must be used as a key to load an appropriate resources from the application's server or a CDN. The example below presumes that the `getInputSourceRenderableModel` call would do the required lookup and caching.
+The WebXR Device API currently does not offer any way to retrieve renderable resources that represent the input devices from the API itself, and as such the `XRInputSource`'s `profiles` must be used to identify and load an appropriate resources. (For example, from the application's server, a CDN, or a local cache.) The `profiles` provides a list of strings that identify the device, given in descending order of preference.
```
NellWaliczek: >given in descending order of preference.

^_^

input-explainer.md
```diff
@@ -228,21 +228,35 @@ For `tracked-pointer` input sources, it is often appropriate for the application
...
+
+For example, the Samsung Odyssey controller is a variant of the standard Windows Mixed Reality controller. As a result, the `profiles` for a that controller could be:
```
NellWaliczek: > As a result, the `profiles` for a that controller could be:

should be
>As a result, the `profiles` for that controller could be:

input-explainer.md
```diff
@@ -228,21 +228,35 @@ For `tracked-pointer` input sources, it is often appropriate for the application
...
 
 ```js
-function loadRenderableInputModels(xrInputSource) {
-  // Don't load renderable models if the input source doesn't have a gamepad. 
-  if (!inputSource.gamepad)
-    return;
+// Exact strings are examples only.
+["samsung-odyssey", "windows-mixed-reality"]
```
NellWaliczek: Why leave off the generic string?

input-explainer.md
```diff
@@ -228,21 +228,35 @@ For `tracked-pointer` input sources, it is often appropriate for the application
...
+```
 
-  // Retrieve a mesh to render based on the gamepad object's id and handedness
-  let renderableModel = getInputSourceRenderableModel(inputSource.gamepad.id, inputSource.handedness);
-  if (!renderableModel)
-    return;
+Applications should iterate through the list until a string is located that corresponds to a known model, which should then be used when rendering the input device. The example below presumes that the `getInputSourceRenderableModel` call would do the required lookup and caching.
```
NellWaliczek: >Applications should iterate through the list until a string is located that corresponds to a known model, which should then be used when rendering the input device.

Can we modify this to also mention that developers are expected to know what to do with the layout implied by the profile if they consider if recognized?  Perhaps calling out that it's important to animate all controller components even if the experience doesn't use the data they provide.

input-explainer.md
```diff
@@ -285,19 +299,11 @@ Examples of input sources that may expose their state this way include Oculus To
...
-  - Inline sessions MUST only expose `id`s of `'unknown'`.
-  - Otherwise the `id` must be a lower-case string that describes the physical input source.
-    - The exact format is [still under discussion](https://github.com/immersive-web/webxr/issues/550), but will probably need to include at least an indication of the input device's manufacturer and model. (Such as `oculus-touch`)
-    - It must not include an indication of the handedness of the input source (such as `oculus-touch-left`), as that can be determined from the `handedness` attribute.
-    - UAs SHOULD make an effort to align on the strings that are returned for any given device.
-
-All other attributes behave as described in the [Gamepad](https://w3c.github.io/gamepad/) specification.
+The exact button and axes layout is given by the `XRInputSource`'s `gamepadLayout` attribute, which contains an array of strings that identify the button and axes layout or subsets of it, ordered from most specific to least specific.
```
NellWaliczek: > The exact button and axes layout is given by the `XRInputSource`'s `gamepadLayout`

wrong attribute name.

index.bs
thetuvix: nit: decending

index.bs
thetuvix: nit: substitue

thetuvix:
Very interesting!

These profiles are attempting to serve a dual purpose:
1. To tell the app which render model to use.
2. To tell the app how to interpret the button and axis arrays they get back from the `Gamepad` API.

However, purpose 2 is only fulfilled if the app is given `buttons` and `axes` arrays that match the particular profile that the app then happens to choose.  If any of the `"touchpad-thumbstick-wand"` profile, the `"windows-mixed-reality"` profile and the  `"valve-index"` profile differ in whether they choose the touchpad or thumbstick as their primary axis, **the app cannot correctly use the `gamepad` attribute across both the Valve Index and Samsung Odyssey controllers.**

Some possible solutions:
* **Enforce a rule that all profiles returned must provide an equivalent meaning for any button or axis IDs that overlap.**  However, that would prevent us from ever implementing both a generic `"touchpad-wand"` profile for HTC Vive-like controllers and a `"thumbstick-wand"` profile for Oculus Touch-like controllers, and then supporting them both on a Windows Mixed Reality controller.  We could decide that kind of thing is out of scope, delegating that fallback support to @NellWaliczek's app library, but then it seems odd to bother with the halfway measure of `"thumbstick-wand"` fallback support in WebXR itself.
* **Define stronger bindings from well-known controller parts to numerical IDs in the `"xr-standard"` mapping.**  For example, we could decide that axis 0 and 1 are always a primary touchpad and axis 2 and 3 are always a primary thumbstick, and so on.  If a controller only has one or the other control, the intervening array elements would be left blank.  This would also have the positive side effect of giving far more predictable button/axis IDs for well-known controller parts for those users that do not use an extra controller mapping library on top, while giving up fairly little in app compat - my experience is that an app which built its logic assuming thumbstick will not behave correctly when presented with touchpad input, and vice versa.  My recollection of why we didn't go that strongly-mapped route was that we wanted to avoid gaps in the gamepad array.  However, per the spec's current [gamepad image](https://immersive-web.github.io/webxr/images/xr-standard-mapping.svg), one of our example controllers leaves buttons[3] blank, and so we're already down that road.
* **Change the `XRInputSource.gamepad` attribute to a `getGamepad(profile)` method**, enabling the UA to adapt the controller that's in use to the controller that the app actually tested against.  This enables the UA to renumber the buttons and axes as needed to meet the app's expectation, even when the set of fallback profiles evolves organically over time in a mutually-exclusive manner.  This could even enable a UA to support a "touchpad-only" app running on a "thumbstick-only" controller, e.g. by injecting synthetic touch/untouch events when the thumbstick is moved off-center.  This is far closer to the similar OpenXR concept of "interaction profiles", which involve an explicit negotiation with the app, although it still allows the app to make the final selection of profile.
* **Change the `XRSession.inputSources` attribute to a `XRSession.getInputSources(supportedProfiles)` method**, enabling the UA to adapt all controllers in advance to the set of controllers that the app actually tested against.  This would enable the UA to be in control of selecting among the profiles that the app supports, expressing the UA's forward-compatible preference for best fit rather than the app's preference.
* **Expose only the UA's single most-preferred `profile` for a given input source**, moving away from the WebXR API supporting multiple profiles.  This relies on apps to use a higher-layer library like the one @NellWaliczek is building to support forward-compatibility to future input devices.  While this provides the least forward-compatibility in the WebXR spec itself, it also avoids some of the problems explored above if the subset of forward-compatibility we're able to tackle in WebXR 1.0 becomes awkward when we then attempt to compose in the other half of forward-compatibility in the app library.

NellWaliczek:
@thetuvix, perhaps I'm misunderstanding something here, but it feels like we can solve this without any signature changes but just ensuring we have the correct set of pre-enumerated "default profiles" for input sources with 'xr-standard' mappings.  (Which is actually the suggestion I'd made to @toji before he posted this PR because I was concerned it would be confusing otherwise....) 

To address your concern more specifically, is there a reason that the final fallback for a windows mixed reality controller wouldn't just be 'thumbstick-touchpad-wand' or whatever we end up calling it?  And for hardware with the primary mechanic being the touchpad rather than the thumbstick, then we could define a 'touchpad-thumbstick-wand'?  Without being super thorough just yet, I was roughly thinking about this list:

* touchpad-thumbstick-wand
* thumbstick-touchpad-wand
* touchpad-wand
* thumbstick-wand
* button-wand

For what it's worth, this approach doesn't preclude our adjusting the ordering of elements in the buttons/axes array.  I'm just still a bit skeptical that "fixing" the ordering will create the fallback behavior you're looking for.  If I'm understanding correctly, you're concerned that people will always assume that the first element is, say, a touchpad and will get messed up interactions because in some cases it is actually a thumbstick.  The counterpoint to that is, if that array entry is empty, developers may not look at the next entry which would leave end users with no way to interact with the content as opposed to a less than ideal way to interact.

But also!  I was kinda assuming that profiles shouldn't be limited to Gamepads!  When we get into hand tracking we'll probably want to distinguish between simple hand gesture selection and articulated hands, yes?  I was sort of assuming that this function would serve the same purpose, but perhaps I'm oversimplifying?

thetuvix:
To me, it seemed like a strong API smell for trying to solve forward-compat at this layer if an app supported `"touchpad-thumbstick-wand"` and was then incompatible with devices that supported `"thumbstick-touchpad-wand"` instead.  That kind of thing strongly suggests to me that we made the wrong tradeoff in allowing touchpads and thumbsticks to differ in their gamepad IDs rather than just assigning well-known IDs to "primary touchpad" and "primary thumbstick".

For example, if we were to do that, `"touchpad-thumbstick-wand"` and `"thumbstick-touchpad-wand"` would collapse to the same profile, and we'd avoid that entire class of app compat issue.  As you point out, the primary concern in doing so is that we would have gaps in the `axes` array, which could cause some apps enumerating axes to do the wrong thing if they stop prematurely.  _**However**_, per the image below, we are coming out of the gate with gaps in well-known controllers like Oculus Rift - at that point, WebXR content will be forced to account for such gaps.  Once content has to account for those gaps anyway, we should just go all-in and assign well-known IDs for well-known parts within the `xr-standard` mapping, avoiding the `"touchpad-thumbstick-wand"`/`"thumbstick-touchpad-wand"` gotchas in the approach above.  Note that this requires no signature changes to the WebXR or Gamepad APIs - it only requires a table in the spec to designate IDs.

xr-standard mapping

index.bs
toji: No, I just over-indexed on reusing text from the gamepadLayout description. Fixing it now.

thetuvix:
Note that the list of 5 profiles you listed actually stresses the key reason I'm interested to align on IDs here:
* touchpad-thumbstick-wand
* thumbstick-touchpad-wand
* touchpad-wand
* thumbstick-wand
* button-wand

All 5 of these profiles could be comfortably mapped to a subset of a Windows Mixed Reality controller.  However, imagine we proceed with this PR and some web app ends up authored for just one of those profiles.  There is only a 1 in 5 chance that the profile the app chose happens to be the one that is ID-compatible with the profile chosen for Windows Mixed Reality controllers. 

If, instead, we assign well-known IDs to touchpad axes and thumbstick axes, the first 4 profiles will all map identically to Windows Mixed Reality controllers, increasing app compat across the web.  For `"button-wand"`, I'd encourage us to see if we really need that one - we already have the `"onselect"` event as the mechanism we introduced to represent a "primary action" across disparate input forms.  I'm not sure we're buying enough for the tradeoffs involved in having `buttons[0]` on each `Gamepad` also represent the same thing.

index.bs
toji: Indeed, the Twitter has spoken! So let it be written, so let it be done.

toji:
I'll try to capture some of the discussion/conclusions from the call today:

Addressing the possible directions Alex laid out [in his comment above](https://github.com/immersive-web/webxr/pull/695#issuecomment-505496061) intent was to have [each profile represent a subset or superset of every other profile](https://github.com/immersive-web/webxr/pull/695/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR1362) in the array. This will definitely be easier to pull off for if we re-organize the `xr-standard` mapping to give joysticks and touchpads static mappings, so my intent is to do that in a separate PR.

In terms of the base "generic" profiles, it's apparent that we should have a list written down for what the expected set is, and that said list should not be part of the spec proper since it's likely to evolve over time. I'll work on producing a starting point for that, possibly as a separate PR. Related to making the `xr-standard` mapping above more rigid, it's also my intent to *not* support both a "touchpad-thumbstick-controller" _and_ a "thumbstick-touchpad-controller" profile, as that would imply that we're trying to communicate a precedence that may not actually be there, and is unlikely to be meaningful to developers who, in the presence of both types of input, will likely make app-specific decisions about how to use them.

Additionally, it's not my expectation that we would have more than one generic profile in the list at any point. ie: While `["touchpad-thumbstick-controller", "touchpad-controller", "thumbstick-controller"]` can certainly be argued to be accurate, it shouldn't be necessary as long as the generic profile set is reasonably sized and well defined. This is something that we can continue to evaluate, though, and also enable after the fact if it becomes necessary.

There's an argument to be made that controller interactions can be handled more robustly by the platform if the API communicates how it wants to use the controller. For example: Touchpads can be silently remapped to buttons if necessary. As such it might be compelling to let the app tell the API what profiles it wants to support and allow the API to report back what it was able to conform to. It's my view currently that while this kind of capability could indeed be useful we should stick to trying to expose as much raw data as we can via the Gamepad API for our initial release and enable libraries to provide that sort of mapping initially. That would allow our group more time to see how OpenXR and similar systems evolve and how well they are adopted by the dev community before committing to a more sophisticated path, likely in a new module or a level 2 of the proposed gamepad module.

toji:
Opened #735 to remove any ambiguities from the `xr-standard` mapping and thus make the profiles a bit more practical.

NellWaliczek:
Overall, great summary.  There's one thing that differed in my recollection

> There's an argument to be made that controller interactions can be handled more robustly by the platform if the API communicates how it wants to use the controller. For example: Touchpads can be silently remapped to buttons if necessary. As such it might be compelling to let the app tell the API what profiles it wants to support and allow the API to report back what it was able to conform to. 

For this, I think the suggestion was to keep the `profiles` array as is, and add an additional API developers can call to ensure that the buttons behave in alignment with the profile they will actually be using.  In otherwords:

```js
let profiles = xrInputSource.profiles;

// use some heuristic that isn't this, but is actually meaningful
let selectedProfile = profiles[profiles.length - 1];

xrInputSource.setSelectedProfile(selectedProfile);
```

@thetuvix, does that match your recollection?

toji:
I don't recall that we reached the point of agreeing on that recommendation on the call, and I"m not in favor of it at this time. That feels more appropriate for a new input module or a level 2 gamepad module down the road. (I can generally see that sort of API being most practical to pair with an action-mapping API, like OpenXR has.)

NellWaliczek:
Ah, sorry I didn't mean we'd made that a *recommendation*.  Just that it was what we were discussing as an option.

toji:
Also opened #738 to outline an initial set of generic profiles.

index.bs
```diff
@@ -1353,6 +1354,24 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
...
+
+Any [=input profile name=]s given after the first entry in the list should provide fallback values that represent alternative representations of the device. This may include a more generic or prior version of the device, a more widely recognized device that is sufficently similar, or a broad description of the device type (such as "touchpad-wand"). If multiple profiles are given, the {{XRInputSource/gamepad}} layouts they describe must all represent a superset or subset of every other profile in the list.
+
+If the {{XRSession}}'s [=mode=] is {{XRSessionMode/inline}}, {{XRInputSource/profiles}} MUST be an empty list.
+
+If the input device cannot be reliably identified, or the user agent wishes to mask the input device being used, it MAY choose to only report generic [=input profile name=]s or an empty list.
```
NellWaliczek: Given that we're now enumerating generic profile names in PR #738, how do we feel about requiring the profiles list to have at least one generic profile name in the list?

toji: We should absolutely strongly recommend it, but I'm not sure if I'm comfortable requiring that for a few reasons:
- If we're going to require that we really should put the profile names in the spec.
- I don't think we'll have a generic profile for _everything_ that might be exposed. (Hands, tracking pucks, stylus, body tracking markers...)
- It's possible that there's some situations in which the underlying native API doesn't give you a sense of what the device looks like? (Especially true if you're using some custom attachment to a tracking puck.)

NellWaliczek: That makes sense, though I do feel like we could reasonably bend the rules on the first point if we wanted to require a generic profile name for "masked" hardware.  Not a blocking issue, though.  If this turns out to be a problem we can always add it later.

index.bs
```diff
@@ -1353,6 +1354,24 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
 
 The {{gripSpace}} MUST be <code>null</code> if the input source isn't inherently trackable such as for input sources with a {{targetRayMode}} of {{XRTargetRayMode/gaze}} or {{XRTargetRayMode/screen}}.
 
+The <dfn attribute for="XRInputSource">profiles</dfn> attribute is a [=/list=] of [=input profile name=]s indicating both the preffered visual representation and behavior of the input source. For example, if the {{XRInputSource/gamepad}} attribute is not <code>null</code> the profile identifies the mapping of the buttons and axes of the attribute of the {{XRInputSource}}. Profiles are given in descending order of preference.
```
NellWaliczek: >Profiles are given in descending order of preference.

Seems like this might make more sense at the end of line  1361, but feel free to ignore

index.bs
NellWaliczek: Eh, I was overthinking this one.

input-explainer.md
```diff
@@ -230,11 +230,11 @@ The majority of `tracked-pointer` input sources will have a non-null `gamepad` a
 
 The WebXR Device API currently does not offer any way to retrieve renderable resources that represent the input devices from the API itself, and as such the `XRInputSource`'s `profiles` must be used to identify and load an appropriate resources. (For example, from the application's server, a CDN, or a local cache.) The `profiles` provides a list of strings that identify the device, given in descending order of preference.
 
-For example, the Samsung Odyssey controller is a variant of the standard Windows Mixed Reality controller. As a result, the `profiles` for a that controller could be:
+For example, the Samsung Odyssey controller is a variant of the standard Windows Mixed Reality controller. As a result, the `profiles` for that controller could be:
 
 ```js
 // Exact strings are examples only.
-["samsung-odyssey", "windows-mixed-reality"]
+["samsung-odyssey", "windows-mixed-reality", "touchpad-thumbstick-controller"]
```
NellWaliczek: @thetuvix , given that this is your hardware, would you prefer the example strings to be different?

toji: I'm curious to hear Alex's take on this, but given the time zone differences at the moment I'm going to assume that we can safely handle any concerns that do come up as a follow-up PR.

thetuvix: "Samsung Odyssey" is a headset and "Windows Mixed Reality" is a platform, so I'd probably go with `samsung-odyssey-controller` and `windows-mixed-reality-controller`, to differentiate from `windows-mixed-reality-clicker`, etc.

However, when we populate the initial registry with the identifiers we expect to be used at WebXR 1.0, we can shake out this example to match the registry then.

PR #699, opened 2019-06-14
Remove context Attribute from XRWebGLLayer

bricetebbs:
I have no idea what I'm doing so I'll explain what I did. I made what seemed like the obvious change in index.bs and tried to find any other references in the file.  I searched through the explainer to see if it needed changes and it didn't seem to. I also checked a few of the code samples and they also don't seem to us this attribute.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -675,7 +675,6 @@ typedef (WebGLRenderingContext or
              XRWebGLRenderingContext context,
              optional XRWebGLLayerInit layerInit)]
 interface XRWebGLLayer {
-  readonly attribute XRWebGLRenderingContext context;
   readonly attribute boolean antialias;
   readonly attribute boolean ignoreDepthValues;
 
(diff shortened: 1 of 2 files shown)


bricetebbs:
doh! forgot about that one. Will fix soon.

toji:
Yes, thanks for the help, Brice!

There's a few references that you missed in the spec doc as well: Basically just do a search for `{{XRWebGLLayer/context}}` and you'll see where it's referenced in spec text or algorithms. In some places the text just may need to be removed, but in others (especially any time it's used in an algorithm or when describing the behavior of other objects) you'll want to replace it with a reference to an internal value.

This means simply that you'll want to add some text to the `XRWebGLLayer` section saying something like:

> Each {{XRWebGLLayer}} has a <dfn for="XRWebGLLayer">context</dfn> object, initially <code>null</code>, which is an instance of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.

At which point you can do a straight replacement of `{{XRWebGLLayer/context}}` with `[=XRWebGLLayer/context=]` in most places and everything will just work out.

I highly recommend running `make` in your repo folder to trigger a build of the bikeshed file when you're done. That'll let you know if there's any warnings or errors. (It's expected to come back clean.)

bricetebbs:
Closing this in favor of #707

Issue #678, opened 2019-05-31
Consider replacing "list of pending render states" with "pending render state"


ddorwin:
Having a list just adds a little more complexity and potential confusion for the user, so it'd be nice to collapse it if possible.

It seems that `updateRenderState()` could maintain a single nullable "pending render state" that gets updated (as in step 5 of https://immersive-web.github.io/webxr/#apply-pending-render-states) and when not null gets applied in https://immersive-web.github.io/webxr/#apply-pending-render-states.

toji:
Editors talked about this and remembered that the current design was due to a previous design that had us resolving a promise every time a change was applied. That's not longer a concern, however, so this simplification is fine.

Manishearth:
https://github.com/immersive-web/webxr/pull/701

PR #701, opened 2019-06-14
Replace 'list of pending render states' with 'pending render state'

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/678

cc @ddorwin

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -455,7 +455,7 @@ The <dfn method for="XRSession">end()</dfn> method provides a way to manually sh
 
 </div>
 
-Each {{XRSession}} has an <dfn>active render state</dfn> which is a new {{XRRenderState}}, a <dfn>list of pending render states</dfn>, which is initially empty.
+Each {{XRSession}} has an <dfn>active render state</dfn> which is a new {{XRRenderState}}, and a <dfn>pending render state</dfn>, which is an {{XRRenderStateInit}} which is initially <code>null</code>.
 
 The <dfn attribute for="XRSession">renderState</dfn> attribute returns the {{XRSession}}'s [=active render state=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -469,23 +469,22 @@ When the <dfn method for="XRSession">updateRenderState(|newState|)</dfn> method
   1. If |session|'s [=ended=] value is <code>true</code>, throw an {{InvalidStateError}} and abort these steps.
   1. If |newState|'s {{XRRenderStateInit/baseLayer}}'s was created with an {{XRSession}} other than |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |newState|'s {{XRRenderStateInit/inlineVerticalFieldOfView}} is set and |session| is an [=immersive session=], throw an {{InvalidStateError}} and abort these steps.
-  1. Append |newState| to |session|'s [=list of pending render states=].
+  1. Set |session|'s [=pending render state=] to |newState|.
```
toji: This doesn't work! Calls to `updateRenderState()` are provided partial state, which is accumulated into the pending state. So, the following:

```js
session.updateRenderState({ depthNear: 1, depthFar: 10 });
session.updateRenderState({ baseLayer: someLayer });
session.updateRenderState({ depthNear: 2 });
```

Will result in a `renderState` on the next frame of:

```js
{
  depthNear:2,
  depthFar: 10,
  baseLayer: someLayer
}
```
In order to accomplish that, only the values that have been explicitly set in the dictionary should be written into the pending state each call. Effectively you'll need to lift the part of the algorithm that says:

> If |newState|'s {{XRRenderStateInit/depthFar}} value is set, set |activeState|'s {{XRRenderState/depthFar}} to |newState|'s {{XRRenderStateInit/depthFar}}...

etc. and place it into the `updateRenderState()` call algorithm instead. Then, during the `apply-pending-render-state` algorithm the state can simply be copied over wholesale.

Manishearth: Oh, good catch. Hmmm.

toji:
Perfect! Thank you!

Issue #607, opened 2019-04-24
When a session is shut down, should this clear the xrCompatible flag?


asajeffrey:
Currently, ending an XRSession (https://immersive-web.github.io/webxr/#shut-down-the-session) doesn't clear the xrCompatible flag of any WebGL rendering context. As a result, ending a session, then beginning a new session on a device with a different GPU, results in the xrCompatible flag being true, even though the context is using an incompatible GPU.

Manishearth:
https://github.com/immersive-web/webxr/pull/705

toji:
Sorry for the confusion, but I don't think this is a concern. If a session shuts down and another is created with the same XR device we definitely want to keep the compatibility bit valid. The only time that it should become invalid is if the underlying device that new sessions would be created against changes, and we already have it specified that making that change does invalidate the compatibility bits. See http://localhost:8080/webxr/#select-an-xr-device

@NellWaliczek: Is there something else about this issue that I'm not remembering that we wanted to address? Feels like we should have caught this during triage, but maybe we did and there was an aspect of it that we still wanted to look into?

asajeffrey:
Ah, so the flag isn't cleared when the session is shut down, but it is cleared when a new session is begun. That works.

PR #705, opened 2019-06-14
Don't check XR compat flag for inline sessions

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/607

r? @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1611,7 +1611,7 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
   1. Let |layer| be a new {{XRWebGLLayer}}
   1. If |session|'s [=ended=] value is <code>true</code>, throw an {{InvalidStateError}} and abort these steps.
   1. If |context| is lost, throw an {{InvalidStateError}} and abort these steps.
-  1. If |context|'s [=XR compatible=] boolean is false, throw an {{InvalidStateError}} and abort these steps.
+  1. If |session| is an [=immersive session=] and |context|'s [=XR compatible=] boolean is <code>false</code>, throw an {{InvalidStateError}} and abort these steps.
   1. Initialize |layer|'s [=XRWebGLLayer/context=] to |context|.
   1. Initialize |layer|'s {{XRWebGLLayer/ignoreDepthValues}} as follows:
     <dl class="switch">


Manishearth:
I don't think this is the complete solution. This doesn't address the case where there are multiple inline sessions present, since it invalidates the other inline sessions. A fix for that is to _only_ unset the flag for `baseLayer`, but that doesn't solve the issue of creating a compatible context and using it for a different device later.

It feels like `makeXRCompatible()` should take a session parameter and mark the context as compatible for that session only.

Alternatively, we can make inline sessions skip this flag completely. I might just do that.

Manishearth:
> Alternatively, we can make inline sessions skip this flag completely. I might just do that.

This was cleaner, ended up doing it.

cc @asajeffrey

toji:
I don't think this is the right approach. As long as the underlying XR device hasn't changed there's no reason that the compatibility bit would become invalid between one XR session and the next. The only time that this would be problematic is if the XR device is changed, and that should already be covered by the algorithm here: http://localhost:8080/webxr/#select-an-xr-device

In retrospect I'm not sure why we left #607 open, as it likely could have been easily explained and closed.

Manishearth:
In that case should I keep the portion of this PR that ignores the flag for inline sessions?

Manishearth:
Since that's a problem regardless of this.

toji:
Editors talked this over and we do want to keep the text about ignoring the flag for inline sessions, but we feel the current context compatibility set/unset logic is correct and should not be changed. If you can refactor the PR to only include the change to the inline check, we'll review it again. Thanks!

Manishearth:
Done!

index.bs
```diff
@@ -1605,7 +1605,7 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
   1. Let |layer| be a new {{XRWebGLLayer}}
   1. If |session|'s [=ended=] value is <code>true</code>, throw an {{InvalidStateError}} and abort these steps.
   1. If |context| is lost, throw an {{InvalidStateError}} and abort these steps.
-  1. If |context|'s [=XR compatible=] boolean is false, throw an {{InvalidStateError}} and abort these steps.
+  1. If |context|'s [=XR compatible=] boolean is false and |session|'s [=XRSession/mode=] is not {{XRSessionMode/"inline"}}, throw an {{InvalidStateError}} and abort these steps.
```
toji: Let's change this to use the [=immersive session=] definition instead, which should be a bit more future-proof.

> If |session| is an [=immersive session=] and |context|'s [=XR compatible=] boolean is &lt;code&gt;false&lt;/code&gt;, throw an {{InvalidStateError}} and abort these steps.

Manishearth:
:rofl: 

done!

toji:
👍

PR #707, opened 2019-06-15
Second try @ Remove context Attribute from XRWebGLLayer 

bricetebbs:
This is a redo of #699 since I made some mistakes in that one I am not sure how to fix.

I think it addressed the comments on that PR

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1583,8 +1583,6 @@ dictionary XRWebGLLayerInit {
              optional XRWebGLLayerInit layerInit)]
 interface XRWebGLLayer {
   // Attributes
-  [SameObject] readonly attribute XRWebGLRenderingContext context;
-
   readonly attribute boolean antialias;
   readonly attribute boolean ignoreDepthValues;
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1600,6 +1598,8 @@ interface XRWebGLLayer {
 };
 </pre>
 
+Each {{XRWebGLLayer}} has a <dfn for="XRWebGLLayer">context</dfn> object, initially <code>null</code>, which is an instance of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
```
Manishearth: You can just say that this is an `{{XRWebGLRenderingContext}}` (but you don't have to, this is fine too)

toji: I think this is actually slightly clearer in this case, since it avoids the need to cross-reference with other parts of the spec. Since it's not a huge difference in the size of the text, nor is it something that gets repeated a lot elsewhere, I'm good with it.

toji:
Perfect, thanks! (And let me know if you want a mini tutorial on updating an existing PR sometime! It's not hard, but it's a process that likes to throw intimidating error messages at you that you.)

Issue #484, opened 2019-01-16
Define the creation of a swap chain, or remove references


cwilso:
The XRWebGLLayer constructor alogorithm (https://immersive-web.github.io/webxr/#dom-xrwebgllayer-xrwebgllayer) refers to the swap chain (steps 11 and 12).  It's not clear that we should get this deep into the process by which the browser takes the pixels and shuffles them off to a device -  the native apis tend to refer to DirectX,  but we can't presuppose a particular underlying API.  We can talk about WebGL buffers but it's not clear if we should define swap chain better or just not go this deep - e.g. "Step 11: Initialize any required render API constructs, and if they are unable to be creating throw an OperationError and abort these steps."

kearwood:
IIUC, the reason for including "initialize the layer's swap chain" here is to ensure that the XRWebGLLayer constructor fails if the UA is unable to allocate GPU-side resources needed to support the layer rendering.  Perhaps we could describe this in a more generic manner:

11. Allocate and initialize resources, including GPU accessible memory buffers, as required to support the compositing of the XRWebGLLayer.
12. If layer’s resources were unable to be created for any reason, throw an OperationError and abort these steps.

Manishearth:
https://github.com/immersive-web/webxr/pull/708

PR #708, opened 2019-06-15
Remove references to swap chain

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/484

(I just used @kearwood's suggested language there, seemed good enough)

r? @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1628,8 +1628,8 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
       <dd>
         1. Initialize |layer|'s {{XRWebGLLayer/antialias}} to |layerInit|'s {{XRWebGLLayerInit/antialias}} value.
         1. Initialize |layer|'s {{XRWebGLLayer/framebuffer}} to a new [=opaque framebuffer=] created with |context|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1628,8 +1628,8 @@ The <dfn constructor for="XRWebGLLayer">XRWebGLLayer(|session|, |context|, |laye
       <dd>
         1. Initialize |layer|'s {{XRWebGLLayer/antialias}} to |layerInit|'s {{XRWebGLLayerInit/antialias}} value.
         1. Initialize |layer|'s {{XRWebGLLayer/framebuffer}} to a new [=opaque framebuffer=] created with |context|.
-        1. Initialize a [=swap chain=] for |layer| that is compatible with |session|'s [=/XR device=].
-        1. If |layer|'s [=swap chain=] was unable to be created for any reason, throw an {{OperationError}} and abort these steps.
+        1. Allocate and initialize resources, including GPU accessible memory buffers, as required to support the compositing of |layer|.
```
toji: I feel it's important to keep the bit about the resources being "compatible with |session|'s [=/XR device=]", so could you add that back in? Otherwise looks good! Definitely want to expand on this more in the future but this is a cleaner way of expressing it in the meantime. Thanks!

toji:
Great, thanks!

Issue #469, opened 2019-01-10
Restrictions on an immersive session


AdaRoseCannon:
BS Issue: https://immersive-web.github.io/webxr/#issue-7782964d

Immersive sessions have exclusive access to the hardware so that other applications do not interfere.

> Additionally, Immersive sessions MUST be given exclusive access to the XR device, meaning that while the immersive session is not blurred the HTML document is not shown on the XR device's display, nor is content from other applications shown on the XR device's display.
https://immersive-web.github.io/webxr/#immersive-session

Whilst this enables great performance it also may result in some restrictions in implementation and what can be built using the API. Which may need to be expanded upon such as:

What does this mean re: Browser and Operating System user interfaces, how much content can the user agent overlay on the render to assist the user or enable normal browsing?

Does this limit the possibility of layered AR apps? I.e. leaving a website as a virtual object in a room to come back and interact with later. Or a weather AR web page to easily check the weather whilst working on other things.

DoDecaDence:
Is this the reason why immersive sessions are not available anymore since the change of requesting them directly `via navigator.xr.requestSession`, instead of requesting an object instance representing the XR device?

As the page currently states itself, the _ar-with-webxr_ codelabs example [does not work anymore](https://googlecodelabs.github.io/ar-with-webxr/).

Since then, even when adapting the respective lines of code that have been altered with the latest API changes, entering an immersive session with an active camera won't work anymore:
![webxr_immersivearsessionnotavailable_chrome_2019-02-12_14-55-12](https://user-images.githubusercontent.com/7498626/52640562-11069080-2ed7-11e9-9a1a-c569475b177f.png)

Only the session mode 'inline' will be accepted. (IIRC, that actually always had been be the case; however it was seemingly ignored, as it did not prevent the camera from working for the Google Pixel.)
As this is not an immersive session, the `session.environmentBlendMode` is set to '`opaque`'. The device orientation is still tracked to update the camera position and thus look around in the rendered scene. But the environment itself won't be visible, and `session.requestHitTest()` will also be rejected.

@alon-grinshpoon appears to have [the same issue](https://github.com/googlecodelabs/ar-with-webxr/issues/8#issuecomment-451639588).

> Whilst this enables great performance it also may result in some restrictions in implementation and what can be built using the API.

You are on point. I did indeed have quite the performance issues with my Google Pixel, where the plane detection features turned out to be the culprit. (Covering the camera fixed the bad framerate.)  
[...]

PR #709, opened 2019-06-17
Clarify that small overlay UIs are allowed in exclusive access

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/469

I don't really address the "immersive session left at a part of the room to be returned to" concern, I'm not sure how that can be handled or if that even should be handled by immersive-ar the same way.

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -373,7 +373,7 @@ An <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or an {{imm
 
 NOTE: Examples of ways [=exclusive access=] may be presented include stereo content displayed on a virtual reality or augmented reality headset, or augmented reality content displayed fullscreen on a mobile device.
 
-NOTE: As an example of overlaid UI, the user-agent or operating system in {{immersive-ar}} mode may overlay mandatory "home" and navigational buttons over the user's wrist.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -369,10 +369,13 @@ enum XRSessionMode {
...
+An <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or an {{immersive-ar}} session. [=Immersive sessions=] MUST provide some level of [=viewer=] tracking, and content MUST be shown at the proper scale relative to the user and/or the surrounding environment. Additionally, [=Immersive sessions=] MUST be given <dfn>exclusive access</dfn> to the [=/XR device=], meaning that while the [=immersive session=] is {{XRVisibilityState/visible}} the HTML document is not shown on the [=/XR device=]'s display, nor does content from any other source have exclusive access. [=Exclusive access=] does not prevent the browser or user agent from overlaying its own UI, however this UI SHOULD be minimal.
 
 NOTE: Examples of ways [=exclusive access=] may be presented include stereo content displayed on a virtual reality or augmented reality headset, or augmented reality content displayed fullscreen on a mobile device.
 
+NOTE: As an example of overlaid UI, the user-agent or operating system in {{immersive-ar}} mode may overlay mandatory "home" and navigational buttons over the user's wrist.
```
toji: I support of this change, but the example given here is pretty AR-centric and only applies to a single device that I'm aware of. We don't have to remove it, but could we add a reference to something more common, like notification messages? I've seen that on a lot more platforms.

Manishearth: Done.

toji:
LGTM!

Issue #467, opened 2019-01-10
Document what happens when the session has ended


AdaRoseCannon:
BS Issue: https://immersive-web.github.io/webxr/#end-the-session

The session may end in a variety of situations such as end method of the XRSession is invoked: https://immersive-web.github.io/webxr/#dom-xrsession-end

When this happens the session permanently shuts down. The process is described here: https://immersive-web.github.io/webxr/#shut-down-the-session

We should elaborate on some of the platform-specific steps to shut down the device’s tracking and rendering capabilities.

AdaRoseCannon:
Should probably include at what step the end event is fired as detailed in: https://immersive-web.github.io/webxr/#event-types

Artyom17:
Another question: should "inline" session fire "end" event when switching to "immersive-vr" occurs? This is not really clear, IMO.

toji:
No. Inline sessions should not be ended by an immersive session starting. They should be suspended till the immersive session ends (firing what's currently the "blur" and "focus" events respectively.) And yes, this should definitely be better documented.

Manishearth:
So, it's not clear to me what the difference between end and shut down is. It seems like they should be the same process.

Is it that "end the session" is what should happen when you call `.end()`, and other ways of shutting down the session are "shut down the system" (with "end the session" calling "shut down the session"). Which ones should set the ended flag? Which ones should trigger onend?

Manishearth:
_tentative_ fix here: https://github.com/immersive-web/webxr/pull/710

PR #710, opened 2019-06-17
Merge 'end the session' with 'shut down the session', clarify, add onend event

Manishearth:
Potential fix for https://github.com/immersive-web/webxr/issues/467

Not quite sure of this, also not sure of what precisely we should list in the platform specific steps.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -446,7 +446,7 @@ When an {{XRSession}} is shut down the following steps are run:
   1. [=Reject=] any outstanding promises returned by |session| with an {{InvalidStateError}}, except for any promises returned by {{XRSession/end()}}.
   1. If no other features of the user agent are actively using them, perform the necessary platform-specific steps to shut down the device's tracking and rendering capabilities. This MUST include:
     - Releasing [=exclusive access=] to the [=/XR device=].
-    - Deallocating any framebuffers attached by the |session| to the [=/XR device=].
+    - Deallocating any graphics resources acquired by |session| for presentation to the [=/XR device=].
     - Putting the [=/XR device=] in a state such that a different source may be able to initiate a session with the same device.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -443,8 +443,12 @@ When an {{XRSession}} is shut down the following steps are run:
   1. Set |session|'s [=ended=] value to <code>true</code>.
   1. If the [=active immersive session=] is equal to |session|, set the [=active immersive session=] to <code>null</code>.
   1. Remove |session| from the [=list of inline sessions=].
-  1. [=Reject=] any outstanding promises returned by |session| with an {{InvalidStateError}}.
-  1. If no other features of the user agent are actively using them, perform the necessary platform-specific steps to shut down the device's tracking and rendering capabilities.
+  1. [=Reject=] any outstanding promises returned by |session| with an {{InvalidStateError}}, except for any promises returned by {{XRSession/end()}}.
```
toji: Ooh, hey! Good catch. 😅

index.bs
```diff
@@ -443,8 +443,12 @@ When an {{XRSession}} is shut down the following steps are run:
...
+  1. If no other features of the user agent are actively using them, perform the necessary platform-specific steps to shut down the device's tracking and rendering capabilities. This MUST include:
+    - Releasing [=exclusive access=] to the [=/XR device=].
+    - Deallocating any framebuffers attached by the |session| to the [=/XR device=].
```
toji: "Attached" doesn't seem like the right word here? Also it may not always be framebuffers specifically, so maybe something more like this (keeping in line with your recent swap chain PR)?

> Deallocating any graphics resources acquired by the |session| for presentation to the [=/XR device=].

Manishearth: That makes sense. Getting on a flight soon, will push this fix later.

toji: Safe travels!

toji:
👏

PR #714, opened 2019-06-17
Simplify handling of unsupported modes in requestSession()

ddorwin:
After #685, there is only one failure path at this point, and these steps can be simplified.

This also addresses incorrectly rejecting with `null`.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -311,20 +311,12 @@ When the <dfn method for="XR">requestSession(|mode|)</dfn> method is invoked, th
   1. If |immersive| is <code>true</code>:
     1. If the algorithm is not [=triggered by user activation=], [=reject=] |promise| with a "{{SecurityError}}" {{DOMException}} and return |promise|.
     1. If [=pending immersive session=] is <code>true</code> or [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}} and return |promise|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -311,20 +311,12 @@ When the <dfn method for="XR">requestSession(|mode|)</dfn> method is invoked, th
...
     1. If [=pending immersive session=] is <code>true</code> or [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}} and return |promise|.
-    1. Set [=pending immersive session=] to be <code>true</code>.
+    1. Set [=pending immersive session=] to <code>true</code>.
   1. Run the following steps [=in parallel=]:
     1. [=ensures an XR device is selected|Ensure an XR device is selected=].
     1. [=Queue a task=] to perform the following steps:
-        1. Follow the steps for the first matching conditions:
-            <dl class="switch">
-              <dt> If the [=XR/XR device=] is <code>null</code>
-              <dd> [=Reject=] |promise| with <code>null</code>.
```
Manishearth: Do we want to lose this distinction between promises rejected as null and notsupported? I don't think it's important.

toji: I don't think it's an important distinction either. I can't imagine that a developer is going to do anything useful with that knowledge besides maybe fingerprinting.

toji:
LGTM, thanks!

Issue #712, opened 2019-06-17
Use "Queue a task *to*..." consistently


ddorwin:
Some algorithms say "Queue a task that fires..." while others say "Queue a task to..." I believe the latter is more common. (Note that not all text includes "Queue a task that" immediately before "fires." Once this issue is resolved, the word "fires" should only appear once in the spec in "When the user agent fires...")

Related, "fire an event" should probably link to https://dom.spec.whatwg.org/#concept-event-fire. "Fire an <custom event>" should probably also link somewhere.

Finally, note that "fire a simple event", which is used in the spec, is now "fire an event": https://github.com/whatwg/html/commit/3d19b31e62debaa6132744a530d3d7e33e0bc269

Manishearth:
https://github.com/immersive-web/webxr/pull/715

PR #715, opened 2019-06-17
Cleanups on event text

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/712

 - Consistently say "Queue a task to fire"
 - Don't say "fire a simple event", just say "fire an event"
 - Link to DOM spec for firing an event

cc @ddorwin


toji:
LGTM, thanks!

PR #716, opened 2019-06-19
Fix detached array in XRRay.matrix algorithm

bialpio:
Fix algorithm for obtaining XRRay.matrix - copy-paste error introduced by PR #684.


toji:

toji:
Thanks! Marked this change as non substantive because it's effectively a typo correction. That way you don't have to go through the IPR process prior to merging.

bialpio:
Thanks! I've linked my W3C account to the github account but only after issuing the PR & seeing some red there. :)

PR #717, opened 2019-06-19
Document how to transition between window and session animation loops.

### LINKED ISSUE #352: Easy for app developers to break render loop initialization on mobile.
Both @mrdoob and I ran into an issue when using Chrome's WebXR implementation with a traditional WebGL app recently, which highlighted a potential API ergonomics issue for me. Given that frequently the difference between a WebGL rendering path and a WebXR one will be relatively minor, it's not unreasonable to expect developers to do something like this:

```js
function onFrame(t, xrFrame) {
  if (xrSession) {
    xrSession.requestAnimationFrame(onFrame);
  } else {
    window.requestAnimationFrame(onFrame);
  }
  // Rendering logic
}
window.requestAnimationFrame(onFrame); 

function onEnterXRButtonClick() {
  xrDevice.requestSession({ exclusive: true }).then((session) => {
    xrSession = session;
    // Additional setup
  });
}
```

The developers expectation is that when they set `xrSession` the next time onFrame is called it'll kick off the XR rendering loop for them (one frame late, but that's not a big deal.) And it _does_ work that way... on Chrome's desktop implementation. On mobile we halt the `window.rAF` render loop once VR starts because the window is no longer visible. This means that it's a race condition whether a final rAF gets called or not, but more often than not the result is that an app which appears to work perfectly well on desktop won't render any frames on mobile.

The solution is to explicitly call `xrSession.requestAnimationFrame()` once when the session is setup, but then that leaves an alternative problem: On desktop the above code with that added initial rAF will now double up on the XR frame callbacks. You could make them separate function calls, but then you'll also be running a `window.rAF` alongside your `session.rAF` on desktop.

So the REAL answer is: Track your last window or session rAF handle and when you switch modes explicitly cancel the previous one before starting the new rAF loop. That's unfortunately non-intuitive and fragile. Is there a reasonable way we can guide users into the right behavior or make this less troublesome at an API level? Or do we collectively think this is something that just needs to be well documented?

toji:
/fixes #352

Due to the variations between how different platforms handle window
animation frames when immersive content is running, making sure that
apps handle the transition between the window and session animation
loops gracefully is unfortunately unintuitive. This adds an example that
demonstrates one way to do it correctly, while also explaining the
problem.

@NellWaliczek: I'm a bit torn on where to put this. It feels out-of-place
in the spec. I can easily move the text over to the explainer if you feel
it would be more appropriate, but I worry that it would be overlooked
there. Thoughts?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -81,6 +81,7 @@ spec: Gamepad; urlPrefix: https://w3c.github.io/gamepad/#
     type: attribute; text: pressed; for: GamepadButton; url: dom-gamepadbutton-pressed
     type: attribute; text: value; for: GamepadButton; url: dom-gamepadbutton-value
 spec:html; type:method; for:HTMLCanvasElement; text:getContext(contextId); url: https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext
+spec:html; type:method; for:Window; text:requestAnimationFrame(callback); url: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-animationframeprovider-requestanimationframe
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: method; text: IsDetachedBuffer; url: sec-isdetachedbuffer
 spec: dom; urlPrefix: https://dom.spec.whatwg.org/#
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -724,6 +725,61 @@ When an {{XRSession}} |session| receives updated [=viewer=] state from the [=/XR
 
 </div>
 
+<div class="example">
+Depending on the device being used, callbacks supplied to {{Window}} {{Window/requestAnimationFrame()}} may not be processed while and [=immersive session=] is active. (For instance, on a mobile or standalone device where the immersive content completely obscures the HTML document.) As such, developers should not rely on {{Window}} {{Window/requestAnimationFrame()}} callbacks to schedule {{XRSession}} {{XRSession/requestAnimationFrame()}} callbacks and visa-versa, even if they share the same rendering logic, or the applications animation loop may not run properly on all platforms. A more effective pattern for applications that wish to transition between these two types of animation loops is demonstrated below:
```
Manishearth: Some browsers support "warnings" in devtools. Should we have a non-normative note suggesting browsers show a warning for `session.rAF()` called within `window.rAF()` or vice versa?

toji: I've wondered about it, but I'm not fully convinced that there's no legitimate use cases for something like that? I don't know, would have to give it some more thought.

NellWaliczek: Heck, could we do something more drastic and reject `session.rAF()` if invoked from `window.rAF()`?  If that's too much, maybe we reject it from there if it's never been hooked before?

Thinking about the consequences of such a choice... in the future, some layers won't ever require processing a rAF (for example a hardware-backed video projection layer).... so what would happen when the video ends and the developer wants to start rendering their own content with a traditional layer type... when would we expect them to hook the `session.rAF()`?  I do worry they might try to use the `window.rAF()` if not prevented from doing so...

Manishearth: > Heck, could we do something more drastic and reject session.rAF() if invoked from window.rAF()? 

I considered suggesting this but it felt a bit too drastic to me. I'm not against it though.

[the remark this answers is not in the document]
toji: We MAY say "must not" but we MUST NOT say "MUST NOT". 😉

toji:
> I'm still a bit confused by the sample code because I'm not sure the purpose of the windowAnimationSuspended variable.

It was to flag that a new `window.rAF` call needed to be scheduled when the session ends, but I agree that it was confusing. I refactored the code a bit to remove the need for that variable at the expense of potentially letting `window.rAF` spin without rendering during the session. Makes the code slightly more concise and focuses it on the core problem more.

Addressed the rest of your feedback as well. Any further thoughts?

toji:
Filed #736. Thanks!

PR #728, opened 2019-06-21
Fix algorithm to compute XRRay.matrix

bialpio:
Fixes for current algorithm to obtain XRRay.matrix:
- Dot product does not result in an angle between 2 vectors but is used as such.
- Magnitude of cross product of 2 vectors can be 0 but rotation can still be necessary.

Additionally, it seems that there used to be some trailing whitespace that got removed by my editor.


toji:
LGTM! Thanks for the whitespace cleanup too. :)

Issue #721, opened 2019-06-19
Duration of user consent


NellWaliczek:
**[Disclaimer: This issue is one of several being filed to capture discussions that began either on #638, on #689, or at the most recent F2F]**

When user consent is required and granted, there are three options I’m aware of for how long user consent can last. (Note, this is not the issue for discussing what triggers UAs to request consent. That can be found in #720 )

* Session, which means each new XRSession would require new prompting
* Browsing-context, which is defined in the HTML Standard and implies user consent would be cumulative when multiple are XRSessions created in the same browsing context.
* Origin, which is defined in the HTML Standard and implies user consent would be cumulative for each XRSession created by that origin

Are there any other options I’ve left out? Is this a decision which can be left up to each User Agent or is this something that must be consistent across all? Are there minimum requirements for the duration based on the type of data being secured?

toji:
Surfacing the relevant text from #638 privacy doc for easy reference:

> #### Lifetime of Consent
> The following guidelines are suggested:
> *   User consent should only be considered valid for the [browsing context](https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#browsing-context) within which it was obtained.
> *   Once a specific consent is obtained for a specific [origin](https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#origin-0) in a [browsing context](https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#browsing-context), that [origin](https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#origin-0) does not need to obtain that specific consent again during the lifetime of that [browsing context](https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#browsing-context). Specifically, if multiple same-origin [XRSession](https://immersive-web.github.io/webxr/#xrsession-interface) objects are created in a [browsing context](https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#browsing-context), and all require the same user consent, then consent should only need to be obtained once.
> *   The user agent must ensure that all mandatory conditions for user consent are met before creating any [XRSession](https://immersive-web.github.io/webxr/#xrsession-interface) object. As a result, the user agent may be required to ask for user consent multiple times in a [browsing context](https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#browsing-context) if it is creating multiple [XRSession](https://immersive-web.github.io/webxr/#xrsession-interface) objects each with different mandatory conditions for user consent.

[...]

Issue #720, opened 2019-06-19
Timing of user consent requests


NellWaliczek:
**[Disclaimer: This issue is one of several being filed to capture discussions that began either on #638, on #689, or at the most recent F2F]**

When getting user consent is required to protect sensitive information, there appear to be three options for when the User Agent can ask for the consent. (Note, this is not the issue for discussing the duration of consent. That can be found in a #721)

1. At the time a session is created
1. At the time an API is invoked which developer intent to use information that requires user consent (e.g. requestReferenceSpace())
1. Some combination of the above depending on the thing consent is being requested for

From what I’ve been able to gather, the following questions need to be answered before we’ll be able to come to a conclusive design:

* Do all User Agents need to align on a single option from the list above? This is really a question of principle rather than how developers would use such an API in practice. In other words, if it would be possible for a developer to write a single predictable code path that would allow User Agents the choice of when to prompt, is that acceptable? Desirable? Concerning?
* Does requesting consent in the middle of an immersive session require the use of Trusted Immersive UI (see #718)? IIRC, John mentioned this might be necessary to ensure end users don’t get confused or tricked by a website spoofing Trusted Immersive UI if they’ve gotten used to consent prompts not being Trusted. (Note, resolving this bullet point may cause a dependency on #719 )

Are there other considerations I’ve missed?

NellWaliczek:
Also related to #423

ddorwin:
Related to the first bullet, I think it's important that developers understand what could happen and are not surprised. For example, if most implementations do one thing and another makes a different decision, developers shouldn't be surprised by a poor user experience, multiple or repeated prompts, etc. In other words, the specification should make it clear to developers - and implementations via normatively specified points where user interaction might be required - when there might be consent prompts so they can avoid patterns that may cause undesirable results. It may also help with compatibility if there are fewer ways things can vary across implementations.

Some additional concerns/considerations:
[...]

PR #733, opened 2019-06-25
Initial explainer text for privacy/security aspects of WebXR

NellWaliczek:
Here we go, folks! This is the first of a series of PRs I'll be posting to address the (omg omg they just keep being created O_O ) large number of issues we've been filing over the past week regarding privacy, security, and user consent.

First off, don't panic!

In this PR I'm just adding an explainer document that will eventually define and address the privacy and security challenges we are attempting to solve.  For right now, this document is more outline than filled in content, and is filled with a lot of **TODOs** that reference the GitHub issues covering each individual topic.  The intention with this PR is to get the bones of the document in place so it will be easier to examine each subsequent change as a diff on top of a solid base.

So for now, the document has a partially filled-in section for the available "protections" (aka mitigations) in WebXR.  It also has a completely empty section for the problems to be addressed and how the protections will be applied to each.

For context, my first step is to get this PR landed, followed by the one I'm about to post which addresses #720 and #721.  After that, I'm going to try to get a PR posted once a day or so to close out on the remaining June issues. *fingers crossed*

Big shout out to @johnpallett for all the groudwork laid in the privacy-and-security repo.  He may notice that much of what's in this document is cribbed and slightly modified from his work over there.  John, if I've mangled something in the process of refactoring it, please let me know!


TrevorFSmith:
It's great that we've reached the point where we can start framing this out. Thanks to everyone who contributed!

nbutko:
Sorry if this is a question with an obvious or well known. I was a little confused about the distinctions around multiple documents and their interaction (responsible document, active document) and the provided links didn't give much in the way of concrete examples for helping understand the issues at stake. Is there a brief example of what would be disallowed under the proposed policies?

privacy-security-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Limiting
+Limiting is when data is reported only when it is within a specific range. For example, it is possible to comfortably limit reporting positional pose data when a user has moved beyond a specific distance away from an approved location. Care should be taken to ensure that the user experience is not negatively affected when employing this mitigation. It is often desireable to avoid a 'hard stop' at the at the end of a range as this may cause disruptive user experiences.
+
+# Protected functionality
+**TODO** Fill this in based on the information in [#638](https://github.com/immersive-web/webxr/pull/638) so that it adequately addresses [#393](https://github.com/immersive-web/webxr/issues/393) and [#485](https://github.com/immersive-web/webxr/issues/485).
+
+# Appendix A: Proposed IDL
+```webidl
+
```
toji: Feels like a `// TODO:` here would be consistent with the rest of the doc.

privacy-security-explainer.md
```diff
@@ -0,0 +1,68 @@
+# Privacy and security
+
+The WebXR Device API enables developers to build content for AR and VR hardware that uses one or more sensors to infer information about the real world, and may then present information about the real world either to developers or directly to the end user. In such systems there are a wide range of input sensor types used (cameras, accelerometers, etc), and a range of real-world data generated. This data is what allows web developers to author WebXR-based experiences. It also enables developers to infer information about people using their sites such as profiling, fingerprinting, and input sniffing. Due to the nature of the Web, WebXR has a higher responsibility to protect users from malicious data usage than XR experiences delivered through closed ecosystem app stores.
```
toji: > It also enables developers to infer information about people using their sites such as profiling, fingerprinting, and input sniffing.

Silly nit: I got caught on this sentence and had to read it several times over before it clicked. I initially read it as "developers, using their sites such as profiling [...], infer information about people."

What about simply:

> It also enables developers to infer information about **users** such as ...

privacy-security-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Trusted UI
+**TODO** Fill this in with what is agreed upon in [#718](https://github.com/immersive-web/webxr/issues/718) and [#719](https://github.com/immersive-web/webxr/issues/719).
+
+## User intention
+It is often necessary to be sure of user intent before exposing sensitive information or allowing actions with a significant effect on the user's experience. This intent may be communicated or observed in a number of ways.
+
+### User activation
+[User activation](https://html.spec.whatwg.org/multipage/interaction.html#activation) is defined within the HTML spec as an action the user can take which can result in certain types of HTML elements becoming activated. For example, a button becomes activated when a user clicks it with a mouse. The concept of user activation, differentiates user-caused events from injected events to prevent pages from spoofing user actions. User activation can be checked by the UA prior to session creation as well as during a session in response to appropriate `XRInputSource` events.
```
toji: Nit: Unnecessary comma after "The concept of user activation"

toji: > User activation can be checked by the UA prior to session creation as well as during a session in response to appropriate `XRInputSource` events.

More accurate to say something like this?

> User activation is required by the UA prior to `immersive` session creation, and is produced during a session in response to appropriate `XRInputSource` events.

privacy-security-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+**TODO** Fill this in with what is agreed upon in [#722](https://github.com/immersive-web/webxr/issues/722) and [#725](https://github.com/immersive-web/webxr/issues/725).
+
+## Data adjustments
+In some cases, security and privacy threats can be mitigated through throttling, quantizing, rounding, limiting, or otherwise adjusting the data reported from the WebXR APIs. This may sometimes be necessary to avoid fingerprinting, even in situations when user intent has been established.  However, data adjustment mitigations can only be used in situations which would not result in user discomfort.
+
+### Throttling
+Throttling is when sensitive data is reported at a lower frequency than otherwise possible. This mitigation has the potential to reduce a site's ability to infer user intent, location, or perform user profiling. However, when not used appropriately throttling runs a significant risk of causing user discomfort. In addition, under many circumstances it may be inadequate to provide a complete mitigation.  For example, 2D touch input data snooping has been proven possible at frequencies as [low as 20Hz](https://arxiv.org/pdf/1602.04115.pdf) via accelerometer data.
```
toji: This section feels like it could use an brief example illustrating how throttling could cause user discomfort, and maybe a counter example of a case where it's not a concern?

privacy-security-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Rounding, quantization, and fuzzing
+Rounding, quantization, and fuzzing are three categories of mitigations that modify the raw data that would otherwise be returned to the developer. Rounding decreases the precision of data by reducing the number of digits used to express it. Quantization constrains continuous data to instead report a discrete subset of values. Fuzzing is the introduction of slight, random errors into the the data. Collectively, these mitigations are useful in WebXR to avoid fingerprinting, and are especially useful when doing so does not cause noticeable impact on user comfort.
```
toji: An example of where the API could apply at least one of these principles would be welcome here too.

privacy-security-explainer.md
Manishearth: probably should put these things in a todo so we can get this landed asap

privacy-security-explainer.md
NellWaliczek: I was actually trying to avoid being specific about how these protections will be applied to WebXR concretely since that's what I'll cover in the Protected Functionality section.  Perhaps this instead:
> Within a WebXR session, `XRInputSource.select`is also considered to be triggered by user activation.  For more information, see [Input events](input-explainer.md#input-events).

NellWaliczek:
@toji, I'm somewhat inclined to leave off the examples you've requested since the intention is to actually specifically state how these protections are applied later in the document. If you're concerned that it will slip through the cracks, how about if I file an issue to track it?

@nbutko, that section in particular was taken almost directly from @johnpallet's #638. John, do you have a suggestion for language that might address Nick's comment?

Separately, @johnpallet, does this PR look ok to you as a starting point?  In other words, nothing blatantly incorrect?  If so, I'm inclined to merge asap.

NellWaliczek:
(gah, re-tagging @johnpallett because I keep mistyping his username)

johnpallett:
> @nbutko, that section in particular was taken almost directly from @johnpallett's #638. John, do you have a suggestion for language that might address Nick's comment?

Generally the sections of #638 that @nbutko mentions were designed to match requirements of the [Generic Sensors API](https://www.w3.org/TR/generic-sensor/#concepts-can-expose-sensor-readings) so the language is very similar.

However, the Generic Sensor API does include more detail about some of these concepts in the [section on Mitigations](https://www.w3.org/TR/generic-sensor/#mitigation-strategies). @nbutko do the explanations in that section address your questions (e.g. explaining Secure Context, Focused Area, etc.?) If so perhaps the same details would make sense here?

privacy-security-explainer.md
```diff
@@ -0,0 +1,68 @@
+# Privacy and security
+
+The WebXR Device API enables developers to build content for AR and VR hardware that uses one or more sensors to infer information about the real world, and may then present information about the real world either to developers or directly to the end user. In such systems there are a wide range of input sensor types used (cameras, accelerometers, etc), and a range of real-world data generated. This data is what allows web developers to author WebXR-based experiences. It also enables developers to infer information about users such as profiling them, fingerprinting their device, and input sniffing. Due to the nature of the Web, WebXR has a higher responsibility to protect users from malicious data usage than XR experiences delivered through closed ecosystem app stores.
```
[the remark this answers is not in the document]
NellWaliczek: Totally agree that we're in this weird middle ground right now since we haven't officially split AR stuff into a separate module.  Given that `immersive-ar` is still in the spec, I'm hesitant to remove references to it (as you'll see in a subsequent PR).  But *yes* I absolutely agree that as it stands, we don't yet have a cohesive story for AR-related threat vectors.
For now, are you ok with my leaving the AR references under the knowledge that there will be a full "culling" of all references when we make the official module split?

privacy-security-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Implied consent
+A User Agent may use implied consent based, for example, on the install status of a web application or frequency and recency of visits. Given the sensitivity of XR data, caution is strongly advised when relying on implicit signals. 
+
+### Explicit consent
+It is often useful to get explicit consent from the user before exposing sensitive information. When gathering explicit user consent, User Agents must present an explanation of what is being requested and provide users the option to decline. Requests for user consent can be presented in many visual forms based on the features being protected and User Agent choice. While often associated with the [Permissions API](https://www.w3.org/TR/permissions/), the concept of user consent does not have exact overlap.
```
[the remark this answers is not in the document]
NellWaliczek: In so much as an explainer can ever be normative, yes it was ;) But you're totally right that it isn't technically necessary for the statement to make sense, so I'll remove it.

privacy-security-explainer.md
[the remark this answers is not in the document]
johnpallett: I'm OK with that, it's not a blocking concern for now and can be addressed later.

NellWaliczek:
I've made all the straightforward changes, but still have a few open questions that I'm waiting for responses from @nbutko, @johnpallett, and @toji.  Since I've now got 3 PRs (2 not yet posted) backed up behind landing this change, I'm aiming to merge this within the next few hours.  If at that point I haven't heard back from y'all or there are still concerns, I'm inclined to track them in new issues.

privacy-security-explainer.md
toji: That phrasing works for me!

privacy-security-explainer.md
toji: I'm fine with Manish's sentiment of landing ASAP.

NellWaliczek:
> However, it wasn't clear to me whether these topics are intended to directly translate into their own normative requirements (and perhaps we don't know yet?)

Oh, I meant to respond to this question from @johnpallett!  The intention was to use that section to inform both normative and non-normative text in the spec (explainers themselves can't really be normative).  More specifically, my plan is to link back to each of these "protections" from the sections in which I describe the parts of the API surface area which require them.  I'm hoping to get that PR posted today and, as you look it over, please do let me know if you think we need to be taking a more normative-like stance in any of the "protections" sections.

nbutko:
@johnpallett Thanks for the pointers. Based on these, I'm understanding that the multi-document case is mostly (exclusively?) related to iframes, and the proposal is to not allow iframes access to sensor data, similar to how this currently works for deviceorienation. That seems reasonable, especially given that there should be no focused iframes while in an immersive session, just a fullscreen framebuffer. Just to round out my understanding, do you know of other types of scenarios in which this might come up?

johnpallett:
@nbutko I believe the Generic Sensors spec allows iframes to access sensor data provided that feature policies allow them to [[eg](https://developers.google.com/web/updates/2017/09/sensors-for-the-web)] and provided that they meet the other requirements (focused, visible, secure origin, etc.) Actual implementations may vary, of course.

Does this address the scenario you were wondering about? If so, do you think this needs additional explainer text to clarify?

PR #734, opened 2019-06-25
Addresses #720 and #721 in the new explainer

### LINKED ISSUE #720: Timing of user consent requests

### LINKED ISSUE #721: Duration of user consent

NellWaliczek:
As mentioned in #733, here's the PR that fixes #720 and fixes #721.  This PR doesn't include spec text just yet. (Yes, that means I'm absolutely breaking my rule about no explainer changes without spec text changes.)  I'm currently working on a follow-up PR that will lay in spec text but it is intertwined with the solution to another issue that is hopefully close to resolution as well.  In the meanwhile, this PR aims ensure we're on the same page regarding desired behavior of explicit consent requests and consent duration.


joshmarinacci:
Looks good to me. Thanks Nell!

NellWaliczek:
Rebased on the latest #733 and added missing information about origns in the section on duration of consent.  @toji  and @johnpallett can you also take a look?

NellWaliczek:
Rebased on master and ready to go as soon as I've gotten sign off

privacy-security-explainer.md
```diff
@@ -36,13 +36,10 @@ It is often necessary to be sure of user intent before exposing sensitive inform
...
-
-### Timing of XR-related consent prompts
-**TODO** Fill this in with what was agreed up on in [#720](https://github.com/immersive-web/webxr/issues/720).
+It is often useful to get explicit consent from the user before exposing sensitive information. When gathering explicit user consent, User Agents present an explanation of what is being requested and provide users the option to decline. Requests for user consent can be presented in many visual forms based on the features being protected and User Agent choice. While often associated with the [Permissions API](https://www.w3.org/TR/permissions/), the concept of user consent does not have exact overlap. If sensitive data is protected by explicit consent and will be used during an `XRSession`, it is strongly recommended that User Agents display the associated consent prompt prior to the session being created.
```
toji: I assume the plan to revisit this once we start working on a 'deferred' feature category, and this text is only intended to apply to the 'required' and 'optional' categories that we've discussed?

privacy-security-explainer.md
```diff
@@ -36,13 +36,10 @@ It is often necessary to be sure of user intent before exposing sensitive inform
...
 
 ### Duration of consent
-**TODO** Fill this in with what was agreed up on in [#721](https://github.com/immersive-web/webxr/issues/721).
+It is strongly recommended that once consent is granted, either implicitly or explicitly, for a specific [origin](https://html.spec.whatwg.org/multipage/origin.html) that this consent persist until the [browsing context](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context) has ended. User agents may also choose to allow consent to persist for all subsequent browsing contexts of that origin. Caution is advised when doing do, especially when done in response to implied consent.
```
toji: My impression previously is that we wanted access to features consent was granted for to be scoped to session lifetime. (As you stated here, the consent itself could persist between sessions in the same browsing context.) Where would be the appropriate place to make that distinction? In this section, or is there another stub that's earmarked for it?

privacy-security-explainer.md
NellWaliczek: Yes.  Filed #740.

privacy-security-explainer.md
NellWaliczek: I'm not sure what you're referring to?  This text was intended to reflect my understanding of the consensus reached in https://github.com/immersive-web/webxr/issues/721#issuecomment-504614488 and https://github.com/immersive-web/webxr/issues/721#issuecomment-505224211.  @johnpallett and @avadacatavra, did I misunderstand?

toji: You know what, it's not actually relevant at this point. Previous conversations were around things like "granting a session access to the camera shouldn't grant that access for the entire page" but that's not something that we're considering at this stage. The features that we ARE discussing (primarily reference space access) are naturally session-bound anyway, so it's a moot point. This is fine.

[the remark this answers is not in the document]
NellWaliczek: Thanks, John!  That's a great suggestion.  And thanks for the exact text.  @avadacatavra, does that work for you?  I'd like to merge this one today if possible.

NellWaliczek:
Ok, I haven't heard back from @avadacatavra yet, but any changes she may have will probably be small in scope, so I'm going to go ahead and merge this one.

PR #735, opened 2019-06-25
Made the xr-standard gamepad mapping more rigid

toji:
Addresses concerns brought up in #695 

This will make cross-platform implementations more compatible by
removing some level of platform option from the mapping process. It will
also make it more practical to expose the discussed "profiles" concept
by making it easier to have inter-compatible sets of profiles.

Also took the opportunity to normalized on the verbiage "thumbstick"
instead of joystick, since that appears to be the official verbiage used
by all the device vendors.

### DIFF
--- a/images/xr-standard-mapping.svg
+++ b/images/xr-standard-mapping.svg
@@ -15,7 +15,7 @@
    version="1.1"
    id="svg8"
    inkscape:version="0.92.4 (5da689c313, 2019-01-14)"
-   sodipodi:docname="controller.svg">
+   sodipodi:docname="xr-standard-mapping.svg">
   <defs
      id="defs2" />
   <sodipodi:namedview
@@ -25,9 +25,9 @@
      borderopacity="1.0"
      inkscape:pageopacity="0.0"
      inkscape:pageshadow="2"
-     inkscape:zoom="2.8"
-     inkscape:cx="381.21983"
-     inkscape:cy="311.364"
+     inkscape:zoom="1.979899"
+     inkscape:cx="399.47711"
+     inkscape:cy="703.73681"
      inkscape:document-units="mm"
      inkscape:current-layer="layer1"
      showgrid="false"
@@ -397,7 +397,7 @@
          x="74.415047"
          y="113.79053"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;stroke-width:0.26458332"
-         id="tspan5165-0-8">(Touchpad Click)</tspan></text>
+         id="tspan5165-0-8">(Touchpad Press)</tspan></text>
     <path
        style="fill:none;stroke:#ff0909;stroke-width:0.46500009;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:0.93000006, 0.46500004;stroke-dashoffset:0;stroke-opacity:1"
        d="m 33.889438,98.965592 4.881146,7.959928 h 20.312485"
@@ -477,44 +477,44 @@
     <text
        xml:space="preserve"
        style="font-style:normal;font-weight:normal;font-size:10.58333302px;line-height:0.5;font-family:sans-serif;text-align:center;letter-spacing:0px;word-spacing:0px;text-anchor:middle;fill:#000000;fill-opacity:1;stroke:none;stroke-width:0.26458332"
-       x="64.945122"
+       x="66.003456"
        y="199.14392"
        id="text5155-1-6-8"><tspan
          sodipodi:role="line"
-         x="64.945122"
+         x="66.003456"
          y="199.14392"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#2109ff;fill-opacity:1;stroke:none;stroke-width:0.26458332;stroke-opacity:1"
-         id="tspan5161-9-84-1">axes[1]</tspan><tspan
+         id="tspan5161-9-84-1">axes[3]</tspan><tspan
          sodipodi:role="line"
-         x="64.945122"
+         x="66.003456"
          y="204.67264"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;stroke-width:0.26458332"
-         id="tspan5165-0-2-8">(Joystick Y)</tspan></text>
+         id="tspan5165-0-2-8">(Thumbstick Y)</tspan></text>
     <path
        style="fill:none;stroke:#2809ff;stroke-width:0.46500012;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:0.93000007, 0.46500005;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 29.157323,215.3292 13.907715,-17.25414 12.532396,0.13364"
+       d="m 29.157323,215.3292 13.907715,-17.25414 13.467839,0.13364"
        id="path5167-2-8-54"
        inkscape:connector-curvature="0"
        sodipodi:nodetypes="ccc" />
     <text
        xml:space="preserve"
        style="font-style:normal;font-weight:normal;font-size:10.58333302px;line-height:0.5;font-family:sans-serif;text-align:center;letter-spacing:0px;word-spacing:0px;text-anchor:middle;fill:#000000;fill-opacity:1;stroke:none;stroke-width:0.26458332"
-       x="57.86248"
+       x="58.920815"
        y="187.55022"
        id="text5155-1-6-4-6"><tspan
          sodipodi:role="line"
-         x="57.86248"
+         x="58.920815"
          y="187.55022"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#2109ff;fill-opacity:1;stroke:none;stroke-width:0.26458332;stroke-opacity:1"
-         id="tspan5161-9-84-2-7">axes[0]</tspan><tspan
+         id="tspan5161-9-84-2-7">axes[2]</tspan><tspan
          sodipodi:role="line"
-         x="57.86248"
+         x="58.920815"
          y="193.07893"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;stroke-width:0.26458332"
-         id="tspan5165-0-2-4-0">(Joystick X)</tspan></text>
+         id="tspan5165-0-2-4-0">(Thumbstick X)</tspan></text>
     <path
        style="fill:none;stroke:#2109ff;stroke-width:0.46500012;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:0.93000008, 0.46500005;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 29.157323,215.3292 6.023264,-29.38238 12.828939,0.13363"
+       d="m 29.157323,215.3292 6.023264,-29.38238 14.298921,0.13363"
        id="path5167-2-8-5-3"
        inkscape:connector-curvature="0"
        sodipodi:nodetypes="ccc" />
@@ -528,12 +528,12 @@
          x="76.370033"
          y="210.47946"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#ff0909;fill-opacity:1;stroke:none;stroke-width:0.26458332;stroke-opacity:1"
-         id="tspan5161-9-8-3">buttons[1]</tspan><tspan
+         id="tspan5161-9-8-3">buttons[3]</tspan><tspan
          sodipodi:role="line"
          x="76.370033"
          y="216.00818"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;stroke-width:0.26458332"
-         id="tspan5361">(Joystick Click)</tspan></text>
+         id="tspan5361">(Thumbstick Press)</tspan></text>
     <path
        style="fill:none;stroke:#ff0909;stroke-width:0.46500012;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:0.93000007, 0.46500005;stroke-dashoffset:0;stroke-opacity:1"
        d="m 29.157323,215.3292 14.24094,-5.91862 h 20.312485"
@@ -688,17 +688,19 @@
          x="103.03984"
          y="283.68451"
          style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#ff0909;fill-opacity:1;stroke-width:0.26458332"
-         id="tspan5165-0-0-0"><tspan
+         id="tspan1063"><tspan
    style="fill:#000000;fill-opacity:1"
-   id="tspan5533">NOTE: </tspan>buttons[3]<tspan
+   id="tspan5533">NOTE: For this device <tspan
+   style="fill:#0000ff;fill-opacity:1"
+   id="tspan1066">axes[0]</tspan>, <tspan
+   style="fill:#0000ff;fill-opacity:1"
+   id="tspan1078">axes[1]</tspan>, and </tspan>buttons[1]<tspan
    style="fill:#000000;fill-opacity:1"
-   id="tspan5539"> is left empty, since it is reserved</tspan></tspan><tspan
+   id="tspan5539"> are</tspan></tspan><tspan
          sodipodi:role="line"
          x="103.03984"
          y="289.21323"
-         style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#ff0909;fill-opacity:1;stroke-width:0.26458332"
-         id="tspan5531"><tspan
-           style="fill:#000000;fill-opacity:1"
-           id="tspan5541">for a secondary joystick or touchpad click</tspan></tspan></text>
+         style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#000000;fill-opacity:1;stroke-width:0.26458332"
+         id="tspan5531">placeholders, since they are reserved for Touchpad inputs</tspan></text>
   </g>
 </svg>
--- a/images/xr-standard-mapping.svg
+++ b/images/xr-standard-mapping.svg
@@ -458,7 +458,7 @@
          id="tspan5268"
          x="12.348122"
          y="60.256302"
-         style="font-size:7.05555534px;stroke-width:0.26458332">Simple '<tspan
+         style="font-size:7.05555534px;stroke-width:0.26458332">Example basic '<tspan
    style="font-style:italic;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:7.05555534px;font-family:Consolas;-inkscape-font-specification:'Consolas Italic'"
    id="tspan5272">xr-standard'</tspan> controller</tspan></text>
     <text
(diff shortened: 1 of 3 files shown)


images/xr-standard-mapping.svg
```diff
@@ -688,17 +688,19 @@
...
          x="103.03984"
          y="289.21323"
-         style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#ff0909;fill-opacity:1;stroke-width:0.26458332"
-         id="tspan5531"><tspan
-           style="fill:#000000;fill-opacity:1"
-           id="tspan5541">for a secondary joystick or touchpad click</tspan></tspan></text>
+         style="font-size:4.23333311px;line-height:0.5;text-align:center;text-anchor:middle;fill:#000000;fill-opacity:1;stroke-width:0.26458332"
+         id="tspan5531">placeholders, since they are reserved for Touchpad inputs</tspan></text>
   </g>
 </svg>
```
NellWaliczek: Not sure where to leave this comment, but can we make sure to explicitly state that the second image is an *example* advanced controller?

index.bs
```diff
@@ -1488,13 +1487,13 @@ The WebXR Device API extends the {{GamepadMappingType}} to describe the mapping
...
+In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} the device MUST report an {{XRInputSource/targetRayMode}} of {{XRTargetRayMode/tracked-pointer}} and MUST have a non-<code>null</code> {{XRInputSource/gripSpace}}. It also MUST have at least one touchpad or thumbstick and MUST have at least one primary button, often a trigger, separate from the touchpad or thumbstick. If a device does not meet the requirements for the {{GamepadMappingType/xr-standard}} mapping it may still be exposed on an {{XRInputSource}} with the <code>""</code> (empty string) mapping. The {{GamepadMappingType/xr-standard}} mapping MUST only be used by {{Gamepad}} instances reported by an {{XRInputSource}}.
```
NellWaliczek: >mapping it may still be exposed on an {{XRInputSource}} with the <code>""</code>

maybe....

>mapping it may still be exposed through {{XRInputSource/gamepad}} with a {{Gamepad/mapping}} of <code>""</code>

index.bs
```diff
@@ -1526,39 +1525,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
 
-Additional buttons or axes may be exposed after these reserved indices, and SHOULD appear in order of decreasing importance. Related axes (such both axes of a joystick) SHOULD be grouped and, if applicable, SHOULD appear in X, Y, Z order. If a device has both a touchpad and a joystick the UA MAY designate whichever it chooses to be the primary axis-based input. Buttons reserved by the UA or platform MUST NOT be exposed.
+Additional buttons or axes may be exposed after these reserved indices, and SHOULD appear in order of decreasing importance. Related axes (such both axes of a thumbstick) SHOULD be grouped and, if applicable, SHOULD appear in X, Y, Z order. Buttons reserved by the UA or platform MUST NOT be exposed.
```
NellWaliczek: nit: (such both axes of a thumbstick) -> (such as both axes of a thumbstick)

NellWaliczek: Can this text be MUST?

toji: I assume you're referring to the SHOULD, and the only reason I left it as such instead of MUST is because it may not always be apparent to the UA what that order would be.

toji:
Feedback addressed aside from the "MUST" recommendation, which I left a comment for.

images/xr-standard-mapping.svg
NellWaliczek: It's the text on the image I was hoping to change....

index.bs
NellWaliczek: Oh, sorry I meant in this part:
> Related axes (such both axes of a thumbstick) SHOULD be grouped and, if applicable, SHOULD appear in X, Y, Z order.

toji: Got it! Done.

images/xr-standard-mapping.svg
toji: Done.

toji:
@thetuvix: I'd appreciate your approval prior to merging as well.

index.bs
```diff
@@ -1526,39 +1525,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
       <th>Required</th>
     </tr>
   </thead>
   <tbody>
     <tr>
       <td>axes[0]</td>
-      <td>Primary Touchpad/Joystick X</td>
-      <td>Yes</td>
+      <td>Touchpad X</td>
+      <td>If thumbstick is not present</td>
```
thetuvix: Side question, since primary axes were already required before: Are we requiring that all `xr-standard` gamepads have either a touchpad or thumbstick?  What about simpler "remote-style" controllers?

index.bs
```diff
@@ -1526,39 +1525,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
 
-Additional buttons or axes may be exposed after these reserved indices, and SHOULD appear in order of decreasing importance. Related axes (such both axes of a joystick) SHOULD be grouped and, if applicable, SHOULD appear in X, Y, Z order. If a device has both a touchpad and a joystick the UA MAY designate whichever it chooses to be the primary axis-based input. Buttons reserved by the UA or platform MUST NOT be exposed.
+Additional buttons or axes may be exposed after these reserved indices, and SHOULD appear in order of decreasing importance. Related axes (such as both axes of a thumbstick) MUST be grouped and, if applicable, MUST appear in X, Y, Z order. Buttons reserved by the UA or platform MUST NOT be exposed.
```
thetuvix: In order to introduce further reserved indices in the future, we'd need some way to reconcile the additional indices. (I suppose this is true before this change as well)

Perhaps apps need to be explicit that they want an `xr-standard` mapping for their `Gamepad` - that way, we can later introduce an `xr-standard-2` mapping or such that pushes out into button/axis 5, 6, etc.

Should the `gamepad` attribute perhaps be `getGamepad(mapping)`?

input-explainer.md
```diff
@@ -330,29 +330,29 @@ The UA may update the `gamepad` state at any point, but it must remain constant
...
-| Button     | `xr-standard` Location            |
-| ---------- | --------------------------------- |
-| buttons[0] | Primary trigger                   |
-| buttons[1] | Primary Touchpad/Joystick click   |
-| buttons[2] | Grip/Secondary trigger            |
-| buttons[3] | Secondary Touchpad/Joystick click |
+| Button     | `xr-standard` Binding  |
+| ---------- | -----------------------|
+| buttons[0] | Primary trigger        |
+| buttons[1] | Touchpad press         |
```
thetuvix: Nit: Should we group the primary/secondary triggers together as 0/1 and touchpad/thumbstick together as 2/3 in the button order?  I believe the reason they were split up like this before was to keep the primary inputs before the secondary inputs in the list.  Does this ordering still imply that touchpad is more "primary" than thumbstick?

index.bs
toji: My thoughts when initially laying out the mapping were that we wanted to ensure that if something was advertised as `xr-standard` developers could depend on it having a certain baseline level of functionality. I still think that's a reasonable goal, though it's maybe less important if paired with the profiles concept. And yes, that would mean that some input devices (Like the Daydream controller) wouldn't qualify for `xr-standard`.

If we wanted to loosen up on that restriction and simply say that `xr-standard` only means that the inputs are mapped to a predictable place, and does not imply which of them are present then we could do so. I think we might want to look at expanding the proposed generic profile set to be a bit more descriptive in that case, with variants like:

 - `button-controller` 
 - `touchpad-controller` // This is basically just the daydream controller and.... anything else?
 - `touchpad-button-controller`
 - `thumbstick-controller` // Never seen a thumbstick controller without a separate trigger
 - `thumbstick-button-controller`
 - `touchpad-thumbstick-controller` // Is this actually a thing anywhere?
 - `touchpad-thumbstick-button-controller`

index.bs
toji: I'm pretty strongly against making that large of a change to this chunk of the API this late. (Though I don't object to introducing a `xr-standard-2` at some point if needed.)

I do think we'll likely outgrow the current shape of the gamepad API at some point, but when that time comes I have a hunch that we'll be better served by introducing something along the lines of an action mapping API rather than continue to hack more robust concepts onto an API that obviously never expected to deal with them. (And hopefully at that point we have a few years of insight into how developers are using OpenXR to help inform our direction.)

input-explainer.md
toji: I wondered about that too. I'm not opposed to it. Would probably help it "read" a bit better. I think I'll make that change.

As for touchpads being "more primary" than thumbsticks, yeah, it kinda does? I don't think it's ultimately going to matter much unless you've got developers who are just blindly responding to axes[0] and [1] and nothing else, which is a recipe for failure on their part right off the bat. I did do a quick inventory of the input devices I could think of off the top of my head and by a fair margin (9-5) touchpads are more common than thumbsticks, so I felt that ordering was justified.

toji:
Okay, pushed an update that swaps the grip and touchpad button indices. (And they did a quick panicked recovery of a prior commit that I accidentally stomped on when pushing, which is why they're out-of-order now. 🙄)

index.bs
```diff
@@ -1592,39 +1591,44 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
 
-NOTE: This diagram demonstrates how two potential controllers would be exposed with the {{GamepadMappingType/xr-standard}} mapping. <img src="images/xr-standard-mapping.svg" alt="Simple 'xr-standard' controller and Advanced 'xr-standard' controller">
+<section class="note">
+<<<<<<< HEAD
+This diagram demonstrates how two potential controllers would be exposed with the {{GamepadMappingType/xr-standard}} mapping. <img src="images/xr-standard-mapping.svg" alt="Simple 'xr-standard' controller and Advanced 'xr-standard' controller">
```
NellWaliczek: Merge fail?

toji: Derp. Fixed!

toji:
@thetuvix: I've got approval from Nell, but I'm going to hold off from merging until I hear that you're OK with the latest changes as well.

input-explainer.md
thetuvix: Yea - I think it makes sense to still keep touchpad (`[2]`) before thumbstick (`[3]`).  Just moving generally away from the notion that this is a sorted list of buttons by "primariness" seems worthwhile.

index.bs
thetuvix: Sounds good.

Even if we later do need to support what I said in the gamepad API, there are slightly less clean ways to do it later without changes now - for example, when requesting your `XRSession`, you could centrally set `xr-standard-2` as your preferring mapping for receiving `xr-standard*` gamepad buttons/axes.

index.bs
thetuvix: > If we wanted to loosen up on that restriction and simply say that xr-standard only means that the inputs are mapped to a predictable place, and does not imply which of them are present then we could do so.

This aligns with what I've been thinking here.  Before, we only had one concept, `xr-standard`, overloaded to serve two purposes:
* `xr-standard` defined which controller parts **map** to which `buttons`/`axes` IDs
* `xr-standard` promised a minimum **profile** of controller parts that developers could assume were present

We've now split these into two distinct `mapping` and `profile` concepts in the API.  It seems reasonable to embrace that, simplifying the meaning of the `xr-standard` mapping to a pure ID mapping, and allowing profiles to make more specific promises.

input-explainer.md
```diff
@@ -336,29 +336,29 @@ The UA may update the `gamepad` state at any point, but it must remain constant
...
 
-| Button     | `xr-standard` Location            |
-| ---------- | --------------------------------- |
-| buttons[0] | Primary trigger                   |
-| buttons[1] | Primary Touchpad/Joystick click   |
-| buttons[2] | Grip/Secondary trigger            |
-| buttons[3] | Secondary Touchpad/Joystick click |
+| Button     | `xr-standard` Binding  |
+| ---------- | -----------------------|
+| buttons[0] | Primary trigger        |
```
thetuvix: Above, `buttons[0]` is defined to possibly be a "primary button" instead.  We should at least mention "Primary trigger/button" here.

Similar to the touchpad/thumbstick ambiguity, there is ambiguity across profiles that would be possible here if two controllers both had a button and a trigger, but differed on whether they considered the trigger or the button to be "primary".  However, since buttons and triggers are far more interchangeable than touchpads and thumbsticks, I'm less worried there.  My gut is that, as we work out the actual generic profiles you sketched out for today's controllers, this ambiguity won't prove to be a problem.

index.bs
thetuvix: Should we continue this thread as part of #738?

index.bs
```diff
@@ -1567,13 +1566,13 @@ The WebXR Device API extends the {{GamepadMappingType}} to describe the mapping
...
+In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} the device MUST report an {{XRInputSource/targetRayMode}} of {{XRTargetRayMode/tracked-pointer}} and MUST have a non-<code>null</code> {{XRInputSource/gripSpace}}. It also MUST have at least one touchpad or thumbstick and MUST have at least one primary button, often a trigger, separate from the touchpad or thumbstick. If a device does not meet the requirements for the {{GamepadMappingType/xr-standard}} mapping it may still expose a {{XRInputSource/gamepad}} with a {{Gamepad/mapping}} of <code>""</code> (empty string). The {{GamepadMappingType/xr-standard}} mapping MUST only be used by {{Gamepad}} instances reported by an {{XRInputSource}}.
 
 <table class="tg">
   <thead>
     <tr>
       <th>Buttons</th>
-      <th><code>xr-standard</code> Location</th>
+      <th><code>xr-standard</code> Binding</th>
```
thetuvix: Nit: Should we use the term "mapping" here, to align with `Gamepad.mapping`?

index.bs
```diff
@@ -1585,18 +1584,18 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
-      <td>Yes</td>
+      <td>Secondary Button/Grip Trigger</td>
+      <td>No</td>
     </tr>
     <tr>
       <td>buttons[2]</td>
-      <td>Secondary Button/Grip Trigger</td>
-      <td>No</td>
+      <td>Touchpad Press</td>
+      <td>If thumbstick is not present</td>
```
thetuvix: This should now be "If touchpad is present".

index.bs
```diff
@@ -1585,18 +1584,18 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
     </tr>
     <tr>
       <td>buttons[3]</td>
-      <td>Secondary Touchpad/Joystick Button</td>
-      <td>No</td>
+      <td>Thumbstick Press</td>
+      <td>If touchpad is not present</td>
```
thetuvix: This should now be "If thumbstick is present".

index.bs
```diff
@@ -1605,39 +1604,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
   <thead>
     <tr>
       <th>Axes</th>
-      <th><code>xr-standard</code> Location</th>
+      <th><code>xr-standard</code> Binding</th>
```
thetuvix: Nit: Should we use the term "mapping" here, to align with `Gamepad.mapping`?

index.bs
```diff
@@ -1605,39 +1604,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
-      <td>Yes</td>
+      <td>Touchpad Y</td>
+      <td>If thumbstick is not present</td>
     </tr>
     <tr>
       <td>axes[2]</td>
-      <td>Secondary Touchpad/Joystick X</td>
-      <td>No</td>
+      <td>Thumbstick X</td>
+      <td>If touchpad is not present</td>
```
thetuvix: This should now be "If thumbstick is present".

index.bs
```diff
@@ -1605,39 +1604,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
     </tr>
     <tr>
       <td>axes[3]</td>
-      <td>Secondary Touchpad/Joystick Y</td>
-      <td>No</td>
+      <td>Thumbstick Y</td>
+      <td>If touchpad is not present</td>
```
thetuvix: This should now be "If thumbstick is present".

index.bs
```diff
@@ -1605,39 +1604,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
-      <td>Yes</td>
+      <td>Touchpad X</td>
+      <td>If thumbstick is not present</td>
     </tr>
     <tr>
       <td>axes[1]</td>
-      <td>Primary Touchpad/Joystick Y</td>
-      <td>Yes</td>
+      <td>Touchpad Y</td>
+      <td>If thumbstick is not present</td>
```
thetuvix: This should now be "If touchpad is present".

index.bs
```diff
@@ -1605,39 +1604,40 @@ In order to report a {{Gamepad/mapping}} of {{GamepadMappingType/xr-standard}} t
...
       <th>Required</th>
     </tr>
   </thead>
   <tbody>
     <tr>
       <td>axes[0]</td>
-      <td>Primary Touchpad/Joystick X</td>
-      <td>Yes</td>
+      <td>Touchpad X</td>
+      <td>If thumbstick is not present</td>
```
thetuvix: This should now be "If touchpad is present".

input-explainer.md
```diff
@@ -336,29 +336,29 @@ The UA may update the `gamepad` state at any point, but it must remain constant
...
 
-Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array. If a device has both a touchpad and a joystick the UA should designate one of them to be the primary axis-based input and expose the other at axes[2] and axes[3] with an associated button at button[3].
+Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array.
 
 In order to make use of the `xr-standard` mapping, a device must meet **at least** the following criteria:
 
  - Is a `tracked-pointer` device. 
  - Has a trigger or similarly accessed button
- - Has at least one touchpad or joystick
+ - Has at least one touchpad or thumbstick
```
thetuvix: It sounds like retracting just this criteria to qualify for `xr-standard` would allow all of the profiles you'd listed.

input-explainer.md
```diff
@@ -336,29 +336,29 @@ The UA may update the `gamepad` state at any point, but it must remain constant
...
+| axes[2] | Thumbstick X          |
+| axes[3] | Thumbstick Y          |
 
-Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array. If a device has both a touchpad and a joystick the UA should designate one of them to be the primary axis-based input and expose the other at axes[2] and axes[3] with an associated button at button[3].
+Additional device-specific inputs may be exposed after these reserved indices, but devices that lack one of the canonical inputs must still preserve their place in the array.
 
 In order to make use of the `xr-standard` mapping, a device must meet **at least** the following criteria:
 
  - Is a `tracked-pointer` device. 
  - Has a trigger or similarly accessed button
```
thetuvix: It seems fine to require at least a primary button/trigger for `buttons[0]`.

Should we more strongly state then that this button MUST also trigger `onselect` for this `XRInputSource`?

index.bs
NellWaliczek: I support this direction.  I'd kinda been feeling the same way as I looked over the actual spec text changes.

toji:
Thanks for the feedback! I agree that the `xr-standard` mapping no longer needs to stand in for profiles. I ran with Alex's suggestion of requiring a primary button (which I still specify should be separate from a touchpad/thumbstick press) but otherwise `xr-standard` in specified here to really just be an indication of which elements are surfaced where, and not which ones are actually present.

Would appreciate one more brief review from @NellWaliczek or @thetuvix before landing, but otherwise I'm going to plan on merging this by EOD.

PR #737, opened 2019-06-25
Add an explicit inline XR device

### LINKED ISSUE #635: Clarify the relationship between "XR device" and "inline"
The language about the "XR device" is a bit confusing as it relates to `"inline"` sessions, especially on clients with an external headset (e.g., desktop).

https://immersive-web.github.io/webxr/#xr-interface says:
> An XR object has an XR device... which... represents the active XR device

and https://immersive-web.github.io/webxr/#xr-device says:
>An XR device is a physical unit of hardware that can present imagery to the user. On desktop clients, this is usually a headset peripheral.

>An XR device has a list of supported modes (a list of strings) that contains "inline"...

On desktop, `"inline"` and `"immersive-vr"` sessions are on two physical units of hardware.
Is the inline capability really a supported mode of the `XR device` representing the HMD?

Note that `"inline"` can't be a provided by a separate XR device because that would prevent using `"inline"` and `"immersive-vr"`sessions at the same time, at least with the current text.

It may be that what we really want is that _user agents_ will always support `"inline"` sessions, but that doesn't necessarily mean every "XR device" needs include it in its "list of supported modes."

### LINKED ISSUE #658: Inconsistency between explainer and spec for inline sessions with no device.
Maybe related to issue #635.

In the spec, the algorithm for RequestSession claims:
If the XR device is null
Reject promise with null.

However, the explainer says:
UAs implementing the WebXR Device API must guarantee that inline sessions can be created, regardless of XR device presence, unless blocked by page feature policy.

Which is correct?  Should all implementations be required to support a "device-less" inline mode?

Manishearth:
Fixes #658, fixes #635

take 2 at https://github.com/immersive-web/webxr/pull/704

This separates the concept of immersive XR devices (which participate in selection, and can be null) from the inline XR device (a guaranteed XR device representing the device containing the screen the browser window renders to). It explicitly adds the concept of the _session_ having an XR device so that you may have different sessions with different devices. There's a bunch of disambiguation done here to clean that up.

r? @toji @NellWaliczek

cc @ddorwin

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -205,7 +205,7 @@ An [=/XR device=] has a <dfn>list of supported modes</dfn> (a [=/list=] of [=/st
 
 The user-agent MUST have an <dfn>inline XR Device</dfn>, which is an [=/XR Device=] that MUST [=list/contains|contain=] {{XRSessionMode/"inline"}} in its [=list of supported modes=]. The [=Inline XR Device=] will report as much pose information of the physical device the browser is rendering to as possible.
 
-NOTE: On phones, the [=inline XR Device=] will report gyroscopic pose information of the phone itself. On desktops and laptops without gyroscopes, the [=inline XR Device=] will not be able to report a pose (and thus {{local}} and {{local-floor}} {{XRSpace}}s will not be [=reference space is supported|supported=]). In case the browser is already running on an [=/XR device=], the [=inline XR device=] will be the same device, and may support multiple [=view|views=].
+NOTE: On phones, the [=inline XR Device=] will report gyroscopic pose information of the phone itself. On desktops and laptops without gyroscopes, the [=inline XR Device=] will not be able to report a pose. In case the browser is already running on an [=/XR device=], the [=inline XR device=] will be the same device, and may support multiple [=view|views=].
 
 Initialization {#initialization}
 ==============
@@ -461,7 +461,7 @@ When an {{XRSession}} is shut down the following steps are run:
   1. If no other features of the user agent are actively using them, perform the necessary platform-specific steps to shut down the device's tracking and rendering capabilities. This MUST include:
     - Releasing [=exclusive access=] to the [=XRSession/XR device=] if |session| is an [=immersive session=].
     - Deallocating any graphics resources acquired by |session| for presentation to the [=XRSession/XR device=].
-    - Putting the [=XRSession/XR device=] in a state such that a different source may be able to initiate a session with the same device.
+    - Putting the [=XRSession/XR device=] in a state such that a different source may be able to initiate a session with the same device if |session| is an [=immersive session=].
   1. [=Queue a task=] that fires an {{XRSessionEvent}} named {{end!!event}} on |session|.
 
 </div>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1720,9 +1737,11 @@ The <dfn method for="XRWebGLLayer">getNativeFramebufferScaleFactor(|session|)</d
...
+In order for a WebGL context to be used as a source for immersive XR imagery it must be created on a <dfn>compatible graphics adapter</dfn> for the [=XR/immersive XR device=]. What is considered a [=compatible graphics adapter=] is platform dependent, but is understood to mean that the graphics adapter can supply imagery to the [=XR/immersive XR device=] without undue latency. If a WebGL context was not already created on the [=compatible graphics adapter=], it typically must be re-created on the adapter in question before it can be used with an {{XRWebGLLayer}}.
+
+Note: On an XR platform with a single GPU, it can safely be assumed that the GPU is compatible with the [=XR/immersive XR device=]s advertised by the platform, and thus any hardware accelerated WebGL contexts are compatible as well. On PCs with both an integrated and discreet GPU the discreet GPU is often considered the [=compatible graphics adapter=] since it generally a higher performance chip. On desktop PCs with multiple graphics adapters installed, the one with the [=XR/immersive XR device=] physically connected to it is likely to be considered the [=compatible graphics adapter=].
```
NellWaliczek: Nit, but not yours: discreet -> discrete

index.bs
```diff
@@ -1942,7 +1961,7 @@ Event Types {#event-types}
 
 The user agent MUST provide the following new events. Registration for and firing of the events must follow the usual behavior of DOM4 Events.
 
-The user agent MAY fire a <dfn event for="XR">devicechange</dfn> event on the {{XR}} object to indicate that the availability of [=/XR device=]s has been changed. The event MUST be of type {{Event}}.
+The user agent MAY fire a <dfn event for="XR">devicechange</dfn> event on the {{XR}} object to indicate that the availability of [=XR/immersive XR device=]s has been changed. The event MUST be of type {{Event}}.
```
NellWaliczek: nit: not you, but should this be
>The user agent MUST fire

index.bs
```diff
@@ -444,9 +459,9 @@ When an {{XRSession}} is shut down the following steps are run:
   1. Remove |session| from the [=list of inline sessions=].
   1. [=Reject=] any outstanding promises returned by |session| with an {{InvalidStateError}}, except for any promises returned by {{XRSession/end()}}.
   1. If no other features of the user agent are actively using them, perform the necessary platform-specific steps to shut down the device's tracking and rendering capabilities. This MUST include:
-    - Releasing [=exclusive access=] to the [=/XR device=].
-    - Deallocating any graphics resources acquired by |session| for presentation to the [=/XR device=].
-    - Putting the [=/XR device=] in a state such that a different source may be able to initiate a session with the same device.
+    - Releasing [=exclusive access=] to the [=XRSession/XR device=] if |session| is an [=immersive session=].
+    - Deallocating any graphics resources acquired by |session| for presentation to the [=XRSession/XR device=].
+    - Putting the [=XRSession/XR device=] in a state such that a different source may be able to initiate a session with the same device.
```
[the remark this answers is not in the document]
Manishearth: I figure it was handwavy enough about "putting it in a state where you can request another session" since that's never not the case for inline sessions, but I made it explicit.

Manishearth:
Addressed

index.bs
```diff
@@ -201,7 +201,11 @@ XR device {#xr-device-concept}
...
+An [=/XR device=] has a <dfn>list of supported modes</dfn> (a [=/list=] of [=/strings=]) that [=list/contains=] the enumeration values of {{XRSessionMode}} that the [=/XR device=] supports.
+
+The user-agent MUST have an <dfn>inline XR Device</dfn>, which is an [=/XR Device=] that MUST [=list/contains|contain=] {{XRSessionMode/"inline"}} in its [=list of supported modes=]. The [=Inline XR Device=] will report as much pose information of the physical device the browser is rendering to as possible.
+
+NOTE: On phones, the [=inline XR Device=] will report gyroscopic pose information of the phone itself. On desktops and laptops without gyroscopes, the [=inline XR Device=] will not be able to report a pose (and thus {{local}} and {{local-floor}} {{XRSpace}}s will not be [=reference space is supported|supported=]). In case the browser is already running on an [=/XR device=], the [=inline XR device=] will be the same device, and may support multiple [=view|views=].
```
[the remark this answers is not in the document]
Manishearth: Removed the parenthetical, but kept the "can't report pose information". Unsure what it should actually be here.

index.bs
```diff
@@ -201,7 +201,11 @@ XR device {#xr-device-concept}
 
 An <dfn for="">XR device</dfn> is a physical unit of hardware that can present imagery to the user. On desktop clients, this is usually a headset peripheral. On mobile clients, it may represent the mobile device itself in conjunction with a viewer harness. It may also represent devices without stereo-presentation capabilities but with more advanced tracking.
 
-An [=/XR device=] has a <dfn>list of supported modes</dfn> (a [=/list=] of [=/strings=]) that [=list/contains=] "<code>inline</code>" and all the other enumeration values of {{XRSessionMode}} that the [=/XR device=] supports.
+An [=/XR device=] has a <dfn>list of supported modes</dfn> (a [=/list=] of [=/strings=]) that [=list/contains=] the enumeration values of {{XRSessionMode}} that the [=/XR device=] supports.
+
+The user-agent MUST have an <dfn>inline XR Device</dfn>, which is an [=/XR Device=] that MUST [=list/contains|contain=] {{XRSessionMode/"inline"}} in its [=list of supported modes=]. The [=Inline XR Device=] will report as much pose information of the physical device the browser is rendering to as possible.
```
toji: We should indicate that the inline XR device MAY be the same as the immersive XR device, if one is present, but it's not required to be.

Manishearth:
Addressed.

Issue #727, opened 2019-06-21
What types of capabilities should applications be allowed to request?


ddorwin:
In addition to specifying when (#720) and how (i.e., #723) applications can request capabilities/features, we will need to determine what are the actual capabilities/features to be exposed. Before we get to that, though, it probably makes sense to explore the categories or dimensions of capabilities that should be exposed. Although the focus has been on consent, there _may_ be other reasons that applications should request capabilities up front (see #723).

The privacy design (#638) details various threats and why consent might be required, but further work is necessary to determine how many different dimensions of consent make sense and what they should be.

As said in response to question 7 in https://github.com/immersive-web/webxr/pull/689#issuecomment-502262866, IMO, something like "spatial-tracking-unbounded" is probably too granular and would become difficult for developers to use. IMO, it probably makes more sense to define what the user is giving up, which may not map 1:1 to a single concept in the API surface.

NellWaliczek:
It's funny, I was actually thinking in a very different direction with #723, though I can see why you'd prefer to separate this discussion from that one.

Do you have a specific proposal for what "defining the user is giving up" would mean?  Are you optimizing for developer difficulty or end-user difficulty?  Because my original though on the topic is developers might find it less confusing if they just indicated which APIs they plan on calling, which is even more granular.  So i'm really interested in seeing what your alternative might look like.

NellWaliczek:
In terms of the way these consent prompts are communicated to the users, I totally agree that the api-level granularity doesn't make much sense.  However, from a developer's perspective of knowing how to ensure necessary prompts are displayed, it does.  In my proposal in #739, the communication from the developer to the UA is done granularly, but because it's all done up front the UA has the option to explain things in the way they thing will make the most sense to their users.

johnpallett:
For context, here is a [suggested list of privacy-sensitive features and associated user communications](https://github.com/immersive-web/webxr/blob/67b0c0992e7d82c383d619900ef110a36c3bfade/designdocs/privacy-design.md#user-communication).

NellWaliczek:
Closed by #739

Issue #723, opened 2019-06-19
Required, optional, deferrable, and not-needed features


NellWaliczek:
**[Disclaimer: This issue is one of several being filed to capture discussions that began either on #638, on #689, or at the most recent F2F]**

This one is sort a redux of a design exploration that @toji and I did many months back (see #433). At the time we were looking for a solution that would allow developers state at the time of session creation what features they absolutely need for it to be worth creating a session and which features would be really nice to have. The idea was to prevent users from having a session start only to have them be immediately told that the experience wouldn't work. Happily, this provided an opportunity for the User Agent to request user consent any of the features that the developer considered required.

I'm working on an updated concrete design proposal which shows how developers might communicate intent to use certain features yet, but in the process of putting it together, I have some questions for you all about what you'd like to see in response to that developer intent being communicated. For example:

1. Should developers be allowed to block session creation if particular "required" features are unsupported by the underlying platform?
2. If a "required" feature depends on user consent, should the User Agent prompt for consent at the time of session creation?
3. If a user does not consent to a "required" feature, should session creation be blocked?
4. If an "optional" feature depends on user consent, should the User Agent prompt for consent at the time of session creation?

NellWaliczek:
Also related to #423 and #424

joshmarinacci:
Is there really a difference between optional and required? From the developer perspective they ask for things and then find out what they were given, either at session start time or api use time. In either case they can render an error or show a message to the user explaining what they can and can't do now.   If there was an optional type in addition to a required permission type, would what the developer does be any different?  When would a developer want to actually use the optional permissions?

Second: from the user perspective would optional vs required make a difference?  If you were given a dialog that says "this app wants required access to the camera and optional access to the microphone, do you want to grant it", would you reasonably be able to answer that question? If you were presented with optional microphone vs required microphone would you understand the distinction? Would you care?

I think this adds extra complexity without providing additional value.

NellWaliczek:
[...]

Issue #591, opened 2019-04-16
When we add more options to session creation, do it with a dictionary.


toji:
This is not an issue so much as a directive. As discussed in #566 it's assumed that we probably will have need of additional arguments beyond the `mode` that will need to be passed into `navigator.xr.requestSession()`, even though we don't have any at this exact moment. When that time comes, the pattern to be used, per #542, is:

```webidl
dictionary XRSessionCreationOptions {	
  XROptionThingy option;
};

partial interface XR {
  Promise<XRSession> requestSession(XRSessionMode mode, optional XRSessionCreationOptions options);
};
```

Called like so:

```js
navigator.xr.requestSession('immersive-vr', { option: value });
```

NellWaliczek:
Closed by #739

PR #739, opened 2019-06-26
Enable optional/required features to be checked before session creation

NellWaliczek:
This PR contains explainer text intended to describe the fixes for #723, #591, #423, #417, and #727.  I'm holding off on the spec text until it's clear we've got consensus on this approach.

First off!  The main purpose of this proposal is to ensure that developers have a way to block session creation if a feature they absolute require is unavailable.  While this may occasionally have to do with consent, it also may simply be due to lack of hardware support.  Following the same pattern, this design allows developers to indicate up front that there are optional features they would very much like to use.  This allows any necessary explicit consent to be requested before a session begins. For now, this PR only lists reference-space related features, but we've heard feedback from folks working on new functionality that a similar system is needed.

Second!  There are several main considerations in the reasoning behind this approach:
1. **Straightforward signals** By aligning features with either the API being called or an indication of the parameters an API will be passed, it makes it very easy for developers to understand what they're requesting.
1. **WebIDL is fiddly** This one is sort of unfortunate.  We originally wanted both arrays to be filled with an enum called `XRFeatureName` (or something).  However that would have meant optional features that the UA did not recognize would have caused sessions to be rejected. Alternatively, we considered using dictionary keys that could be ignored when processing the optional array.  However, that would have meant required features the UA did not recognize would have allowed sessions to be created anyway.  We considered having these two arrays be made up of different things (as was tried and abandoned in #433), but the difference seemed to cause unnecessary cognitive load and confusion for developers.
1. **Simple to write** We explored a LOT of variations in IDL signatures and,  due to the previous bullet, many were very verbose or confusing.  There may be a bit of confusion about being able to add a reference space type in the same list as something unrelated (e.g. 'geo orientation'?), but after looking at the alternatives the tradeoff seemed worth it.
1. **Extensible**  In the future if we find that `DOMStrings` are insufficient to communicate feature requirements, we can easily and transparently switch the arrays to be of type `Any` or a union type with whatever new configuration requirements we have identified.

(And yes, there's a few more PRs ready to be posted once we have consensus established on this one)


explainer.md
```diff
@@ -373,21 +373,9 @@ The UA may choose to present the immersive AR session's content via any type of
 
 ## Inline sessions
 
-There are several scenarios where it's beneficial to render a scene whose view is controlled by device tracking within a 2D page. For example:
+When authoring content to be viewed immersively, it may be beneficial to use an `inline` session to view the same content in a 2D browser window. Using an inline session enables content to use a single rendering path for both inline and immersive presentation modes. It also makes switching between inline content and immersive presentation of that content easier.
```
toji: 👍 Thanks for re-aligning this statement with the current design priorities.

NellWaliczek:
Missed one
/fixes #424

explainer.md
```diff
@@ -405,20 +393,67 @@ function beginInlineXRSession() {
...
+ - Making use of head-tracking features for devices like [zSpace](http://zspace.com/) systems.
+
+## Advanced functionality
+
+Beyond the core APIs described above, the WebXR Device API also exposes several options for taking greater advantage of the XR hardware's capabilities.
+
+### Feature dependencies
+Once developers have mastered session creation and rendering, they will often be interested in using additional WebXR features that may not be universally available. While developers are generally encouraged to design for progressive enhancement, some experiences may have requirements on features that are not guaranteed to be universally available. For example, a guided tour would not function on an Oculus Go because it is unable to provide an [`unbounded` reference space](spatial-tracking-explainer.md#unbounded-reference-space). If an experience is completely unusable without a specific feature, it would be a poor user experience to initialize the underlying XR platform and create a session only to immediately notify the user it won't work.
```
toji: > For example, a guided tour would not function on an Oculus Go because it is unable to provide an `unbounded` reference space.

This example sounded weird to me on first read, because a certain form of guided tour (Navigating between photospheres, street view-style) is reasonably popular on 3DoF headsets like the Go. What about: 

>  **an experience which requires users to move around a large physical space, such as** a guided tour, would not function on an Oculus Go because...

explainer.md
```diff
@@ -405,20 +393,67 @@ function beginInlineXRSession() {
...
+* **Optional** The experience would like to use this feature for the entire session, but can function without it. Again, if [explicit consent](privacy-security-explainer.md#explicit-consent) is necessary, users will be prompted in response to `xr.requestSession()`. However, session creation will succeed regardless of the feature's hardware support or user intent. Developers must not assume optional features are available in the session and check the result from attempting to use them.
+
+Features recognized by the UA that are not explicitly listed in these arrays will be implicitly added to `XRSessionInit.optionalFeatures` if they do not require consent or [consent has been implied](privacy-security-explainer.md#implied-consent). For example, creating an `immersive-vr` session implies consent for `local` and `local-floor` reference spaces even if not explicitly listed in `XRSessionInit`. If the UA does not recognize an entry in `XRSession.requiredFeatures` the session will not be created.
```
toji: > Features recognized by the UA that are not explicitly listed in these arrays will be implicitly added to `XRSessionInit.optionalFeatures` if they do not require consent or consent has been implied.

I'm sure how to interpret this line. Are you stating that if user agent A that has determined it can safely offer, say, eye-tracking without user consent should implicitly append it to the optional array? (*Terrible* example, I know, but work with me.) That seems problematic, because then if user agent B decides that eye-tracking really *does* require explicit consent, experiences written only against A will fail because the feature was not requested.

I'm onboard with `local` and `local-floor` being implicit for `immersive-vr`, but I feel like they probably need to be part of an explicit list of implied features rather than something left to the user agent.

NellWaliczek: >I'm sure how to interpret this line. Are you stating that if user agent A that has determined it can safely offer, say, eye-tracking without user consent should implicitly append it to the optional array? (Terrible example, I know, but work with me.) That seems problematic, because then if user agent B decides that eye-tracking really does require explicit consent, experiences written only against A will fail because the feature was not requested.

Perhaps this can be fixed by changing the sentence to say "if they do not require a signal of user intent"?  For context, I added this sentence based on our noodling about the idea that maybe there are features that have no consent requirements at all, such as the ability to support an xr-standard gamepad.  At the time you were hesitant to include it because if gamepad based motion controllers are present, they should always appear in the inputSources array.  My case for including it on the list was that there are some experiences which will not work if the hardware cannot support a multibutton input source.  I tried to strike the balance by suggesting only automatically enabling features that are normatively specified to not be dependent on a signal of user intent.  Does that make more sense?

NellWaliczek: > I'm onboard with local and local-floor being implicit for immersive-vr, but I feel like they probably need to be part of an explicit list of implied features rather than something left to the user agent.

My next PR calls this out a lot more concretely as it digs into threats and mitigations, but I'm ok with making some sort of a list here if it helps.  I only left it out of here because the list is already located in the [Spatial tracking explainer](spatial-tracking-explainer.md#xrreferencespace-availability)

NellWaliczek: Perhaps both issues can be address with a single list of "exceptions" added to this section?

toji: > Perhaps this can be fixed by changing the sentence to say "if they do not require a signal of user intent"?

Ah, I'm see what you were trying to say. I'm OK with this mental model of "everything that doesn't require user intent is automatically optional". It does help clarify the gamepad case, and I'm sure others. However, we'd still want an explicit list of which features those are so that we don't get UA disagreements.

NellWaliczek: Ok, how about I switch the whole paragraph around to this:

> If the UA does not recognize an entry in `XRSession.requiredFeatures` the session will not be created. Under some circumstances, features recognized by the UA that are not explicitly listed in these arrays may be implicitly added to `XRSessionInit.optionalFeatures`. For example, if a feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent). Or, if the requested session mode implies consent, such as `immersive-vr` does for `local` and `local-floor` reference spaces.

toji: Better, thanks!

NellWaliczek:
Thanks, Brandon!  This is a fairly important PR, so I plan to leave this open for a few days to give folks a chance to review and leave feedback.  If nothing showstopping comes up, I'm aiming to merge by Friday.

explainer.md
```diff
@@ -405,20 +393,67 @@ function beginInlineXRSession() {
...
+* **Optional** The experience would like to use this feature for the entire session, but can function without it. Again, if [explicit consent](privacy-security-explainer.md#explicit-consent) is necessary, users will be prompted in response to `xr.requestSession()`. However, session creation will succeed regardless of the feature's hardware support or user intent. Developers must not assume optional features are available in the session and check the result from attempting to use them.
+
+If the UA does not recognize an entry in `XRSession.requiredFeatures` the session will not be created. Under some circumstances, features recognized by the UA but not explicitly listed in these arrays may be implicitly added to `XRSessionInit.optionalFeatures`. For example, if a feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent). Or, if the requested session mode implies consent, such as `immersive-vr` does for `local` and `local-floor` reference spaces.
```
johnpallett: I'd suggest adding two additional concepts here:
- Performance might affect whether a feature is added implicitly
- Which features get added implicitly might vary by user agent. Don't count on anything not `required`

Also, I think I know what the last sentence means, but the phrase 'implies consent' could be interpreted as 'implies user consent' so I'm not certain. I think the intention is that session mode could imply that certain features are available.

In case it helps, here's a suggested alternative including the above concepts and clarifying the meaning around session mode (note: this isn't intended to be blocking, I'm just including it in case it's useful):

> If the UA does not recognize an entry in `XRSession.requiredFeatures` the session will not be created.
>
> Features not listed in either array may or may not be available during the session. Under some circumstances the UA may implicitly add features to `XRSessionInit.optionalFeatures`, enabling them during the session.
>
>For example, a user agent may choose to enable a feature that does not require a signal of [user intent](privacy-security-explainer.md#user-intent) and which has no adverse performance implications. This behavior is likely to vary between user agents, and developers should not depend upon the availability of features not explicitly listed as `required`.
>
> The user agent may also enable features that are implied by session type, or implied by other requested features. For example, as any immersive session must support both `local` and `local-floor` reference spaces, a user agent may add both references spaces to `XRSessionInit.optionalFeatures` when an `immersive-vr` session is created.

On the last bit, I'm curious whether this is a firm requirement. For example, consider a scenario where a developer requests an immersive session with a `local` reference space, but not `local-floor`. Suppose the user has configured their viewer height. #638 proposes that exposing actual viewer height requires [explicit consent](privacy-security-explainer.md#explicit-consent) because it is sensitive data. In this case, should the user agent:

1. Require [explicit consent](privacy-security-explainer.md#explicit-consent) because `local-floor` is implied? 
2. Add an implicit `local-floor` reference space, but use emulated floor height to avoid [explicit consent](privacy-security-explainer.md#explicit-consent)?
3. Not add `local-floor` at all, even though it's an immersive session?
4. Any of the above, at the discretion of the user agent (and if a developer needs `local-floor` they should make it `required`?)

NellWaliczek: Good point about potential performance feedback!  And I definitely didn't make it clear enough, that it is not a choice to be left to the UA discretion about which features qualify for being automatically enabled.  Without consistency on this, we'd end up in a world of hurt.  

I also, better understand your concern now about `local-floor` technically having a different threat profile than simply `local`.  While it is a requirement that `local` and `local-floor` be *possible* in any session that supports `immersive-vr` and `immersive-ar`, it doesn't mean we necessarily want to expose user height without the UA having the opportunity to manage that risk differently.

Hopefully my latest commit has addressed both your concerns and @toji's. If so, I'd appreciate an "Approve" :)

explainer.md
```diff
@@ -405,20 +393,70 @@ function beginInlineXRSession() {
...
+* **Optional** The experience would like to use this feature for the entire session, but can function without it. Again, if [explicit consent](privacy-security-explainer.md#explicit-consent) is necessary, users will be prompted in response to `xr.requestSession()`. However, session creation will succeed regardless of the feature's hardware support or user intent. Developers must not assume optional features are available in the session and check the result from attempting to use them.
+
+If the UA does not recognize an entry in `XRSession.requiredFeatures` the session will not be created. Some features recognized by the UA but not explicitly listed in these arrays may be enabled by default for a session. This is only allowed if the feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent) nor impact performance or the behavior of other features when enabled. At this time, only the following features may be enabled by default:
```
toji: > If the UA does not recognize an entry in `XRSession.requiredFeatures` the session will not be created.

Nit: Upon re-reading with fresh eyes, this sentence feels out-of-place with respect to the rest of the paragraph. Would it fit better at the end of the **Required:**  paragraph above?

explainer.md
toji: The table definitely helps clarify things, thanks!

To make sure I understand your above comment, Nell: Sounds like the intent is that `local-floor` is no longer one of the implicit features, but (as is currently required by the spec) would still be available to any immersive session via an emulated floor height. The only way to get a non-emulated floor height would be to explicitly request `local-floor` in either the `requiredFeatures` or `optionalFeatures` lists. Is that correct?

explainer.md
```diff
@@ -580,6 +618,11 @@ enum XRSessionMode {
   "immersive-ar"
 }
 
+dictionary XRSessionInit {
```
klausw: This still needs to be added to the requestSession argument list above?

```webidl
Promise<XRSession> requestSession(XRSessionMode mode, optional XRSessionInit options);
```

Speaking of this, is the intent that these init options only apply to `requestSession`, not `supportsSession`? If yes, I think it would be helpful to call that out specifically in the explainer. For example, asking for options may require prompting the user to get an answer, and `supportsSession` is intended to be non-interactive.

toji: Yes, the intent is these options only apply to `requestSession`. Not only would it violate the desire for `supportsSession` to be non-interactive to properly resolve some of the requirements, it would also make it a much more potent fingerprinting venue. As such `supportsSession` will remain `mode`-only.

explainer.md
```diff
@@ -405,20 +393,70 @@ function beginInlineXRSession() {
...
+WebXR allows the following features to be requested:
+* `local`
+* `local-floor`
+* `bounded-floor`
+* `unbounded`
+
+This list is currently limited to a subset of reference space types, but in the future will expand to include additional features. Some potential future features under discussion that would be candidates for this list are: eye tracking, plane detection, geo alignment, etc.
+
+Developers communicate their feature requirements by categorizing them into one of the following groups:
+* **Required** This feature must be available in order for the experience to function at all. If [explicit consent](privacy-security-explainer.md#explicit-consent) is necessary, users will be prompted in response to `xr.requestSession()`. Session creation will be rejected if the feature is unavailable for the XR device or if the UA determines the user does not wish the feature enabled.
```
johnpallett:
```suggestion
* **Required** This feature must be available in order for the experience to function at all. If [explicit consent](privacy-security-explainer.md#explicit-consent) is necessary, users will be prompted in response to `xr.requestSession()`. Session creation will be rejected if the feature is unavailable for the XR device or if the UA determines the user does not wish the feature enabled. If the UA does not recognize an entry in `XRSession.requiredFeatures` the session will not be created. 
```

explainer.md
johnpallett: @toji I agree, that sentence probably makes more sense in the 'required' section. I've added a suggestion above in case it helps.

@NellWaliczek the use of 'may' and 'allowed' here could be interpreted as user agent discretion - i.e. some UAs might add 'local' and others might not. Was that the intent? If not I'd suggest '... will be **enabled** by default...' and 'This is only **done** if the feature...' as follows:

```suggestion
Some features recognized by the UA but not explicitly listed in these arrays will be enabled by default for a session. This is only done if the feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent) nor impact performance or the behavior of other features when enabled. At this time, only the following features will be enabled by default:

```

explainer.md
NellWaliczek: This change came from my discussion with @johnpallett on Friday, so please correct me if I've misunderstood.  I was under the impression that we couldn't automatically allow a local-floor with even an emulated height because it could be used for fingerprinting.

explainer.md
NellWaliczek: >I've added a suggestion above in case it helps.

@johnpallett, above where?

NellWaliczek: Oh, GitHub.  It put your comments in two places.  Found it.

explainer.md
NellWaliczek: Thanks for catching that!

NellWaliczek:
Updated.  Waiting on the response to https://github.com/immersive-web/webxr/pull/739#discussion_r299256167 from @johnpallett

explainer.md
johnpallett: The concern with `local-floor` is when the viewer height is _not_ emulated. The user's height is sensitive data (i.e. 'are you a child') so would likely require explicit consent on some platforms.

In contrast, `local-floor` with _emulated_ height doesn't have this concern, and the UA can round the data to mitigate the fingerprinting risk.

@NellWaliczek If I understood on Friday the primary concern was making sure user agents were consistent in their behavior; it seems like requiring an _emulated_ `local-floor` in this scenario could achieve that consistency.

NellWaliczek: Oh, right now I remember.  We have two competing problems to solve:

1. We definitely don't want `local-floor` to be forced to use an emulated height when a real one is available (either via sensor data or a user configuration setting).
1. We definitely don't want to have inconsistency about some hardware needing to have `local-floor` requested up front and others not.  

The intent behind this approach was simply to settle on the lowest common denominator to ensure consistency and user safety.

@toji, you on board with that?

johnpallett: I'm not @toji but that makes sense to me and I agree with the approach as written

toji: To make sure I'm on the same page: My reading of this is that `local-floor` must be explicitly requested to be used, and `local` is the only reference space that will be implicitly available to immersive sessions. If that's correct, I'm on board with it.

NellWaliczek: That's correct

NellWaliczek:
Ok, this PR's been open for a bit and has been iterated based on the feedback provided.  At this point, I'm inclined to merge and address any subsequent feedback in a separate PR so that we are unblocked on finishing the "VR complete" milestone.

toji:
doit

joshmarinacci:
Engage!

PR #742, opened 2019-06-27
Require that sensitive UI hides WebXR content

### LINKED ISSUE #724: Behavior when XRSession.visibility is 'visible-blurred'

The current spec text states
> XRSessions MUST have their visibility state set to hidden or visible-blurred when the user is interacting with potentially sensitive UI from the user agent (such as entering a URL) in the trusted environment.

It also states

> A state of visible-blurred indicates that imagery rendered by the XRSession may be seen by the user, but is not the primary focus. requestAnimationFrame() callbacks MAY be throttled. Input is not processed by the XRSession.

What other, if any, concrete requirements should be added to the behavior of the session RAF, getViewerPose(), getPose(), and the data these APIs return while in this state? For example, should callbacks be forced to throttle down to, uh I don't know, 10fps? Should pose data be quantized or rounded?  PR #638 touches on this topic a bit, but doesn't seem to provide any concrete guidance.

toji:
/fixes #724

Until we get some better guidelines in place regarding what kind of
mitigations are appropriate for using the `visible-blurred` state when
sensitive UI is being displayed, simply require that the content is
hidden instead. `visible-blurred` may still be used in less sensitive
cases where the XR content is not in the foreground but still visible.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -80,8 +80,10 @@ spec: Gamepad; urlPrefix: https://w3c.github.io/gamepad/#
     type: attribute; text: touched; for: GamepadButton; url: dom-gamepadbutton-touched
     type: attribute; text: pressed; for: GamepadButton; url: dom-gamepadbutton-pressed
     type: attribute; text: value; for: GamepadButton; url: dom-gamepadbutton-value
-spec:html; type:method; for:HTMLCanvasElement; text:getContext(contextId); url: https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext
-spec:html; type:method; for:Window; text:requestAnimationFrame(callback); url: https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-animationframeprovider-requestanimationframe
+spec:html; urlPrefix: https://html.spec.whatwg.org/multipage/
+    type:method; for:HTMLCanvasElement; text:getContext(contextId); url: canvas.html#dom-canvas-getcontext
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2067,7 +2067,7 @@ Gaze Tracking {#gazetracking-security}
...
 
-To prevent this risk the user agent MUST set the [=visibility state=] of all {{XRSession}}s to {{XRVisibilityState/hidden}} or {{XRVisibilityState/visible-blurred}} when the users is interacting with sensitive, trusted UI such as URL bars or system dialogs. Additionally, to prevent a malicious page from being able to monitor input on a other pages the user agent MUST set the [=visibility state=] of all {{XRSession}}s on non-focused pages to {{XRVisibilityState/hidden}}.
+To prevent this risk the user agent MUST set the [=visibility state=] of all {{XRSession}}s to {{XRVisibilityState/hidden}} when the users is interacting with sensitive, trusted UI such as URL bars or system dialogs. Additionally, to prevent a malicious page from being able to monitor input on a other pages the user agent MUST set the [=visibility state=] of all {{XRSession}}s on non-focused pages to {{XRVisibilityState/hidden}}.
```
NellWaliczek: Can we add a link to something that's a concrete definition of the "non-focused pages"

NellWaliczek: nit: "when the users is"
nit: "on a other pages"

toji:
Thanks Nell! New issue is #743.

PR #745, opened 2019-06-27
Removed an outdated gamepad issue reference

toji:
Thanks to @alcooper91 for pointing out the oversight on my part!


toji:
Already is. (5 lines above the removed issue, just far enough away for GitHub to fold it up.)

PR #749, opened 2019-07-02
Spec text for required and optional features

toji:
This is the corresponding spec text for the functionality described in
PR #739. There's some hand waving going on around user consent and such
because those will be better defined in a follow-up PR. The primary goal
of this PR is to establish the algorithims that drive how the feature
requests are processed.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -458,6 +458,7 @@ function onSessionStarted(session) {
 Some features recognized by the UA but not explicitly listed in these arrays will be enabled by default for a session. This is only done if the feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent) nor impact performance or the behavior of other features when enabled. At this time, only the following features will be enabled by default:
 | Feature | Circumstances |
 | ------ | ------- |
+| `viewer` | Successfully created any session |
 | `local` | Successfully created session of mode `immersive-ar` or `immersive-vr` |
 
 ### Controlling rendering quality
--- a/explainer.md
+++ b/explainer.md
@@ -456,6 +456,7 @@ function onSessionStarted(session) {
 ```
 
 Some features recognized by the UA but not explicitly listed in these arrays will be enabled by default for a session. This is only done if the feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent) nor impact performance or the behavior of other features when enabled. At this time, only the following features will be enabled by default:
+
 | Feature | Circumstances |
 | ------ | ------- |
 | `viewer` | Requested `XRSessionMode` is `inline`, `immersive-vr`, or `immersive-ar` |
--- a/explainer.md
+++ b/explainer.md
@@ -458,8 +458,8 @@ function onSessionStarted(session) {
 Some features recognized by the UA but not explicitly listed in these arrays will be enabled by default for a session. This is only done if the feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent) nor impact performance or the behavior of other features when enabled. At this time, only the following features will be enabled by default:
 | Feature | Circumstances |
 | ------ | ------- |
-| `viewer` | Successfully created any session |
-| `local` | Successfully created session of mode `immersive-ar` or `immersive-vr` |
+| `viewer` | Requested `XRSessionMode` is `inline`, `immersive-vr`, or `immersive-ar` |
+| `local` | Requested `XRSessionMode` is `immersive-vr` or `immersive-ar` |
 
 ### Controlling rendering quality
 
--- a/index.bs
+++ b/index.bs
@@ -400,7 +400,7 @@ Some features of an {{XRSession}} may not be universally available for a number 
 
 Since it is a poor user experience to to initialize the underlying XR platform and create an {{XRSession}} only to immediately notify the user that the applications cannot function correctly, developers can indicate <dfn>required features</dfn> by passing an {{XRSessionInit}} dictionary to {{XR/requestSession()}}. This will block the creation of the {{XRSession}} if any of the [=required features=] are unavailable due to device limitations or in the absense of a clear signal of [=user intent=] to expose [=sensitive information=] related to the feature.
 
-Additionally, developers are encouraged to design experiences which progressively enchance their funcionality when run one more capable devices. <dfn>Optional features</dfn> which the experience does not required but will take advantage of when available must also be indicated in an {{XRSessionInit}} dictionary to ensure that [=user intent=] can be determined before enabling the feature if necessary.
+Additionally, developers are encouraged to design experiences which progressively enhance their functionality when run one more capable devices. <dfn>Optional features</dfn> which the experience does not require but will take advantage of when available must also be indicated in an {{XRSessionInit}} dictionary to ensure that [=user intent=] can be determined before enabling the feature if necessary.
 
 <pre class="idl">
 dictionary XRSessionInit {
--- a/index.bs
+++ b/index.bs
@@ -438,7 +438,7 @@ Some {{XRSessionMode}}s enable certain [=feature names=] as [=optional features=
   </tbody>
 </table>
 
-ISSUE: Define <dfn>sensitive information</dfn>, what qualifies as a signal of <dfn>user intent</dfn>, when <dfn>explicit consent</dfn> is necessary, and how user agents can <dfn>request user consent</dfn>.
+ISSUE: Define <dfn>sensitive information</dfn>, what qualifies as a signal of <dfn>user intent</dfn>, when <dfn>explicit consent</dfn> is necessary and how user agents can request it.
 
 <div class="algorithm" data-algorithm="resolve-features">
 
@@ -460,7 +460,7 @@ To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optio
     1. If |session|'s [=XRSession/mode=] is not one of the modes associated with |feature| in the [=default features=], continue to the next entry.
     1. If the functionality described by |feature| is not supported by |session|'s [=XRSession/XR Device=] or the user agent has otherwise determined to reject the feature, continue to the next entry.
     1. Else append |feature| to |session|'s [=list of enabled features=].
-  1. If |consentRequired| or |consentOptional| are not empty, [=request user consent=] to use the functionality described by those features.
+  1. If |consentRequired| or |consentOptional| are not empty, request [=explicit consent=] to use the functionality described by those features.
   1. For each |feature| in |consentRequired| perform the following steps:
     1. If a clear signal of [=user intent=] to enable |feature| has not been given, return <code>false</code>.
     1. Else append |feature| to |session|'s [=list of enabled features=].
--- a/index.bs
+++ b/index.bs
@@ -340,12 +340,12 @@ When the <dfn method for="XR">requestSession(|mode|, |options|)</dfn> method is 
             1. If |immersive| is <code>true</code>, set [=pending immersive session=] to <code>false</code>.
             1. Abort these steps.
         1. Let |session| be a new {{XRSession}} object.
+        1. [=Initialize the session=] with |session|, |mode|, and |device|.
         1. [=Resolve the requested features=] given by |options|' {{XRSessionInit/requiredFeatures}} and |options|' {{XRSessionInit/optionalFeatures}} values for |session|, and let |resolved| be the returned value.
         1. If |resolved| is <code>false</code>, run the following steps:
             1. [=Reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}}.
             1. If |immersive| is <code>true</code>, set [=pending immersive session=] to <code>false</code>.
             1. Abort these steps.
-        1. [=Initialize the session=] with |session|, |mode|, and |device|.
         1. Potentially set the [=active immersive session=] based on the following:
             <dl class="switch">
               <dt> If |immersive| is <code>true</code>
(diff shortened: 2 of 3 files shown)


index.bs
```diff
@@ -340,6 +340,11 @@ When the <dfn method for="XR">requestSession(|mode|)</dfn> method is invoked, th
             1. If |immersive| is <code>true</code>, set [=pending immersive session=] to <code>false</code>.
             1. Abort these steps.
         1. Let |session| be a new {{XRSession}} object.
```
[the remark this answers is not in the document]
toji: No, because the new line will actually populate an internal variable of the session with the enabled features.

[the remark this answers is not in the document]
NellWaliczek: Oh I see.  [=Resolve the requested features=] actually changes the session, yes?  If so, then line 348 needs to go above so that the algorithm can reference those properties ;)

index.bs
```diff
@@ -388,6 +393,77 @@ NOTE: Examples of ways [=exclusive access=] may be presented include stereo cont
...
+      <td>{{XRSessionMode/"immersive-vr"}}</td>
+    </tr>
+  </tbody>
+</table>
+
+ISSUE: Define what it means when a feature <dfn>requires consent</dfn>, how user agents can <dfn>request user consent</dfn>, and what it means when a user <dfn>denied consent</dfn>.
+
+<div class="algorithm" data-algorithm="resolve-features">
+
+To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optionalFeatures| for an {{XRSession}} |session|, the user agent MUST run the following steps:
```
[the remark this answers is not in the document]
toji: Yup. 🤷‍♀

toji:
Okay, did another pass on this to address all of @NellWaliczek's feedback and in general try to adhere better to the verbiage established in #739.

index.bs
```diff
@@ -388,6 +393,77 @@ NOTE: Examples of ways [=exclusive access=] may be presented include stereo cont
...
+  </thead>
+  <tbody>
+    <tr>
+      <td>{{XRReferenceSpaceType/"local"}}</td>
+      <td>{{XRSessionMode/"immersive-vr"}}</td>
+    </tr>
+  </tbody>
+</table>
+
+ISSUE: Define what it means when a feature <dfn>requires consent</dfn>, how user agents can <dfn>request user consent</dfn>, and what it means when a user <dfn>denied consent</dfn>.
```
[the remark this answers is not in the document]
toji: My plan was generally to address it in an immediate follow-up PR, and I'm not entirely sure if these definitions would take the form of an algorithm or not?

index.bs
```diff
@@ -405,16 +409,15 @@ dictionary XRSessionInit {
...
+The <dfn dict-member for="XRSessionInit">optionalFeatures</dfn> array contains any [=Optional features=] for the experience. If any value in the list is not a recognized [=feature name=] it will be ignored. Features listed in the {{XRSessionInit/optionalFeatures}} array will be enabled if supported by the [=XRSession/XR Device=] and, if necessary, given a clear signal of [=user intent=], but will not block creation of the {{XRSession}} if absent. 
 
 The feature lists accept any string. However, in order to be a considered a valid <dfn>feature name</dfn>, the string must be a value from the following enums:
 
  - {{XRReferenceSpaceType}}
 
-Some {{XRSessionMode}}s request certain [=feature names=] as optional features by default. The following table described the <dfn>default features</dfn> associated with each {{XRSessionMode}}:
+Some {{XRSessionMode}}s request certain [=feature names=] as [=optional features=] by default. This is only done if the feature does not require a signal of [=user intent=] nor impact performance or the behavior of other features when enabled. The following table described the <dfn>default features</dfn> associated with each {{XRSessionMode}}:
```
NellWaliczek: Nit: this sounds a bit weird to me...
>Some {{XRSessionMode}}s request

What about 

>Some {{XRSessionMode}}s enable certain ...

NellWaliczek: nit: described -> describes

index.bs
```diff
@@ -424,14 +427,18 @@ Some {{XRSessionMode}}s request certain [=feature names=] as optional features b
     </tr>
   </thead>
   <tbody>
+    <tr>
+      <td>{{XRReferenceSpaceType/"viewer"}}</td>
+      <td>{{XRSessionMode/"inline"}}, {{XRSessionMode/"immersive-vr"}}</td>
```
NellWaliczek: I've generally been including "immersive-ar" everywhere until we do the module split.  Mostly because I want to make sure I don't miss anything at the time that will need to be called out in the new documents.

index.bs
```diff
@@ -442,24 +449,23 @@ To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optio
   1. For each |feature| in |requiredFeatures| perform the following steps:
     1. If |feature| is not a valid [=feature name=], return <code>false</code>.
     1. If the functionality described by |feature| is not supported by |session|'s [=XRSession/XR Device=], return <code>false</code>.
-    1. If the functionality described by |feature| [=requires consent=], append it to |consentRequired|.
+    1. If the functionality described by |feature| requires [=explicit consent=], append it to |consentRequired|.
```
NellWaliczek: This is a bit quirky and we might need to look at other spec text to get it right... That said, I think what we're trying to say here (and also below) is that if implied consent is sufficient and the UA has determined to reject it, then it should be rejected.  Otherwise, if explicit consent is required, then prompt.

NellWaliczek:
Oh, bother.  I just noticed that line spatial-tracking-explainer.md line 329 should have been updated to say "local" not "local-floor".  Do you mind fixing that up in this PR while you're at it?

toji:
Done. :) Any further thoughts? I know this is a tricky one to get the verbiage right on.

explainer.md
```diff
@@ -458,6 +458,7 @@ function onSessionStarted(session) {
 Some features recognized by the UA but not explicitly listed in these arrays will be enabled by default for a session. This is only done if the feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent) nor impact performance or the behavior of other features when enabled. At this time, only the following features will be enabled by default:
 | Feature | Circumstances |
 | ------ | ------- |
+| `viewer` | Successfully created any session |
```
NellWaliczek: I probably should have originally phrased this a bit differently.  This (and changes to the next line) should probably read...

>Requested `XRSessionMode` is `inline`, `immersive-vr`, or `immersive-ar`

toji: Done!

toji:
Thanks! Took care of the last nit, so now I'm just patiently awaitng feedback from @johnpallett or @avadacatavra. 👀

explainer.md
```diff
@@ -458,7 +458,8 @@ function onSessionStarted(session) {
 Some features recognized by the UA but not explicitly listed in these arrays will be enabled by default for a session. This is only done if the feature does not require a signal of [user intent](privacy-security-explainer.md#user-intent) nor impact performance or the behavior of other features when enabled. At this time, only the following features will be enabled by default:
 | Feature | Circumstances |
```
johnpallett: @toji the preview doesn't render this table correctly for me, are you seeing that as well?

index.bs
```diff
@@ -388,6 +393,83 @@ NOTE: Examples of ways [=exclusive access=] may be presented include stereo cont
...
+
+Since it is a poor user experience to to initialize the underlying XR platform and create an {{XRSession}} only to immediately notify the user that the applications cannot function correctly, developers can indicate <dfn>required features</dfn> by passing an {{XRSessionInit}} dictionary to {{XR/requestSession()}}. This will block the creation of the {{XRSession}} if any of the [=required features=] are unavailable due to device limitations or in the absense of a clear signal of [=user intent=] to expose [=sensitive information=] related to the feature.
+
+Additionally, developers are encouraged to design experiences which progressively enchance their funcionality when run one more capable devices. <dfn>Optional features</dfn> which the experience does not required but will take advantage of when available must also be indicated in an {{XRSessionInit}} dictionary to ensure that [=user intent=] can be determined before enabling the feature if necessary.
```
johnpallett: typos. I like the word 'enchance' though. :)

```suggestion
Additionally, developers are encouraged to design experiences which progressively enhance their functionality when run one more capable devices. <dfn>Optional features</dfn> which the experience does not require but will take advantage of when available must also be indicated in an {{XRSessionInit}} dictionary to ensure that [=user intent=] can be determined before enabling the feature if necessary.
```

explainer.md
toji: Thanks. Just needed to add a line of whitespace above it.

index.bs
```diff
@@ -388,6 +393,83 @@ NOTE: Examples of ways [=exclusive access=] may be presented include stereo cont
...
+};
+</pre>
+
+The <dfn dict-member for="XRSessionInit">requiredFeatures</dfn> array contains any [=Required features=] for the experience. If any value in the list is not a recognized [=feature name=] the {{XRSession}} will not be created. If any feature listed in the {{XRSessionInit/requiredFeatures}} array is not supported by the [=XRSession/XR Device=] or, if necessary, has not received a clear signal of [=user intent=] the {{XRSession}} will not be created.
+
+The <dfn dict-member for="XRSessionInit">optionalFeatures</dfn> array contains any [=Optional features=] for the experience. If any value in the list is not a recognized [=feature name=] it will be ignored. Features listed in the {{XRSessionInit/optionalFeatures}} array will be enabled if supported by the [=XRSession/XR Device=] and, if necessary, given a clear signal of [=user intent=], but will not block creation of the {{XRSession}} if absent. 
+
+The feature lists accept any string. However, in order to be a considered a valid <dfn>feature name</dfn>, the string must be a value from the following enums:
+
+ - {{XRReferenceSpaceType}}
```
billorr-google: I like the overall shape, but have a couple question about the list of features.  I apologize if these are discussed or answered in the explainer already, but I didn't see an answer here.

Devices that support local reference spaces MUST support local-floor reference spaces, through emulation if necessary, and vice versa. <- If I require local-floor, am I guaranteed real floor height or emulated is sufficient to satisfy this?  Do I need to require local-floor for requestReferenceSpace to succeed later with emulated floor-level?

It seems there isn't any way to require features like "6dof" or "6dof controllers".  Is this intentional?  I see demand for this (some apps/games may not work well in 3dof).  Sites can get a somewhat reasonable approximation by requiring "bounded-floor", but this may exclude hardware that supports 6dof but not bounds.

Its possible that for some features (ie bounds), hardware capability/calibration may not be known until after the session is started from the browser's perspective.  Is the required features optimistic, and the session succeeds even though bounds may never be available, or should the browser fail the requestSession call?

NellWaliczek: > Devices that support local reference spaces MUST support local-floor reference spaces, through emulation if necessary, and vice versa. <- If I require local-floor, am I guaranteed real floor height or emulated is sufficient to satisfy this? Do I need to require local-floor for requestReferenceSpace to succeed later with emulated floor-level?

> It seems there isn't any way to require features like "6dof" or "6dof controllers". Is this intentional? I see demand for this (some apps/games may not work well in 3dof). Sites can get a somewhat reasonable approximation by requiring "bounded-floor", but this may exclude hardware that supports 6dof but not bounds.

The decision not to allow "require 6DOF" is actually core to the spatial tracking design.  "Local" and "local-floor" experiences are the lowest common denominator and should work for either 3dof or 6dof.  Real height is not guaranteed, but developers shouldn't be differentiating it anyway because people come in lots of heights and the UAs may be adding overrides for accessibility purposes.  If an experience requires 6dof, then it should request "bounded-floor" by is very nature.  Controller requirements are described in #755.

> Its possible that for some features (ie bounds), hardware capability/calibration may not be known until after the session is started from the browser's perspective. Is the required features optimistic, and the session succeeds even though bounds may never be available, or should the browser fail the requestSession call?

Perhaps I'm misunderstanding, but the exact bounds aren't necessary to do a session check.  The capability to create a bounded reference space is what's being checked.

(ok, seriously.. woods time now)

billorr-google: Understood about local and local-floor being lowest common denominator that should generally work with 3dof.  Also agree with the accessibility concerns.  It sounds like we're ok with requiring bounded-floor ~= requiring 6dof, which is probably fine for VR headsets.  As AR is added we may want to revisit this.  Thanks for the link to the input issue.

For the last point, there are several degrees of capability - the runtime may support bounds for some hardware, the hardware may support bounds, there may be data (calibration?) that actually specify bounds, the bounds may be located relative to the user.  Should we be clear which we mean?  There may be times where we don't know which degree of capability we have without starting the runtime - OpenXR doesn't seem to specify a way to inquire about bounds without starting a session for example.

More specifically - if bounded-floor is required, should requestReferenceSpace('bounded-floor') necessarily succeed?  I think my comment/question is at most a clarification, so feel free to land.

Have fun in the woods.

toji: Going to land this now prior to heading into the US holiday weekend, but to try and answer your question:

Requirements are always going to be a best effort endeavor for certain things, especially if connectivity or calibration of the functionality in question is fluid. In a fair number of cases I'd imagine that the user agent will be running on known hardware (like a Quest) and can confidently answer capability questions without any API interaction. In some cases it may actually be appropriate to spin up a platform-level session equivalent to do some basic checks as long as you can delay visual transitions. (I think OpenXR probably allows for this.) It may also be appropriate for the UA/platform to intervene at the point of session creation to ask the user to do some setup. For example:

> This VR experience requires you to configure your room boundaries before launching. Do you wish to do so now, or exit this experience?

I would also say in some cases it's probably acceptable to, if you know the platform is capable of a feature but it takes a moment to initialize, return dummy data until you have the real stuff. Such as a `bounded-floor` reference space returning a ~1m circle of bounds initially and then firing a `reset` event when the actual data kicks in.

> if bounded-floor is required, should requestReferenceSpace('bounded-floor') necessarily succeed?

I'm not 100% confident about this, as I can see the need for some wiggle room in some cases, but in general I would say yes? I also feel like that's a policy that may require different answers for different features as we add new ones in the future. (If we add a theoretical `6dof-controllers` feature, it would be impractical to insist that said controllers be continuously available from the very first frame.)

PR #760, opened 2019-07-08
Fix typo in requestReferenceSpace algorithm

bricetebbs:
Value seemed missing

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1060,7 +1060,7 @@ To check if a <dfn>reference space is supported</dfn> for a given reference spac
   1. If |type| is {{viewer}}, return <code>true</code>.
   1. If |type| is {{local}} or {{local-floor}}, and |session| is an [=immersive session=], return <code>true</code>.
   1. If |type| is {{local}} or {{local-floor}}, and the [=XRSession/XR device=] supports reporting orientation data, return <code>true</code>.
-  1. If |type| is {{bounded-floor}} and |session| is an [=immersive session=], return if [=bounded reference spaces are supported=] by the [=XRSession/XR device=].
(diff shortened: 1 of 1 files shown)


toji:
Hm... I see why this doesn't read well and think it needs to be fixed, but your change actually subtly changes the logic, so we need to be careful here. The original text is:

> If |type| is {{bounded-floor}} and |session| is an [=immersive session=], return if [=bounded reference spaces are supported=] by the [=XRSession/XR device=].

Which, stated differently, intends to say "If `bounded-floor` has been requested run the algorithm named `bounded reference spaces are supported` and return the value that it returns." Crucially, this statement will return with either true or false any time the type is `bounded-floor` (and the session is an immersive one)

Your proposed change:

> If |type| is {{bounded-floor}} and |session| is an [=immersive session=], **return true** if [=bounded reference spaces are supported=] by the [=XRSession/XR device=].

Actually indicates that if bounded spaces aren't supported then the logic should fall through to the next statement, because it only specifies a return value when bounded spaces are supported. A minor but important distinction. 😄 

What we need for both readability and correctness is something along the lines of:

> 1.  If |type| is {{bounded-floor}} and |session| is an [=immersive session=]:
>     1. Let |boundedSupported| be whether or not [=bounded reference spaces are supported=] by the [=XRSession/XR device=]
>     1. Return |boundedSupported|

That seems a bit awkward, but at least it's unambiguous. Suggestions for more elegant wording are welcome!

bricetebbs:
Hmm. If that's how the spec code works then its probably fine as is maybe? I was just trying to make the cases all look similar as the others also rely on fall-through right?

PR #761, opened 2019-07-09
Spec: Pose privacy considerations

toji:
Describes the mitigations and checks that must be used when returning
pose data from the API. Also includes discussion of data adjustments
and the privacy considerations for returning immersive sessions. Pulls
heavily from privacy-security-explainer.md.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -325,6 +325,8 @@ When the <dfn method for="XR">requestSession(|mode|, |options|)</dfn> method is 
     1. Check if an [=immersive session request is allowed=], and if not [=reject=] |promise| with a "{{SecurityError}}" {{DOMException}} and return |promise|.
     1. If [=pending immersive session=] is <code>true</code> or [=active immersive session=] is not <code>null</code>, [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}} and return |promise|.
     1. Set [=pending immersive session=] to <code>true</code>.
+  1. If |immersive| is <code>false</code>:
+    1. Check if an [=inline session request is allowed=], and if not [=reject=] |promise| with a "{{SecurityError}}" {{DOMException}} and return |promise|.
   1. Run the following steps [=in parallel=]:
     1. Choose |device| based on the following:
         <dl class="switch">
@@ -439,6 +441,8 @@ Some {{XRSessionMode}}s enable certain [=feature names=] as [=optional features=
     </tr>
   </tbody>
 </table>
+
+The combined list of [=feature names=] given by the {{XRSessionInit/requiredFeatures}}, {{XRSessionInit/optionalFeatures}}, and [=default features=] are collectively considered the <dfn>requested features</dfn> for an {{XRSession}}.
 
 <div class="algorithm" data-algorithm="resolve-features">
 
@@ -2109,6 +2113,19 @@ Sensitive Information {#sensitive-information-header}
 
 In the context of XR, <dfn>sensitive information</dfn> includes, but is not limited to, user configurable data such as interpupillary distance (IPD) and sensor-based data such as {{XRPose}}s. All {{XRSessionMode/"immersive-vr"}} and {{XRSessionMode/"immersive-ar"}} sessions will expose some amount of sensitive data, due to the user's pose being necessary to render anything. However, in some cases, the same sensitive information will also be exposed via {{XRSessionMode/"inline"}} sessions.
 
+### Active and focused document ### {#active-and-focused-document}
+A document MUST be [=active and focused=] at the time that [=sensitive information=] is requested.
+
+<div class="algorithm" data-algorithm="document-is-active-and-focused">
+
+To determine if a given {{Document}} |document| is <dfn>active and focused</dfn> the user agent MUST run the following steps:
+
+  1. If the [=currently focused area=] does not belong to |document|, return <code>false</code>
+  1. If |document| is not of the [=same origin-domain=] as the [=active document=], return <code>false</code>
+  1. Return <code>true</code>
+
+</div>
+
 ### Trustworthy documents and origins ### {#trustworthy-documents}
 In order to expose any [=sensitive information=] the requesting document MUST be considered [=trustworthy=].
 
@@ -2117,21 +2134,8 @@ In order to expose any [=sensitive information=] the requesting document MUST be
 To determine if a given {{Document}} |document| is <dfn>trustworthy</dfn> the user agent MUST run the following steps:
 
   1. If |document| is not a [=responsible=] document, return <code>false</code>
-  1. If |document| is not of a [=secure context=], return <code>false</code>
-  1. If |document|'s origin is blocked by [[#feature-policy|feature policy]], return <code>false</code>
-  1. Return <code>true</code>
-
-</div>
-
-### Active and focused document ### {#active-and-focused-document}
-In addition to being [=trustworthy=], a document MUST be [=active and focused=] at the time that [=sensitive information=] is requested.
-
-<div class="algorithm" data-algorithm="document-is-active-and-focused">
-
-To determine if a given {{Document}} |document| is <dfn>active and focused</dfn> the user agent MUST run the following steps:
-
-  1. If the [=currently focused area=] does not belong to |document|, return <code>false</code>
-  1. If |document| is not of the [=same origin-domain=] as the [=active document=], return <code>false</code>
+  1. If |document| is not [=active and focused=], return <code>false</code>
+  1. If |document|'s origin is not allowed to use the WebXR [[#feature-policy|feature policy]] or the [=underlying sensors=]' feature policies, return <code>false</code>
   1. Return <code>true</code>
 
 </div>
@@ -2188,10 +2192,20 @@ Users must be in control of when immersive sessions are created because the crea
 To determine if an <dfn>immersive session request is allowed</dfn> the user agent MUST run the following steps:
 
   1. If the request was not [=triggered by user activation=], return <code>false</code>
-  1. If the request does not originate from a [=trustworthy=] document and origin, return <code>false</code>
-  1. If the request does not originate from a document that is [=active and focused=], return <code>false</code>
-  1. If the request does not originates from a document that is allowed to use the WebXR [[#feature-policy|feature policy]] as well as the [=underlying sensors=]' feature policies, return <code>false</code>
-  1. If [=user intent=] is not well understood, either via [=explicit consent=] or [=implicit consent=], return <code>false</code>
+  1. If the requesting document is not considered [=trustworthy=], return <code>false</code>
+  1. If [=user intent=] to begin an [=immersive session=] is not well understood, either via [=explicit consent=] or [=implicit consent=], return <code>false</code>
+  1. Return <code>true</code>
+
+</div>
+
+Starting an {{XRSessionMode/"inline"}} session does not implicitly carry the same requirements, though additional requirements may be imposed depending on the session's [=requested features=].
+
+<div class="algorithm" data-algorithm="inline-session-allowed">
+
+To determine if an <dfn>inline session request is allowed</dfn> the user agent MUST run the following steps:
+
+  1. If the requesting document is not [=responsible=], return <code>false</code>
+  1. If requesting document's origin is not allowed to use the WebXR [[#feature-policy|feature policy]] or the [=underlying sensors=]' feature policies, return <code>false</code>
   1. Return <code>true</code>
 
 </div>
@@ -2205,19 +2219,23 @@ To determine if <dfn>poses may be reported</dfn> to an {{XRSession}} |session|, 
 
   1. Let |document| be the document that owns |session|.
   1. If the request does not originate from |document|, return <code>false</code>.
-  1. If |document| is not [=active and focused=], return <code>false</code>.
+  1. If |document| is not considered [=trustworthy=], return <code>false</code>.
   1. If |session|'s {{XRSession/visibilityState}} in not {{XRVisibilityState/"visible"}}, return <code>false</code>.
   1. Determine if the pose data can be returned as follows:
     <dl class="switch">
-      <dt> If [=user intent=] is well understood, either via [=explicit consent=] or [=implicit consent=]
+      <dt> If the pose data is known by the user agent to not expose fingerprintable sensor data
       <dd> Return <code>true</code>.
       <dt> If [=data adjustments=] will be applied to the underlying sensor data to prevent fingerprinting or profiling
+      <dd> Return <code>true</code>.
+      <dt> If [=user intent=] is well understood, either via [=explicit consent=] or [=implicit consent=]
       <dd> Return <code>true</code>.
       <dt> Otherwise
       <dd> Return <code>false</code>.
     </dl>
 
 </div>
+
+Note: The method by which a user agent determines that poses do not expose fingerprintable data is left to the user agent's discretion.
 
 The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion of {{XRView}} information. When more than one view is present and the physical relationship between these views is configurable by the user, the relationship between these views is considered [=sensitive information=] as it can be used to fingerprint or profile the user.
 
--- a/index.bs
+++ b/index.bs
@@ -2221,7 +2221,7 @@ To determine if <dfn>poses may be reported</dfn> to an {{XRSession}} |session|, 
 
 The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion of {{XRView}} information. When more than one view is present and the physical relationship between these views is configurable by the user, the relationship between these views is considered [=sensitive information=] as it can be used to fingerprint or profile the user.
 
-If the relationship between {{XRView}}s could uniquely identify the [=/XR device=], then the user agent MUST anonymized the {{XRView}} data to prevent fingerprinting. The method of anonymization is at the discretion of the user agent. Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] prior to reporting any {{XRView}} data.
+If the relationship between {{XRView}}s could uniquely identify the [=/XR device=], then the user agent MUST anonymized the {{XRView}} data to prevent fingerprinting. The method of anonymization is at the discretion of the user agent. Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] during session creation, prior to reporting any {{XRView}} data.
 
 <section class="unstable">
 Gaze Tracking {#gazetracking-security}
--- a/index.bs
+++ b/index.bs
@@ -2135,7 +2135,7 @@ To determine if a given {{Document}} |document| is <dfn>trustworthy</dfn> the us
 
   1. If |document| is not a [=responsible=] document, return <code>false</code>
   1. If |document| is not [=active and focused=], return <code>false</code>
-  1. If |document|'s origin is not allowed to use the WebXR [[#feature-policy|feature policy]] or the [=underlying sensors=]' feature policies, return <code>false</code>
+  1. If |document|'s origin is not allowed to use the WebXR [[#feature-policy|feature policy]], return <code>false</code>
   1. Return <code>true</code>
 
 </div>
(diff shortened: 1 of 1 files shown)


toji:
@johnpallett: Can't add you as a reviewer directly, but could you take a look at this?

index.bs
```diff
@@ -2153,6 +2154,68 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+
+Protected functionality {#protected-functionality}
+-----------------------
+
+The [=sensitive information=] exposed by the API can be divided into categories that share threat profiles and necessary protections against those threats.
+
+### Immersiveness ### {#protect-immersiveness}
+Users must be in control of when immersive sessions are created because the creation causes invasive changes on a user's machine. For example, starting an [=immersive session=] will engage the [=/XR device=] sensors, take over access to the device's display, and begin presentating immersive content which may terminate another application's access to the XR hardware. It may also incur significant power or performance overhead on some systems or trigger the launching of a status tray or storefront.
+
+<div class="algorithm" data-algorithm="immersive-session-allowed">
```
johnpallett: Do we need a similar algorithm for inline sessions? 

For context:
- The spec has sensitive data access requirements (#sensitive-information-header and #active-and-focused-document)
- An inline session only needs to meet sensitive data access requirements if it will actually expose sensitive data. This isn't always the case, e.g. An inline session with a viewer reference space and no inputs doesn't expose sensitive data
- _User intent_ requirements for sensitive data are spelled out in #761 and #762. Inline sessions can obtain user consent mid-session, so that seems to be covered.

However, AFAICT:
- The reference space feature requirements in #762 do not include sensitive data access requirements
- The pose data requirements in #761 only partially address sensitive data access requirements
- The algorithm below addresses sensitive data access requirements, but only for immersive sessions
- There doesn't appear to be an algorithm or set of requirements that would cause them to apply for inline sessions

Does this mean there is a chance that an inline session could be created that exposed sensitive data, but did not meet all the sensitive data access requirements? 

For example, could an insecure-origin inline session be created that exposes pose data in a local reference space? Where is the algorithm step that prevents that? 

Note: It's possible that this is covered, and I missed it. I'll review again tomorrow to double check.

index.bs
```diff
@@ -2153,6 +2154,68 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+      <dd> Return <code>true</code>.
+      <dt> Otherwise
+      <dd> Return <code>false</code>.
+    </dl>
+
+</div>
+
+The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion of {{XRView}} information. When more than one view is present and the physical relationship between these views is configurable by the user, the relationship between these views is considered [=sensitive information=] as it can be used to fingerprint or profile the user.
+
+If the relationship between {{XRView}}s could uniquely identify the [=/XR device=], then the user agent MUST anonymized the {{XRView}} data to prevent fingerprinting. The method of anonymization is at the discretion of the user agent. Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] prior to reporting any {{XRView}} data.
```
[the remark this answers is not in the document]
johnpallett: _Note: Inline sessions might wish to do this later than session creation, but immersive sessions shouldn't, which is why I'm wondering about the timing here._

index.bs
```diff
@@ -2153,6 +2154,68 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+
+  1. Let |document| be the document that owns |session|.
+  1. If the request does not originate from |document|, return <code>false</code>.
+  1. If |document| is not [=active and focused=], return <code>false</code>.
+  1. If |session|'s {{XRSession/visibilityState}} in not {{XRVisibilityState/"visible"}}, return <code>false</code>.
+  1. Determine if the pose data can be returned as follows:
+    <dl class="switch">
+      <dt> If [=user intent=] is well understood, either via [=explicit consent=] or [=implicit consent=]
+      <dd> Return <code>true</code>.
+      <dt> If [=data adjustments=] will be applied to the underlying sensor data to prevent fingerprinting or profiling
```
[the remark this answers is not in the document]
johnpallett: Note: The explainer captures this by using the phrase 'may' instead of 'will'

index.bs
```diff
@@ -2153,6 +2154,68 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+  1. Return <code>true</code>
+
+</div>
+
+### Poses ### {#protect-poses}
+When based on sensor data, {{XRPose}} and {{XRViewerPose}} will expose [=sensitive information=] that may be misused in a number of ways, including input sniffing, gaze tracking, or fingerprinting.
+
+<div class="algorithm" data-algorithm="poses-allowed">
+
+To determine if <dfn>poses may be reported</dfn> to an {{XRSession}} |session|, the user agent MUST run the following steps:
```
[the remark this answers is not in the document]
toji: I disagree with the suggestion in this case specifically because we're defining an algorithm here that is explicitly called by the algorithm which populates the pose, which implies it'll be called every time.

index.bs
```diff
@@ -2196,7 +2259,7 @@ The feature identifier for this feature is <code>"xr"</code>.
 
 The [=default allowlist=] for this feature is <code>["self"]</code>.
 
-In addition to the <code>"xr"</code> feature policy, feature policies for underlying sensors must also be respected if a site could isolate and extract sensor data that would otherwise be blocked by those feature policies.
+In addition to the <code>"xr"</code> feature policy, feature policies for <dfn>underlying sensors</dfn> must also be respected if a site could isolate and extract sensor data that would otherwise be blocked by those feature policies.
```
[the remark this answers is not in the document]
toji: SGTM, though I'd still like to hear from @ddorwin and @NellWaliczek on the matter.

toji:
Okay, restructured the doc a bit based on John's comments and a quick call with him (thanks!) Hopefully this addressed the concerns raised and puts us in a bit better position for applying definitions like the document's "trustworthiness" in more places.

index.bs
```diff
@@ -2153,6 +2158,82 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+When based on sensor data, {{XRPose}} and {{XRViewerPose}} will expose [=sensitive information=] that may be misused in a number of ways, including input sniffing, gaze tracking, or fingerprinting.
+
+<div class="algorithm" data-algorithm="poses-allowed">
+
+To determine if <dfn>poses may be reported</dfn> to an {{XRSession}} |session|, the user agent MUST run the following steps:
+
+  1. Let |document| be the document that owns |session|.
+  1. If the request does not originate from |document|, return <code>false</code>.
+  1. If |document| is not considered [=trustworthy=], return <code>false</code>.
+  1. If |session|'s {{XRSession/visibilityState}} in not {{XRVisibilityState/"visible"}}, return <code>false</code>.
```
johnpallett: @toji sorry if I already asked this, should we also be checking be 'active and focused' ie:

```suggestion
  1. If |session|'s {{XRSession/visibilityState}} in not {{XRVisibilityState/"visible"}}, return <code>false</code>.
  1. If |document| is not [=active and focused=], return <code>false</code>.
```

toji: The `trustworthy` algorithm (referenced on line 2215 here) includes a check that the document is active and focused already.

index.bs
```diff
@@ -2196,7 +2277,7 @@ The feature identifier for this feature is <code>"xr"</code>.
 
 The [=default allowlist=] for this feature is <code>["self"]</code>.
 
-In addition to the <code>"xr"</code> feature policy, feature policies for underlying sensors must also be respected if a site could isolate and extract sensor data that would otherwise be blocked by those feature policies.
+In addition to the <code>"xr"</code> feature policy, feature policies for <dfn>underlying sensors</dfn> must also be respected if a site could isolate and extract sensor data that would otherwise be explicitly blocked by those feature policies' origin allow list.
```
johnpallett: Per #768 we may want to hold off on the underlying sensor requirement to avoid the anti-pattern of developers enabling sensors for hardware that doesn't need them. I realize this is the opposite of what I'd proposed before.

I'd suggest we remove this line.

index.bs
johnpallett: You're correct! I was looking at an older version of the algorithm. Thanks, you can resolve this.

index.bs
```diff
@@ -2153,6 +2158,82 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
 
 Regardless of how long the user agent chooses to persist the user's consent, [=sensitive information=] MUST only be exposed by an {{XRSession}} which has not [=ended=]. 
 
+Data adjustments {#data-adjustments-header}
+----------------
+
+In some cases, security and privacy threats can be mitigated through <dfn>data adjustment</dfn>s such as throttling, quantizing, rounding, limiting, or otherwise manipulating the data reported from the [=/XR device=]. This may sometimes be necessary to avoid fingerprinting, even in situations when [=user intent=] has been established.  However, [=data adjustment=] mitigations MUST only be used in situations which would not result in user discomfort.
+
+### Throttling ### {#throttling}
+Throttling is when [=sensitive information=] is reported at a lower frequency than otherwise possible. This mitigation has the potential to reduce a site's ability to infer user intent, location, or perform user profiling. However, when not used appropriately throttling runs a significant risk of causing user discomfort. In addition, under many circumstances it may be inadequate to provide a complete mitigation.
```
NellWaliczek: nit: Grammar is a bit wrong in this sentence.  "location" probably needs to be prefaced with "infer"

>This mitigation has the potential to reduce a site's ability to infer user intent, location, or perform user profiling.

index.bs
```diff
@@ -2153,6 +2158,82 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+The [=sensitive information=] exposed by the API can be divided into categories that share threat profiles and necessary protections against those threats.
+
+### Immersiveness ### {#protect-immersiveness}
+Users must be in control of when immersive sessions are created because the creation causes invasive changes on a user's machine. For example, starting an [=immersive session=] will engage the [=/XR device=] sensors, take over access to the device's display, and begin presentating immersive content which may terminate another application's access to the XR hardware. It may also incur significant power or performance overhead on some systems or trigger the launching of a status tray or storefront.
+
+<div class="algorithm" data-algorithm="immersive-session-allowed">
+
+To determine if an <dfn>immersive session request is allowed</dfn> the user agent MUST run the following steps:
+
+  1. If the request was not [=triggered by user activation=], return <code>false</code>
```
NellWaliczek: Do we still need to expand the definition of user activation to include things like launching an installed web app?

index.bs
```diff
@@ -2108,29 +2113,29 @@ Sensitive Information {#sensitive-information-header}
...
+<div class="algorithm" data-algorithm="document-is-trustworthy">
 
-To determine if a given {{Document}} |document| is <dfn>active and focused</dfn> the user agent MUST run the following steps:
+To determine if a given {{Document}} |document| is <dfn>trustworthy</dfn> the user agent MUST run the following steps:
 
-  1. If the [=currently focused area=] does not belong to |document|, return <code>false</code>
-  1. If |document| is not of the [=same origin-domain=] as the [=active document=], return <code>false</code>
+  1. If |document| is not a [=responsible=] document, return <code>false</code>
+  1. If |document| is not [=active and focused=], return <code>false</code>
+  1. If |document|'s origin is not allowed to use the WebXR [[#feature-policy|feature policy]] or the [=underlying sensors=]' feature policies, return <code>false</code>
```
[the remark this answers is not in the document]
NellWaliczek: @johnpallett Is there already an issue tracking the need to make a decision on this topic?

index.bs
```diff
@@ -2153,6 +2158,82 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+
+</div>
+
+Starting an {{XRSessionMode/"inline"}} session does not implicitly carry the same requirements, though additional requirements may be imposed depending on the session's [=requested features=].
+
+<div class="algorithm" data-algorithm="inline-session-allowed">
+
+To determine if an <dfn>inline session request is allowed</dfn> the user agent MUST run the following steps:
+
+  1. If the requesting document is not [=responsible=], return <code>false</code>
```
NellWaliczek: As written this will allow inline sessions to trigger consent prompts without a user interaction.  Since that might slip through the cracks I'd like to confirm it is intentional. For context, the new api to prompt for access to accelerometer data was added since a bunch of sites already silently sniff this data.  However camera/mic access doesn't appear to have a interaction requirement.

index.bs
```diff
@@ -2153,6 +2158,82 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+### Poses ### {#protect-poses}
+When based on sensor data, {{XRPose}} and {{XRViewerPose}} will expose [=sensitive information=] that may be misused in a number of ways, including input sniffing, gaze tracking, or fingerprinting.
+
+<div class="algorithm" data-algorithm="poses-allowed">
+
+To determine if <dfn>poses may be reported</dfn> to an {{XRSession}} |session|, the user agent MUST run the following steps:
+
+  1. Let |document| be the document that owns |session|.
+  1. If the request does not originate from |document|, return <code>false</code>.
+  1. If |document| is not considered [=trustworthy=], return <code>false</code>.
```
NellWaliczek: Does this actually need to be called out given that immersive session creation and inline w/special features already gate on trustworthy documents?

index.bs
```diff
@@ -2153,6 +2158,82 @@ It is recommended that once [=explicit consent=] is granted for a specific [=/or
...
+      <dd> Return <code>false</code>.
+    </dl>
+
+</div>
+
+Note: The method by which a user agent determines that poses do not expose fingerprintable data is left to the user agent's discretion.
+
+The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion of {{XRView}} information. When more than one view is present and the physical relationship between these views is configurable by the user, the relationship between these views is considered [=sensitive information=] as it can be used to fingerprint or profile the user.
+
+If the relationship between {{XRView}}s could uniquely identify the [=/XR device=], then the user agent MUST anonymized the {{XRView}} data to prevent fingerprinting. The method of anonymization is at the discretion of the user agent. Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] during session creation, prior to reporting any {{XRView}} data.
```
NellWaliczek: nit: typo "anonymized" -> "anonymize"

NellWaliczek: >Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] during session creation, prior to reporting any {{XRView}} data.

Does this need to be non-normative?

index.bs
```diff
@@ -2108,29 +2113,29 @@ Sensitive Information {#sensitive-information-header}
...
+<div class="algorithm" data-algorithm="document-is-active-and-focused">
 
-To determine if a given {{Document}} |document| is <dfn>trustworthy</dfn> the user agent MUST run the following steps:
+To determine if a given {{Document}} |document| is <dfn>active and focused</dfn> the user agent MUST run the following steps:
 
-  1. If |document| is not a [=responsible=] document, return <code>false</code>
-  1. If |document| is not of a [=secure context=], return <code>false</code>
-  1. If |document|'s origin is blocked by [[#feature-policy|feature policy]], return <code>false</code>
+  1. If the [=currently focused area=] does not belong to |document|, return <code>false</code>
+  1. If |document| is not of the [=same origin-domain=] as the [=active document=], return <code>false</code>
```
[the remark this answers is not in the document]
johnpallett: A primary consideration of this requirement is input sniffing. I could imagine a scenario where a user in an immersive session was presented with an OS-level virtual keyboard or notification input (possibly in response to an event triggered by a different origin).

index.bs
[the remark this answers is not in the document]
johnpallett: Good question; #768 contains the considerations and raises the question. It could be used either to track a decision about whether we should remove the requirement in the short-term, or could be used to find a longer-term solution. @ddorwin @NellWaliczek do you have a preference for how that conversation is structured?

index.bs
toji: This was because trustworthy status included active and focused, but I agree it reads a bit weird give that that's the only part of "trustworthiness" that can change. Replaced with "active and focused" directly.

index.bs
toji: How would we want to frame that? First impression is to say that if any of the required features will require explicit consent and the request was not triggered by user activation then reject, but in order for that to be predictable it assumes that all browsers consider the same features to require explicit consent, and I don't think that's practical. Another approach would be to say that if _any_ optional or required features are present then user activation is required, but that feels overly aggressive. Thoughts?

index.bs
toji: Ah, thank you for the reminder. Done.

@johnpallett, do you want to sanity check this new chunk of text for unintended consequences?

toji:
Updated the PR to address any feedback that isn't pending a separate decision.

index.bs
NellWaliczek: I'm not actually sure we'd even want to make a change here.  For example, camera/microphone don't require a user interaction before prompting.  Neither do notifications (though ugh that's sort of painful).

If we *did* decide to do this, the safest approach might just be to say we require a user action if any features are requested.  Alternatively, developers could call requestSession and only add a button to cause a user action if the initial call fails with a specific error.

index.bs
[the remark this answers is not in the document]
toji: I'm going to accept John's suggestion here because it seems like the overall lowest impact way to proceed forward for the following reason:

The intent behind this was to prevent pages from "backdooring" certain bits of accelerometer/gyro sensor data that may reveal things like pin entry on a 2D page. I think it's reasonable to say that this should not be a concern for `immersive-vr` because the context in which the sensors are being used is significantly different and has a different set of threat vectors (the handling of which is what this entire section is about.) The biggest concerning scenario, as such, is sensor-based tracking in `inline` mode. Fortunately for us this is not something with guaranteed availability. It may be rejected by the user or the UA, or simply not be supported by the device at all (desktop.) As such, if we were to introduce a similar restriction down the line the actual impact of it would be that some mobile devices start displaying the same inline behavior as a desktop device. That seems like a pretty soft failure, as these things go. Plus, it gives a clear indication that something has stopped working and gives us an opportunity to communicate why in the console at that point. If we went the other way and decided that these additional feature policies were not, in fact, necessary we would have a much more difficult time communicating to developers that they have now exposed certain features unnecessarily, since nothing would indicate that there was a problem to look into and it would be a mighty heuristics challenge to identify scenarios in which the developer ONLY was using those feature policies for WebXR and as such we could safely tell them to remove them.

index.bs
toji: > the safest approach might just be to say we require a user action if any features are requested.

Thinking this through: In an ideal world developers could throw up inline content with no sensor data immediately, and then at some point in the future when the user chooses to allow it they could upgrade in-place to start using sensors. I don't really see how that exact pattern could work out with how the API is structured, though (and have no desire to restructure it to work around this one issue.)

The problem is that even if you specify a feature as optional you'd have to either have consent granted or denied before the session was created, which would block the inline session in the "perfect" scenario. And even if you somehow proceeded without waiting for consent most apps would query the `local` space immediately, fail, and never check again. You could build out features for allowing consent to be async but... No.

So it seems like the "right" developer pattern here is to create a viewer-only inline session right away, then wait for consent to be granted for sensor data and, if granted, swap out the sessions at the point that the sensor-enabled session is created. That pattern works with either a immediate prompt via a butter bar or similar OR waiting on user interaction. One of them is far less annoying to the user, though, so I'd probably prefer to have any features require user interaction. Plus it's a pattern that's well-reflected in the native world where apps like Google Maps have little icons in their 360 views to toggle accelerometer-based rotation on and off. (It's also an easier restriction to loosen in the future, whereas adding it after the fact would be a significant breaking change.)

toji:
Okay, updated the PR to take into account the underlying sensor discussion and inline feature discussion, both of which I've dived into in greater detail above. I have not explicitly addressed the concern about iframes on desktop devices because I'm still not sure from the conversation if that's a scenario that we aren't concerned about. I do think it's something that we could loosen restrictions on in the future with no ill effects, though.

index.bs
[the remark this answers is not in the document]
NellWaliczek: @johnpallett I'm not sure we're talking about the same thing? In that case, wouldn't the previous bullet be the one that cause the algorithm to return false?

index.bs
NellWaliczek: Works for me

PR #762, opened 2019-07-09
Spec: Reference space privacy considerations

toji:
Describes the mitigations and checks that must be used when supporting
the various reference spaces. Pulls heavily from text originally in
privacy-security-explainer.md.

Posted as a draft because I'm not sure if I'm happy with the format yet.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2267,7 +2267,6 @@ If {{XRReferenceSpaceType/"bounded-floor"}} is included in an {{XRSession}}'s [=
 Any {{XRReferenceSpaceType/"bounded-floor"}} reference spaces MUST respect the following restrictions:
   - Each point in the {{XRBoundedReferenceSpace/boundsGeometry}} must be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=]; the suggested distance is 15 meters in all directions.
   - Each point in the {{XRBoundedReferenceSpace/boundsGeometry}} must be [=rounding|rounded=] sufficiently to prevent fingerprinting. Rounding to the nearest 5cm is suggested. For user's safety, rounded points values MUST NOT fall outside the original bounds.
-  - The <code>Y</code> value of the [=native origin=] must be [=rounding|rounded=] sufficiently to prevent fingerprinting of lower-order bits; rounding to the nearest 1cm is suggested.
   - All poses returned relative to a {{XRReferenceSpaceType/"bounded-floor"}} reference space MUST be [=limiting|limited=] to a reasonable distance beyond the {{XRBoundedReferenceSpace/boundsGeometry}} in all directions; the suggested distance is 1 meter beyond the bounds in all directions.
 
 #### local-floor #### {#local-floor-privacy}
@@ -2276,13 +2275,15 @@ On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local-floor"
 If {{XRReferenceSpaceType/"local-floor"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local-floor"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
 
 Any {{XRReferenceSpaceType/"local-floor"}} reference spaces MUST respect the following restrictions:
-  - If the floor level is based on sensor data or is set to a non-default emulated value, the <code>Y</code> value of the [=native origin=] must be [=rounding|rounded=] sufficiently to prevent fingerprinting of lower-order bits; rounding to the nearest 1cm is suggested.
+  - If the [=viewer=]'s offset from the floor is determined with a non-default emulated value, the {{XRRigidTransform/position}}'s {{DOMPointReadOnly/y}} value of poses returned relative to the reference space must be [=rounding|rounded=] sufficiently to prevent fingerprinting of lower-order bits; rounding to the nearest 1cm is suggested.
   - All poses returned relative to a {{XRReferenceSpaceType/"local-floor"}} reference space MUST be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=]; the suggested distance is 15 meters in all directions.
 
 #### local #### {#local-privacy}
 On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting.
 
-If {{XRReferenceSpaceType/"local"}} is included in an {{XRSession}}'s [=requested features=] and the {{XRSession}}'s [=XRSession/mode=] is {{XRSessionMode/"inline"}}, the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
+If {{XRReferenceSpaceType/"local"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
+
+Note: {{XRReferenceSpaceType/"local"}} is always included in the [=requested features=] of {{XRSessionMode/"immersive-vr"}} and {{XRSessionMode/"immersive-ar"}} sessions as a [=default feature=], as as such {{XRSessionMode/"immersive-vr"}} or {{XRSessionMode/"immersive-ar"}} sessions always need to obtain [=explicit consent=] or [=implicit consent=] to use {{XRReferenceSpaceType/"local"}} reference spaces.
 
 Any {{XRReferenceSpaceType/"local"}} reference spaces MUST respect the following restrictions:
   - All poses returned relative to a {{XRReferenceSpaceType/"local"}} reference space MUST be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=]; the suggested distance is 15 meters in all directions.
--- a/index.bs
+++ b/index.bs
@@ -444,6 +444,37 @@ Some {{XRSessionMode}}s enable certain [=feature names=] as [=optional features=
 
 The combined list of [=feature names=] given by the {{XRSessionInit/requiredFeatures}}, {{XRSessionInit/optionalFeatures}}, and [=default features=] are collectively considered the <dfn>requested features</dfn> for an {{XRSession}}.
 
+Some [=feature names=], when present in the [=requested features=] list, require that [=user intent=] to use the feature is well understood, via either [=explicit consent=] or [=implicit consent=]. The following table describes which features <dfn>require consent</dfn> prior to being enabled:
+
+<table class="tg">
+  <thead>
+    <tr>
+      <th>Feature</th>
+      <th>Criteria</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>{{XRReferenceSpaceType/"local"}}</td>
+      <td>Always requires consent</td>
+    </tr>
+    <tr>
+      <td>{{XRReferenceSpaceType/"local-floor"}}</td>
+      <td>Always requires consent</td>
+    </tr>
+    <tr>
+      <td>{{XRReferenceSpaceType/"bounded-floor"}}</td>
+      <td>Always requires consent</td>
+    </tr>
+    <tr>
+      <td>{{XRReferenceSpaceType/"unbounded"}}</td>
+      <td>Always requires consent</td>
+    </tr>
+  </tbody>
+</table>
+
+Note: {{XRReferenceSpaceType/"local"}} is always included in the [=requested features=] of {{XRSessionMode/"immersive-vr"}} and {{XRSessionMode/"immersive-ar"}} sessions as a [=default feature=], and as such {{XRSessionMode/"immersive-vr"}} or {{XRSessionMode/"immersive-ar"}} sessions always need to obtain [=explicit consent=] or [=implicit consent=].
+
 <div class="algorithm" data-algorithm="resolve-features">
 
 To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optionalFeatures| for an {{XRSession}} |session|, the user agent MUST run the following steps:
@@ -2259,10 +2290,20 @@ If the relationship between {{XRView}}s could uniquely identify the [=/XR device
 Note: Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] during session creation, prior to reporting any {{XRView}} data.
 
 ### Reference spaces ### {#protect-reference-spaces}
-Various reference space types have restrictions placed on their creation to ensure [=sensitive information=] is exposed safely.
+Depending on the reference spaces used, several different types of [=sensitive information=] may be exposed to the application.
 
-For any {{XRReferenceSpaceType}} included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure the document is allowed to use all policy-controlled features associated with the sensor types used to track the [=native origin=] of the {{XRReferenceSpaceType}} or the feature MUST be rejected.
+ - On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting.
 
+ - On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local-floor"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting. In addition, because the {{XRReferenceSpaceType/"local-floor"}} reference spaces provide an established floor level, it may be possible for a site to infer the user's height, allowing user profiling and fingerprinting.
+
+ - {{XRReferenceSpaceType/"bounded-floor"}} reference spaces, when sufficiently constrained in size, do not enable developers to determine geographic location. However, because the floor level is established and users are able to walk around, it may be possible for a site to infer the user's height or perform gait analysis, allowing user profiling and fingerprinting. In addition, it may be possible perform fingerprinting using the bounds reported by a bounded reference space.
+
+ - {{XRReferenceSpaceType/"unbounded"}} reference spaces reveal the largest amount of spatial data and may result in user profiling and fingerprinting. For example, this data may enable determining user's specific geographic location or to perform gait analysis.
+
+As a result the various reference space types have restrictions placed on their creation to ensure the [=sensitive information=] expose is handled safely:
+
+Most reference spaces require that [=user intent=] to use the reference space is well understood, either via [=explicit consent=] or [=implicit consent=]. See the [=require consent|features which require consent=] table for details.
+ 
 Any group of {{XRReferenceSpaceType/"local"}}, {{XRReferenceSpaceType/"local-floor"}}, and {{XRReferenceSpaceType/"bounded-floor"}} reference spaces that are capable of being related to one another must share a common [=native origin=]; This restriction does not apply when the creation of {{XRReferenceSpaceType/"unbounded"}} reference spaces has not been restricted.
 
 <div class="algorithm" data-algorithm="poses-limited">
@@ -2278,28 +2319,6 @@ To determine if <dfn>poses must be limited</dfn> between two spaces, |space| and
 Note: Is is suggested that poses reported relative to a {{XRReferenceSpaceType/"local"}} or {{XRReferenceSpaceType/"local-floor"}} reference space be [=limiting|limited=] to a distance of 15 meters from the {{XRReferenceSpace}}'s [=native origin=].
 
 Note: Is is suggested that poses reported relative to a {{XRBoundedReferenceSpace}} be [=limiting|limited=] to a distance of 1 meter outside the {{XRBoundedReferenceSpace}}'s [=native bounds geometry=].
-
-#### unbounded #### {#unbounded-privacy}
-{{XRReferenceSpaceType/"unbounded"}} reference spaces reveal the largest amount of spatial data and may result in user profiling and fingerprinting. For example, this data may enable determining user's specific geographic location or to perform gait analysis.
-
-If {{XRReferenceSpaceType/"unbounded"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"unbounded"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
-
-#### bounded-floor #### {#bounded-floor-privacy}
-{{XRReferenceSpaceType/"bounded-floor"}} reference spaces, when sufficiently constrained in size, do not enable developers to determine geographic location. However, because the floor level is established and users are able to walk around, it may be possible for a site to infer the user's height or perform gait analysis, allowing user profiling and fingerprinting. In addition, it may be possible perform fingerprinting using the bounds reported by a bounded reference space.
-
-If {{XRReferenceSpaceType/"bounded-floor"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"bounded-floor"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
-
-#### local-floor #### {#local-floor-privacy}
-On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local-floor"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting. In addition, because the {{XRReferenceSpaceType/"local-floor"}} reference spaces provide an established floor level, it may be possible for a site to infer the user's height, allowing user profiling and fingerprinting.
-
-If {{XRReferenceSpaceType/"local-floor"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local-floor"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
-
-#### local #### {#local-privacy}
-On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting.
-
-If {{XRReferenceSpaceType/"local"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
-
-Note: {{XRReferenceSpaceType/"local"}} is always included in the [=requested features=] of {{XRSessionMode/"immersive-vr"}} and {{XRSessionMode/"immersive-ar"}} sessions as a [=default feature=], and as such {{XRSessionMode/"immersive-vr"}} or {{XRSessionMode/"immersive-ar"}} sessions always need to obtain [=explicit consent=] or [=implicit consent=].
 
 <section class="unstable">
 Gaze Tracking {#gazetracking-security}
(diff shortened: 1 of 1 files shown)


toji:
@johnpallett: Can't add you as a reviewer directly, but could you take a look at this?

index.bs
```diff
@@ -2216,6 +2218,46 @@ The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion
...
+
+If {{XRReferenceSpaceType/"bounded-floor"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"bounded-floor"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
+
+Any {{XRReferenceSpaceType/"bounded-floor"}} reference spaces MUST respect the following restrictions:
+  - Each point in the {{XRBoundedReferenceSpace/boundsGeometry}} must be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=]; the suggested distance is 15 meters in all directions.
+  - Each point in the {{XRBoundedReferenceSpace/boundsGeometry}} must be [=rounding|rounded=] sufficiently to prevent fingerprinting. Rounding to the nearest 5cm is suggested. For user's safety, rounded points values MUST NOT fall outside the original bounds.
+  - The <code>Y</code> value of the [=native origin=] must be [=rounding|rounded=] sufficiently to prevent fingerprinting of lower-order bits; rounding to the nearest 1cm is suggested.
```
johnpallett: I might have missed something when reviewing the privacy & security explainer here... apologies. IIUC correctly the Y value of the native origin will be 0 at the floor height, so it doesn't need to be rounded.

I believe the value that needs rounding is an _emulated viewer height_, if it's set by the user to a non-default value. That's the only situation where it would be a unique, reproducible number (for the very first pose) between sessions.

@NellWaliczek is that your understanding as well?

index.bs
```diff
@@ -2216,6 +2218,46 @@ The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion
...
+
+#### local-floor #### {#local-floor-privacy}
+On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local-floor"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting. In addition, because the {{XRReferenceSpaceType/"local-floor"}} reference spaces provide an established floor level, it may be possible for a site to infer the user's height, allowing user profiling and fingerprinting.
+
+If {{XRReferenceSpaceType/"local-floor"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local-floor"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
+
+Any {{XRReferenceSpaceType/"local-floor"}} reference spaces MUST respect the following restrictions:
+  - If the floor level is based on sensor data or is set to a non-default emulated value, the <code>Y</code> value of the [=native origin=] must be [=rounding|rounded=] sufficiently to prevent fingerprinting of lower-order bits; rounding to the nearest 1cm is suggested.
```
johnpallett: See note above about this potentially being the emulated viewer height, not the native origin Y value

index.bs
```diff
@@ -2216,6 +2218,46 @@ The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion
...
+  - All poses returned relative to a {{XRReferenceSpaceType/"local-floor"}} reference space MUST be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=]; the suggested distance is 15 meters in all directions.
+
+#### local #### {#local-privacy}
+On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting.
+
+If {{XRReferenceSpaceType/"local"}} is included in an {{XRSession}}'s [=requested features=] and the {{XRSession}}'s [=XRSession/mode=] is {{XRSessionMode/"inline"}}, the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
```
johnpallett: Note: User intent is also required for 'local' for immersive sessions; the difference is that the feature is added implicitly. That doesn't appear to be covered here or in the session creation section.

I wonder if that needs an extra step in the session creation algorithm above, something like:

> When creating any immersive session on a device that supports 6DOF tracking, the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.

index.bs
```diff
@@ -2216,6 +2218,46 @@ The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion
 
 If the relationship between {{XRView}}s could uniquely identify the [=/XR device=], then the user agent MUST anonymized the {{XRView}} data to prevent fingerprinting. The method of anonymization is at the discretion of the user agent. Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] prior to reporting any {{XRView}} data.
 
+### Reference spaces ### {#protect-reference-spaces}
+Various reference space types have restrictions places on their use to prevent the unintentional exposure of [=sensitive information=].
+
+For any {{XRReferenceSpaceType}} included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure the document is allowed to use all policy-controlled features associated with the sensor types used to track the [=native origin=] of the {{XRReferenceSpaceType}} or the feature MUST be rejected.
```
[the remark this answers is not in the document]
toji: The specific text from the explainer I'm pulling from here is:

> In response the UA must ensure that:
> * The document is allowed to use all the policy-controlled features associated with the sensor types used to track the native origin of an unbounded reference space

Which is much closer to the text in question. The text you point out is actually captured more explicitly in #761 (Specifically [this line](https://github.com/immersive-web/webxr/pull/761/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR2262)) which reads:

> In addition to the <code>"xr"</code> feature policy, feature policies for <dfn>underlying sensors</dfn> must also be respected if a site could isolate and extract sensor data that would otherwise be blocked by those feature policies. 

That's not to say that we can't or shouldn't make this text clearer if needed, I just want to make sure we're on the same page regarding where this text was pulled from.

index.bs
toji: That meshes with what I recall from previous conversations as well, thanks for calling it out. I'll make the change now, and allow Nell to intervene if she has concerns.

index.bs
toji: Adjusted this text to account for that and added a non-normative note reminding the reader that this is always included as a default feature of immersive sessions.

toji:
Thanks for the feedback, John! Left some responses and made some tweaks.

> Are we comfortable that this approach gives sufficient direction to implementers? ...

This has been my primary source of angst around this text. (And #761, to a slightly lesser degree.) It comes down to a couple of things, I think:

 - There's some of these privacy principles that must be validated at runtime (are you visible and focused?) and those slot well into spec algorithms. Then there's other limits that feel more like intrinsic properties of the implementation, and as such it sounds extremely weird to ask for it to be validated in an algorithm. (ie: "`If the pose will not be sufficiently rounded to prevent fingerprinting return null and abort the rest of these steps.`" Um, no? Because if you're not doing that then it just mean's you're not a valid implementation?) It's not as clear what the appropriate mechanism is to surface those kind of behaviors is.
 - Some of this content could/should be moved up in the document to be closer to the definition of the mechanism in question. Like, we could move the `bounded-floor` bounds limitations up into the section dedicated to discussing that interface easily, and it would likely read well. I worry that doing so may dilute the overall privacy message, though? When you have all of the privacy-centric dicsussion in one place like this it helps sell that it's important and gives a better mental picture of how all the limits relate rather than just being a jumble of somewhat confusing lines scattered throughout the doc?

If you have any thoughts on how to ease that tension, I'm happy to hear it!

index.bs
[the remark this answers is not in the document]
johnpallett: Good clarification, thanks... my questions still stand (though I should probably have asked them with relation to the explainer instead of now).

For me it boils down to whether the phrase "tracking the native origin" sufficiently covers the possibility that gyro data could be used for device orientation, including the viewer reference space. WDYT?

index.bs
```diff
@@ -2247,13 +2246,15 @@ On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local-floor"
...
   - All poses returned relative to a {{XRReferenceSpaceType/"local-floor"}} reference space MUST be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=]; the suggested distance is 15 meters in all directions.
 
 #### local #### {#local-privacy}
 On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting.
 
-If {{XRReferenceSpaceType/"local"}} is included in an {{XRSession}}'s [=requested features=] and the {{XRSession}}'s [=XRSession/mode=] is {{XRSessionMode/"inline"}}, the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
+If {{XRReferenceSpaceType/"local"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
```
johnpallett: Thanks Brandon, this is clearer. A different observation now...

Line 422 (the section on implicit feature additions) says that features (i.e. "local") are only added if ""if the feature does not require a signal of [=user intent=]"

However, this line now says that "local" requires a signal of user intent which might be in conflict with line 422. I believe that's why this was restricted to "inline" originally.

One way of addressing this apparent conflict is to add text here saying something like "With immersive sessions it is assumed that this consent is obtained during session creation, as all immersive sessions must allow local tracking." WDYT?

index.bs
johnpallett: > If you have any thoughts on how to ease that tension, I'm happy to hear it!

index.bs
```diff
@@ -2216,6 +2218,47 @@ The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion
 
 If the relationship between {{XRView}}s could uniquely identify the [=/XR device=], then the user agent MUST anonymized the {{XRView}} data to prevent fingerprinting. The method of anonymization is at the discretion of the user agent. Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] prior to reporting any {{XRView}} data.
 
+### Reference spaces ### {#protect-reference-spaces}
+Various reference space types have restrictions places on their use to prevent the unintentional exposure of [=sensitive information=].
+
+For any {{XRReferenceSpaceType}} included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure the document is allowed to use all policy-controlled features associated with the sensor types used to track the [=native origin=] of the {{XRReferenceSpaceType}} or the feature MUST be rejected.
```
johnpallett: Per comment in #761 I wonder if this text needs to be adjusted slightly:

```suggestion
For any {{XRReferenceSpaceType}} included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure the document is not explicitly blocked from using any of the policy-controlled features associated with the sensor types used to track the [=native origin=] of the {{XRReferenceSpaceType}}, or the feature MUST be rejected.
```

@ddorwin WDYT?

index.bs
```diff
@@ -2234,6 +2234,47 @@ The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion
...
+
+#### bounded-floor #### {#bounded-floor-privacy}
+{{XRReferenceSpaceType/"bounded-floor"}} reference spaces, when sufficiently constrained in size, do not enable developers to determine geographic location. However, because the floor level is established and users are able to walk around, it may be possible for a site to infer the user's height or perform gait analysis, allowing user profiling and fingerprinting. In addition, it may be possible perform fingerprinting using the bounds reported by a bounded reference space.
+
+If {{XRReferenceSpaceType/"bounded-floor"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"bounded-floor"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
+
+Any {{XRReferenceSpaceType/"bounded-floor"}} reference spaces MUST respect the following restrictions:
+  - Each point in the {{XRBoundedReferenceSpace/boundsGeometry}} must be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=]; the suggested distance is 15 meters in all directions.
```
NellWaliczek: These two points feel like they should be in an algorithm for defining the bounds which should maybe be called from the algorithm for computing the effective bounds.

toji:
Moved the reference space requirement bullet points around to the appropriate contexts and/or made them part of an algorthim. This feels like it "dilutes" the privacy-centric portion of the doc, because now some privacy driven limits are scattered throughout the doc, but it is probably easier for implementers this way and ultimately that's what counts.

index.bs
```diff
@@ -969,6 +969,7 @@ To <dfn>populate the pose</dfn> of an {{XRSpace}} |space| in an {{XRSpace}} |bas
   1. If |space|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |baseSpace|'s [=XRSpace/session=] does not equal |session|, throw an {{InvalidStateError}} and abort these steps.
   1. Check if [=poses may be reported=] and, if not, throw a {{SecurityError}} and abort these steps.
+  1. If [=poses must be limited=] between |space| and |baseSpace|, return <code>null</code>.
```
NellWaliczek: I suspect we should follow the tracking loss behavior here rather than just returning null? @thetuvix

index.bs
```diff
@@ -1027,7 +1028,9 @@ An {{XRReferenceSpace}} is most frequently obtained by calling {{XRSession/reque
...
+  - Passing a type of <dfn enum-value for="XRReferenceSpaceType">local-floor</dfn> creates an {{XRReferenceSpace}} instance. It represents a tracking space with a [=native origin=] at the floor in a safe position for the user to stand. The <code>y</code> axis equals <code>0</code> at floor level, with the <code>x</code> and <code>z</code> position and orientation initialized based on the conventions of the underlying platform. If the floor level isn't known it MUST be estimated. If the estimated floor level is determined with a non-default value, it MUST be [=rounding|rounded=] sufficiently to prevent fingerprinting. When using this reference space the user is not expected to move beyond their initial position much, if at all, and tracking is optimized for that purpose. For devices with [=6DoF=] tracking, {{local-floor}} reference spaces should emphasize keeping the origin stable relative to the user's environment.
+
+    Note: It the floor level of a {{XRReferenceSpaceType/"local-floor"}} reference space is adjusted to prevent fingerprinting, [=rounding|rounded=] to the nearest 1cm is suggested.
```
NellWaliczek: nit: It the floor -> If the floor

NellWaliczek: This restriction should also be shared with `bounded-floor`.  Given that, should we add an algorithm for determining the floor level and put this text there?

index.bs
```diff
@@ -2246,6 +2257,49 @@ If the relationship between {{XRView}}s could uniquely identify the [=/XR device
...
+### Reference spaces ### {#protect-reference-spaces}
+Various reference space types have restrictions placed on their creation to ensure [=sensitive information=] is exposed safely.
+
+For any {{XRReferenceSpaceType}} included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure the document is allowed to use all policy-controlled features associated with the sensor types used to track the [=native origin=] of the {{XRReferenceSpaceType}} or the feature MUST be rejected.
+
+Any group of {{XRReferenceSpaceType/"local"}}, {{XRReferenceSpaceType/"local-floor"}}, and {{XRReferenceSpaceType/"bounded-floor"}} reference spaces that are capable of being related to one another must share a common [=native origin=]; This restriction does not apply when the creation of {{XRReferenceSpaceType/"unbounded"}} reference spaces has not been restricted.
+
+<div class="algorithm" data-algorithm="poses-limited">
+
+To determine if <dfn>poses must be limited</dfn> between two spaces, |space| and |baseSpace|, the user agent MUST run the following steps:
```
NellWaliczek: Gah I hate to be pedantic, but the way this is written means the `local` and `local-floor` restrictions will take precedence even when the other parameter is a `bounded-floor`.  Is that intentional?

index.bs
```diff
@@ -2246,6 +2257,49 @@ If the relationship between {{XRView}}s could uniquely identify the [=/XR device
...
+
+#### local #### {#local-privacy}
+On devices which support [=6DoF=] tracking, {{XRReferenceSpaceType/"local"}} reference spaces may be used to perform gait analysis, allowing user profiling and fingerprinting.
+
+If {{XRReferenceSpaceType/"local"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"local"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
+
+Note: {{XRReferenceSpaceType/"local"}} is always included in the [=requested features=] of {{XRSessionMode/"immersive-vr"}} and {{XRSessionMode/"immersive-ar"}} sessions as a [=default feature=], as as such {{XRSessionMode/"immersive-vr"}} or {{XRSessionMode/"immersive-ar"}} sessions always need to obtain [=explicit consent=] or [=implicit consent=] to use {{XRReferenceSpaceType/"local"}} reference spaces.
```
NellWaliczek: nit: as as such  -> and as such

NellWaliczek: > to use {{XRReferenceSpaceType/"local"}} reference spaces

I think we can strike this part of the sentence

index.bs
```diff
@@ -2246,6 +2257,49 @@ If the relationship between {{XRView}}s could uniquely identify the [=/XR device
...
+
 <section class="unstable">
 Gaze Tracking {#gazetracking-security}
```
NellWaliczek: Are we going to have a separate PR to remove the privacy sections that we've replaced with all this new text?

index.bs
```diff
@@ -2234,6 +2234,47 @@ The primary difference between {{XRViewerPose}} and {{XRPose}} is the inclusion
...
+
+Any group of {{XRReferenceSpaceType/"local"}}, {{XRReferenceSpaceType/"local-floor"}}, and {{XRReferenceSpaceType/"bounded-floor"}} reference spaces that are capable of being related to one another must share a common [=native origin=]; This restriction does not apply when {{XRReferenceSpaceType/"unbounded"}} reference spaces are also able to be created.
+
+#### unbounded #### {#unbounded-privacy}
+{{XRReferenceSpaceType/"unbounded"}} reference spaces reveal the largest amount of spatial data and may result in user profiling and fingerprinting. For example, this data may enable determining user's specific geographic location or to perform gait analysis.
+
+If {{XRReferenceSpaceType/"unbounded"}} is included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure [=user intent=] to allow {{XRReferenceSpaceType/"unbounded"}} tracking is well understood, either via [=explicit consent=] or [=implicit consent=], or the feature MUST be rejected.
```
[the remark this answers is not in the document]
NellWaliczek: Or maybe putting a table  near the algorithm for determining consent?

index.bs
```diff
@@ -2246,6 +2257,49 @@ If the relationship between {{XRView}}s could uniquely identify the [=/XR device
 
 Note: Furthermore, if the relationship between {{XRView}}s is affected by a user-configured interpupillary distance, then it is strongly recommended that the user agent require [=explicit consent=] during session creation, prior to reporting any {{XRView}} data.
 
+### Reference spaces ### {#protect-reference-spaces}
+Various reference space types have restrictions placed on their creation to ensure [=sensitive information=] is exposed safely.
+
+For any {{XRReferenceSpaceType}} included in an {{XRSession}}'s [=requested features=] the user agent MUST ensure the document is allowed to use all policy-controlled features associated with the sensor types used to track the [=native origin=] of the {{XRReferenceSpaceType}} or the feature MUST be rejected.
+
+Any group of {{XRReferenceSpaceType/"local"}}, {{XRReferenceSpaceType/"local-floor"}}, and {{XRReferenceSpaceType/"bounded-floor"}} reference spaces that are capable of being related to one another must share a common [=native origin=]; This restriction does not apply when the creation of {{XRReferenceSpaceType/"unbounded"}} reference spaces has not been restricted.
```
NellWaliczek: Any chance we can make this one a bit more algorithmic too?

index.bs
```diff
@@ -1108,6 +1111,14 @@ Note: Other XR platforms sometimes refer to the type of tracking offered by a {{
...
 
+Each point in the [=native bounds geometry=] MUST be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=].
+
+Note: It is suggested that points of the [=native bounds geometry=] be [=limiting|limited=] to 15 meters from the [=XRSpace/native origin=] in all directions.
+
+Each point in the [=native bounds geometry=] MUST also be [=rounding|rounded=] sufficiently to prevent fingerprinting.  For user's safety, rounded points values MUST NOT fall outside the bounds reported by the platform.
+
+Note: It is suggested that points of the [=native bounds geometry=] be [=rounding|rounded=] to the nearest 5cm.
```
NellWaliczek: Nit: is this rounding or quantization?

index.bs
```diff
@@ -1108,6 +1111,14 @@ Note: Other XR platforms sometimes refer to the type of tracking offered by a {{
...
 
+Each point in the [=native bounds geometry=] MUST be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=].
+
+Note: It is suggested that points of the [=native bounds geometry=] be [=limiting|limited=] to 15 meters from the [=XRSpace/native origin=] in all directions.
+
+Each point in the [=native bounds geometry=] MUST also be [=rounding|rounded=] sufficiently to prevent fingerprinting.  For user's safety, rounded points values MUST NOT fall outside the bounds reported by the platform.
```
NellWaliczek: @johnpallett It just occurred to me... is rounding a sufficient mitigation here or should we also be applying fuzzing?

index.bs
toji: `bounded-floor` will never have an emulated floor height by definition, and while real heights may be used in profiling they're considered too noisy for fingerprinting.

index.bs
toji: It's probably not a big deal either way, but the `bounded-floor` limits have the potential to be more precise (and, in the most extreme cases, larger if the PRs suggested limits were followed). So I swapped the order.

index.bs
toji: Updated this to make hitting the limit act more like tracking loss.

toji:
Thanks for the review! Made a partial update to hit some of the simpler changes. I'll get the remaining change on or before Monday.

index.bs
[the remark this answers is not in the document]
toji: Moved to a table up higher in the doc, and restructured the lower section as a result. Reads better now.

index.bs
toji: Yeah, a separate PR sounds best for this.

index.bs
toji: Take a look at where this falls in the flow of things now and see if it reads better? If not I can try to agorithimify it.

toji:
Okay, everything should be addressed one way or another now. PTAL!

index.bs
```diff
@@ -444,6 +444,37 @@ Some {{XRSessionMode}}s enable certain [=feature names=] as [=optional features=
...
+  <thead>
+    <tr>
+      <th>Feature</th>
+      <th>Criteria</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>{{XRReferenceSpaceType/"local"}}</td>
+      <td>Always requires consent</td>
```
NellWaliczek: Shouldn't this be for inline sessions only?

index.bs
```diff
@@ -1108,6 +1142,14 @@ Note: Other XR platforms sometimes refer to the type of tracking offered by a {{
...
 
+Each point in the [=native bounds geometry=] MUST be [=limiting|limited=] to a reasonable distance from the reference space's [=native origin=].
+
+Note: It is suggested that points of the [=native bounds geometry=] be [=limiting|limited=] to 15 meters from the [=XRSpace/native origin=] in all directions.
+
+Each point in the [=native bounds geometry=] MUST also be [=rounding|rounded=] sufficiently to prevent fingerprinting.  For user's safety, rounded points values MUST NOT fall outside the bounds reported by the platform.
```
NellWaliczek: > [=rounding|rounded=]

This is now out of sync with the non-normative note below.

index.bs
```diff
@@ -2247,6 +2289,37 @@ If the relationship between {{XRView}}s could uniquely identify the [=/XR device
...
+
+ - {{XRReferenceSpaceType/"unbounded"}} reference spaces reveal the largest amount of spatial data and may result in user profiling and fingerprinting. For example, this data may enable determining user's specific geographic location or to perform gait analysis.
+
+As a result the various reference space types have restrictions placed on their creation to ensure the [=sensitive information=] expose is handled safely:
+
+Most reference spaces require that [=user intent=] to use the reference space is well understood, either via [=explicit consent=] or [=implicit consent=]. See the [=require consent|features which require consent=] table for details.
+ 
+Any group of {{XRReferenceSpaceType/"local"}}, {{XRReferenceSpaceType/"local-floor"}}, and {{XRReferenceSpaceType/"bounded-floor"}} reference spaces that are capable of being related to one another must share a common [=native origin=]; This restriction does not apply when the creation of {{XRReferenceSpaceType/"unbounded"}} reference spaces has not been restricted.
```
NellWaliczek: > another must share a common

MUST?

NellWaliczek: >This restriction does not apply when the creation of {{XRReferenceSpaceType/"unbounded"}} reference spaces has not been restricted.

I know this was my suggestion, but ugh this double negative sentence is a bit hard to parse... How about:

> This restriction only applies when the creation of {{XRReferenceSpaceType/"unbounded"}} reference spaces has been restricted.

index.bs
NellWaliczek: Before merging, can you file that?  Are we intending to do it before VR complete (*looks skeptically at the calendar*)

index.bs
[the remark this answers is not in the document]
NellWaliczek: Yeah that works

index.bs
NellWaliczek: Maybe I'm remember this wrong, but weren't we considering bounded-floor references spaces on devices like the orginal hololens to have an emulated height?

index.bs
toji: #776. And while it would be nice, I don't think this is necessary for VR complete because it should mostly entail removing redundant information.

index.bs
toji: I don't recall that? And I'm highly skeptical of a system that would use an actual emulated height (as in: an arbitrary value with no user configuration) for something like this.

toji:
Nits addressed. Thanks!

index.bs
```diff
@@ -1146,7 +1146,7 @@ Each point in the [=native bounds geometry=] MUST be [=limiting|limited=] to a r
 
 Note: It is suggested that points of the [=native bounds geometry=] be [=limiting|limited=] to 15 meters from the [=XRSpace/native origin=] in all directions.
 
-Each point in the [=native bounds geometry=] MUST also be [=rounding|rounded=] sufficiently to prevent fingerprinting.  For user's safety, rounded points values MUST NOT fall outside the bounds reported by the platform.
+Each point in the [=native bounds geometry=] MUST also be [=quantization|quantized=] sufficiently to prevent fingerprinting.  For user's safety, rounded points values MUST NOT fall outside the bounds reported by the platform.
```
NellWaliczek: "rounded points" -> "quantized points"

index.bs
NellWaliczek: My bad, then!

index.bs
NellWaliczek: Ok, let's just make sure we didn't preemptively remove the 'unstable' banner from around it before we cut the next working draft.

index.bs
NellWaliczek: ping @johnpallett.  If you think this is a concern, please file an issue!

johnpallett: No blocking concerns here. Thanks @NellWaliczek

PR #767, opened 2019-07-17
Block mid-session consent requests

### LINKED ISSUE #751: For now, block mid-session consent requests for non-xr features
We're still figuring out what we'd like to do about mid-session requests for user consent for features that have nothing directly to do with XR (e.g. usb, geolocation, etc) in #726.  Given that issue isn't a "VR Complete" issue, it might be prudent to temporarily add spec text which blocks allowing any form of consent from being requested mid-session.  Perhaps we can allow for the exception of if the UA has a Trusted Immersive UI solution, even though the actual requirements for such a solution have not been finalized (see #718)

toji:
/fixes #751

Details that consent dialogs may not be displayed while an immersive
session is active, and explains that this is a short-term restriction.

I chose to go with terminating the session in this case rather than
suppressing the consent dialog altogether, since it provides very clear
feedback about the error case and doesn't require us to make up new
behavior for the permissions in question.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2151,7 +2151,14 @@ It is often useful to get <dfn>explicit consent</dfn> from the user before expos
 ### Duration of consent ### {#consent-duration}
 It is recommended that once [=explicit consent=] is granted for a specific [=/origin=] that this consent persist until the [=/browsing context=] has ended. User agents may choose to lengthen or shorten this consent duration based upon implicit or explicit signals of [=user intent=], but implementations are advised to exercise caution when deviating from this recommendation, particularly when relying on implicit signals. For example, it may be appropriate for a web application installed with the express intent of running immersive content to persist the user's consent, but not for an installed web application where immersive content is a secondary feature.
 
-Regardless of how long the user agent chooses to persist the user's consent, [=sensitive information=] MUST only be exposed by an {{XRSession}} which has not [=ended=]. 
+Regardless of how long the user agent chooses to persist the user's consent, [=sensitive information=] MUST only be exposed by an {{XRSession}} which has not [=ended=].
+
+Mid-session consent {#mid-session-consent}
+-------------------
+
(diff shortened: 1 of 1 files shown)


toji:
Also pinging @blairmacintyre, though I couldn't add you as a reviewer since you haven't been explicitly added as a collaborator for this repo. (We can change that if you'd like!)

index.bs
```diff
@@ -2151,7 +2151,14 @@ It is often useful to get <dfn>explicit consent</dfn> from the user before expos
...
 
-Regardless of how long the user agent chooses to persist the user's consent, [=sensitive information=] MUST only be exposed by an {{XRSession}} which has not [=ended=]. 
+Regardless of how long the user agent chooses to persist the user's consent, [=sensitive information=] MUST only be exposed by an {{XRSession}} which has not [=ended=].
+
+Mid-session consent {#mid-session-consent}
+-------------------
+
+There are multiple non-XR APIs which cause user agents to request [=explicit consent=] to use a feature. If the user agent will request the user's consent while there is an [=active immersive session=], the user agent MUST [=shut down the session=] prior to displaying the consent request to the user. If the user's consent for the feature had been granted prior to the [=active immersive session=] being created the session does not need to be terminated. 
+
+Note: This limitation is to ensure that there is behavioral parity between all user agents until consensus is reached about how to securely display these types of prompts. It is not expected to be a long term requirement.
```
blairmacintyre: Looks good to me!

blairmacintyre:
Looks good.  (And I'm happy to be a collaborator)

index.bs
johnpallett: LGTM. I particularly appreciate that the motivation is spelled out. Thank you!

johnpallett: PS: Non-blocking suggestion to avoid prescribing a solution:

```suggestion
Note: This limitation is to ensure that there is behavioral parity between all user agents until consensus is reached about how user agents should manage mid-session [=explicit consent=]. It is not expected to be a long term requirement. 
```

PR #782, opened 2019-07-30
Removes references to XRInputSource.gamepad from explainer and index.bs

NellWaliczek:
This is the first of two PRs to execute the decision to split the WebXR API into separate modules.  This PR handles removing the content moving to the new WebXR Gamepads Module. Removed content can be found in the partner PR in the new module repo https://github.com/immersive-web/webxr-gamepads-module/pull/1


index.bs
```diff
@@ -1526,31 +1511,25 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
...
+Profiles are given in descending order of specificity. Any [=input profile name=]s given after the first entry in the list should provide fallback values that represent alternative representations of the device. This may include a more generic or prior version of the device, a more widely recognized device that is sufficently similar, or a broad description of the device type (such as "touchpad-wand"). If multiple profiles are given, the layouts they describe must all represent a superset or subset of every other profile in the list.
 
 If the {{XRSession}}'s [=mode=] is {{XRSessionMode/"inline"}}, {{XRInputSource/profiles}} MUST be an empty list.
 
 If the input device cannot be reliably identified, or the user agent wishes to mask the input device being used, it MAY choose to only report generic [=input profile name=]s or an empty list.
 
 <p class="note">
```
NellWaliczek: Should this note be moved to the xr-gamepads-module instead?

toji: I think these should stay here, because they describe the expected behavior of the profiles moreso than the gamepad behavior. Your modifications feel appropriate in that regard.

index.bs
```diff
@@ -1526,31 +1511,25 @@ The <dfn attribute for="XRInputSource">gripSpace</dfn> attribute is an {{XRSpace
...
-
- - Multiple buttons
- - A button which reports analog values
- - An axis
-
-Note: {{XRInputSource}}s in an {{XRSession}}'s {{XRSession/inputSources}} array are "live". As such values within them, such as the {{gamepad}} button and axis state, are updated in-place. This means that it doesn't work to save a reference to an {{XRInputSource}}'s {{gamepad}} on one frame and compare it to the same {{XRInputSource}}'s {{gamepad}} from a subsequent frame to test for state changes, because they will be the same object. Therefore developers that wish to compare input state from frame to frame should copy the state in question.
+Note: {{XRInputSource}}s in an {{XRSession}}'s {{XRSession/inputSources}} array are "live". As such values within them, such as the {{gripSpace}}, are updated in-place. This means that it doesn't work to save a reference to an {{XRInputSource}}'s {{gripSpace}} on one frame and compare it to the same {{XRInputSource}}'s {{gripSpace}} from a subsequent frame to test for state changes, because they will be the same object. Therefore developers that wish to compare input state from frame to frame should copy the state in question.
```
toji: This is slightly less effective now, because there's not really an observable state on the `gripSpace` until you query it's pose, which isn't the same type of "live" behavior. I'm wondering if this whole paragraph should just move to the gamepads module?

input-explainer.md
```diff
@@ -226,8 +226,6 @@ For `tracked-pointer` input sources, it is often appropriate for the application
 
 
 #### Choosing renderable models
-The majority of `tracked-pointer` input sources will have a non-null `gamepad` attribute on the `XRInputSource` object. The `Gamepad`'s `id` is used to determine what should be rendered if the app intends to visualize the input source itself, rather than an alternative virtual object. (See the section on [Button and Axis State](#button-and-axis-state) for more details.)
```
toji: Ewe. This was just flat out wrong, wasn't it? (The use of the Gamepad.id for models)

index.bs
NellWaliczek: yeah I was going back and forth on that because the actual behavior isn't different in the gamepads module.  I tried to split the difference by keeping the original text in that repo and making a small modification here.  Perhaps if I just remove the example here?

input-explainer.md
NellWaliczek: yeeeeeep.

index.bs
toji: Yeah, that seems best. You've got an explanation of the "liveness" of the gamepad attrib in that module anyway.

NellWaliczek: If this modification doesn't work for you, unresolve and I'll take another stab at it.

index.bs
NellWaliczek: I'm inclined to use the, soon to be renamed, xr-gamepad-mappings repo to be the authoritative list of profile names.  If so, we should probably remove this text at that point.  I'll leave it for now.

toji:
Latest updates still LGTM. Thanks!

PR #787, opened 2019-08-05
Correction of article(a/an) in index.bs

takahirox:
This PR replaces "a" with "an" for the followings in index.bs

- a environment blending mode value
- a enum
- a active boolean
- a inline session's output context
- a end event
- a inputsourceschange event


toji:
Thank you!

PR #789, opened 2019-08-06
Add IPD abbreviation to another location

ddorwin:
This might be useful for anyone searching for the term. The abbreviation was already present for the other reference to interpupillary distance.


toji:
Thanks!

Issue #792, opened 2019-08-08
[Editorial] fix on bikeshed and spell


himorin:
check was performed to c0b4743, will raise PR shortly

- bikeshed, two link error "No 'dfn' refs found for 'extend'." on lines 664 and 696 (need to add in 'anchors'?)
- in example 5, 2nd paragraph "lik so:" -> "like so:"
- in 10.1, mixing of prefered or preferred (two items, like: "The profiles attribute is a list of input profile names indicating both the prefered visual representation and behavior of the input source.")
- in 10.1, sufficently -> sufficiently at "This may include a more generic or prior version of the device, a more widely recognized device that is sufficently similar, "
- in 13.4.3, desireable -> desirable at "It is often desireable to avoid a 'hard stop' at the at the end of a range as this may cause disruptive user experiences. "

PR #793, opened 2019-08-08
editorial fix on bikeshed and spell

himorin:
for #792


toji:
Thanks you very much!

Issue #783, opened 2019-07-30
Should XRInputSource.profiles changes fire an XRInputSourcesChangeEvent?


jacobcdewitt:
The spec says "When the handedness, targetRayMode, or presence of a gripSpace or gamepad for any XR input sources change for XRSession session, the user agent MUST run the following steps". Would it also make sense to re-create the input source and fire an event when the contents of the profiles array changes?

jacobcdewitt:
It's worth noting that the profiles is a FrozenArray, which means it probably shouldn't be updated in-place by the browser. Since it also has the [SameObject] attribute, re-creating the XRInputSource seems like the only compliant solution for when the profiles array changes.

@toji , what are your thoughts on this?

toji:
Yes.

Manishearth:
https://github.com/immersive-web/webxr/pull/795

PR #795, opened 2019-08-13
Fire input source events when profiles changes

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/783

r? @toji


toji:
LGTM!

Issue #769, opened 2019-07-18
Should getViewport return null if there is no matching viewport?


asajeffrey:
The spec at https://immersive-web.github.io/webxr/#dom-xrwebgllayer-getviewport says "Let glViewport be the WebGL viewport from the list of viewports associated with view," burt doesn't say what to if there isn't one. Should there be text "If there is no such viewport, return `null`."?

toji:
So in general if you can't produce a valid viewport for a given `XRView` then you shouldn't be reporting that `XRView` in the first place. There's no practical use for a view that cannot be rendered, and as such a view without a corresponding viewport is not a thing the spec anticipates. (Note that the viewport is a layer-specific concept, and some future layers may not have need for viewports at all, or have an alternative way of associating a view with a renderable area. Regardless of approach, however, there _must_ always be a way to map a view to the thing that it needs to render into.)

That said, this DID highlight a different but related inconsistency for us while reviewing it. In the algorithm for `getViewport` we indicated that if the view was from a different `XRSession` than the layer then the function should return `null`, but elsewhere in the spec for the same class of error we throw an `InvalidStateError`, which seems like a more fitting response to what is clearly an invalid call. Posted #771 to normalize that behavior. Thanks!

asajeffrey:
Oh I see, if a device changes which views are available, then its expected to restart the session? So the list of views for a given session is immutable?

toji:
No, the list of views may change between frames and the app is expected to iterate through the list of views on a per frame basis.

But now I see where you're coming from. If the app holds a reference to a view and the view configuration switches mid-session so that the old view is no longer active then it won't have a viewport associated with it any longer, but may still be passed into this function. We should probably throw an InvalidStateError in that situation as well.

Thanks for bringing that to our attention!

Manishearth:
fixed in https://github.com/immersive-web/webxr/pull/796

PR #796, opened 2019-08-13
Forbid using stale XRViews in getViewport()

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/769

r? @toji @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1581,6 +1581,8 @@ interface XRWebGLLayer {
 
 Each {{XRWebGLLayer}} has a <dfn for="XRWebGLLayer">context</dfn> object, initially <code>null</code>, which is an instance of either a {{WebGLRenderingContext}} or a {{WebGL2RenderingContext}}.
 
+Each {{XRWebGLLayer}} has an associated <dfn for="XRWebGLLayer">session</dfn>, which is the {{XRSession}} it was created with.
+
 <div class="algorithm" data-algorithm="construct-webgl-layer">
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1649,10 +1652,11 @@ Each {{XRWebGLLayer}} MUST have a <dfn>list of viewports</dfn> which is a [=/lis
 
 <div class="algorithm" data-algorithm="get-viewport">
 
-The <dfn method for="XRWebGLLayer">getViewport(|view|)</dfn> method, when invoked, MUST run the following steps:
+The <dfn method for="XRWebGLLayer">getViewport(|view|)</dfn> method, when invoked on an {{XRSession}} |session|, MUST run the following steps:
```
toji: `getViewport()` is a method of an `XRWebGLLayer`, so saying "invoked on an XRSession" is incorrect. The `XRWebGLLayer` accepts an `XRSession` in the constructor, and that's the one that we should be checking against here, but it turns out that we're not explicitly keeping a reference to that? We do check it in [the `updateRenderState()` algorithm](https://immersive-web.github.io/webxr/#dom-xrsession-updaterenderstate), but it kinda cheats:

> If newState’s baseLayer's was created with an XRSession other than session, throw an InvalidStateError and abort these steps.

We could do the same thing here and just reference "the XRSession the layer was created with" or we could actually start keeping an explicit reference to it. I'll leave that to you.

Manishearth: Oh, for some reason i've been thinking that it's a method on XRSession all this time!

toji:
You, sir, are absurdly fast at responding to feedback! 😮 Thanks!

Issue #711, opened 2019-06-17
Link "fire an XRInputSourceEvent" text to the related algorithm


ddorwin:
https://immersive-web.github.io/webxr/#xrinputsourceevent-interface contains an algorithm that begins with:
> When the user agent fires an `XRInputSourceEvent` _event_ it MUST run the following steps:

I believe that all of the references to "... fire an `XRInputSourceEvent`..." in other algorithms should link (those three words) to this algorithm.

ddorwin:
Related: Should there be guidance related to ensuring these events are delivered before the next rAF callback? That may be implied/specified by text like the following, but it would be nice to make the intent obvious, potentially with a non-normative note.
> When an XR input source for XRSession session begins its primary action the UA MUST run the following steps:

Manishearth:
This is a tiny bit more than just an editorial fix, because as it stands we never create the `frame` attribute, and _when_ that gets created is kind of important.

Manishearth:
Fixed in #797

PR #797, opened 2019-08-13
Link to 'fire an input source' algorithm, explicitly construct frame

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/711

This ended up being a bit trickier than I thought; I can also do the very simple fix originally proposed.

r? @NellWaliczek

cc @ddorwin

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1455,11 +1455,12 @@ When an [=XR input source=] for {{XRSession}} |session| begins its [=primary act
 
 <div class="algorithm" data-algorithm="on-input-end">
 
-When an [=XR input source=] for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
+When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1455,21 +1455,23 @@ When an [=XR input source=] for {{XRSession}} |session| begins its [=primary act
 
 <div class="algorithm" data-algorithm="on-input-end">
 
-When an [=XR input source=] for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
+When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
 
+  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| for the time the event occurred.
```
Manishearth: It's a bit tedious to create the frame separately each time. I could also create the frame in the algorithm, however we may need to more strongly imply that the frame refers to poses from the time of the event, given that the algorithm is always called within a queued task (and thus the frame should not refer to poses from the time of the _task_)

toji: Yeah, I can see that this is a bit of a pain, but I think your logic here is sound and that this captures the intended timing well.

toji:
LGTM. Thanks!

Issue #413, opened 2018-10-10
User agents should provide headset activation instructions


ddorwin:
While some VR platforms provide instructions on starting a VR session (e.g., inserting the phone into a Daydream View), others do not (e.g., Cardboard). Some authors created their own phone-insertion UI for Cardboard.

With WebVR, authors were able to detect the type of device via `displayName` before starting presentation. We've heard from authors that are looking to transition to WebXR that they can no longer differentiate the headsets (by design) and thus cannot determine whether to display such UI, especially before an immersive session starts and more data is available.

It seems, however, that applications should never have been trying to display such UI in the first place, and this was probably just filling a gap in what the implementation provided. While this approach may have been reasonable when there were few form factors, it's not reasonable for applications to provide appropriate instructions for all form factors or even know whether they need to. There are also additional cases to consider, such as when the user is already in a headset looking at a 2D page - it wouldn't make sense for the page to show such UI, but it might not be able to detect this case.

Thus, user agents should be responsible for showing such UI when appropriate. There probably isn't a lot of spec text to write for this, but authors should know that they do not need to and should not include such UI in their applications.

ddorwin:
Note that since the UA is responsible for this, implementers have a lot of flexibility for what, if anything to display. For example, perhaps this does not make sense for desktop scenarios or maybe the desktop UI is displayed outside browser window. Also, implementations could allow users to disable/enable such UI, allow users to choose "Don't show this again," only display it the first N times, etc.

Manishearth:
Fixed in #799

PR #799, opened 2019-08-13
Require UAs to show manual device activation steps

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/413

cc @ddorwin

r? @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -546,7 +546,9 @@ To <dfn>initialize the session</dfn>, given |session|, |mode|, and |device|, the
   1. Set |session|'s [=XRSession/mode=] to |mode|.
   1. Set |session|'s [=XRSession/XR device=] to |device|.
   1. [=Initialize the render state=].
-  1. If no other features of the user agent have done so already, perform the necessary platform-specific steps to initialize the device's tracking and rendering capabilities.
+  1. If no other features of the user agent have done so already, perform the necessary platform-specific steps to initialize the device's tracking and rendering capabilities, including showing any necessary instructions to the user.
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -546,7 +546,9 @@ To <dfn>initialize the session</dfn>, given |session|, |mode|, and |device|, the
   1. Set |session|'s [=XRSession/mode=] to |mode|.
   1. Set |session|'s [=XRSession/XR device=] to |device|.
   1. [=Initialize the render state=].
-  1. If no other features of the user agent have done so already, perform the necessary platform-specific steps to initialize the device's tracking and rendering capabilities.
+  1. If no other features of the user agent have done so already, perform the necessary platform-specific steps to initialize the device's tracking and rendering capabilities, including showing any necessary instructions to the user.
+
+NOTE: Some devices require additional user instructions for activation, for example going into immersive mode on a phone-based headset device requires inserting the phone into the headset, and doing so on a desktop browser connected to an external headset requires wearing the headset. It is the responsibility of the user agent &mdash; not the author &mdash; to show any such instructions.
```
toji: NOTE: => Note:

Also, minor wording quibble:

> It is the responsibility of the user agent &mdash; not the author &mdash; to show any such instructions.

Even though this text is non-normative, we don't want to imply that the UA must be the one showing the instructions instead of delegating that to the platform when appropriate. (IE: For Daydream it's the Daydream service, not Chrome, that shows the harness instructions.) Perhaps it would be more accurate to say:

> It is the responsibility of the user agent &mdash; not the author &mdash; to **ensure any such instructions are shown**.

Regardless, the most important part is simply that content authors shouldn't be trying to show instructions and you've captured that well.

Manishearth: > NOTE: => Note:

fwiw both work in bikeshed syntax

Manishearth: Resolved.

toji: > fwiw both work in bikeshed syntax

Yup, and we've used both intermixed in the past. I did a pass over the whole doc to normalize it recently, though, and so this is just to keep with style conventions.

toji:
Thanks!

Issue #671, opened 2019-05-28
Define how a WebGLFramebuffer is allocated from XRWebGLLayerInit options 


toji:
Called out by @NellWaliczek during review of #656. We should be more explicit about exactly how the options specified in `XRWebGLLayerInit` contribute to the `framebuffer` that is allocated for an `XRWebGLLayer`. Specifically the `alpha`, `depth`, and `stencil` options.

Manishearth:
fixed in https://github.com/immersive-web/webxr/pull/800

PR #800, opened 2019-08-13
Explicitly mention how depth/alpha/stencil values get used

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/671

There's actually no direct algorithm for WebGL framebuffers that can handle these, typically you control these via `framebufferRenderbuffer` and `framebufferTexture2D` which our code forbids.

Instead, I looked at how the primary drawing buffer handles this in WebGL and updated the opaque framebuffer text to work with that, since its purpose is to match the drawing buffer anyway.

r? @toji


toji:
Perfect, thank you!

Issue #698, opened 2019-06-14
Provide limits to depthNear/Far


cabanier:
Currently, the spec defines that depthNear is initialized with .1 and doesn't call out any minimum (or maximum) values.
The Magic Leap platform has a minimum depth of .37 and won't render any content if a smaller near clip was specified.

Can the browser silently move the clip range on the c++ side to a supported range? 
Or should we enhance the API so the platform can return a clipping range and automatically adjust values to be within the clipping range?

As an aside, [WebGL does not support](https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.14) a inverted depth range:
> The WebGL API does not support depth ranges with where the near plane is mapped to a value greater than that of the far plane. A call to depthRange will generate an INVALID_OPERATION error if zNear is greater than zFar.

Should we remove it from the spec, or should we keep it in case WebGPU supports it?

We currently don't enforce a maximum depth for the far clip, but maybe we should have it?

toji:
My intent was to allow implementations to adjust the depthNear and depthFar as needed. I seem to remember one other vendor mentioning that they had some fairly strict limits on this as well. Obviously that needs some additional spec text to make this explicit. (It'd look a lot like how we specify the [minimum/maximum inline field of view limits](https://immersive-web.github.io/webxr/#minimum-inline-field-of-view).)

As for WebGL not supporting an inverted depth range, I'm actually not clear on that point. The `depthRange` function doesn't actually set the near and far planes, but instead the depth values that they are correlated with (0..1 by default). I believe it's perfectly possible to flip the near/far planes when computing your projection matrix and have it continue to map correctly to that range, it's just now your near plane == a depth of 1 and your far plane == a depth of 0. Combining that with a call to `gl.depthFunc(gl.GREATER);` will give you the expected depth testing behavior for the flipped range.

HOWEVER, because WebGL/OpenGL use a depth coordinates range of -1 to 1 post-projection (which is what's actually mapped to the glDepthRange discussed above) you lose out on most of the benefits of inverting your depth as described with the "What about OpenGL" diagram on this page: https://developer.nvidia.com/content/depth-precision-visualized. So it's not clear if Reverse-Z techniques are really useful for WebGL. But they may indeed be useful for WebGPU, so I think it's reasonable to not want to shut down that possibility prematurely.

toji:
(Setting this as a July milestone work item, only because it wont change any API signatures. Feel free to assume that implementations can clip the developer-specified depth range if needed, though.)

[...]

PR #802, opened 2019-08-13
Allow UA to constrain clip planes

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/698

For enforcing `near < far` I don't think we should be throwing an error due to the two-step nature of applying render states. Instead, perhaps we should set near=far or something? (and encourage UAs to warn to the console). Alternatively, we can throw an error when the pending render state is created, but that can be a bit unpredictable.

cc @cabanier

r? @toji


toji:
LGTM. And as the comments in #698 point out it's not actually clear that we want to prevent inverted depth ranges, so I think it's appropriate that there's not a restriction enforcing that `near < far` here.

Issue #481, opened 2019-01-15
Expand upon the XR Compositor


AdaRoseCannon:
BS Issue: https://immersive-web.github.io/webxr/#issue-6d0cdba2

Section: https://immersive-web.github.io/webxr/#compositor

The XRCompositor is needs some detail such as the XRCompositor Interface, how the list of layer images works and explain the frame lifecycle.

Manishearth:
The target of this bug has changed significantly since it was filed, but I tried to fix remaining concerns in https://github.com/immersive-web/webxr/pull/805

PR #805, opened 2019-08-13
Clarify the compositor a bit

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/481

Most of what we need is already there, but added a bit of text about what the compositor can do, and emphasized that it should protect rendered content from leaking across pages.

r? @NellWaliczek


toji:
LGTM!

PR #807, opened 2019-08-14
Change features to a sequence of 'any'

### LINKED ISSUE #791: XRSessionInit has generic types and is not future proof
`XRSessionInit` takes a sequence of DOMString but wouldn't it be better if it takes an enumeration? There's a predefined set of allowed values so it seems it should be restricted to those.

I think `XRSessionInit` should also be made a bit more forward looking. We might want to ask for additional features (anchors, meshing, hit testing, etc) and this would be a good place to do it.
What if we change the IDL to the following:
```
enum XRSpaceType {
  "local",
  "local-floor",
  "bounded-floor",
  "unbounded"
};

dictionary XRSessionFeatures {
  XRSpaceType space;
}

dictionary XRSessionInit {
  XRSessionFeatures requiredFeatures;
  XRSessionFeatures optionalFeatures;
};

```

Future extension spec can then extend the 'XRSessionFeatures' object with more than just a simple value ie

```
dictionary XRWorldMeshFeatures {
  readonly attribute FrozenArray<DOMPointReadOnly> boundsGeometry; 
  XRWorldMeshQuality quality;
};

dictionary XRSessionFeatures {
  XRSpaceType space;
  XRWorldMeshFeatures worldMesh;
};
```

Any thoughts @toji ?

Sorry that I didn't chime in earlier. I was on PTO in early July and didn't notice all these changes going in.

toji:
/fixes #791 

Updates the `requiredFeatures`/`optionalFeatures` dictionary keys to accept sequences of `any` values for forwards compatibility purposes. (Namely, enabling `optionalFeatures` to always ignore unknown values even if non-enum feature are introduced in the future.)

After the discussion in the above issue, I talked to @bfgeek about how to approach this particular concern. From that conversation it seemed that we either wanted to make this a array of [`object`](https://heycam.github.io/webidl/#idl-object) or an array of [`any`](https://heycam.github.io/webidl/#idl-any), depending on which seemed to fit our use case better. I tested both in Chrome, and both value types worked. I chose `any` for the PR here because in Javascript strings are not considered to be "objects" (ie: `"string" instanceof object == false`). The WebIDL definition of `object` appears to be a distinct concept from the Javascript `object` type, but even reading through the WebIDL spec that's not readily apparent. As such using `any` felt like it was easier to understand on initial reading.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -395,8 +395,8 @@ Additionally, developers are encouraged to design experiences which progressivel
 
 <pre class="idl">
 dictionary XRSessionInit {
-  sequence&lt;DOMString&gt; requiredFeatures;
-  sequence&lt;DOMString&gt; optionalFeatures;
+  sequence&lt;any&gt; requiredFeatures;
+  sequence&lt;any&gt; optionalFeatures;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -392,19 +392,21 @@ Additionally, developers are encouraged to design experiences which progressivel
...
 
 The <dfn dict-member for="XRSessionInit">optionalFeatures</dfn> array contains any [=Optional features=] for the experience. If any value in the list is not a recognized [=feature name=] it will be ignored. Features listed in the {{XRSessionInit/optionalFeatures}} array will be enabled if supported by the [=XRSession/XR Device=] and, if necessary, given a clear signal of [=user intent=], but will not block creation of the {{XRSession}} if absent.
 
-The feature lists accept any string. However, in order to be a considered a valid <dfn>feature name</dfn>, the string must be a value from the following enums:
+The feature lists accept any value. However, in order to be a considered a valid <dfn>feature name</dfn>, the value must be a string from the following enums:
```
Manishearth: Should we have a named list of acceptable values so that other specs may extend this easily?

NellWaliczek: I understand the intent, but this wording seems a bit contradictory to me.  Can we adjust it to somehow imply that *for now* the value must be a string.

toji: Latest iteration makes this a bit more explicit, but I'm reluctant to duplicate the actual values into a list here because it's harder to maintain, especially if future modules extend the enums in question.

toji:
Agreed that "feature value" sounds weird. Changed to "feature descriptor".

PR #809, opened 2019-08-16
Ensure an immersive device is selected in makeXRCompatible()

### LINKED ISSUE #778: What should creating xrCompatible GL context do if there is no active immersive device?
The spec text for `makeXrCompatible()` (https://immersive-web.github.io/webxr/#dom-webglrenderingcontextbase-makexrcompatible) and for `getContext({ xrCompatible: true })` refer to "the immersive XR device" (https://immersive-web.github.io/webxr/#xr-immersive-xr-device) which is the active immersive XR device. It's not obvious what is expected to happen if no device is currently active.

This has the possibility to be a footgun, for example the code:
```js
navigator.xr.requestSession("immersive-vr").then((session) => {
  let glCanvas = document.createElement("canvas");
  let gl = glCanvas.getContext("webgl", { xrCompatible: true });
  ...
});
```
is fine because there's an active immersive device when the GL context is created, but the code:
```js
let glCanvas = document.createElement("canvas");
let gl = glCanvas.getContext("webgl", { xrCompatible: true });
navigator.xr.requestSession("immersive-vr").then((session) => {
  ...
});
```
is not.

If `getContext` took the session object as a parameter, it would be more obvious what the indended semantics was, e.g. something like:
```js
navigator.xr.requestSession("immersive-vr").then((session) => {
  let glCanvas = document.createElement("canvas");
  let gl = glCanvas.getContext("webgl", { xrCompatible: session });
  ...
});
```

Manishearth:
Fixes #778

cc @asajeffrey

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1775,8 +1775,12 @@ When this method is invoked, the user agent MUST run the following steps:
   1. Let |promise| be [=a new Promise=].
     1. Run the following steps [=in parallel=]:
     1. Let |context| be the target {{WebGLRenderingContextBase}} object.
-    1. If |context|'s [=WebGL context lost flag=] is set, [=reject=] |promise| with an {{InvalidStateError}} and abort these abort these steps.
+    1. If |context|'s [=WebGL context lost flag=] is set, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
     1. If |context|'s [=XR compatible=] boolean is <code>true</code>, [=/resolve=] |promise| and abort these steps.
+    1. [=ensures an immersive XR device is selected|Ensure an immersive XR device is selected=].
+    1. If the [=XR/immersive XR device=] is <code>null</code>:
+        1. Set |context|'s [=XR compatible=] boolean to <code>true</code>.
+        1. [=Reject=] |promise| with an {{InvalidStateError}} and abort these steps.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1775,8 +1775,12 @@ When this method is invoked, the user agent MUST run the following steps:
   1. Let |promise| be [=a new Promise=].
     1. Run the following steps [=in parallel=]:
     1. Let |context| be the target {{WebGLRenderingContextBase}} object.
-    1. If |context|'s [=WebGL context lost flag=] is set, [=reject=] |promise| with an {{InvalidStateError}} and abort these abort these steps.
+    1. If |context|'s [=WebGL context lost flag=] is set, [=reject=] |promise| with an {{InvalidStateError}} and abort these steps.
     1. If |context|'s [=XR compatible=] boolean is <code>true</code>, [=/resolve=] |promise| and abort these steps.
+    1. [=ensures an immersive XR device is selected|Ensure an immersive XR device is selected=].
+    1. If the [=XR/immersive XR device=] is <code>null</code>:
+        1. Set |context|'s [=XR compatible=] boolean to <code>true</code>.
```
toji: This should be `false`, shouldn't it? (It would still allow for `inline` sessions, which don't check for context compatibility.)

Manishearth: copy paste bug :smile:

PR #811, opened 2019-08-22
Add dictionary default value

saschanaz:
Required after heycam/webidl#750. (Found from web-platform-tests/wpt#18382)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -215,7 +215,7 @@ XR {#xr-interface}
 [SecureContext, Exposed=Window] interface XR : EventTarget {
   // Methods
   Promise&lt;void&gt; supportsSession(XRSessionMode mode);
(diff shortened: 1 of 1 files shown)


saschanaz:

saschanaz:
> What's the expected behavior of the default value being {} if the dictionary is defined as having required values?

AFAIK that shouldn't happen as the default empty dictionary value `{}` does not include the required values. But does the spec explicitly forbid it, @bzbarsky?

> For example, lines 373 and 471 make the assumption that the {{XRSessionInit/optionalFeatures}} and {{XRSession/requiredFeatures}} are present even though we don't have them marked as "required" in the dictionary. I'm not sure if we want to file a separate issue for those?

That looks erroneous, those should probably get an explicit default value `[]`.

PR #818, opened 2019-08-26
Some edits and questions.

alice:
As promised: I took an editing pass to suggest ways to make it easier for a clueless bystander in a hurry (i.e. me) to follow all the information in the explainer.

In particular, I tried to cut down on word count where possible (in the end I only managed to cut about 300 words), and on paragraph length and word complexity otherwise.

As I went through, I left 
[//]: # "comments"
and/or
[//]: # (comments) 
(writing over a couple of separate sessions and I guess I switched conventions)
with suggestions or questions I didn't feel were answered by the explainer in its current form.

Please don't feel like you need to pull this PR directly; this is more of a way for me to understand the API and to suggest ways that you might be able to make the explainer more clear, concise, and readable by someone who doesn't necessarily have any existing context.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -1,104 +1,180 @@
 # WebXR Device API Explained
 
-## What is WebXR?
-The [WebXR Device API](https://immersive-web.github.io/webxr/) provides access to input and output capabilities commonly associated with Virtual Reality (VR) and Augmented Reality (AR) devices. It allows you develop and host VR and AR experiences on the web.
+The [WebXR Device API](https://immersive-web.github.io/webxr/) provides access to input and output capabilities commonly associated with Virtual Reality (VR) and Augmented Reality (AR) devices. It allows you develop and host VR and AR (collectively, XR) experiences on the web.
 
-Examples of supported devices include (but are not limited to):
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+### Contents:
 
- - [ARCore-compatible devices](https://developers.google.com/ar/discover/supported-devices)
- - [Google Daydream](https://vr.google.com/daydream/)
- - [HTC Vive](https://www.htcvive.com/)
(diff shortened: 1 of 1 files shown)


AdaRoseCannon:
Adding the facetoface label, to make sure we schedule some time to speak with the TAG at TPAC.

NellWaliczek:
Thanks for all the thorough questions and feedback.  We filed a handful of individual issues based on this PR, and the rest of the content I believe was addressed during our joint WG/TAG meeting at TPAC.  From what I recall there were three remaining topics to resolve:
1. *Should `supportsSession` be changed to return a bool?*  At TPAC it was agreed to change this behavior, as well as the API name to avoid developers not being able to detect the change.  The newly proposed name is 'isSessionSupported' and can be found in PR #841 
1. *How should WebXR's feature policy interact with the feature policies of the underlying sensors?*  During our discussion at TPAC, @dontcallmedom suggested that a solution to this problem would be changing the purpose and name of the 'xr' feature policy to something that instead protects spatial tracking.  There are a number of benefits of this suggestion and, after discussing further, @toji, @manishearth, and I are supportive of this approach.  The exact details can be found in PR #842
1. *What is the appropriate way to handle enum values that are required by a specification other than the one that originally defined the eum?* We talked about a few different options but didn't come to a concrete conclusion.  After further discussion with the Gamepad API folks, I'm still not entirely sure what the right approach should be.  We've gotten several suggestions on how to go about this, but they all have different drawbacks and there doesn't appear to be consensus on the approach.  Given that this isn't a problem unique to WebXR, we'd really love to get a more definitive answer from the TAG about which approach is best for web platform consistency.
  1. When the secondary spec nears CR, move nearly all references to the enum and its purpose into the original spec
  1. When the secondary spec nears CR, move the value to the original spec and point to the secondary spec for the explanation of it's purpose and use.
  1. Investigate adding partial enums to webidl
  1. Change the enum to be a DOM string

NellWaliczek:
Moving discussion of the one remaining open issue to the topic dedicated to it.  Please reopen if something was missed!

Issue #756, opened 2019-07-02
Spatial tracking on inline sessions when viewed within an XR headset


NellWaliczek:
First up, I'm not suggesting the core spec include a diorama reference space.  I filed that in proposals a while back, which is absolutely the right place to track it.  (See https://github.com/immersive-web/proposals/issues/31) 

I'm assuming we're expecting inline sessions to be allowed to be created for a 2D page being viewed from within an XR-headset.  If we allow spatial tracking to be enabled for that session, what does that mean for view/projection matrices?

The current designs have developers rendering directly into the canvas's default framebuffer, which means no stereoscopy for now. (We'll definitely need to address this in the potential layers module being considered for design).  However, a bunch of questions now occur to me, some of which might have impact on the core spec, and I'd super appreciate a quick sanity check

The questions that occurred to me are:
* With the reference space types we have today, what is expected to happen to the view/projection matrices provided in response to the user moving to the left or right of the floating browser window?
* In the future, would we require all reference spaces for an inline session being viewed within a head-worn XR device be required to report at least 2 views?
* If not, does that mean we'd want to create a specific reference space for that purpose?  (for ease of discussion, let's call it `diorama`).  Let's assume such a reference space will be it's own feature in the `XRSessionInit` feature arrays.
* If a developer creates both a `local` and `diorama` reference space, would that cause `getViewerPose()` to  return a different number of views depending on the supplied reference space? If so, we probably need to update the spec text so that the list of views isn't owned (even indirectly) as a single property of the session.
* Are there other implications of inline sessions with spatial tracking that are different when in a head-worn XR device?
* Should we be considering disallowing spatial tracking for inline sessions inside an XR headset for now?

ddorwin:
I think the assumed answer to the last question has been "yes" as the privacy implications/considerations have not been discussed or addressed.

Related to this, the last sentence of the non-normative note in https://immersive-web.github.io/webxr/#xr-device-concept seems to imply that this _is_ supported and may need to be updated depending on the outcome of this issue.
>In case the browser is already running on an XR device, the inline XR device will be the same device, and may support multiple views.

toji:
Bumping from Aug to Oct so we can close that (woefully late) milestone. What was the decision here again?

rcabanier:
I thought we were going to remove spatial tracking for inline sessions since noone had interest in supporting them

NellWaliczek:
[...]

PR #829, opened 2019-09-05
Disallow stereo inline sessions for now

Manishearth:
With https://github.com/immersive-web/webxr/issues/636 stereo inline sessions aren't possible anymore since you need a separate presentation context (or at least a separate framebuffer) to make that work. This PR removes some of the older text allowing stereo inline sessions.

In the future we may add diorama support (see https://github.com/immersive-web/webxr/issues/756, https://github.com/immersive-web/proposals/issues/31), but for now the spec shouldn't allow a thing that it renders impossible anyway.

r? @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -371,7 +371,7 @@ enum XRSessionMode {
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


toji:
LGTM, thanks!

Issue #808, opened 2019-08-16
Should XRRigidTransform.matrix be [SameObject]?


jacobcdewitt:
The other `readonly Float32Array` attribute in the spec, `XRView.projectionMatrix`, is marked as `[SameObject]`, but `XRRigidTransform.matrix` is not.

toji:
Yes, it should be. The matrix values are not expected to change for the lifetime of the `XRRigidTransform` object. We do expect it to be lazily evaluated in many cases, but that can be done without violating the `[SameObject]` behavior.

Manishearth:
So while the values are always the same, the matrix array object itself may change, see https://immersive-web.github.io/webxr/#xrrigidtransform-obtain-the-matrix

We recompute the array whenever its buffer is "detached", so this can't be `SameObject`.

That said, we should be _removing_ `SameObject` from `projectionMatrix` and giving it a similar "obtain" algorithm.

Manishearth:
https://github.com/immersive-web/webxr/pull/830

PR #830, opened 2019-09-05
Handle detached buffers in projectionMatrix

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/808, all `Float32Array` attributes should consistently not be `[SameObject]` and should hadle detached buffers.

r? @toji


toji:
LGTM, thanks for the speedy turnaround!

PR #833, opened 2019-09-06
Replace "user's head" with "viewer"

ddorwin:
The former is incorrect for devices that are not head-mounted.


toji:
Great clarification. Thanks!

Issue #817, opened 2019-08-23
WebXR phrasing is not extensible & potentially incorrectly treats “immersive session” as synonymous with “immersive-vr”


bialpio:
It seems to me that the phrasing of parts of WebXR spec is not extensible. One example is related to requesting reference spaces, during the [check](https://immersive-web.github.io/webxr/#reference-space-is-supported) if the reference space is supported:
“3. If type is `local` or `local-floor`, and session is an `immersive session`, return `true`.”

“Immersive session” text links to a [section](https://immersive-web.github.io/webxr/#immersive-session):
“In this document, the term inline session is synonymous with an inline session and the term immersive session is synonymous with an immersive-vr session.”

The above text does not allow for extending the main WebXR spec (see for example issue immersive-web/webxr-ar-module#21) w/o causing potentially breaking changes (or at least might make life harder for the early implementers).

rcabanier:
Can we add phrasing in the AR spec that changes the definition of "Immersive session"?

Are there other extension specs so we can see how this is usually done?

Manishearth:
IMO this issue should be moved to the AR module for modifying the immersive session definition. That's kind of why it exists, so there's one "patch" that needs to be applied.

Manishearth:
This was fixed by #869

PR #834, opened 2019-09-06
Check whether a session is inline rather than immersive when appropriate

ddorwin:
For steps that are really related to inline sessions, check that directly for clarity.

This also avoids needing to consider things like #817 for these steps.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -770,10 +770,10 @@ When an {{XRRenderState}} object is created for an {{XRSession}} |session|, the 
   1. Initialize |state|'s {{XRRenderState/depthFar}} to <code>1000.0</code>.
   1. Initialize |state|'s {{XRRenderState/inlineVerticalFieldOfView}} based on the following:
     <dl class="switch">
-      <dt> If |session| is an [=immersive session=]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1089,7 +1089,7 @@ To check if a <dfn>reference space is supported</dfn> for a given reference spac
   1. If |type| is {{viewer}}, return <code>true</code>.
   1. If |type| is {{local}} or {{local-floor}}, and |session| is an [=immersive session=], return <code>true</code>.
   1. If |type| is {{local}} or {{local-floor}}, and the [=XRSession/XR device=] supports reporting orientation data, return <code>true</code>.
-  1. If |type| is {{bounded-floor}} and |session| is an [=immersive session=], return if [=bounded reference spaces are supported=] by the [=XRSession/XR device=].
+  1. If |type| is {{bounded-floor}} and |session| is an [=immersive session=], return the result of whether [=bounded reference spaces are supported=] by the [=XRSession/XR device=].
```
ddorwin: Another way to phrase this is "the result of executing the Foo algorithm." However, that doesn't seem consistent with how the algorithms are referenced in this spec.

toji:
LGTM, thanks!

PR #836, opened 2019-09-06
Replace repeated variable check with If-Otherwise

ddorwin:


toji:
Thank you for all of these cleanups. They really help the flow of the spec!

PR #841, opened 2019-09-12
supportsSession->isSessionSupported, resolves w/ bool

### LINKED ISSUE #824: TAG Feedback: Should supportsSession resolve with a boolean?
From the [TAG feedback for the WebXR API](https://github.com/w3ctag/design-reviews/issues/403#issuecomment-527299905):

> Why does `navigator.xr.supportsSession()` not return a `Promise<boolean>` rather than rejecting in the case that the session type is not supported? That would seem like a better semantic match to the wording of the method, as well as not requiring the author to program a simple feature detection step in a try/catch style.

(A brief technical tangent: This method is asynchronous because we expect that in many cases this call will be the first indication a page has that XR capabilities are desired, and initializing the resources required to answer whether or not a given mode is supported is likely heavyweight enough that it's appropriate to not block the main thread waiting for an answer.)

The current behavior (rejecting when an unsupported mode is passed) was discussed with @bfgeek to review ergonomics a while back, but it's probably worth briefly revisiting to ensure that we haven't broken any assumptions about the functionality since then.

It's worth noting that several other bits of API surface (`requestSession`, `requestReferenceSpace`, etc.) use this same style, where rejecting the promise indicates failure, and we likely want to remain consistent in how we interact with promises throughout the API. Also worth noting that, as far as I can see, the current behavior is consistent with the use of other `Promise<void>`s in APIs such as WebUSB and WebAudio.

toji:
Not intended for merge until discussed at TPAC /facetoface

Would fix #824 if this is the direction the working group agrees on.

Acts on feedback from the TAG that supportsSession should
resolve with a boolean rather than reject when a session mode is not
supported. In order to aid developers the method name was changed so
that previous uses of the old method will be clearly identified as
failing, rather than silently report that a session was supported that
shouldn't be because the app failed to check the resolved boolean value.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -72,7 +72,7 @@ The basic steps most WebXR applications will go through are:
 
 The UA will identify an available physical unit of XR hardware that can present imagery to the user, referred to here as an "XR device". On desktop clients this will usually be a headset peripheral; on mobile clients it may represent the mobile device itself in conjunction with a viewer harness (e.g., Google Cardboard/Daydream or Samsung Gear VR). It may also represent devices without stereo-presentation capabilities but with more advanced tracking, such as ARCore/ARKit-compatible devices. Any queries for XR capabilities or functionality are implicitly made against this device.
 
-> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to expose. The UA is allowed to use any criteria it wishes to select which device is used, including settings UI that allow users to manage device priority. Calling `navigator.xr.supportsSession` or `navigator.xr.requestSession` with `'inline'` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
+> **Non-normative Note:** If there are multiple XR devices available, the UA will need to pick which one to expose. The UA is allowed to use any criteria it wishes to select which device is used, including settings UI that allow users to manage device priority. Calling `navigator.xr.sessionSupported` or `navigator.xr.requestSession` with `'inline'` should **not** trigger device-selection UI, however, as this would cause many sites to display XR-specific dialogs early in the document lifecycle without user activation.
 
 It's possible that even if no XR device is available initially, one may become available while the application is running, or that a previously available device becomes unavailable. This will be most common with PC peripherals that can be connected or disconnected at any time. Pages can listen to the `devicechange` event emitted on `navigator.xr` to respond to changes in device availability after the page loads. (XR devices already available when the page loads will not cause a `devicechange` event to be fired.) `devicechange` fires an event of type `Event`.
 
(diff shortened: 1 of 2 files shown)


thetuvix:
For name bikeshedding, what about `XR.isSessionSupported(mode)`?

This matches [`MediaRecorder.isTypeSupported(mimeType)`](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/isTypeSupported).

toji:
> For name bikeshedding, what about `XR.isSessionSupported(mode)`?

I have no qualms with that name, assuming it meets the naming guidelines that were brought up by the TAG.

toji:
Updated the PR with Alex's suggested name.

explainer.md
```diff
@@ -137,7 +139,7 @@ function beginXRSession() {
...
 In this sample, the `beginXRSession` function, which is assumed to be run by clicking the "Enter VR" button in the previous sample, requests an `XRSession` that operates in `immersive-vr` mode. The `requestSession` method returns a promise that resolves to an `XRSession` upon success. In addition to the `XRSessionMode`, developers may supply an `XRSessionInit` dictionary containing the capabilities that the returned session must have. For more information, see [Feature dependencies](#feature-dependencies).
 
-If `supportsSession` resolved for a given mode, then requesting a session with the same mode should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.) The UA is ultimately responsible for determining if it can honor the request.
+If `isSessionSupported` resolved to _true_ for a given mode, then requesting a session with the same mode should be reasonably expected to succeed, barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.) The UA is ultimately responsible for determining if it can honor the request.
```
NellWaliczek: >barring external factors (such as `requestSession` not being called in a user activation event for an immersive session.)

Not this PR, but might be good to add the example of requesting required features as another reason the requestSession call might fail

toji:
Have https://github.com/immersive-web/webxr-samples/pull/41 and https://github.com/immersive-web/webxr-polyfill/pull/93 ready to roll immediately after merging this to ensure that the change propagates as quickly as possible.

/agenda to ensure that all WG members have an opportunity to look over this change prior to landing it, which will ideally happen this week.

Artyom17:
Ship it!

dontcallmedom:
With regard to naming guidelines, I think we're good if I interpret the [relevant guideline](https://w3ctag.github.io/design-principles/#naming-is-hard) correctly:
> Boolean properties, options, or API parameters which are asking a question about their argument should not be prefixed with is, while methods that serve the same purpose, given that it has no side effects, should be prefixed with is to be consistent with the rest of the platform.

toji:
No concerns raised, so merged!

Issue #770, opened 2019-07-18
[Feature Policy] Establish guidance about policy-controlled features for new XR features


ddorwin:
The solution for Feature Policy for the core ("VR Complete") WebXR Device API spec (see #308) is fairly simple and defers a lot of the complexity to the future. While we don't need to solve the reset now, we will fairly soon as incubations and/or additional modules mature. As noted at the end of https://github.com/immersive-web/webxr/issues/729#issuecomment-513027584, we can't retroactively give developers control later, so it's important that new features/APIs/extensions of WebXR do the right thing before anyone ships them.

Thus, as much as possible, we should try to formalize the idea that new features / extensions of WebXR must carefully consider whether they also depend on additional policy-controlled feature(s) and provide guidance on choosing/defining a new policy-controlled feature (or reusing an existing one) in a way that is consistent with the overall design/intent.

Relevant discussions that may feed into this guidance include #308, #729, #763, and #765.

Issue #768, opened 2019-07-17
[Feature Policy] Consider implications and practicality of requiring authors to enable policy-controlled features for underlying sensors


ddorwin:
https://immersive-web.github.io/webxr/#feature-policy currently says:

>In addition to the `"xr"` feature policy, feature policies for underlying sensors must also be respected if a site could isolate and extract sensor data that would otherwise be blocked by those feature policies.

While the motivation ([in the explainer](https://github.com/immersive-web/webxr/blob/master/privacy-security-explainer.md#underlying-sensors-feature-policy) and more details in https://github.com/immersive-web/webxr/issues/732#issuecomment-505670527) makes sense, the implications and practical considerations might mean this is not the best decision for users, developers, and implementations.

Perhaps most importantly, should user agents implement this requirement and application developers grant the necessary policies (i.e., the union of [policy-controlled features](https://wicg.github.io/feature-policy/#policy-controlled-feature) required for all implementations devices), those developers would be allowing (direct) access to all such features (i.e., for cross-origin iframes) _regardless_ of whether the underlying XR implementation needs it.

As an illustrative example, consider the case where a specific underlying sensor is used by a small portion of all user devices. In order to support them, an application must allow (direct) access to that underlying sensor data across all user agents. Or consider the case where an underlying sensor's data was previously accessible but this "back door" has been closed by the implementation or such devices are no longer in use; most applications that supported such devices would likely continue to allow (direct) access to that underlying sensor data across all user agents (either because the application is no longer maintained or the developer is unaware that it is safe to make the change).

<br>In addition, it is worth considering the implications for both developers and implementers and whether this is the best solution.

For application developers:
* What developers really want to say is "I want this site to be able to use XR hardware."
* Developers must know all such underlying sensors that might be used and include all of them.
* As new underlying sensor technologies are used, developers will need to update their code to support new devices that use them.
* Similarly, as underlying sensors that aren't currently directly exposed by the web platform get exposed, new policies will become required for implementations and devices that previously worked without them.
* If developers fail to do the three previous items, their applications may not work - or stop working - across user agents and devices.
* Developers may not want to allow the dedicated API as they only care about WebXR, which has explicit intent, usage indication, use case-specific privacy considerations, etc. (Caveat: It's unclear whether Feature Policy enables APIs or capabilities - see #765.)
* Feature Policy for WebXR is already going to be complex (i.e., #763).
[...]

Issue #765, opened 2019-07-15
[Feature Policy] Confirm whether WebXR features can/should reuse existing feature names


ddorwin:
I, at least, have been assuming that XR features should reuse existing [policy-controlled feature names](https://github.com/w3c/webappsec-feature-policy/blob/master/features.md). For example, `"camera"` for frame-aligned camera frame access in AR and maybe `"ambient-light-sensor"` for [lighting estimation](https://github.com/immersive-web/lighting-estimation/). We should confirm whether this is the case and maybe document some guidelines because it might affect the Feature Policy solution for WebXR and will definitely affect future XR features.

On the one hand:
* The existing [feature names](https://github.com/w3c/webappsec-feature-policy/blob/master/features.md) read more like capabilities than specification names or APIs (i.e., methods).
* It seems undesirable to duplicate features.
* Reusing feature names makes it easier for polyfills to also work.

On the other hand:
* The specification is called _Feature_ Policy, not capability policy.
* The specification [says](https://w3c.github.io/webappsec-feature-policy/#policy-controlled-feature), "A policy-controlled feature is an **API or behaviour** which can be enabled or disabled in a document by referring to it in a feature policy" (emphasis added).
* Each feature name is defined by a specification, and the [Policy Controlled Features](https://github.com/w3c/webappsec-feature-policy/blob/master/features.md) doc is just a non-normative collection of those.
  - Note that the _default allowlist_ for each such feature is defined in the individual specifications as well.
* What would it mean for another specification to come along and redefine the meaning of a string from another specification (even if it is in a consistent way)?
   - For example, for an AR frame-aligned camera frame access API to _also_ define behavior related to `"camera"`, which is already defined at https://w3c.github.io/mediacapture-main/#feature-policy-integration.

/cc @clelland

Issue #763, opened 2019-07-12
[Feature Policy] Validate the plan for dependent policy-controlled features


ddorwin:
One of the outcomes of #308 (see discussion starting at https://github.com/immersive-web/webxr/issues/308#issuecomment-504610297) was a plan to have sub-features - policy-controlled features that depend on the main `"xr"` policy-controlled feature. As noted in that comment, this appears to be unique.

Before we commit to this plan, we should:
1. Ensure that such dependent policy-controlled features are acceptable from a Feature Policy and web platform point of view.
1. Figure out how the sub-feature mechanics would be specified.

The following are more details on (2).

Note that each policy-controlled feature has a state, and there isn't really a mechanism for checking other features. Is it sufficient that `"xr"` would make `navigator.xr` inaccessible or would the WebXR algorithms need to check combinations at each feature?

Does text related to the default allowlist remain unchanged? Specifically, text like the following:
>The [default allowlist](https://w3c.github.io/webappsec-feature-policy/#default-allowlist) for this feature is `["self"]`.

/cc @clelland

PR #842, opened 2019-09-12
Update behavior, name of Feature Policy

### LINKED ISSUE #823: [Feature Policy] should not control exposure of navigator.xr
https://immersive-web.github.io/webxr/#feature-policy currently says that the "policy-controlled feature... controls whether the `xr` attribute is exposed on the `Navigator` object."

According to @clelland, controlling exposure of an IDL member is not a common pattern ([context](https://chromium-review.googlesource.com/c/chromium/src/+/1783670/1/third_party/blink/renderer/modules/xr/xr.cc#493)). For consistency with other APIs, the feature check should probably be moved to the top-level methods on the `navigator.xr` object.

<br>FWIW, the decision to use `"xr"` was somewhat motivated (i.e., https://github.com/immersive-web/webxr/issues/308#issuecomment-504610297 and related conversations) by the simplicity of not exposing the `xr` attribute. Changing _how_ the control works would enable make other options possible, though that's not required or blocking this change.

toji:
Not intended for merge until discussed at the TPAC /facetoface

Would fix #823 if the working group agreed to the suggested behavioral change from that issue. For a broader look at the editors proposed approach to Feature Policy handling within the Immersive Web APIs, please see https://docs.google.com/document/d/1RZTL69JsTxoJUyXNnu_2v0PPILqrDpYW3ZxDjMAqQ-M/edit?usp=sharing, which will also be discussed at the face to face.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2221,7 +2221,7 @@ Integrations {#integrations}
 Feature Policy {#feature-policy}
 --------------
 
-This specification defines a [=policy-controlled feature=] that controls whether the any {{XRSession}}s may be returned by {{XR/requestSession()}} that makes use of spatial tracking, and whether support for sessions modes that require spatial tracking may be indicated by either {{XR/isSessionSupported()}} or {{devicechange}} events on the {{Navigator/xr|navigator.xr}} object.
+This specification defines a [=policy-controlled feature=] that controls whether any {{XRSession}} that makes use of spatial tracking may be returned by {{XR/requestSession()}}, and whether support for session modes that require spatial tracking may be indicated by either {{XR/isSessionSupported()}} or {{devicechange}} events on the {{Navigator/xr|navigator.xr}} object.
 
 The feature identifier for this feature is <code>"xr-spatial-tracking"</code>.
 
--- a/index.bs
+++ b/index.bs
@@ -2215,7 +2215,7 @@ Integrations {#integrations}
 Feature Policy {#feature-policy}
 --------------
 
-This specification defines a [=policy-controlled feature=] that controls whether the {{Navigator/xr}} attribute is exposed on the {{Navigator}} object.
+This specification defines a [=policy-controlled feature=] that controls whether the any {{XRSession}}s may be returned by {{XR/requestSession()}} and whether {{XR/isSessionSupported()}} indicates that any session mode is supported.
 
 The feature identifier for this feature is <code>"xr"</code>.
 
--- a/index.bs
+++ b/index.bs
@@ -422,7 +422,7 @@ Future iterations of this specification and additional modules may expand the li
 
 Note: Features are accepted as an array of <code>any</code> values to ensure forwards compatibility. It allows unrecognized optional values to be properly ignored as new [=feature descriptor=] types are added.
 
-Depending on the {{XRSessionMode}} requested, certain [=feature descriptors=] are added to the {{XRSessionInit/requiredFeatures}} or {{XRSessionInit/optionalFeatures}} lists by default. This is only done if the feature does not require a signal of [=user intent=] nor impact performance or the behavior of other features when enabled. The following table describes the <dfn>default features</dfn> associated with each session type and feature list:
+Depending on the {{XRSessionMode}} requested, certain [=feature descriptors=] are added to the {{XRSessionInit/requiredFeatures}} or {{XRSessionInit/optionalFeatures}} lists by default. The following table describes the <dfn>default features</dfn> associated with each session type and feature list:
 
 <table class="tg">
   <thead>
@@ -497,14 +497,14 @@ To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optio
   1. Add every [=feature descriptor=] in the [=default features=] table associated with |session|'s [=XRSession/mode=] to the indicated feature list if it is not already present.
   1. For each |feature| in |requiredFeatures| perform the following steps:
     1. If |feature| is not a valid [=feature descriptor=], return <code>false</code>.
-    1. If |session|'s [=XRSession/XR Device=] is not [=capable of supporting=] the functionality described by |feature| or the user agent has otherwise determined to reject the feature, return <code>false</code>.
     1. If the requesting document's origin is not allowed to use any [[#feature-policy|feature policy]] required by |feature| as indicated by the [=feature requirements=] table, return <code>false</code>.
+    1. If |session|'s [=XRSession/XR Device=] is not [=capable of supporting=] the functionality described by |feature| or the user agent has otherwise determined to reject the feature, return <code>false</code>.
     1. If the functionality described by |feature| requires [=explicit consent=], append it to |consentRequired|.
     1. Else append |feature| to |session|'s [=list of enabled features=].
   1. For each |feature| in |optionalFeatures| perform the following steps:
     1. If |feature| is not a valid [=feature descriptor=], continue to the next entry.
-    1. If |session|'s [=XRSession/XR Device=] is not [=capable of supporting=] the functionality described by |feature| or the user agent has otherwise determined to reject the feature, continue to the next entry.
     1. If the requesting document's origin is not allowed to use any [[#feature-policy|feature policy]] required by |feature| as indicated by the [=feature requirements=] table, continue to the next entry.
+    1. If |session|'s [=XRSession/XR Device=] is not [=capable of supporting=] the functionality described by |feature| or the user agent has otherwise determined to reject the feature, continue to the next entry.
     1. If the functionality described by |feature| requires [=explicit consent=], append it to |consentOptional|.
     1. Else append |feature| to |session|'s [=list of enabled features=].
   1. If |consentRequired| or |consentOptional| are not empty, request [=explicit consent=] to use the functionality described by those features.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2205,7 +2206,7 @@ Integrations {#integrations}
 Feature Policy {#feature-policy}
 --------------
 
-This specification defines a [=policy-controlled feature=] that controls whether the {{Navigator/xr}} attribute is exposed on the {{Navigator}} object.
+This specification defines a [=policy-controlled feature=] that controls whether the any {{XRSession}}s may be returned by {{XR/requestSession()}} and whether {{XR/supportsSession()}} indicates that any session mode is supported.
```
alcooper91: We should also consider the behavior of device change events

toji: Great catch! I've updated the PR to also indicate that `devicechange` events shouldn't fire when the feature policy is not set.

index.bs
```diff
@@ -273,6 +273,7 @@ The <dfn method for="XR">supportsSession(|mode|)</dfn> method queries if a given
 When this method is invoked, it MUST run the following steps:
 
   1. Let |promise| be [=a new Promise=].
+  1. If the requesting document's origin is not allowed to use the WebXR [[#feature-policy|feature policy]], [=reject=] |promise| with a "{{SecurityError}}" {{DOMException}} and abort these steps.
```
alcooper91: It looks like this is the supportsSession algorithm, should this be done for requestSession as well?

alcooper91: Ahh, requestSession I see that it's buried in the "Trustworthy" and "request inline" sessions.  Perhaps it should be unified up here to match the early bail out of supportsSession?
Also, if the decision is made to *not* make this change for whatever reason, should that logic be removed?

toji:
After discussion at TPAC, updated this to conform to what sounds like the general direction the group wants to go. Proposed feature policy is now `xr-spatial-tracking` and, as the name implies, more narrowly only blocks features that attempt to use spatial tracking. (This happens to be all immersive sessions, as should be clearer from the refactored required features text.) This has the side effect of allowing `inline` sessions without tracking to be used everywhere, which we're asserting is the desired behavior. (This following comments from the TAG that relying on a polyfill for these scenarios as I had described earlier was incredibly strange.) PTAL!

index.bs
```diff
@@ -2206,9 +2212,9 @@ Integrations {#integrations}
 Feature Policy {#feature-policy}
 --------------
 
-This specification defines a [=policy-controlled feature=] that controls whether the {{Navigator/xr}} attribute is exposed on the {{Navigator}} object.
+This specification defines a [=policy-controlled feature=] that controls whether the any {{XRSession}}s may be returned by {{XR/requestSession()}} that makes use of spatial tracking, and whether support for sessions modes that require spatial tracking may be indicated by either {{XR/supportsSession()}} or {{devicechange}} events on the {{Navigator/xr|navigator.xr}} object.
```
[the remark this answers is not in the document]
toji: Good catch, thanks!

index.bs
```diff
@@ -481,20 +490,19 @@ To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optio
 
   1. Let |consentRequired| be an empty [=/list=] of {{DOMString}}.
   1. Let |consentOptional| be an empty [=/list=] of {{DOMString}}.
+  1. Add every [=feature descriptor=] in the [=default features=] table associated with |session|'s [=XRSession/mode=] to the indicated feature list if it is not already present.
   1. For each |feature| in |requiredFeatures| perform the following steps:
     1. If |feature| is not a valid [=feature descriptor=], return <code>false</code>.
     1. If |session|'s [=XRSession/XR Device=] is not [=capable of supporting=] the functionality described by |feature| or the user agent has otherwise determined to reject the feature, return <code>false</code>.
+    1. If the requesting document's origin is not allowed to use any [[#feature-policy|feature policy]] required by |feature| as indicated by the [=feature requirements=] table, return <code>false</code>.
```
alcooper91: For this and the the optional features, it feels a little cleaner if we do all of our static checks first (not valid, feature policy blocked) and then require anything that causes checks out to the device.

NellWaliczek: I agree with flipping the ordering of this step with the one before it

index.bs
```diff
@@ -481,20 +490,19 @@ To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optio
...
     1. If the functionality described by |feature| requires [=explicit consent=], append it to |consentRequired|.
     1. Else append |feature| to |session|'s [=list of enabled features=].
   1. For each |feature| in |optionalFeatures| perform the following steps:
     1. If |feature| is not a valid [=feature descriptor=], continue to the next entry.
     1. If |session|'s [=XRSession/XR Device=] is not [=capable of supporting=] the functionality described by |feature| or the user agent has otherwise determined to reject the feature, continue to the next entry.
+    1. If the requesting document's origin is not allowed to use any [[#feature-policy|feature policy]] required by |feature| as indicated by the [=feature requirements=] table, continue to the next entry.
```
NellWaliczek: Same as above

index.bs
```diff
@@ -2206,9 +2212,9 @@ Integrations {#integrations}
 Feature Policy {#feature-policy}
 --------------
 
-This specification defines a [=policy-controlled feature=] that controls whether the {{Navigator/xr}} attribute is exposed on the {{Navigator}} object.
+This specification defines a [=policy-controlled feature=] that controls whether any {{XRSession}} that makes use of spatial tracking may be returned by {{XR/requestSession()}}, and whether support for session modes that require spatial tracking may be indicated by either {{XR/supportsSession()}} or {{devicechange}} events on the {{Navigator/xr|navigator.xr}} object.
```
NellWaliczek: > makes use of spatial tracking

perhaps this statement should be stronger?  "requires the use of spatial tracking"?

index.bs
```diff
@@ -417,53 +418,61 @@ Future iterations of this specification and additional modules may expand the li
 
 Note: Features are accepted as an array of <code>any</code> values to ensure forwards compatibility. It allows unrecognized optional values to be properly ignored as new [=feature descriptor=] types are added.
 
-Some {{XRSessionMode}}s enable certain [=feature descriptors=] as [=optional features=] by default. This is only done if the feature does not require a signal of [=user intent=] nor impact performance or the behavior of other features when enabled. The following table describes the <dfn>default features</dfn> associated with each session:
+Depending on the {{XRSessionMode}} requested, certain [=feature descriptors=] are added to the {{XRSessionInit/requiredFeatures}} or {{XRSessionInit/optionalFeatures}} lists by default. This is only done if the feature does not require a signal of [=user intent=] nor impact performance or the behavior of other features when enabled. The following table describes the <dfn>default features</dfn> associated with each session type and feature list:
```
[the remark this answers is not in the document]
toji: Feeling like we should just drop that line entirely? We can put whatever criteria we want in the table below.

index.bs
toji: Good call. Flipping!

index.bs

toji:
Updated to address review feedback. Thanks!

index.bs
```diff
@@ -2212,12 +2212,14 @@ Integrations {#integrations}
...
+This specification defines a [=policy-controlled feature=] that controls whether any {{XRSession}} that requires the use of spatial tracking may be returned by {{XR/requestSession()}}, and whether support for session modes that require spatial tracking may be indicated by either {{XR/supportsSession()}} or {{devicechange}} events on the {{Navigator/xr|navigator.xr}} object.
 
 The feature identifier for this feature is <code>"xr-spatial-tracking"</code>.
 
 The [=default allowlist=] for this feature is <code>["self"]</code>.
 
+Note: If the document's origin is not allowed to use the <code>"xr-spatial-tracking"</code> feature policy any [=immersive sessions=] will be blocked, because all [=immersive sessions=] require some use of spatial tracking. [=Inline sessions=] will still be allowed, but restricted to only using the {{XRReferenceSpaceType/"viewer"}} {{XRReferenceSpace}}.
```
NellWaliczek: > [=Inline sessions=] will still be allowed, but restricted to only using the {{XRReferenceSpaceType/"viewer"}} {{XRReferenceSpace}}.

how about 

>[=Inline sessions=] will still be allowed, but requestSession will only succeed if no reference spaces are listed in the [=required features=] array.  These sessions will only be capable of creating an {{XRReferenceSpaceType/"viewer"}} {{XRReferenceSpace}}.

dontcallmedom:
Just a thought following one of my comments during the meeting: I wonder if we can and should grant feature-id that are needed to run a WebXR polyfill (e.g. accelerometer, magnetomer?, gyroscope) when xr-spatial-tracking is granted - this would enable the polyfill to run in that situation.

("should" is whether this is a useful use case to cater for, "can" is whether this is a concept that either already exists or would be considered by the feature policy team)

NellWaliczek:
/fixes #823, #770, #768, #765, #763, #729

toji:
Updated to incorporate Nell's suggestion.

> I wonder if we can and should grant feature-id that are needed to run a WebXR polyfill (e.g. accelerometer, magnetomer?, gyroscope) when xr-spatial-tracking is granted - this would enable the polyfill to run in that situation.

That would only work for browsers that actually have a WebXR implementation, and my impression from talking with the TAG last week was that using polyfills to prop up the functionality of a native implementation was "weird". :)

/agenda to ensure that all WG members have an opportunity to look over this change prior to landing it, which will ideally happen this week.

rcabanier:
Is there a tracking bug to add this to the [feature policy spec](https://w3c.github.io/webappsec-feature-policy/)?
I looked their [github repo](https://github.com/w3c/webappsec-feature-policy/issues/) but I didn't find any references to WebXR

toji:
That's a good point. Given how the feature policy spec is structured, we'll actually want to update https://github.com/w3c/webappsec-feature-policy/blob/master/features.md, but regardless we should put up a PR to make that change. I'll file it now.

PR #851, opened 2019-09-20
Update example profile names to match real names

thetuvix:
This PR updates the example input profile names mentioned in the spec to align with the real profile names standardized in the [input profile registry](https://github.com/immersive-web/webxr-input-profiles).  This avoids confusion if developers believe those example input profile names are actually what will show up in practice.


index.bs
```diff
@@ -1470,16 +1470,16 @@ The <dfn attribute for="XRInputSource">profiles</dfn> attribute is a [=/list=] o
...
+For example, the Samsung HMD Odyssey's controller is a design variant of the standard Windows Mixed Reality controller. Both controllers share the same input layout. As a result, the {{XRInputSource/profiles}} for a Samsung HMD Odyssey controller could be: <code>["samsung-odyssey", "microsoft-mixed-reality", "generic-trigger-squeeze-touchpad-thumbstick"]</code>. The appearance of the controller is most precisely communicated by the first profile in the list, with the second profile describing an acceptable substitute, and the last profile a generic fallback that describes the device in the roughest sense. (It's a wand with a trigger, squeeze button, touchpad and thumbstick.)
```
NellWaliczek: Should we also remove the word "wand" from this paragraph now that it's no longer being used as part of profile names?

index.bs
```diff
@@ -1470,16 +1470,16 @@ The <dfn attribute for="XRInputSource">profiles</dfn> attribute is a [=/list=] o
...
+Similarly, the Valve Index controller is backwards compatible with the HTC Vive controller, but the Index controller has additional buttons and axes. As a result, the {{XRInputSource/profiles}} for the Valve Index controller could be: <code>["valve-index", "htc-vive", "generic-trigger-squeeze-touchpad-thumbstick"]</code>. In this case the input layout described by the <code>"valve-index"</code> profile is a superset of the layout described by the <code>"htc-vive"</code> profile. Also, the <code>"valve-index"</code> profile indicates the precise appearance of the controller, while the <code>"htc-vive"</code> controller has a significantly different appearance. In this case the UA would have deemed that difference acceptable. And as in the first example, the last profile is a generic fallback.
```
NellWaliczek: Ok, so here's a question... is this actually the desired behavior for the valve index?  Or would it be better to skip the `htc-vive` fallback since there is actually a generic variant that can be used?

thetuvix: Good question!  This fallback to HTC Vive is a pre-existing assertion, so I'll carve off a separate issue to make that call.

thetuvix: Filed as immersive-web/webxr-input-profiles#98.

thetuvix:
@toji: This should be ready to merge.

thetuvix:
@toji: This should be ready to merge.

toji:
LGTM!

PR #853, opened 2019-09-21
Removed errant 's'

fordacious:
Found this while perusing the spec.


dontcallmedom:

PR #857, opened 2019-09-23
Fix missing and in populate the pose

takahirox:
This PR fixes missing "and" in populate the pose section.

```
Else if limit is false "and" the tracking system provides....
```


dontcallmedom:

Manishearth:
Thanks!

Issue #863, opened 2019-09-30
[editorial] Prepare for updated WD (2019/Sep)


himorin:
- [ ] modify publication date of updated WD

PR #867, opened 2019-10-03
as 2019/10/07 WD

himorin:
for #863


toji:
Thanks! So, to be clear about procedure at this point, the next change that we merge into the spec should revert this to and editors draft, correct?

PR #869, opened 2019-10-08
Move "immersive-ar" mode back into the core spec

toji:
Half of the fix for https://github.com/immersive-web/webxr-ar-module/issues/32, will have a corresponding PR in the WebXR AR module to remove the WebIDL definition from that spec.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -386,7 +386,7 @@ enum XRSessionMode {
 
   - A session mode of <dfn enum-value for="XRSessionMode">inline</dfn> indicates that the session's output will be shown as an element in the HTML document. {{inline}} session content MUST be displayed in mono (i.e., with a single [=view=]). It MAY allow for [=viewer=] tracking. User agents MUST allow {{inline}} sessions to be created.
   - A session mode of <dfn enum-value for="XRSessionMode">immersive-vr</dfn> indicates that the session's output will be given [=exclusive access=] to the [=XR/immersive XR device=] display and that content <b>is not</b> intended to be integrated with the user's environment.
-  - The behavior of the <dfn enum-value for="XRSessionMode">immersive-ar</dfn> session mode is defined in the <a href="https://immersive-web.github.io/webxr-ar-module/">WebXR AR Module</a> and is not valid unless the UA implements that module.
+  - The behavior of the <dfn enum-value for="XRSessionMode">immersive-ar</dfn> session mode is defined in the <a href="https://immersive-web.github.io/webxr-ar-module/">WebXR AR Module</a> and MUST NOT be added to the [=XR/immersive XR device=]'s [=list of supported modes=] unless the UA implements that module.
 
 In this document, the term <dfn>inline session</dfn> is synonymous with an {{inline}} session and the term <dfn>immersive session</dfn> refers to either an {{immersive-vr}} or {{immersive-ar}} session.
 
--- a/index.bs
+++ b/index.bs
@@ -312,7 +312,7 @@ The <dfn method for="XR">requestSession(|mode|, |options|)</dfn> method attempts
 When this method is invoked, the user agent MUST run the following steps:
 
   1. Let |promise| be [=a new Promise=].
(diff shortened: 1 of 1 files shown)


toji:
AR module portion of this PR is https://github.com/immersive-web/webxr-ar-module/pull/38

index.bs
```diff
@@ -379,14 +379,16 @@ The {{XRSessionMode}} enum defines the modes that an {{XRSession}} can operate i
...
   "inline",
-  "immersive-vr"
+  "immersive-vr",
+  "immersive-ar"
 };
 </pre>
 
   - A session mode of <dfn enum-value for="XRSessionMode">inline</dfn> indicates that the session's output will be shown as an element in the HTML document. {{inline}} session content MUST be displayed in mono (i.e., with a single [=view=]). It MAY allow for [=viewer=] tracking. User agents MUST allow {{inline}} sessions to be created.
   - A session mode of <dfn enum-value for="XRSessionMode">immersive-vr</dfn> indicates that the session's output will be given [=exclusive access=] to the [=XR/immersive XR device=] display and that content <b>is not</b> intended to be integrated with the user's environment.
+  - The behavior of the <dfn enum-value for="XRSessionMode">immersive-ar</dfn> session mode is defined in the <a href="https://immersive-web.github.io/webxr-ar-module/">WebXR AR Module</a> and is not valid unless the UA implements that module.
```
thetuvix: We should be more specific around "not valid" here, specifying that even if a UA doesn't support the AR module, it must still let the page call `isSessionSupported("immersive-ar")` and get false rather than an exception.

rcabanier: yes, `isSessionSupported` should not throw an exception if you pass in an unknown value

rcabanier: this creates a spec loop. WebXR can't make a normative reference to the AR spec which is not in CR.
We should either move the enum to `DOMString` and define the behavior in each spec, or we merge the specs

toji:
Also /fixes #817, FWIW.

index.bs
thetuvix: The pushback on partial enums seemed to be around not having spooky behavior at a distance, where many other specs modify the stated behavior of a core spec.  However, if we make this a `DOMString` purely to work around the spec loop, we'd end up with the worst of both worlds - the other spec still modifies what's logically an enum here, but now we have no mention of it at all in the core spec, so someone doesn't even have a breadcrumb to know about other values.

A `DOMString` could make sense if that's the only mechanism to have `isSessionSupported` return false for unknown future values.  Is throwing an exception on unknown enum values required by WebIDL?  If so, are input enums basically non-extensible in WebIDL?

toji: > We should be more specific around "not valid" here

Agreed. I've updated the PR to now state that the enum MUST NOT be added to the list of supported modes unless the UA implements the AR module, which I think gives us all the clarification we need.

> enum vs. DOMString

At this point honestly I'm looking for the smallest delta I can get from what was previously specced to what WebIDL allows that will let us move forward. As stated previously, I can't see this particular enum growing much larger than this over the lifetime of the API, so I don't feel that it warrants as much collective brainpower as we've already expended on it.

thetuvix: Sounds good!  The use of "list of supported modes" elsewhere makes clear that this will return false.

rcabanier: > Agreed. I've updated the PR to now state that the enum MUST NOT be added to the list of supported modes unless the UA implements the AR module, which I think gives us all the clarification we need.

You are still referring normatively to the other spec which you **can't** do

> I can't see this particular enum growing much larger than this over the lifetime of the API

as @dino mentioned, other places on the web platform use DOMString if they want to keep things forward compatible. Having an enum doesn't buy us anything except unwanted spec and runtime behavior.

rcabanier: > A `DOMString` could make sense if that's the only mechanism to have `isSessionSupported` return false for unknown future values. Is throwing an exception on unknown enum values required by WebIDL? 

Yes
If we put a DOMString in, almost nothing will caused an exception (See javascript's [toString](https://tc39.es/ecma262/#sec-tostring))

> If so, are input enums basically non-extensible in WebIDL?

Yes.

PR #882, opened 2019-10-16
Note explaining why API shares canvas GL contexts

### LINKED ISSUE #861: Add explainer discussion of why webxr shares its WebGL context with a canvas
Working through the implementation of opaque framebuffers in Servo, which is make more complex by webxr sharing WebGL contexts with canvas. Any canvas element with an XR-compatible WebGL context now has two potential targets: the canvas and the webxr device, and since we're using double-buffering and the cadence for webGL rAF and webxr rAF are independent, this means four buffers to juggle.

Looking at the explainer, I don't see any discussion of this design choice, can we add one?

toji:
/fixes #861

Explains the rational behind why XRWebGLlayer uses an existing WebGL
context rather than creating it's own.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -244,6 +244,8 @@ let gl = glCanvas.getContext("webgl", { xrCompatible: true });
 
 Ensuring context compatibility with an XR device through either method may have side effects on other graphics resources in the page, such as causing the entire user agent to switch from rendering using an integrated GPU to a discrete GPU.
 
+> **Note:** The `XRWebGLLayer` uses a WebGL context created by a Canvas element or `OffscreenCanvas` rather than creating it's own to both allow for the same content to be rendered to the XR device and the page, as well as allowing the page to load it's WebGL resorces prior to the session being created.
+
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -244,6 +244,8 @@ let gl = glCanvas.getContext("webgl", { xrCompatible: true });
 
 Ensuring context compatibility with an XR device through either method may have side effects on other graphics resources in the page, such as causing the entire user agent to switch from rendering using an integrated GPU to a discrete GPU.
 
+> **Note:** The `XRWebGLLayer` uses a WebGL context created by a Canvas element or `OffscreenCanvas` rather than creating it's own to both allow for the same content to be rendered to the XR device and the page, as well as allowing the page to load it's WebGL resorces prior to the session being created.
```
Manishearth:
```suggestion
> **Note:** The `XRWebGLLayer` uses a WebGL context created by a Canvas element or `OffscreenCanvas` rather than creating its own to both allow for the same content to be rendered to the XR device and the page, as well as allowing the page to load it's WebGL resources prior to the session being created.
```

PR #883, opened 2019-10-16
Add a link to the Gamepads module

### LINKED ISSUE #803: Main spec should link to Gamepad module
It would be helpful to have the main spec link to the WebXR Gamepad module spec documentation, probably in the XRInputSource section: https://immersive-web.github.io/webxr/#xrinputsource-interface

Otherwise, people reading the main spec might be unaware that there can potentially be Gamepads on the XRInputSource objects. Even if they are aware of this, the official documentation is difficult to find.

toji:
/fixes #803

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1465,6 +1465,8 @@ interface XRInputSource {
 };
 </pre>
 
+Note: The {{XRInputSource}} interface is also extended by the <a href="https://immersive-web.github.io/webxr-gamepads-module/">WebXR Gamepads Module</a>
+
 The <dfn attribute for="XRInputSource">handedness</dfn> attribute describes which hand the [=XR input source=] is associated with, if any. Input sources with no natural handedness (such as headset-mounted controls) or for which the handedness is not currently known MUST set this attribute {{XRHandedness/"none"}}.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1465,6 +1465,8 @@ interface XRInputSource {
 };
 </pre>
 
+Note: The {{XRInputSource}} interface is also extended by the <a href="https://immersive-web.github.io/webxr-gamepads-module/">WebXR Gamepads Module</a>
```
Manishearth: bikeshed lets you do ``[[!webxr-gamepads-1]]`` or ``[[!webxr-gamepads-1|WebXR Gamepads Module]]`

Manishearth: I think this makes it work better in the references section

toji: Ah, good to know! It appears that the gamepads module (who's identifier should be `webxr-gamepads-module-1`, FWIW) isn't known to bikeshed yet, though.

Manishearth: I'm seeing it in my local bikeshed data, going to poke at this a bit

Manishearth: looks like they need to be in biblio-keys.json

Manishearth: Yeah so it seems like bikeshed is aware of the spec but not aware _enough_ to generate proper bibliography entries. Pinged Tab on the issue, they'll probably know why

Manishearth: For now we should merge this

PR #885, opened 2019-10-16
Export a dfn for input-source-change

Manishearth:
Makes it easier to refer to from the gamepads module

r? @toji


toji:
LGTM! Merging.

PR #888, opened 2019-10-17
Better define how depthNear and depthFar are used.

### LINKED ISSUE #881: Define "near clip plane" and "far clip plane"
These terms are used without being defined. Is the plane parallel to the display surface (which is what I initially assumed) or is it also transformed so that it is perpendicular to the floor (which I suspect might be the case)?

By "transformed so that it is perpendicular to the floor" I mean that if the clip plane is parallel to the display surface when the user is looking straight ahead without any rotation up or down, it does not change if the user looks up or down.

Or is there some other definition?

toji:
/fixes #881

Provides a bit more explanation as to what depthNear/Far mean and why
they're specified when other aspects of the projection matrix are not.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -801,7 +801,7 @@ When an {{XRRenderState}} object is created for an {{XRSession}} |session|, the 
 
 </div>
 
-The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the distance, in meters, of the near clip plane from the viewer. The <dfn attribute for="XRRenderState">depthFar</dfn> attribute defines the distance, in meters, of the far clip plane from the [=viewer=].
+The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the distance, in meters, of the near clip plane from the [=viewer=]. The <dfn attribute for="XRRenderState">depthFar</dfn> attribute defines the distance, in meters, of the far clip plane from the [=viewer=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -801,9 +801,11 @@ When an {{XRRenderState}} object is created for an {{XRSession}} |session|, the
 
 </div>
 
-The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the distance, in meters, of the near clip plane from the viewer. The <dfn attribute for="XRRenderState">depthFar</dfn> attribute defines the distance, in meters, of the far clip plane from the viewer.
+The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the distance, in meters, of the near clip plane from the viewer. The <dfn attribute for="XRRenderState">depthFar</dfn> attribute defines the distance, in meters, of the far clip plane from the [=viewer=].
```
Manishearth:
```suggestion
The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the distance, in meters, of the near clip plane from the [=viewer=]. The <dfn attribute for="XRRenderState">depthFar</dfn> attribute defines the distance, in meters, of the far clip plane from the [=viewer=].
```

toji: Wow... gold star to me for missing that one. Especially when I managed to properly link the same term **on the same line**. 🙄

PR #893, opened 2019-10-21
Define squeeze events

### LINKED ISSUE #876: Expose events for a squeeze(grip) action
(See https://github.com/immersive-web/webxr-gamepads-module/issues/2 for the "squeeze" naming if unfamiliar)

A couple of us discussed this a bit at TPAC: We think it would be useful to expose a set of `squeeze` events like we have `select` events, for a "primary squeeze" action (currently [defined in the gamepads module](https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping)). This basically exposes an event for `buttons[1]` (if nonempty) just as `select` is an event for `buttons[0]`. Devices would not be required to trigger this event (unlike the primary action, which is mandatory).

Select and squeeze seem to be the two main webxr-exposed buttons that are common across most devices and have familiar and uniform meanings (it wouldn't make as much sense to expose, say, a "B button pressed" event)

This is also good for AR devices that wish to expose hands as input sources: these hand-input-sources do not have gamepads, but can have select or squeeze events. We eventually need to come up with a way to expose articulated hand tracking, but a lot of applications need just one or both of these events to work, without needing gamepad support.

cc @thetuvix @NellWaliczek

Manishearth:
/fixes #876

This adds an optional "primary squeeze" mirroring the "primary action". Once this lands we can backreference it [from the gamepads module](https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping).

This also cleans up the primary action spec text to make the primary action a requirement.

cc @thetuvix

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1536,7 +1536,7 @@ When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=prim
 
 Sometimes platform-specific behavior can result in a [=primary action=] being interrupted or cancelled. For example, a [=/XR input source=] may be removed from the [=XRSession/XR device=] after the [=primary action=] is started but before it ends.
 
-Each [=XR input source=] MAY define a <dfn>primary squeeze action</dfn>. The [=primary squeeze action=] is a platform-specific action that, when engaged, produces {{XRSession/squeezestart}}, {{XRSession/squeezeend}}, and {{XRSession/squeeze}} events. The [=primary squeeze action=] should be used for actions rougly mapping to squeezing or grabbing. Examples of possible [=primary squeeze action=]s are pressing a grip trigger or making a grabbing hand gesture. If the platform guidelines define a recommended primary squeeze action then it should be used as the [=primary squeeze action=], otherwise the user agent MAY select one.
+Each [=XR input source=] MAY define a <dfn>primary squeeze action</dfn>. The [=primary squeeze action=] is a platform-specific action that, when engaged, produces {{XRSession/squeezestart}}, {{XRSession/squeezeend}}, and {{XRSession/squeeze}} events. The [=primary squeeze action=] should be used for actions roughly mapping to squeezing or grabbing. Examples of possible [=primary squeeze action=]s are pressing a grip trigger or making a grabbing hand gesture. If the platform guidelines define a recommended primary squeeze action then it should be used as the [=primary squeeze action=], otherwise the user agent MAY select one.
 
 <div class="algorithm" data-algorithm="on-squeeze-start">
 
--- a/index.bs
+++ b/index.bs
@@ -1515,8 +1515,6 @@ Note: {{XRInputSource}}s in an {{XRSession}}'s {{XRSession/inputSources}} array 
 
 Each [=XR input source=] SHOULD define a <dfn>primary action</dfn>. The [=primary action=] is a platform-specific action that, when engaged, produces {{XRSession/selectstart}}, {{XRSession/selectend}}, and {{XRSession/select}} events. Examples of possible [=primary action=]s are pressing a trigger, touchpad, or button, speaking a command, or making a hand gesture. If the platform guidelines define a recommended primary input then it should be used as the [=primary action=], otherwise the user agent is free to select one.
 
-
-
 <div class="algorithm" data-algorithm="on-input-start">
 
 When an [=XR input source=] for {{XRSession}} |session| begins its [=primary action=] the UA MUST run the following steps:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1526,6 +1538,29 @@ When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=prim
 
 Sometimes platform-specific behavior can result in a [=primary action=] being interrupted or cancelled. For example, a [=/XR input source=] may be removed from the [=XRSession/XR device=] after the [=primary action=] is started but before it ends.
 
+Each [=XR input source=] MAY define a <dfn>primary squeeze</dfn>. The [=primary squeeze=] is a platform-specific action that, when engaged, produces {{XRSession/squeezestart}}, {{XRSession/squeezeend}}, and {{XRSession/squeeze}} events. The [=primary squeeze=] should be used for actions rougly mapping to squeezing or grabbing. Examples of possible [=primary squeeze=]s are pressing a grip trigger or making a grabbing hand gesture. If the platform guidelines define a recommended primary squeeze then it should be used as the [=primary squeeze=], otherwise the user agent is free to select one.
```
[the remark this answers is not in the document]
Manishearth: oh don't worry i have been finding the terminology "primary squeeze" hilarious as well, but i'm not too worried about it.

the retcon seems fine, though this may require further edits to the gamepad module as well as the input profiles repo for that. "primary squeeze action" seems okay though, i'll make that change

Manishearth:
Addressed

index.bs
```diff
@@ -553,10 +553,13 @@ enum XRVisibilityState {
...
   // Events
   attribute EventHandler onend;
-  attribute EventHandler onselect;
   attribute EventHandler oninputsourceschange;
+  attribute EventHandler onselect;
   attribute EventHandler onselectstart;
   attribute EventHandler onselectend;
+  attribute EventHandler onsqueeze;
+  attribute EventHandler onsqueezestart;
+  attribute EventHandler onsqueezeend;
```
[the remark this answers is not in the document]
Aadjou: it will also be hard for non-native developers to write. Especially without the aid of camelcase or similar.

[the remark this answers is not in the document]
Aadjou: it will also be hard for non-native developers to write. Especially without the aid of camelcase or similar.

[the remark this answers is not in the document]
NellWaliczek: I agree they're hard to read.  Unfortunately, IIRC, this casing convention isn't one our WG decided on; we're just following the general web platform guidance.  Given that, I'm not sure what else can be done...

index.bs
```diff
@@ -1526,6 +1536,29 @@ When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=prim
 
 Sometimes platform-specific behavior can result in a [=primary action=] being interrupted or cancelled. For example, a [=/XR input source=] may be removed from the [=XRSession/XR device=] after the [=primary action=] is started but before it ends.
 
+Each [=XR input source=] MAY define a <dfn>primary squeeze action</dfn>. The [=primary squeeze action=] is a platform-specific action that, when engaged, produces {{XRSession/squeezestart}}, {{XRSession/squeezeend}}, and {{XRSession/squeeze}} events. The [=primary squeeze action=] should be used for actions rougly mapping to squeezing or grabbing. Examples of possible [=primary squeeze action=]s are pressing a grip trigger or making a grabbing hand gesture. If the platform guidelines define a recommended primary squeeze action then it should be used as the [=primary squeeze action=], otherwise the user agent is free to select one.
```
NellWaliczek: >Each [=XR input source=] MAY define 

Why is this event "MAY" and the other "SHOULD"?

Manishearth: We require a primary action, but I don't think we need to require a primary squeeze action.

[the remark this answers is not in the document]
Manishearth: Yes, the gamepads module already defines "primary squeeze", i'd like to edit it to point to this once this lands.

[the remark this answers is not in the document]
Manishearth: https://github.com/immersive-web/webxr-gamepads-module/pull/24

NellWaliczek: Does that mean the trigger action should be MUST?

Manishearth: Hmm. Come to think of it, I thought it already was MUST somewhere in the spec. I don't mind changing it here, given that we're kinda requiring this in the gamepads spec.

index.bs
```diff
@@ -1526,6 +1536,29 @@ When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=prim
 
 Sometimes platform-specific behavior can result in a [=primary action=] being interrupted or cancelled. For example, a [=/XR input source=] may be removed from the [=XRSession/XR device=] after the [=primary action=] is started but before it ends.
 
+Each [=XR input source=] MAY define a <dfn>primary squeeze action</dfn>. The [=primary squeeze action=] is a platform-specific action that, when engaged, produces {{XRSession/squeezestart}}, {{XRSession/squeezeend}}, and {{XRSession/squeeze}} events. The [=primary squeeze action=] should be used for actions rougly mapping to squeezing or grabbing. Examples of possible [=primary squeeze action=]s are pressing a grip trigger or making a grabbing hand gesture. If the platform guidelines define a recommended primary squeeze action then it should be used as the [=primary squeeze action=], otherwise the user agent MAY select one.
```
thetuvix: Nit: "rougly"

thetuvix: In the gamepad spec, will we then specify that for `"xr-standard"`, the squeeze action must map to button 1?

Manishearth: Yep! https://github.com/immersive-web/webxr-gamepads-module/pull/24

PR #895, opened 2019-10-25
Corrections of "it's" to "its"

a2sheppy:
Somewhat nitpicky I suppose. If it had just been one, I might have ignored it, but once I saw the second and then more, I started getting twitchy. Ah, the live of a professional writer with a severe case of OAD (Obsessive Apostrophe Disorder). All places which previously used "it's" that should have been "its" are corrected by this change.


toji:
I'm going to just assume these were all my fault and apologize profusely. 😞 

Thanks! 😉

PR #897, opened 2019-10-28
Tighten up spec text around frame timings, add hook for frame updates

Manishearth:
This makes it easier to do things like in https://github.com/immersive-web/webxr-gamepads-module/pull/22 , where we want to be able to insert an action that occurs on the main thread whenever a frame is about to be processed.

r? @NellWaliczek

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1536,7 +1536,7 @@ Each [=XR input source=] MUST define a <dfn>primary action</dfn>. The [=primary 
 
 When an [=XR input source=] |source| for {{XRSession}} |session| begins its [=primary action=] the UA MUST run the following steps:
 
-  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the event occurred.
+  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the action occurred.
   1. [=Queue a task=] to [=fire an input source event=] with name {{selectstart!!event}}, frame |frame|, and source |source|.
 
 </div>
@@ -1545,7 +1545,7 @@ When an [=XR input source=] |source| for {{XRSession}} |session| begins its [=pr
 
 When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
 
-  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the event occurred.
+  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the action occurred.
   1. [=Queue a task=] to perform the following steps:
     1. [=Fire an input source event=] with name {{select!!event}}, frame |frame|, and source |source|.
     1. [=Fire an input source event=] with name {{selectend!!event}}, frame |frame|, and source |source|.
@@ -1560,7 +1560,7 @@ Each [=XR input source=] MAY define a <dfn>primary squeeze action</dfn>. The [=p
 
 When an [=XR input source=] |source| for {{XRSession}} |session| begins its [=primary squeeze action=] the UA MUST run the following steps:
 
-  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the event occurred.
+  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the action occurred.
   1. [=Queue a task=] to [=fire an input source event=] with name {{squeezestart!!event}}, frame |frame|, and source |source|.
 
 </div>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -863,7 +863,7 @@ When this method is invoked, the user agent MUST run the following steps:
 
 <div class="algorithm" data-algorithm="run-animation-frames">
 
-When an {{XRSession}} |session| receives updated [=viewer=] state from the [=XRSession/XR device=], it runs an <dfn>XR animation frame</dfn> with a timestamp |now| and an {{XRFrame}} |frame|, which MUST run the following steps regardless of if the [=list of animation frame callbacks=] is empty or not:
+When an {{XRSession}} |session| receives updated [=viewer=] state from the [=XRSession/XR device=], it runs an <dfn>XR animation frame</dfn> with a timestamp |now| and an {{XRFrame}} |frame| with [=XRFrame/time=] |now|, which MUST run the following steps regardless of if the [=list of animation frame callbacks=] is empty or not:
```
toji: This doesn't feel like the right way to specify this. `|now|` and `|frame|` are both inputs to the algorithm, but the new text feels like an assignment happening in the algorithm description.

I'd suggest either:
 - Assigning `[=XRFrame/time=]` to `|now|` explicitly in the algorithm or
 - Removing |now| from the algorithm description altogether and using [=XRFrame/time=] in it's place elsewhere in the algorithm.

Manishearth: Done

toji: Hm... I think we went to far in the other direction. We still need to have a declaration of the `XRFrame` somewhere.

Manishearth: Not sure what you mean by that, it's constructed in the algorithm itself (as it should be, IMO, previously it just appeared out of the ether)

Manishearth: Oh, my bad, i pushed up a partial commit because I forgot to save

toji: 🤣

index.bs
```diff
@@ -2004,7 +2021,9 @@ The <dfn attribute for="XRInputSourceEvent">frame</dfn> attribute is an {{XRFram
 When the user agent has to <dfn>fire an input source event</dfn> with name |name|, {{XRFrame}} |frame|, and {{XRInputSource}} |source| it MUST run the following steps:
 
   1. Create an {{XRInputSourceEvent}} |event| with {{Event/type}} |name|, {{XRInputSourceEvent/frame}} |frame|, and {{XRInputSourceEvent/inputSource}} |source|
+  1.
```
toji: Accidental +1?

index.bs
```diff
@@ -995,6 +999,17 @@ When this method is invoked, the user agent MUST run the following steps:
...
+
+<div class="algorithm" data-algorithm="apply-frame-updates">
+
+To <dfn for=XRFrame>apply frame updates</dfn> for an {{XRFrame}}, the user agent MUST run the following steps:
+
+ 1. Do nothing.
+
+</div>
+
+NOTE: Further specs may add new kinds of frame updates to these steps, e.g. for gamepad state.
```
NellWaliczek: This seems somewhat odd to me, but I'm not sure what else to suggest.  Is there someone with experience with this sort of problem that we could get a sanity check from?

index.bs
```diff
@@ -1519,17 +1534,18 @@ Each [=XR input source=] MUST define a <dfn>primary action</dfn>. The [=primary
 
 <div class="algorithm" data-algorithm="on-input-start">
 
-When an [=XR input source=] for {{XRSession}} |session| begins its [=primary action=] the UA MUST run the following steps:
+When an [=XR input source=] |source| for {{XRSession}} |session| begins its [=primary action=] the UA MUST run the following steps:
 
-  1. [=Queue a task=] to [=fire an event|fire=] an {{XRInputSourceEvent}} named {{selectstart!!event}} on |session|.
+  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the event occurred.
```
NellWaliczek: The sentence above calls it an "action" so perhaps this should be:
> the time the action occurred

index.bs
```diff
@@ -1519,17 +1534,18 @@ Each [=XR input source=] MUST define a <dfn>primary action</dfn>. The [=primary
...
+  1. [=Queue a task=] to [=fire an input source event=] with name {{selectstart!!event}}, frame |frame|, and source |source|.
 
 </div>
 
 <div class="algorithm" data-algorithm="on-input-end">
 
 When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
 
-  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| for the time the event occurred.
+  1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the event occurred.
```
NellWaliczek: Same as above

index.bs
Manishearth: An alternate plan was to have a "list of frame updates" (a list of callbacks) on the session and session with gamepads append a frame update to it. I felt that might get a bit unweildy though.

Manishearth:
Updated.

PR #902, opened 2019-11-08
Fix indentation

Manishearth:
The IsDetachedBuffer line isn't rendering as indented, which changes the meaning of the spec.

r? @toji


toji:
LGTM. I'll merge it.

PR #903, opened 2019-11-08
Unset animationFrame boolean at end of frame

lojjic:
The XRFrame's private animationFrame boolean is set to true during the frame handler but never set back to false. This seems like the right place to add it.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -881,6 +881,7 @@ When an {{XRSession}} |session| receives updated [=viewer=] state from the [=XRS
     1. If the entry's [=cancelled=] boolean is <code>true</code>, continue to the next entry.
     1. [=Invoke the Web IDL callback function=], passing |now| and |frame| as the  arguments
     1. If an exception is thrown, [=report the exception=].
+  1. Set |frame|'s [=animationFrame=] boolean to <code>false</code>.
   1. Set |frame|'s [=active=] boolean to <code>false</code>.
 
 </div>


Manishearth:
r+, not sure if this can be marked as non substantiative for IPR purposes: https://labs.w3.org/repo-manager/pr/id/immersive-web/webxr/903

toji:
LGTM. The logic previously would have held regardless, because the one function where this matters (`getViewerPose()`) would have given the same error due to the `active` boolean being false, which may have been my original reasoning for not explicitly flipping this boolean back. It feels like a reasonable clarifiaction regardless.

That said, according to https://labs.w3.org/repo-manager/pr/id/immersive-web/webxr/903 I don't think we can consider this non-substantiative due to the following:

> if the contribution does not concern a normative part of a specification, or is editorial in nature (e.g. fixing typos or examples), the contribution can be marked as non-substantive

Since this actually alters a normative algorithm, I think we'd have to satisfy the IPR requirement before accepting the change. Sorry, I know that's a pain. Are you a member of the W3C (or are a member of an organization that is?) If so you need to link your GitHub account with a W3C account.

lojjic:
Thanks @toji, no I'm not a W3C member. I could probably get linked up as an employee of Verizon, but that hardly seems worth the effort for this one-liner. Perhaps I'll just open this as an issue instead and let someone else add it?

lojjic:
I've opened issue #905 for further discussion, and am closing this PR.

PR #917, opened 2019-11-14
removed Querying consent section per #900

avadacatavra:


toji:
LGTM. Merging since we have both my approval and @Manishearth's

Issue #696, opened 2019-06-13
Specify the relationship between session visibilityState and window state


toji:
A concern that we've talked about before which reemerged during #687 is specifying exactly how XRSessions should react to various changes in the browser state and, conversely, how the browser should react to sessions being displayed. This should be specced as normatively as possible to ensure consistent behavior between browsers.

For example: What happens to the window visibilityState and rAF loop when an immersive session starts on a single-display device (mobile or standalone)? What about on multi-display devices (desktop tethered headsets)? What happens to the XRSession when you minimize a window? What about when you switch windows? Switch tabs? Does the session end, does it's visibilityState change, or does it just keep running normally?

billorr-google:
Are there already thoughts on this issue, and it just needs to be written up, or does it need thinking through?

For sensor-backed inline sessions, I think we should require frame focus to give out pose data (or match the requirements that sensor APIs have).  I could see an argument for still calling raf callbacks when the frame is not focused.

For immersive sessions with external monitor, I think visibility in-headset should be sufficient to give out pose data and continue the session.  A key reason that window focus shouldn't be required is that several VR systems will take focus to display their own UI.  If someone minimizes the window while a session is occurring, the user in-headset doesn't know that and doesn't care.  However, if the headset has a proximity sensor, and detects no user in-headset, we could save resources by bluring/hiding the session.  Similarly, page visibility should report whether the 2d page is visible, not whether the vr page is visible in-headset.

For immersive sessions without an external monitor, I think the page is not visible while the session is visible.  Window raf could be paused while the session is visible instead of the page.

If we're eventually going to add things like DOM overlay or DOM layers, it may make sense to keep window raf firing, and keep the page reporting as visible for immersive sessions.

thetuvix:
Specifically, do we expect UAs to transition to a `session.visibilityState` of `"hidden"` when a user of a desktop VR peripheral takes off their headset?

Per the spec text, UAs may wire things up this way, since the headset's view at that point "cannot be seen by the user".  This is how Chromium has currently [wired things up on Windows MR headsets](https://chromium-review.googlesource.com/c/chromium/src/+/1790422), for example.  However, I would not be surprised if this differs across other XR backends in Chromium (and further, across other UAs).

[...]

PR #919, opened 2019-11-15
summarized #696 (no spec enforced relationship) and #724 for focus/vis

avadacatavra:
I didn't really capture anything from #696 based on @toji's recent comments.

Happy to add more information if anyone has suggestions


toji:
LGTM! Thank you.

Issue #913, opened 2019-11-14
Context loss handling in makeXRCompatible does not make sense


bzbarsky:
https://immersive-web.github.io/webxr/#dom-webglrenderingcontextbase-makexrcompatible step 1.8.1 says:

> Force context to be lost and handle the context loss as described by the WebGL specification.

where "handle the context loss" links to https://www.khronos.org/registry/webgl/specs/latest/1.0/#CONTEXT_LOST

Those steps set some flags, then queue a task to fire an event.  But step 1.8.2 seems to assume the event has already fired (it hasn't; there's just a task queued to fire it).  Furthermore, when that task runs, it will restore the drawing buffer, but step 1.8.3 does that too.

This section seems like it wants to reuse parts of the default webgl behavior, but not all of it.  The right way to do this is probably to factor out the to-be-shared behavior (if possible), or something.  But the current setup really doesn't make sense.

PR #921, opened 2019-11-15
Multiple fixes to XR compatibility algorithms

### LINKED ISSUE #908: makeXRCompatible mutates object state directly in a parallel section
https://www.w3.org/TR/webxr/#dom-webglrenderingcontextbase-makexrcompatible step 1.6.1 is:

> Set context’s XR compatible boolean to false.

but that state is synchronously observable (e.g. via whether the `XRWebGLLayer` constructor throws).  As a result, if the web page does:

    ctx.makeXRCompatible();
    var layer = new XRWebGLLayer(session, ctx);

this will sometimes throw and sometimes not, depending on the scheduling of the parallel section vs the main thread of execution.  Or maybe it will read a partially-updated value, depending on the exact hardware architecture and memory model...

In any case, parallel sections should not be synchronously touching (reading _or_ writing) state.  They should queue tasks (and say which document and task queue those are associated with!) to do that.

### LINKED ISSUE #910: What happens in getContext if there is no immersive XR device?
If after "Ensure an immersive XR device is selected" the immersive XR device is null, what should happen?  Should the result still return a WebGL context with the "XR compatible" boolean set to true?  That seems pretty odd, but is what's specced right now...

### LINKED ISSUE #911: makeXRCompatible seems to have an unnecessary parallel section
https://immersive-web.github.io/webxr/#dom-webglrenderingcontextbase-makexrcompatible has a parallel section that does the following:

* Check the context lost flag; this needs to happen outside a parallel section anyway.
* Check the XR compatible flag.  This needs to happen outside a parallel section anyway, so there is no race with mutators of the flag.
* "Ensure an immersive XR device is selected", which can be done without going parallel as in the changes for the `getContext` method.  Though this part doesn't even make sense; see https://github.com/immersive-web/webxr/issues/909
* _Sets_ the XR compatible flag.  This needs to happen outside a parallel section.
* Queues a task.  That clearly doesn't need to be in a parallel section, though it could be...

### LINKED ISSUE #912: makeXRCompatible doesn't specify what task source its task uses
https://immersive-web.github.io/webxr/#dom-webglrenderingcontextbase-makexrcompatible step 1.8 says nothing about the task source.  It should.  It should also say something about the document association.

This is a general problem through the whole spec: none of the task bits say what sources are used.

toji:
Based on feedback from @bzbarsky, this pull request contains fixes for the following issues:

/fixes #910

Ensures that the logic that the getContext algorithm follows more closely
the logic and results of the makeXRCompatible algorithm.

/fixes #911
/fixes #908

Removes the requirement that a part of the makeXRCompatible algorithm
be performed in parallel, which both simplifies the logic and
solves some concurrency issues with setting the XR compatibile boolean.

/fixes #912

Specifies the task source for makeXRCompatible task.
We'll also want to do a larger review of the spec to determine the task
sources for any other tasks that we're queueing.

In addition to the usual editors cross review, I'd especially appreciate @bzbarsky's feedback and believe that @foolip may also want to take a look.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -52,6 +52,7 @@ spec: WebGL; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.0/
     type: dfn; text: Restore the context; url: #restore-the-drawing-buffer
     type: dfn; text: actual context parameters; url: #actual-context-parameters
     type: dfn; text: create a drawing buffer; url: #create-a-drawing-buffer
+    type: dfn; text: WebGL task source; url: #5.15
 spec: WebGL 2.0; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/2.0/
     type: interface; text: WebGL2RenderingContext; url: WebGL2RenderingContext
 spec: Orientation Sensor; urlPrefix: https://w3c.github.io/orientation-sensor/
@@ -1921,6 +1922,7 @@ When this method is invoked, the user agent MUST run the following steps:
               1. [=Restore the context=] on a [=compatible graphics adapter=] for the [=XR/immersive XR device=].
               1. Set |context|'s [=XR compatible=] boolean to <code>true</code>.
               1. [=/Resolve=] |promise|.
+           The [=/task source=] for this task is the [=/WebGL task source=].
     </dl>
   1. Return |promise|.
 
--- a/index.bs
+++ b/index.bs
@@ -1916,13 +1916,12 @@ When this method is invoked, the user agent MUST run the following steps:
       <dt> If |context| was created on a [=compatible graphics adapter=] for the [=XR/immersive XR device=]
       <dd> Set |context|'s [=XR compatible=] boolean to <code>true</code> and [=/resolve=] |promise|.
       <dt> Otherwise
-      <dd> [=Queue a task=] to perform the following steps:
+      <dd> [=Queue a task=] on the [=/WebGL task source=] to perform the following steps:
               1. Force |context| to be lost and [=handle the context loss=] as described by the WebGL specification.
               1. If the [=canceled flag=] of the "webglcontextlost" event fired in the previous step was not set, [=reject=] |promise| with an {{AbortError}} and abort these steps.
               1. [=Restore the context=] on a [=compatible graphics adapter=] for the [=XR/immersive XR device=].
               1. Set |context|'s [=XR compatible=] boolean to <code>true</code>.
               1. [=/Resolve=] |promise|.
-           The [=/task source=] for this task is the [=/WebGL task source=].
     </dl>
   1. Return |promise|.
 
--- a/index.bs
+++ b/index.bs
@@ -1867,9 +1867,14 @@ The [=XR compatible=] boolean can be set either at context creation time or afte
 When the {{HTMLCanvasElement}}'s {{HTMLCanvasElement/getContext()}} method is invoked with a {{WebGLContextAttributes}} dictionary with {{xrCompatible}} set to <code>true</code>, run the following steps:
 
   1. [=ensures an immersive XR device is selected|Ensure an immersive XR device is selected=].
-  1. [=Create the WebGL context=] as usual, ensuring it is created on a [=compatible graphics adapter=] for the [=XR/immersive XR device=].
-  1. Let |context| be the newly created WebGL context.
-  1. Set |context|'s [=XR compatible=] boolean to true.
+  1. If the [=XR/immersive XR device=] is <code>null</code>:
+    1. [=Create the WebGL context=] as usual.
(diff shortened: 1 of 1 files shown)


bzbarsky:
@toji Is there a draft version of the spec with these changes merged in where I could see them in context?

toji:
There is not (and unfortunately it's not as trivial to do so as it should be), but I can fix that. Give me a moment.

index.bs
```diff
@@ -1899,23 +1905,27 @@ The <dfn method for="WebGLRenderingContextBase">makeXRCompatible()</dfn> method
...
+      <dd> Set |context|'s [=XR compatible=] boolean to <code>false</code> and [=reject=] |promise| with an {{InvalidStateError}}.
+      <dt> If the [=XR/immersive XR device=] is <code>null</code>
+      <dd> Set |context|'s [=XR compatible=] boolean to <code>false</code> and [=reject=] |promise| with an {{InvalidStateError}}.
+      <dt> If |context|'s [=XR compatible=] boolean is <code>true</code>
+      <dd> [=/Resolve=] |promise|.
+      <dt> If |context| was created on a [=compatible graphics adapter=] for the [=XR/immersive XR device=]
+      <dd> Set |context|'s [=XR compatible=] boolean to <code>true</code> and [=/resolve=] |promise|.
+      <dt> Otherwise
+      <dd> [=Queue a task=] to perform the following steps:
+              1. Force |context| to be lost and [=handle the context loss=] as described by the WebGL specification.
```
bzbarsky: OK, so https://github.com/immersive-web/webxr/issues/913 is still outstanding, right?  That's fine, since this PR is not claiming to fix that.

index.bs
```diff
@@ -1899,23 +1905,27 @@ The <dfn method for="WebGLRenderingContextBase">makeXRCompatible()</dfn> method
...
+      <dt> If |context|'s [=WebGL context lost flag=] is set
+      <dd> Set |context|'s [=XR compatible=] boolean to <code>false</code> and [=reject=] |promise| with an {{InvalidStateError}}.
+      <dt> If the [=XR/immersive XR device=] is <code>null</code>
+      <dd> Set |context|'s [=XR compatible=] boolean to <code>false</code> and [=reject=] |promise| with an {{InvalidStateError}}.
+      <dt> If |context|'s [=XR compatible=] boolean is <code>true</code>
+      <dd> [=/Resolve=] |promise|.
+      <dt> If |context| was created on a [=compatible graphics adapter=] for the [=XR/immersive XR device=]
+      <dd> Set |context|'s [=XR compatible=] boolean to <code>true</code> and [=/resolve=] |promise|.
+      <dt> Otherwise
+      <dd> [=Queue a task=] to perform the following steps:
```
bzbarsky: It'd be better to say "Queue a task on the WebGL task source to ...", I think, instead of putting the task source part completely separate from this operation.

Also, this still needs to define the document for this task...

index.bs
toji: That's correct. Trying to keep the reviews manageable in size.

index.bs
toji: Ah, interesting. I was primarily mimicking the pattern that I saw repeated throughout the HTML standard. For example:
 - https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-toblob
 - https://html.spec.whatwg.org/multipage/interactive-elements.html#details-notification-task-steps
 - https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-fragid:task-source

Happy to change it, though. (Also none of those examples define a document, unless I'm overlooking something, so it wasn't clear what was required in that case.)

bzbarsky: The HTML standard has large parts that predate the "modern" formulation of queueing tasks, and also some sections where lots of tasks get queued and it tries to coalesce the "which task source" verbiage.

In practice, this is really annoying to deal with and makes it hard to find what task source to use for tasks, and I wish the HTML standard would stop doing that.

And yes, most things don't documents at the moment, but that's being worked on.  See https://github.com/whatwg/html/issues/4980

toji: Thanks for the explanation! If you don't mind providing a bit more guidance, I've got a followup question.

In this specific case it seems like the appropriate document is the document which created the canvas element which created the WebGL context, unless the context was created by an OffscreenCanvas, in which case I'm not sure how to reason about what the appropriate document is? (It may have been created in a worker.)

bzbarsky: Right, if we have an element it's easy.  If we don't, then I don't know enough about the relevant APIs to tell you.  What does Chrome implement, since they actually do document-association of tasks?

toji:
I'm going to merge this now since it clearly gets us to a better place with regards to the issues @bzbarsky has raised, with the intent to follow up with the remaining issues (including the declaration of the task document) in a future PR.

PR #922, opened 2019-11-15
Deduplicate paragraph for primary action cancel

klausw:
This paragraph appeared twice, both before and after the "on-squeeze-end" algorithm, likely as a result of a merge conflict. Delete the first copy of it, keeping the second one before the "on-input-cancelled" algorithm.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1541,8 +1541,6 @@ When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=prim
 
 </div>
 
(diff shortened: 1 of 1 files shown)


klausw:
FYI, the retained copy (which also covers squeeze events) is at line 1567, just outside the default context view:

https://github.com/immersive-web/webxr/pull/922/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dL1567

Manishearth:
My fault, thanks for the fix!

PR #929, opened 2019-11-20
Split input sources into primary/auxiliary

### LINKED ISSUE #920: Support input sources without primary action?
https://immersive-web.github.io/webxr/#xrinputsource-interface currently says:

> Each XR input source MUST define a primary action. The primary action is a platform-specific action that, when engaged, produces selectstart, selectend, and select events. Examples of possible primary actions are pressing a trigger, touchpad, or button, speaking a command, or making a hand gesture. If the platform guidelines define a recommended primary input then it should be used as the primary action, otherwise the user agent is free to select one.

Does this really need to be so restrictive? I think there are valid use cases for XR input sources that don't have primary actions.

For example, the [VIVE Tracker](https://www.vive.com/us/vive-tracker/) provides 6DoF poses, but doesn't have any gamepad-style input at all. It would be useful to be able to support such trackers in WebXR, for example for tracked props or as body/leg trackers.

Separately, for transient touch input, I think it would be useful to provide multitouch input, but I think it would be confusing if each finger would produce distinct select events, especially if those end up arriving in orders that applications aren't expecting. (For example, the event sequence could be selectstart#0, selectstart#1, selectend#0, selectend#1). If additional fingers were reported as additional transient input sources, but without select events, multitouch-aware applications could opt in to using them based on inputsourceschange events, but they wouldn't confuse applications that are designed around a single primary input source.

Would it be OK to rephrase that? Maybe something along these lines?

XR input sources are either *primary input sources* or *auxiliary input sources*. The system MUST provide at least one primary input source. A primary input source MUST define a primary action. The primary action is [...]

An XR input device is an *auxiliary input source* if it does not have buttons or other affordances suitable as a primary action. They MAY be provided in addition to primary input sources.

Note: while an overall XR system is required to support at least one primary input source, users are of course free to design experiences that use only auxiliary controllers, or no controllers at all.

klausw:
This introduces new definitions for primary/auxiliary XR input sources, and "transient action" for transient input device to support a transient input device that isn't a primary input source.

Fixes #920


index.bs
```diff
@@ -1589,43 +1591,43 @@ When an [=XR input source=] |source| for {{XRSession}} |session| has its [=prima
...
 
-[=Transient input sources=] follow a slightly different sequence when firing [=primary action=] events:
+[=Transient input sources=] follow a slightly different sequence when firing [=transient action=] events:
 
 <div class="algorithm" data-algorithm="on-transient-input-start">
 
-When a [=transient input source=] |source| for {{XRSession}} |session| begins its [=primary action=] the UA MUST run the following steps:
+When a [=transient input source=] |source| for {{XRSession}} |session| begins its [=transient action=] the UA MUST run the following steps:
```
Manishearth: The text below seems to imply that transient actions can also be primary actions, but this means that select events would be fired twice since this line no longer "replaces" the primary action spec text the way it used to.

Is there ever a case for a non-primary transient action?

klausw: I've rephrased it in 9d55efa. I wasn't aware that this was a replacement, that's rather subtle...

An example for a non-primary transient action would be adding additional fingers to a multitouch screen, these would create transient input sources and corresponding `inputsourceschange` events, but no `select{,start,end}` events, similar to how non-primary pointer events don't generate compatibility mouse events. I had updated the example in the previous paragraph to cover that.

As a side note, one additional reason for introducing this distinction is that I'm currently working on the DOM Overlay draft spec, where it would be useful to have a concept of transient sources that don't necessarily trigger XR select events, i.e. because they were handled as a DOM event. That's beyond the scope of this PR, and we can revisit this issue as part of that discussion, but I thought that the concept of non-primary transient sources would also be useful on its own.

klausw:
As requested in the discussion, bf439b4 removes the auxiliary input source changes and tracker example, keeping just the transient input source changes. @Manishearth , how does this look?

index.bs
```diff
@@ -1611,43 +1611,43 @@ When an [=XR input source=] |source| for {{XRSession}} |session| has its [=prima
...
+Some [=/XR device=]s may support <dfn>transient input sources</dfn>, where the [=XR input source=] is only meaningful while performing a <dfn>transient action</dfn>, either the [=primary action=] for a [=primary input source=], or a device-specific action for an [=auxiliary input source=]. An example would be mouse, touch, or stylus input against an {{XRSessionMode/"inline"}} {{XRSession}}, which MUST produce a transient {{XRInputSource}} with a {{targetRayMode}} set to {{screen}}, treated as a [=primary action=] for the [[POINTEREVENTS#dfn-primary-pointer]], and as a non-primary [=transient action=] for a non-primary pointer. [=Transient input sources=] are only present in the session's [=list of active XR input sources=] for the duration of the [=transient action=].
```
Manishearth: You've deleted the definition of auxiliary input source but still refer to it here. We should still be defining it somewhere.

klausw: Sorry about the mismatch. I've reverted the deletion, and made a new change that basically just updates the example text to remove the tracker reference and links to transient input sources for secondary screen touches.

index.bs
```diff
@@ -1544,7 +1544,9 @@ Similarly, the Valve Index controller is backwards compatible with the HTC Vive
...
+An [=XR input source=] is a <dfn>primary input source</dfn> if it supports a <dfn>primary action</dfn>. The [=primary action=] is a platform-specific action that, when engaged, produces {{XRSession/selectstart}}, {{XRSession/selectend}}, and {{XRSession/select}} events. Examples of possible [=primary action=]s are pressing a trigger, touchpad, or button, speaking a command, or making a hand gesture. If the platform guidelines define a recommended primary input then it should be used as the [=primary action=], otherwise the user agent is free to select one. The device MUST support at least one [=primary input source=].
+
+An [=XR input source=] is a <dfn>auxiliary input source</dfn> if it does not support a [=primary action=], for example [=transient input sources=] associated with secondary screen touches on a multitouch device.
```
Manishearth:
```suggestion
An [=XR input source=] is an <dfn>auxiliary input source</dfn> if it does not support a [=primary action=], for example [=transient input sources=] associated with secondary screen touches on a multitouch device.
```

klausw: Done in 0527c57

index.bs
klausw: I've made a small change to this section in commit a1723b3, using the new term "auxiliary action" to refer to a device-specific non-primary action. It was too easy to misintepret "non-primary [=transient action=]" as implying that all transient actions are non-primary, but that's not the case - they include both types.

klausw:
Can this be merged, or are we waiting for additional approvals? Nell is still listed as pending reviewer, but she's said she won't be available for reviews.

toji:
I think we're good to merge given that Manish and I have approved, and Nell is going to be out for a while.

PR #931, opened 2019-12-02
Update .pr-preview.json to use the right index.bs

mounirlamouri:
It will avoid to have PR author sending rendered output as in https://github.com/immersive-web/webxr/pull/900


mounirlamouri:
@toji PTAL

toji:
Thanks! I was trying to figure this out before Thanksgiving but coming up empty.

PR #934, opened 2019-12-05
Mark XRWebGLLayer.framebuffer as nullable

Manishearth:
The text defining it already allows it to be null


toji:
Good catch, thanks!

PR #938, opened 2019-12-10
Clarify bounded reference space behavior

### LINKED ISSUE #937: Clarify that XRBoundedReferenceSpace may be created when bounds aren't currently available
There's a bit of confusion that come up in the Chrome implementation around exactly when a `bounded-floor` reference space is allowed to be created. Till now we had assumed that if the bounds geometry wasn't available at the time the reference space was created the request should be rejected, which in turn would cause applications to have to poll for the reference space on devices where the bounds may not be available immediately upon startup (such as OpenVR or OpenXR). This is especially problematic for developers since there's no way to know if the device is incapable of supporting bounds or if it simply doesn't have them initialize at the moment, so you would have to poll forever in hopes that it would eventually resolve.

Looking through the current spec text, there's nothing to explicitly indicate that this is the desired behavior. In fact, the algorithm to check if [bounded reference spaces are supported](https://immersive-web.github.io/webxr/#bounded-reference-spaces-are-supported) is simply:

> 1. If the XR device cannot report boundaries, return false.
> 1. If the XR device cannot identify the height of the user’s physical floor, return false.

The wording here of "cannot report/cannot identify" suggests that the test should not be for whether or not the data is currently available but whether or not the device is capable of supporting it at all.

Still, there's obviously some room for confusion on this point and there's no clarity given to what should happen if the bounds aren't currently present (empty array? null?) so there's room for clarification.

toji:
/fixes #937 

Adds some clarifying text to the `XRBoundedReferenceSpace` definition to indicate that the `boundsGeometry` array must be set to an empty array (`[]`) if the bounds are not known, and that `bounded-floor` reference spaces are allowed to be returned even if the floor level or bounds geometry hasn't yet been resolved as long as the device is known to support it.

CC @thetuvix

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1207,6 +1207,8 @@ Note: It is suggested that points of the [=native bounds geometry=] be [=quantiz
 
 The <dfn attribute for="XRBoundedReferenceSpace">boundsGeometry</dfn> attribute is an array of {{DOMPointReadOnly}}s such that each entry is equal to the entry in the {{XRBoundedReferenceSpace}}'s [=XRBoundedReferenceSpace/native bounds geometry=] premultiplied by the {{XRRigidTransform/inverse}} of the [=XRSpace/origin offset=]. In other words, it provides the same border in {{XRBoundedReferenceSpace}} coordinates relative to the [=XRSpace/effective origin=].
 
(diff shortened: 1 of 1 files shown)


toji:
Since Manish has given this his approval and Nell is on leave, I'm merging.

PR #939, opened 2019-12-10
Correctly resolve promise when end() is called on an ended session

Manishearth:
Servo was timing out on some tests because they'd call `end()` multiple times and the second time it would never resolve the promise.

r? @toji


Manishearth:
Wait, this approach isn't correct.

Manishearth:
Okay, the spec as-is is correct, it's just easy to get wrong, since the standard way to do this is to queue up end promises and wait till the backend tells you to fire the end event, but if that's already happened you will get a time out.

One way to improve this is to add a `completely shut down` flag that is only set when it's completely shut down.

PR #941, opened 2019-12-11
Improve code style in example

surma:
Declaring a `let` variable globally and then setting its value inside a `.then()` is bad practice and can lead less experienced programmers down a faulty (or even racey) path.

I’d recommend using modern JS with async/await. I only updated the first example here to start a discussion, I am happy to update the later example if the spec editors find this beneficial and/or valuable :D


toji:
Thanks for this, and sorry it took so long to merge. Basically wanted to leave it here as a reminder to myself to update the rest of the code samples, which I've now done in #959.

PR #947, opened 2019-12-30
Potentially update the inline device when the device changes

Manishearth:
This was kind of missed in https://github.com/immersive-web/webxr/pull/737

Basically, if the user "switches into" viewing the page on an immersive XR device after the page has already loaded, inline sessions should be able to use and request tracking.

This does mean you can request an inline session while viewing the page in immersive mode, remove and disconnect the headset, and then have an inline session that was previously allowed to use tracking spaces but no longer can. This is a bit of a weird wrinkle but I think in the current situation it will just refuse to return poses, which is fine.

r? @toji


toji:
I'm sorry I've missed reviewing this for so long! Yes, this was definitely my intent and it's good to have it stated explicitly.

PR #948, opened 2020-01-06
Rename XR interface to XRSystem

### LINKED ISSUE #936: Use a longer name for the XR interface?
This issue courtesy of a day's worth of debugging time from @NellWaliczek, thanks!

In Nell's scenario she was attempting to use the WebXR Polyfill on a page which happened to use a minified library which just so happened to select `XR` as the name of one of its minified globabl functions. This added the `XR` function to `window`, which then conflicted with trying to polyfill the `XR` interface and subsequently caused the polyfill to fail. That specific failure has been fixed in the polyfill, but even in native implementations the accidental overwriting of the `XR` interface could cause some minor issues, primarily for tests, polyfills, or extensions.

For example, a script that tests to ensure that `navigator.xr` is available and of the right type may call:

```js
if (navigator.xr instanceof XR) {
  //...
}
```

Which would fail if the `XR` interface had been accidentally replaced (though calls to `navigator.xr` will continue to function normally.)

This leads to the question of if we should potentially rename the `XR` interface to something longer to avoid potential conflicts. For example: `XRNavigator` or `XRSessionManager`. This would potentially cause some _very_ minor backwards compat issues, but since most pages are expected to only interact with the `navigator.xr` instance changing the interface name should leave 99.999% of WebXR code intact. The question is if we feel like the potential for such a short name to conflict with minified libraries is enough of a problem for us to bother with a change.

(To be clear: I am **not** suggesting that the `navigator.xr` object be changed, since that would be a major backwards compatibility break on a now-shipped API.)

toji:
/fixes #936 

If we're going to change the name of the `XR` interface to avoid conflicts we ought to do it soon. Of the alternate name suggestions put out thus far I liked @nbutko's `XRSystem` best by a large margin, not least of which because it aligns well with OpenXR's use of the term System for a similar concept.

To reiterate, while this does introduce some very minor backwards compatibility issues it's not expected to affect the vast majority of pages, may fix problems for a small number of them, and will generally be opaque to developers not working on, say, the polyfill.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -180,10 +180,10 @@ Application flow {#applicationflow}
 
 Most applications using the WebXR Device API will follow a similar usage pattern:
 
-  * Query {{XR/isSessionSupported()|navigator.xr.isSessionSupported()}} to determine if the desired type of XR content is supported by the hardware and UA.
+  * Query {{XRSystem/isSessionSupported()|navigator.xr.isSessionSupported()}} to determine if the desired type of XR content is supported by the hardware and UA.
   * If so, advertise the XR content to the user.
   * Wait for the user to [=triggered by user activation|trigger a user activation event=] indicating they want to begin viewing XR content.
-  * Request an {{XRSession}} within the user activation event with {{XR/requestSession()|navigator.xr.requestSession()}}.
+  * Request an {{XRSession}} within the user activation event with {{XRSystem/requestSession()|navigator.xr.requestSession()}}.
   * If the {{XRSession}} request succeeds, use it to run a [[#frame|frame loop]] to respond to XR input and produce images to display on the [=XRSession/XR device=] in response.
   * Continue running the [[#frame|frame loop]] until the [=shut down the session|session is shut down=] by the UA or the user indicates they want to exit the XR content.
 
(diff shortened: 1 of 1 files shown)


rolivo:
Hi, isn't it possible to reconsider this? Chrome has implemented the API by default since version 79, so there could be sites that already use it, thanks

Manishearth:
It's very unlikely that sites will be directly accessing `window.XR` (and not `navigator.xr`) since there's nothing useful directly on that interface object.

toji:
Yeah, to clarify: This does _not_ change the way developers interact with the API (It's still accessed at `navigator.xr`, for example.) It only changes the name of one of the interfaces to avoid accidental conflicts due to the shortness of the name. Since the interface in question happens to be one that developers never call directly (you don't instantiate a `new XR();`) the name change will be completely transparent and backward compatible to everyone except:

 - The polyfill developers
 - A couple of Web Platform Tests
 - Maybe the WebXR emulator?

And maybe an extension or two that I don't know about that's doing some pretty deep hooks. Effectively 0% of WebXR content will be affected.

toji:
No complaints raised at the Seattle F2F. Merging.

sideshowbarker:
Is there are related https://crbug.com/ bug for this?

toji:
There is now: https://bugs.chromium.org/p/chromium/issues/detail?id=1050600

PR #966, opened 2020-02-25
Turn XRPermissionStatus into an interface

### LINKED ISSUE #965: XRPermissionStatus is a dictionary but inherits from an interface
https://github.com/immersive-web/webxr/pull/900 added two dictionaries, `XRPermissionDescriptor` and `XRPermissionStatus`, which inherit from `PermissionDescriptor` and `PermissionStatus` respectively.

Unfortunately, `PermissionStatus` is an interface not a dictionary ([spec](https://w3c.github.io/permissions/#permissionstatus))! This caused the IDL import into WPT ([PR](https://github.com/web-platform-tests/wpt/pull/21949)) to start failing with `object "XRPermissionStatus inherits PermissionStatus, but XRPermissionStatus is not an interface."`

I'm not an IDL expert, but I assume a dictionary isn't meant to be able to inherit from an interface, and `XRPermissionStatus` should be an interface :)

Manishearth:
Unsure if I should use a FrozenArray or an `any[]` here, I think it's fine to use a frozen array even though the attribute _can_ change because then the entire attribute will change (it's not `[SameObject]`)

/fixes #965 
r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2378,8 +2378,8 @@ dictionary XRPermissionDescriptor: PermissionDescriptor {
 <dt>[=permission result type=]</dt>
 <dd>
 <pre class=idl>
-dictionary XRPermissionStatus: PermissionStatus {
-  sequence&lt;any&gt; granted;
+interface XRPermissionStatus: PermissionStatus {
+  attribute FrozenArray<any> granted;
 };
 </pre>
 


stephenmcgruer:
I think you need to use `&lt;` `&gt;` here. The `<any>` has been lost in the spec, which ends up with invalid IDL still: https://immersive-web.github.io/webxr/#xrpermissionstatus

Manishearth:
https://github.com/immersive-web/webxr/pull/967 , thanks for pointing it out!

Issue #877, opened 2019-10-15
Note somewhere that you should be always be queueing up the next rAF within rAF


Manishearth:
Currently, we run an [XR animation frame](https://immersive-web.github.io/webxr/#xr-animation-frame) regardless of whether or not you've triggered rAF.

> When an XRSession session receives updated viewer state from the XR device, it runs an XR animation frame with a timestamp now and an XRFrame frame, which MUST run the following steps regardless of if the list of animation frame callbacks is empty or not:

We also clear framebuffers before each "XR animation frame":

> The buffers attached to an opaque framebuffer MUST be cleared to the values in the table below when first created, or prior to the processing of each XR animation frame. This is identical to the behavior of the WebGL context’s default framebuffer.

Both of these things are good decisions.

However, put together, this means that an application that does not queue up the next rAF within the current rAF callback will experience flashes of black.

This is kind of different from HTML rAF, in HTML you can choose to not queue up a rAF for a bit and that just makes the scene static until you do, whether your application is moving elements around or canvas stuff. This is less jarring.

It seems to me that the only time you should _not_ queue up a rAF within your rAF callback is when you don't intend to render anything -- e.g. when your visibility state becomes `hidden`.

It may be worth noting this non normatively somewhere in the spec, for authors (and doc writers) to pick up.

toji:
Hm... I don't agree with this, though you bring up a good point about the flash of black. That's unintended.

It's a valid technique in some cases to render a single frame (like a splash screen logo) and then skip several frames while other work is happening, letting the device's reprojection system (if it has one) handle any tracking. I don't think we want to prevent that from working here, and so "forcing" developers to keep a full rAF loop scheduled seems problematic.

I'd rather that we find some way to express that the compositor should only update it's imagery if the layer framebuffer is modified during the course of the rAF callbacks (not including the initial clear). This is, as far as I recall, what Chrome is doing anyway.

Manishearth:
cc @asajeffrey 

I suspect that might be pretty tricky to specify? Not sure if webgl gives us the terms necessary to spec that.

I'm okay with us specifying that -- it is a nicer solution -- but I'd check in with implementations first.

Manishearth:
Either way, I think we should definitely explicitly spec this if we're going this way, or add the note.

toji:
The WebGL spec already [states this pretty clearly](https://www.khronos.org/registry/webgl/specs/latest/1.0/#2.2):

[...]

Issue #969, opened 2020-03-03
WebXR should describe what happens if no draw calls happen during RequestAnimationFrame


cabanier:
As mentioned in https://github.com/immersive-web/layers/issues/20, the spec needs to be updated to describe reprojection behavior in case the author didn't draw into the WebGL framebuffer.

Currently the specification is silent which might cause UAs to implement different behavior.

raviramachandra:
Update: On Magicleap One, if we don't explicitly draw anything in RAF we render black. @toji @cabanier

cabanier:
> Update: On Magicleap One, if we don't explicitly draw anything in RAF we render black.

Is this something that you can fix on the ML side? 
Also is there a minimum refresh rate on your platform?

Manishearth:
We discussed this a bunch in https://github.com/immersive-web/webxr/issues/877 as well

Manishearth:
/agenda get quick platform feedback on what their expected behavior is

cabanier:
FYI the customer experience was skipping frames between scenes while displaying a loading screen.

cabanier:
The spec was updated to cover this case. Closing.

PR #970, opened 2020-03-04
Defining when an opaque framebuffer is considered dirty

toji:
This cribs very directly from the equivalent [WebGL spec text](https://www.khronos.org/registry/webgl/specs/latest/1.0/#2.2), with appropriate verbiage changes. Related to #969, but doesn't actually fix that one since the issue itself is more about what happens when the users doesn't dirty the framebuffer for a while.


toji:
Just realized that this /fixes #877

asajeffrey:
LGTM FWIW

PR #973, opened 2020-03-04
Updated the spec to match the latest user activation definitions

toji:
Recently the definition of [User Activation](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation) in the HTML standard has changed. Most notably, it was divided into concepts of ["sticky activation"](https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation) and ["transient activation"](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation), and specs that rely on activation behavior (like ours) need to update to indicate which them mean. The chromium team also put together a [brief guide for specs that need to migrate](https://docs.google.com/document/d/14wT89JZ0qeRehXGkcn3_meXxjvlHKgM9d7aJj80kQcQ/edit).

In our case I think it's pretty clear that transient activation is the right choice, since that's generally what we had assumed before and we want to retain the behavior that session launches should be associated with a specific action (like a button click) rather than allowing new sessions at any point after the page has been interacted with. I've updated the spec language to reflect that in this PR.

Additionally, the transient activation is scoped to a specific Window and we need to specify which one we intend to evaluate the activation on. I'm on much shakier ground when it comes to this type of language, but in this case it seems pretty clear to me that it ought to be the `Window` that hosts the `Navigator` object that hosts the `XRSystem` that `requestSession()` was called on. I've tried to express that here, but I would really love feedback from someone like @asajeffrey that has a better understanding of these things.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -187,7 +187,7 @@ Most applications using the WebXR Device API will follow a similar usage pattern
 
   * Query {{XRSystem/isSessionSupported()|navigator.xr.isSessionSupported()}} to determine if the desired type of XR content is supported by the hardware and UA.
   * If so, advertise the XR content to the user.
-  * Wait for the user to [=triggered by user activation|trigger a user activation event=] indicating they want to begin viewing XR content.
+  * Wait for the window to have [=transient activation=]. This is most commonly indicated by the user clicking a button on the page indicating they want to begin viewing XR content.
   * Request an {{XRSession}} within the user activation event with {{XRSystem/requestSession()|navigator.xr.requestSession()}}.
   * If the {{XRSession}} request succeeds, use it to run a [[#frame|frame loop]] to respond to XR input and produce images to display on the [=XRSession/XR device=] in response.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -328,16 +328,17 @@ When this method is invoked, the user agent MUST run the following steps:
 
   1. Let |promise| be [=a new Promise=].
   1. Let |immersive| be <code>true</code> if |mode| is an [=immersive session=] mode, and <code>false</code> otherwise.
+  1. Let |global object| be the [=relevant Global object=] for the {{Navigator}} object which the {{XRSystem}} on which this method was invoked is a member of.
```
asajeffrey: The relevant global is probably the right global here. I think the language can be simplified to "the [=relevant Global object=] for the {{XRSystem}} " (all platform objects have a relevant global, you don't need to say the navigator).

toji:
Thanks Alan and Manish! Updated the text to be the simplified version Alan suggested. Merging.

PR #980, opened 2020-03-09
Add XRInputSourceEvent::{screenX, screenY}

### LINKED ISSUE #978: Add x,y coordinates to input source events for screen-space
Currently if a handheld AR session gets a `screen` touch event, the `x, y` coordinates of the event aren't surfaced [without doing some backcalculation](https://github.com/immersive-web/webxr/issues/977#issuecomment-595364947)

We should perhaps add something like `readonly double? x; readonly double? y;` to `InputSourceEvent` (or make a subclass) so that this can be surfaced directly.

This could potentially also be useful for touch events on layers (https://github.com/immersive-web/layers/issues/21)

See https://github.com/immersive-web/webxr/issues/977

cc @toji @cabanier

Manishearth:
Putting a strawman proposal out there so folks can discuss it better.

Fixes #978

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/980.html" title="Last updated on May 7, 2020, 1:49 PM UTC (4ab2bd4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/980/9d3ccaf...Manishearth:4ab2bd4.html" title="Last updated on May 7, 2020, 1:49 PM UTC (4ab2bd4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1542,7 +1542,8 @@ An [=XR input source=] is an <dfn>auxiliary input source</dfn> if it does not su
 When an [=XR input source=] |source| for {{XRSession}} |session| begins its [=primary action=] the UA MUST run the following steps:
 
   1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the action occurred.
-  1. [=Queue a task=] to [=fire an input source event=] with name {{selectstart!!event}}, frame |frame|, and source |source|.
+  1. Let |x| and |y| be the coordinates of the event if |source| has {{XRInputSource/targetRayMode}} {{XRTargetRayMode/"screen"}}.
+  1. [=Queue a task=] to [=fire an input source event=] with name {{selectstart!!event}}, frame |frame|, source |source|, and coordinates |x| and |y|.
 
 </div>
 
@@ -1551,9 +1552,10 @@ When an [=XR input source=] |source| for {{XRSession}} |session| begins its [=pr
 When an [=XR input source=] |source| for {{XRSession}} |session| ends its [=primary action=] the UA MUST run the following steps:
 
   1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the action occurred.
+  1. Let |x| and |y| be the screen coordinates of the touch event if |source| has {{XRInputSource/targetRayMode}} {{XRTargetRayMode/"screen"}}.
   1. [=Queue a task=] to perform the following steps:
-    1. [=Fire an input source event=] with name {{select!!event}}, frame |frame|, and source |source|.
-    1. [=Fire an input source event=] with name {{selectend!!event}}, frame |frame|, and source |source|.
+    1. [=Fire an input source event=] with name {{select!!event}}, frame |frame|, source |source|, and coordinates |x| and |y|.
+    1. [=Fire an input source event=] with name {{selectend!!event}}, frame |frame|, source |source|, and coordinates |x| and |y|.
 
 </div>
 
@@ -1586,7 +1588,8 @@ Sometimes platform-specific behavior can result in a [=primary action=] or [=pri
 When an [=XR input source=] |source| for {{XRSession}} |session| has its [=primary action=] cancelled the UA MUST run the following steps:
 
   1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| with [=XRFrame/time=] being the time the action occurred.
-  1. [=Queue a task=] to [=fire an input source event=] an {{XRInputSourceEvent}} with name {{selectend!!event}}, frame |frame|, and source |source|.
+  1. Let |x| and |y| be the screen coordinates of the touch event if |source| has {{XRInputSource/targetRayMode}} {{XRTargetRayMode/"screen"}}.
+  1. [=Queue a task=] to [=fire an input source event=] an {{XRInputSourceEvent}} with name {{selectend!!event}}, frame |frame|, source |source|, and coordinates |x| and |y|.
 
 </div>
 
@@ -1611,10 +1614,11 @@ Some [=/XR device=]s may support <dfn>transient input sources</dfn>, where the [
 When a [=transient input source=] |source| for {{XRSession}} |session| begins its [=transient action=] the UA MUST run the following steps:
 
   1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| for the time the action occurred.
+  1. Let |x| and |y| be the screen coordinates of the touch event if |source| has {{XRInputSource/targetRayMode}} {{XRTargetRayMode/"screen"}}.
   1. [=Queue a task=] to perform the following steps:
     1. Fire any <code>"pointerdown"</code> events produced by the [=XR input source=]'s action, if necessary.
     1. [=add input source|Add the XR input source=] to the [=list of active XR input sources=].
-    1. If the [=transient action=] is a [=primary action=], [=fire an input source event=] with name {{selectstart!!event}}, frame |frame|, and source |source|.
+    1. If the [=transient action=] is a [=primary action=], [=fire an input source event=] with name {{selectstart!!event}}, frame |frame|, source |source|, and coordinates |x| and |y|.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2019,23 +2025,29 @@ interface XRInputSourceEvent : Event {
...
+  double x;
+  double y;
 };
 </pre>
 
 The <dfn attribute for="XRInputSourceEvent">inputSource</dfn> attribute indicates the {{XRInputSource}} that generated this event.
 
 The <dfn attribute for="XRInputSourceEvent">frame</dfn> attribute is an {{XRFrame}} that corresponds with the time that the event took place. It may represent historical data. Any {{XRViewerPose}} queried from the {{XRInputSourceEvent/frame}} MUST have an empty {{XRViewerPose/views}} array.
 
+The <dfn attribute for="XRInputSourceEvent">x</dfn> and <dfn attribute for="XRInputSourceEvent">y</dfn> attributes define screen coordinates for events with {{XRInputSourceEvent/inputSource}}s that have a {{XRInputSource/targetRayMode}} of {{XRTargetRayMode/"screen"}}. They are <code>null</code> for other kinds of events.
```
Manishearth: Is "screen coordinates" the right term to use? Perhaps "viewport-relative coordinates"? Except it's possible for the viewports to be different in multi-view situations (I can't think of a screen-space multi-view situation, though).

Manishearth: Also, should we be allowing this for other target ray modes? This may potentially be useful for the layers spec as well cc @cabanier

index.bs
```diff
@@ -2019,23 +2025,29 @@ interface XRInputSourceEvent : Event {
...
   [SameObject] readonly attribute XRInputSource inputSource;
+  readonly attribute double? x;
+  readonly attribute double? y;
 };
 
 dictionary XRInputSourceEventInit : EventInit {
   required XRFrame frame;
   required XRInputSource inputSource;
+  double x;
+  double y;
```
toji, line 2036: If the values in the event are nullable these should be as well, right?

index.bs
```diff
@@ -2019,23 +2025,29 @@ interface XRInputSourceEvent : Event {
   constructor(DOMString type, XRInputSourceEventInit eventInitDict);
   [SameObject] readonly attribute XRFrame frame;
   [SameObject] readonly attribute XRInputSource inputSource;
+  readonly attribute double? x;
+  readonly attribute double? y;
```
toji, line 2029: I feel like these new attributes should be slightly better qualified, the way that mouse events have moved on from a simple `x`/`y` to `clikent`/`screen`/`page`/`etc` versions. Given that they're currently only produced from "screen" input sources it seems like `screenX`/`screenY` would be appropriate, but I'm very open to other suggestions.

index.bs
toji: You could have stereo screens via shutter/polarized glasses or lenticular displays. You'd still be interacting on the screen surface and projecting inward, though, so we should always have a single point of interaction on the screen surface.

Manishearth:
Sorry for dropping this, updated!

Manishearth:
/agenda to run by the group again before merging

Manishearth:
/agenda to run by the group again before merging

looks like the bot didn't pick this up?

cabanier:
> /agenda to run by the group again before merging
> 
> looks like the bot didn't pick this up?

This week is the CG call

Manishearth:
Right, but the bot still picks up and tags things immediately, it doesn't know about the choice of meeting.

cc @AdaRoseCannon

AdaRoseCannon:
Hmm sometimes the bot misses things I really haven't worked out why yet?!, that should have worked, if it ever forgets to add the label please chase me or an editor up to manually add the label sorry I missed this til today.

index.bs
```diff
@@ -2019,23 +2025,29 @@ interface XRInputSourceEvent : Event {
   constructor(DOMString type, XRInputSourceEventInit eventInitDict);
   [SameObject] readonly attribute XRFrame frame;
   [SameObject] readonly attribute XRInputSource inputSource;
+  readonly attribute double? screenX;
+  readonly attribute double? screenY;
```
thetuvix, line 2029: What kind of logic do we expect apps to write with `screenX`/`screenY`?  We already give the app a `targetRaySpace` for raycasting into the 3D scene, and so the primary scenario I see here is activating a 2D button that the app rendered in `"viewer"` space.  However, apps that do that will not be usable on AR headsets.

Instead, we are solving the screen-space button problem in a cross-form-factor manner with [DOM overlays](https://immersive-web.github.io/dom-overlays/).  DOM overlays handle 2D input raycasting for you, and critically, they support headsets by allowing those overlays to be rendered on non-head-locked UI when using headsets.

Does this change undo that by assisting apps in directly rendering 2D content in `"viewer"` space?

Manishearth: I would expect apps to us [XRInteractionMode](https://immersive-web.github.io/webxr-ar-module/#xrinteractionmode-enum) to do custom UI instead here. I can imagine an app using gamepads in world-space and input clicks on GL-rendered buttons in screen-space.

This does assist apps in directly rendering 2D content to the viewer space. I'm not sure if that really undoes this, since not everything supports DOM overlay, and I'd like there to be support for custom UI patterns that are not picked by the user-agent.

index.bs
```diff
@@ -1623,10 +1627,11 @@ When a [=transient input source=] |source| for {{XRSession}} |session| begins it
 When a [=transient input source=] |source| for {{XRSession}} |session| ends its [=transient action=] the UA MUST run the following steps:
 
   1. Let |frame| be a new {{XRFrame}} with {{XRFrame/session}} |session| for the time the action occurred.
+  1. Let |x| and |y| be the screen coordinates of the touch event if |source| has {{XRInputSource/targetRayMode}} {{XRTargetRayMode/"screen"}}.
```
thetuvix, line 1630: Would we need to further define "screen coordinates" here?
* Is this in raw screen pixels or normalized coordinates?
* How do the coordinate axes react to screen rotation?
* How does (0, 0) react to browser chrome or other DOM elements that may take up part of the screen during an `"inline"` session?

If the goal is to allow interaction with WebGL-rendered screen-space UI, would it still be simplest for the app to raycast into their scene in 3D to avoid worrying about all of those 2D conventions that just cancel out in 3D?

Manishearth: @bialpio what coordinates is Chrome using in the gamepad implementation currently?

My impression would be that this is in the rotated screen space, using screen px coordinates (like DOM APIs), and is relative to the top left of the rendered area.

thetuvix: Would DOM pixel coordinates be useful for an app that is raycasting against WebGL-based screen-space UI? (the stated purpose of an app using `screenX`/`screenY` rather than DOM overlays for screen-space UI)

For an `"inline"` session, I can imagine an app doing its own normalization of `screenX`/`screenY` relative to their canvas width/height, so they can ultimately target scene geometry in their game engine's logical screen space.  However, for an `"immersive-ar"` session with no output canvas, does the app have enough information to reliably reason about the implicit "canvas size" that gives meaning to those coordinates?  Across UAs that make different decisions about fitting within system UI or notches on the edge of the screen, is there a predictable X and Y value that represents the right/bottom of the screen, so apps can do that normalization math?  And would any of that ever be more convenient than just raycasting with `targetRaySpace`?

Perhaps `screenX`/`screenY` are only usable reliably in `"inline"` sessions.  If so, should we spec them to be null in immersive sessions?

Manishearth: Hmmm.. I was actually imagining this mostly to be useful for handheld AR, especially given that Chromium is exposing this via gamepad already (apparently at least one user wanted this -- we should find out why they needed it?)

I was was thinking this would be based on the output framebuffer size.

Manishearth:
@bialpio would you be able to clarify what coordinates Chrome is exposing here?

I think my base position here is that if Chrome feels the need to ship away to expose these coordinates, such an API is probably necessary and needs to exist _somewhere_. It can be here, or it can be on gamepad, it doesn't matter that much to me, but we should specify it somehow.

I'm considering writing a contrastive PR to show what this would look like on the gamepad spec.

bialpio:
For touch screen input, we expose the normalized coordinates in the [-1;1] range as axes on a gamepad object available on an input source.

Manishearth:
Ah, thanks. That makes sense.

Manishearth:
https://github.com/immersive-web/webxr/issues/978

PR #982, opened 2020-03-10
Make XRReferenceSpaceInit.transform nullable

### LINKED ISSUE #981: XRReferenceSpaceEvent.transform does not always have the same type as XRReferenceSpaceEventInit.transform
The WebIDL for `XRReferenceSpaceEventInit` describes a `transform` dictionary member that is optional, but non-nullable:

```
dictionary XRReferenceSpaceEventInit : EventInit {
  required XRReferenceSpace referenceSpace;
  XRRigidTransform transform;
};
```

If a value for `transform` is not included in `XRReferenceSpaceEventInit`, it will result in a value of `undefined`.

The `XRReferenceSpaceEvent` interface describes a `transform` member that is nullable:

```
[SameObject] readonly attribute XRRigidTransform? transform;
```

It is unnatural to have a value of `undefined` used in the dictionary when constructing the event become `null`.

We should define the `transform` member of the `XRReferenceSpaceEventInit` dictionary as nullable, with a default value of `null`:

```
dictionary XRReferenceSpaceEventInit : EventInit {
  required XRReferenceSpace referenceSpace;
  XRRigidTransform? transform = null;
};
```

kearwood:
Fixes #981

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2089,7 +2089,7 @@ interface XRReferenceSpaceEvent : Event {
 
 dictionary XRReferenceSpaceEventInit : EventInit {
   required XRReferenceSpace referenceSpace;
-  XRRigidTransform transform;
+  XRRigidTransform? transform = null;
 };
 </pre>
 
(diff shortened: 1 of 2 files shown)


Manishearth:
I feel like it is really fine as is: the correct fix is to actually specify this for the constructor (which we haven't done), but this fix is also good.

kearwood:
Rebased...

kearwood:
I don't seem to have write access any more.  Would anyone be able to merge and/or check my privilege?

PR #985, opened 2020-03-16
Clarify when tracking/input data is exposed via inline devices

### LINKED ISSUE #984: Clarify whats learn-able through an inline session re privacy
To my read, the spec seems ambiguous on whether the user needs to give permission to start an inline session (in chromium and similar browsers, it currently doesn't).  The spec is similarly unclear what devices are accessible from the `inputSources ` property of the returned, un-permission-gated `XRSession` object.

From conversation with implementors, it seems like this will be the same for all platforms (and so doesn't have a privacy risk), but that doesn't seem to be required by the spec.

The spec should either require that implementations never return device information through this ungated path, or require permission to start an inline session

toji:
/fixes #984

This makes some changes to how we pick inline devices that make the spec line up more closely with the group's intent for when inline sessions are allowed to expose tracking data, as well as add a bit of non-normative text to more clearly explain the purpose of the default inline session that can be accessed without first gaining user consent.

CC: @pes10k

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -211,9 +211,13 @@ An [=/XR device=] has a <dfn>list of supported modes</dfn> (a [=/list=] of [=/st
 
 Each [=/XR device=] has a <dfn for="XR device">list of enabled features</dfn> for each {{XRSessionMode}} in its [=list of supported modes=], which is a [=/list=] of [=feature descriptors=] which MUST be initially an empty [=/list=].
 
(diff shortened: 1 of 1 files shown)


toji:
I'm not sure I follow, @Manishearth? You mean 2D page inputs generated by an XR controller? The UA  should be handling the management of those input into pointer events already, and those pointer events should be the only thing exposed. From this PR:

> MUST NOT report [=XR input source=]s or events other than those created by pointer events.

Is there a case I'm overlooking here?

Manishearth:
> Is there a case I'm overlooking here?

Yes, the case that the inline XR device that gets selected is not the default inline XR device. The spec is structured such that e.g. if you're in a 2D browser on a device, the inline XR device that gets selected is the same as the immersive one, so you can expose head poses.

cabanier:
> > Is there a case I'm overlooking here?
> 
> Yes, the case that the inline XR device that gets selected is not the default inline XR device. The spec is structured such that e.g. if you're in a 2D browser on a device, the inline XR device that gets selected is the same as the immersive one, so you can expose head poses.

Do you know of a UA that implements this?

toji:
I think that should be addressed by the change to `obtaining the current device`, which now states that if the features arrays are empty you always get the default device.

Issue #983, opened 2020-03-12
WebXR API allows 2 or 3 bits of fingerprintability through isSessionSupported


pes10k:
As called out by the spec, the API currently reveals fingerprint-sensitive information about the user's device, in a way thats not gated (i.e. [isSessionSupported](https://immersive-web.github.io/webxr/#dom-xrsystem-issessionsupported)).  I'm not sure if this is 2 or 3 bits b/c its not clear to me if `inline` ever returns false.

I think the best solution here would be to make `isSessionSupported` gated behind the permission too.

toji:
First off, `inline` should never return false, so it's not a fingerprintable bit.

Aside from that the 2 bits of fingerprintability that comes from `isSessionSupported()` is very intentional and was given a lot of discussion within the group. The API shape is designed such that sites can query that boolean to determine whether or not they should advertise XR capabilities, which is a crucial user ergonomics capability. Imagine a large site like YouTube wants to use the API to display 360 videos. They would most likely want to put an icon for triggering VR presentation next to the fullscreen icon, etc, but only if the user has some hope of actually being able to view the content in VR. It's simply not going to be acceptable to sites that strive for a polished user experience to show the button to all users only to immediately say "Nope, sorry" to the majority of them when they click it. Similarly, if a permission was required to even be able to show the button, no site would use it because it's incredibly intrusive, confusing, and unwanted to have a page immediately pop up a "I want access to your VR headset!" prompt when it first opens. Instead, we want the permissions prompts to be associated with the actual creation of a session, at which point the user has presumably asked for the VR capabilities to activate (it requires transient user activation) and thus a permission prompt won't be surprising.

The API has been designed such that doing anything useful with it requires creating an XRSession, which DOES make use of permissions prior to exposing anything else fingerprintable. This was the tradeoff the group made between making the API useful and usable for pages vs. revealing as little as possible until the user consents to it. I feel pretty comfortable with the balance that we struck.

pes10k:
Hmm, i don't follow. Surely if YouTube wants to present a "view in webxr" button, they can either ask "would you like youtube to know about webxr" or similar, or YouTube can have some "enable WebXR flow", just as with notifications or anything else.

I'm not following whats unique about WebXR that should expose users to fingerprinting risk

toji:
[...]

PR #986, opened 2020-03-17
Allow more flexibility in what isSessionSupported returns.

toji:
WIP.

Accommodates some of the mitigation strategies discussed in #983, but does not resolve that issue.


avadacatavra:
This looks really good to me. I like the note a lot

PR #987, opened 2020-03-18
Allow depth&&stencil result if depth||stencil requested.

jdashg:
We don't want to encourage !depth&&!stencil result if depth xor stencil.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1764,8 +1764,8 @@ An <dfn>opaque framebuffer</dfn> functions identically to a standard {{WebGLFram
  - An [=opaque framebuffer=] MAY support antialiasing, even in WebGL 1.0.
  - An [=opaque framebuffer=]'s attachments cannot be inspected or changed. Calling {{framebufferTexture2D}}, {{framebufferRenderbuffer}}, {{deleteFramebuffer}}, or {{getFramebufferAttachmentParameter}} with an [=opaque framebuffer=] MUST generate an {{INVALID_OPERATION}} error.
  - An [=opaque framebuffer=] is considered incomplete outside of a {{XRSession/requestAnimationFrame()}} callback. When not in a {{XRSession/requestAnimationFrame()}} callback calls to {{checkFramebufferStatus}} MUST generate a {{FRAMEBUFFER_UNSUPPORTED}} error and attempts to clear, draw to, or read from the [=opaque framebuffer=] MUST generate an {{INVALID_FRAMEBUFFER_OPERATION}} error.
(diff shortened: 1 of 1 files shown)


cabanier:
That sounds reasonable and likely already the way that this is implemented.

jdashg:
Actually I had forgotten, but WebGL 1 went the other way on this for portability reasons:
> The depth, stencil and antialias attributes, when set to true, are requests, not requirements. The WebGL implementation should make a best effort to honor them. When any of these attributes is set to false, however, the WebGL implementation must not provide the associated functionality. 

I was remembering the original behavior before we ran into the portability issues that caused the wording we have today.

WebGL 2 made them strict, non-optional:
> Different from WebGL 1.0, the depth, stencil, and antialias attributes in WebGL 2.0 must be obeyed by the WebGL implementation. 

Behind the scenes depth-xor-stencil are almost always DEPTH24_STENCIL8 with special handling.
I want to follow up on this PR to propose the barely-backwards-incompatible depth||stencil => depth&&stencil requirement instead.

toji:
I'm fine with this, as it trends towards my overall desire to have the opaque framebuffer mimic the default framebuffer behavior. Especially since any backwards compatibility issues introduced by the change would be vanishingly rare in practice. You're only going to be surprised by having a stencil buffer when you didn't expect one if you've enabled stencil test, which you really shouldn't be doing if you don't think you have a stencil buffer. On the flip side, I have a hard time imagining ANY type of XR content that will ask for a stencil buffer but explicitly not a depth buffer.

I'll take a pass at updating the spec text like @jdashg suggested above.

Worth noting that it seems like the WebGL spec is mildly self-contridictary on this subject, as in one location it says:

> The depth, stencil and antialias attributes, when set to true, are requests, not requirements. The WebGL implementation should make a best effort to honor them.

Then further down, in the Context creation parameters section, states:

>  **depth**
> If the value is true, the drawing buffer has a depth buffer of at least 16 bits. If the value is false, no depth buffer is available.
> **stencil**
> If the value is true, the drawing buffer has a stencil buffer of at least 8 bits. If the value is false, no stencil buffer is available.

Which implies more strictness than the prior paragraph.

toji:
Hey Jeff, can you make sure your GitHub account is linked to the Mozilla org so that our ipr process is happy with the change? [Details here](https://labs.w3.org/repo-manager/pr/id/immersive-web/webxr/987)

jdashg:
Yep, sorry!

jdashg:
Should work now, though I can't retrigger myself. ("Pull Request Loading"
hits " NetworkError when attempting to fetch resource.")

toji:
Great, thanks! Revalidate was finicky but it looks like it worked.

PR #1002, opened 2020-04-21
Defer initial inputsourcechange event till after the promise resolves

### LINKED ISSUE #961: Timing of initial inputsourceschange event
https://immersive-web.github.io/webxr/#list-of-active-xr-input-sources

> Each XRSession has a list of active XR input sources (a list of XRInputSource) which MUST be initially an empty list.

This means that if the device already has input sources connected from the start, it needs to immediately fire an `inputsourceschange` event after the session is created. [Threejs relies on this happening](https://github.com/mrdoob/three.js/issues/18638).

However, there's a period of time in which the session exists but is not yet available to user code (when the promise hasn't been resolved yet, or the promise has resolved and the resolution microtasks haven't triggered yet). In such a scenario, firing an `inputsourceschange` event is not useful since nothing can catch it. The current design of the spec seems to have a bit of a pit of failure where codebases can reasonably assume that all input sources will arrive through an `inputsourceschange` event, but it is technically legal for implementors to fire these events before JS code is practically able to catch it.

I feel like we should _either_: 

 - Change the spec wording to allow the input sources list to be non empty at the beginning, and urge framework authors to not make this assumption
 - Add spec text that _explicitly_ triggers the [add input source](https://immersive-web.github.io/webxr/#xrsession-add-input-source) algorithm for existing input sources right after the "resolve" line in [the `requestSession()` algorithm](https://immersive-web.github.io/webxr/#dom-xrsystem-requestsession)

Servo _currently_ does the former, but that's currently incorrect. Chrome _seems_ to be doing the latter, but it's not explicit, so it might be possible that there's actually a race there.

cc @alcooper91

Manishearth:
Fixes #961

As mentioned in #961, this is not the only way to fix this issue; the other way to fix it involves changing the spec so that `inputSources` can start off non-null, however that involves ensuring the ecosystem respects that as well.

r? @toji

cc @cabanier

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -382,6 +382,13 @@ When this method is invoked, the user agent MUST run the following steps:
               <dd> Append |session| to the [=list of inline sessions=].
             </dl>
         1. [=/Resolve=] |promise| with |session|.
+        1. [=Queue a task=] to perform the following steps:
+            NOTE: These steps ensure that initial <code>inputsourceschange</code> events occur after the initial session is resolved.
+            1. Set |session|'s [=XRSession/promise resolved=] flag to <code>true</code>.
+            1. Let |sources| be any existing input sources attached to |session|.
+            1. If |sources| is non-empty, perform the following steps:
+                1. Set |session|'s [=list of active XR input sources=] to |sources|.
+                1. Fire an {{XRInputSourcesChangeEvent}} named {{inputsourceschange!!event}} on |session| with {{XRInputSourcesChangeEvent/added}} set to |sources|.
   1. Return |promise|.
 
 </div>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -382,6 +382,13 @@ When this method is invoked, the user agent MUST run the following steps:
               <dd> Append |session| to the [=list of inline sessions=].
             </dl>
         1. [=/Resolve=] |promise| with |session|.
+        1. [=Queue a task=] to perform the following steps:
```
cabanier, line 385: Does the `Resolve` step wait until the task is done?

Manishearth: It's the other way around, `resolve` uses a microtask, and `queue a task` uses a regular task, so we can guarantee that the callback attached in `requestSession(..).resolve(callback)` will execute before the task below does.

cabanier: ok. Resolve actually calls the `then` function. I think this looks good!

Manishearth: Yep!

er, i should have written `then` instead of `resolve` in my comment above, but you get the idea

Manishearth:
@klausw you might want to ensure the impl in chrome does this deferring properly

PR #1003, opened 2020-04-21
Handle null features and features with stringifiers

### LINKED ISSUE #994: Need to describe what happens when a non-string object is used as a XRSessionInit.requiredFeatures value
It is not clearly defined what should happen when values such as `null`, `nothing`, or objects with a `ToString` function are included within the `XRSessionInit.requiredFeatures` array.

I would like to propose that objects that can be converted to a string, with `ToString` be converted and then compared with the enum values.

For `null` and `nothing`, perhaps they should be ignored without preventing the XRSession from being created?

Manishearth:
Fixes #994

may also address https://github.com/immersive-web/webxr/issues/860

cc @kearwood

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -86,6 +86,8 @@ spec: PageVisibility; urlPrefix: https://www.w3.org/TR/page-visibility-2/#
 
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: method; text: IsDetachedBuffer; url: sec-isdetachedbuffer
+    type: dfn; text: ?; url: sec-returnifabrupt-shorthands
+    type: abstract-op; text: ToString; url: sec-tostring
 spec: dom; urlPrefix: https://dom.spec.whatwg.org/#
     type:algorithm; text:fire an event; url: concept-event-fire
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2506,14 +2511,22 @@ To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optio
     1. Return the [=tuple=] <code>(|consentRequired|, |consentOptional|, |granted|)</code>
   1. Add every [=feature descriptor=] in the [=default features=] table associated with |mode| to the indicated feature list if it is not already present.
   1. For each |feature| in |requiredFeatures| perform the following steps:
-    1. If |feature| is not a valid [=feature descriptor=], return <code>null</code>.
+    1. If the |feature| is <code>null</code>, [=continue=] to the next entry.
```
cabanier, line 2514: Maybe the [language in the Permissions spec](https://www.w3.org/TR/permissions/#dom-permissions-query) is a bit more concise

Manishearth: Hmmmm. This is tricky to do since we need to convert it to multiple things, potentially.

PR #1005, opened 2020-04-21
Allow cancelling rAF callbacks from within rAF

### LINKED ISSUE #923: cancelAnimationFrame: clarify behavior when called from a frame callback
As discussed in https://github.com/immersive-web/webxr-polyfill/pull/116#discussion_r346593146 calling `xrSession.cancelAnimationFrame` from within a frame callback should be able to cancel another subsequent callback within that same frame's callbacks list. A strict reading of the current algorithm as written does not allow for this; it says to search through the "list of animation frame callbacks" for the handle to cancel, but that "list of animation frame callbacks" is explicitly set to an empty list prior to callback execution.

Manishearth:
Fixes #923

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -857,7 +857,7 @@ callback XRFrameRequestCallback = void (DOMHighResTimeStamp time, XRFrame frame)
 
 Each {{XRFrameRequestCallback}} object has a <dfn for="XRFrameRequestCallback">cancelled</dfn> boolean initially set to <code>false</code>.
 
-Each {{XRSession}} has a <dfn>list of animation frame callbacks</dfn>, which is initially empty, and an <dfn>animation frame callback identifier</dfn>, which is a number which is initially zero.
+Each {{XRSession}} has a <dfn>list of animation frame callbacks</dfn>, which is initially empty, a  <dfn>list of currently running animation frame callbacks</dfn>, which is also initially empty, and an <dfn>animation frame callback identifier</dfn>, which is a number which is initially zero.
 
 <div class="algorithm" data-algorithm="request-animation-frame">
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -879,7 +879,7 @@ The <dfn method for="XRSession">cancelAnimationFrame(|handle|)</dfn> method canc
 When this method is invoked, the user agent MUST run the following steps:
 
   1. Let |session| be the target {{XRSession}} object.
-  1. Find the entry in |session|'s [=list of animation frame callbacks=] that is associated with the value |handle|.
+  1. Find the entry in |session|'s [=list of animation frame callbacks=] and |session|'s [=list of currently running animation frame callbacks=] that is associated with the value |handle|.
```
toji, line 882: Nit: This could be an "or", since the value should not exist in both at once.

index.bs
```diff
@@ -892,15 +892,16 @@ When an {{XRSession}} |session| receives updated [=viewer=] state for timestamp
   1. If |session|'s [=pending render state=] is not <code>null</code>, [=apply the pending render state=].
   1. If |session|'s {{XRSession/renderState}}'s {{XRRenderState/baseLayer}} is <code>null</code>, abort these steps.
   1. If |session|'s  [=XRSession/mode=] is {{XRSessionMode/"inline"}} and |session|'s {{XRSession/renderState}}'s [=XRRenderState/output canvas=] is <code>null</code>, abort these steps.
-  1. Let |callbacks| be a list of the entries in |session|'s [=list of animation frame callback=], in the order in which they were added to the list.
+  1. Set  |session|'s [=list of currently running animation frame callbacks=] to be a copy of the list of the entries in |session|'s [=list of animation frame callbacks=], in the order in which they were added to the list.
```
toji, line 895: What's the reasoning behind the new copy here? We should be able to just move the pointer around, so to speak, right?

Manishearth: Trying to preserve the clause that that list is in order of append, but really that's how lists work anyway so i don't think this is necessary.

Manishearth:
Addressed

PR #1009, opened 2020-04-24
Make XRLayer an eventTarget

cabanier:
This adds a new event that can be set to the session by the compositor for the case where the underlying swapchain was reset.

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/webxr-1/pull/1009.html" title="Last updated on May 6, 2020, 4:00 AM UTC (ea55fd1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1009/fbc4499...cabanier:ea55fd1.html" title="Last updated on May 6, 2020, 4:00 AM UTC (ea55fd1)">Diff</a>


cabanier:
See https://github.com/immersive-web/layers/issues/106 for more information.

Manishearth:
cc @asajeffrey

toji:
I'm not really in favor of the event name here. I understand the intent and I think that we do need a way of signalling this scenario, but `contextrestored` seems easily conflated with the page's WebGL context being lost/restored, and the fact that it fires from the session doesn't help suggest the course of action the page should take.

Let's continue the conversation on immersive-web/layers#106 a bit more before we land anything.

RafaelCintron:
I agree with @toji on the name of the event.  

I presume that when the underlying swap chain is "reset", all of the WebGL resources the developer has allocated up to this point (buffers, shaders, programs, etc) are still good to use. If so, the event name should reflect that only XR related textures need to be redrawn.  If the WebGL context itself has been lost, then web developers should hook the already existing WebGL [context lost](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.15.2) and [context restored](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.15.3) event.

cabanier:
> I presume that when the underlying swap chain is "reset", all of the WebGL resources the developer has allocated up to this point (buffers, shaders, programs, etc) are still good to use. 

I asked @Artyom17 and he said that this can't happen during a rAF call (at least for Oculus).
If the swapchain is reallocated, everything will good to use during the next rAF.

> If so, the event name should reflect that only XR related textures need to be redrawn. 

Do you have a suggestion? We added [needsRedraw](https://immersive-web.github.io/layers/#dom-xrcompositionlayer-needsredraw)to the spec so maybe `redraw`, `forceRedraw`, `needsRedraw`?

toji:
Based on the conversation in immersive-web/layers#106, I thought we were leaning towards communicating this on a per-layer basis? If so, do we need to add something to the session at all? I would imagine a system-level context loss would simply flag all layers as `needsRedraw` or whatever the final attrib/event is.

cabanier:
Yes, we still need some sort of event in case we have an experience that doesn't always have logic in the rAF call (ie a video player with an cube layer skybox).

toji:
Right, I agree that we need events, but I was under the impression that we we're leaning towards them being per-layer events?

cabanier:
> Right, I agree that we need events, but I was under the impression that we we're leaning towards them being per-layer events?

We added a new attribute [needsRedraw](https://immersive-web.github.io/layers/#dom-xrcompositionlayer-needsredraw) on each layer which becomes `true` if a redraw is needed.
Authors should listen for the event and if it firs, they should look at all layers (during the rAF) and redraw the ones where the flag was set to `true`

Manishearth:
It seems cleaner to instead have an event on each layer that fires when this happens. Then you don't need needsRedraw either

cabanier:
> It seems cleaner to instead have an event on each layer that fires when this happens. Then you don't need needsRedraw either

See https://github.com/immersive-web/layers/issues/106#issuecomment-619575621 for @toji 's thinking about the attribute.
With the combination event/flag, we can support UA's that allow for some redraws to be missed since they can invent the missing frame with reprojection.
For Chrome AR, the flag would always be true since they don't have reprojection
For Magic Leap, it would be for a couple of seconds
For Oculus, we never switch it to true

Manishearth:
Hmm, when talking with him earlier it seemed like he felt it should be on the layer, i'll let him put forth an opinion.

RafaelCintron:
+1 to using a combination event and flag.  Since resources are exposed on the layer, putting a flag there seems prudent.

cabanier:
Does that mean that layers should become event targets?

cabanier:
@Manishearth If layers become event target, I think we will need to update the WebXR spec since the base class `XRLayer` is defined there.

Manishearth:
That seems fine :) we will probably need that for input anyway

cabanier:
ok. I will update the PR.

cabanier:
@Manishearth @toji I changed to PR so XRLayer can receive events. After this goes in, I will create   a PR for layers.

cabanier:
Thanks @Manishearth 
Can you merge it? I don't have that power :-)

Manishearth:
I typically wait for Brandon to also approve before merging :smile:

PR #1012, opened 2020-04-30
Session feature requests no longer need the session parameter

### LINKED ISSUE #1011: Consider requesting the xr permission before creating the XRSession
The [current](https://immersive-web.github.io/webxr/#dom-xrsystem-requestsession) `requestSession()` algorithm specifies that the [permission request](https://immersive-web.github.io/webxr/#request-the-xr-permission) should be done after creating the `XRSession`.

However I think the permission request algorithm does not have any dependency with the session. We do indeed need access to the `XRSession`'s device, but that was obtained [a bit earlier](https://immersive-web.github.io/webxr/#obtain-the-current-device) so we don't need the session for that. It's also mentioned that the `XRPermissionDescriptor` should be 

> initialized with session, requiredFeatures, and optionalFeatures

but it turns out that it just requires the `XRSessionMode` not the session itself.

My proposal would be to move the [permission request](https://immersive-web.github.io/webxr/#request-the-xr-permission) step up and put it just before creating the new `XRSession`. I believe this fits very well with what's mentioned in the [feature dependencies section](https://immersive-web.github.io/webxr/#feature-dependencies)

> This will block the creation of the XRSession if any of the required features are unavailable due to device limitations or in the absence of a clear signal of user intent to expose sensitive information related to the feature.

WDYT?

Manishearth:
This is a leftover from an older draft of my permissions changes.

Fixes #1011

r? @toji

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -370,7 +370,7 @@ When this method is invoked, the user agent MUST run the following steps:
             1. Abort these steps.
         1. Let |session| be a new {{XRSession}} object.
         1. [=Initialize the session=] with |session|, |mode|, and |device|.
-        1. Let |descriptor| be an {{XRPermissionDescriptor}} initialized with |session|, |requiredFeatures|, and |optionalFeatures|
+        1. Let |descriptor| be an {{XRPermissionDescriptor}} initialized with |mode|, |requiredFeatures|, and |optionalFeatures|
         1. Let |status| be an {{XRPermissionStatus}}, initially <code>null</code>
         1. [=Request the xr permission=] with |descriptor| and |status|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -384,6 +382,8 @@ When this method is invoked, the user agent MUST run the following steps:
               <dt> Otherwise
               <dd> Append |session| to the [=list of inline sessions=].
             </dl>
+        1. Let |session| be a new {{XRSession}} object.
+        1. [=Initialize the session=] with |session|, |mode|, and |device|.
```
svillar, line 386: This should be moved one step up as you might potentially set the active immersive session to session (and or adding it to the list of inline sessions) just above

PR #1016, opened 2020-05-01
Add non-normative note about view orientations

### LINKED ISSUE #1013: Clarify whether each view is required to have the same z-axis as the viewer?
Should content be allowed to assume that each of the views shares a z-axis with the viewer? Some content might, for example, z-sort surfaces according to their z-position in viewer space, then use that z-sorting to render to each eye. This is fine if sort order is consistent, but not if the views have different z-axes.

The spec is silent on this topic, though there is a note in https://immersive-web.github.io/webxr/#xrviewerpose-interface that says that spectator mode is supported, which implies that not all views share a z axis.

Manishearth:
Fixes #1013

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/1016.html" title="Last updated on May 1, 2020, 6:15 PM UTC (e37a6e5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1016/86839d2...Manishearth:e37a6e5.html" title="Last updated on May 1, 2020, 6:15 PM UTC (e37a6e5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1268,6 +1268,8 @@ Each [=view=] corresponds to a display or portion of a display used by an XR dev
 
 A [=view=] has an associated internal <dfn>view offset</dfn>, which is an {{XRRigidTransform}} describing the position and orientation of the [=view=] in the [=XRSession/viewer reference space=]'s [=coordinate system=].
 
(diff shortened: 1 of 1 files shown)


toji:
Great, thanks!

PR #1019, opened 2020-05-06
Add Texture (slice) support to XRWebGLLayer

jdashg:
Allow for rendering into app-created framebuffers using TEXTURE_2D_ARRAY WebGLTextures. (particularly for use with OVR_multiview2)

***
<a href="https://pr-preview.s3.amazonaws.com/jdashg/webxr/pull/1019.html" title="Last updated on May 6, 2020, 10:46 PM UTC (1c273ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1019/a64a328...jdashg:1c273ff.html" title="Last updated on May 6, 2020, 10:46 PM UTC (1c273ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1888,6 +1888,7 @@ interface XRWebGLLayer: XRLayer {
 
   // Static Methods
   static double getNativeFramebufferScaleFactor(XRSession session);
+  static boolean supportsViewType(XRWebGLLayerViewType viewType);
 };
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -1329,6 +1329,7 @@ An {{XRViewport}} object describes a viewport, or rectangular region, of a graph
   readonly attribute long y;
   readonly attribute long width;
   readonly attribute long height;
+  readonly attribute unsigned long textureSlice;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


toji:
This is an interesting proposal, but I'm not sure what it offers that isn't being captured in the [Layers Module](https://immersive-web.github.io/layers/)? Given that this relies on manual blitting of the multisampled renderbuffer it seems roughly equivalent to the methods that module will provide to create texture arrays.

jdashg:
It's a lot quicker to make this minor modification to the existing API, rather than the much more detailed layers proposal. This is based on feedback that doing the full layers implementation would take considerably more time, compared to extending the existing API.

cabanier:
The layers proposal as currently written allows for a UA to just implement a single projection layer which should be significantly easier to implement.

toji:
@jdashg: To clarify @cabanier's previous comment, as discussed in https://github.com/immersive-web/layers/issues/127 the current direction we're taking for the layers module is that any browser that implements it _must_ support projection layers (including exposing them as a texture array). Every other feature, including support for multi-layer compositing and quad/cylinder/equirect/cube layers is gated behind a feature descriptor that must be passed at session creation and which the UA is free to reject.

This division allows UAs to narrowly focus on exposing just the mechanics needed for more efficient stereo rendering, which really should be a fairly minor change on top of the existing layer plumbing, while still moving us towards a more generalized layer system. The complexity of the rest of the layers system is something that the UA will only need to implement if they feel it's beneficial to their platform.

Does that give you confidence that we can close this PR in favor of the Layers module path?

toji:
Since there's been no response for three weeks and we seem to be addressing the use case via the layers module, closing.

Issue #497, opened 2019-01-24
Which globals apply to security policy?


asajeffrey:
The current spec language for context isolation (https://immersive-web.github.io/webxr/#contextisolation-security) is phrased in terms of "page", for example:

> Also, to prevent CORS-related vulnerabilities each page will see a new instance of objects returned by the API, such as XRSession.

it is not obvious how to map this to HTML security policy (https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects), which is phrased in terms of "realm"s or "global"s. In particular, the spec should make clear which global is being referred to by "page", does this mean incumbent global?

NellWaliczek:
@Manishearth, is this one something you'd be comfortable picking up?

Manishearth:
I'm not very clear on what changes need to be made here, @asajeffrey is it just updating "page" to "incumbent global" with the appropriate xrefs?

(I'm not 100% familiar with all the globals stuff, especially how it pertains to security)

asajeffrey:
@Manishearth will that section will be non-normative when it stabilizes? If so, it's not as crucial to get the spec language 100% tight. The important thing is to make sure that whatever security goals we put in that section are actually met by the text!

The sentence "Also, to prevent CORS-related vulnerabilities each page will see a new instance of objects returned by the API, such as XRSession" seems to be validated by the requirement that each `Naviagator` gets an unique `XR` object.

The sentence "Similarly, methods invoked on the API MUST NOT cause an observable state change on other pages" seems weird, if security domain A opens an XR session, then domain B tries to, will there not be failures that B can observe which indicate that some other domain has opened a session?

Manishearth:
> will that section will be non-normative when it stabilizes?

I don't know, I'm not a spec editor :smile: @NellWaliczek might

toji:
I like to consider myself reasonably well versed in the workings of the web, but I've got to admit that trying to parse through what the various "realms" and "globals" actually correspond to here is making my brain melt. @asajeffrey: Could you help us resolve what globals we should be using if I describe the intent a bit better?

Some of the text in the example section could probably be dropped. For instance:

> Also, to prevent CORS-related vulnerabilities each page will see a new instance of objects returned by the API, such as XRSession. 

I think it would be extremely unusual (if not impossible) for Javascript objects created, for instance, in two separate tabs to receive the same object instance in both places. As such we probably don't need to specify that "our API should function like Javscript is normally expected to function."

[...]

PR #1023, opened 2020-05-11
Made it explicit which realm each object gets created in

asajeffrey:
Added text for each object creation saying which realm they get created in.

***
<a href="https://pr-preview.s3.amazonaws.com/asajeffrey/webxr-spec/pull/1023.html" title="Last updated on May 11, 2020, 8:47 PM UTC (a7f27e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1023/a64a328...asajeffrey:a7f27e6.html" title="Last updated on May 11, 2020, 8:47 PM UTC (a7f27e6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -28,6 +28,8 @@ spec:infra;
     type:dfn; text:continue
 spec:permissions-1;
     type:dfn; text:powerful feature
+spec:webidl;
+    type:dfn; text:new
 </pre>
 
 <pre class="anchors">
@@ -74,6 +76,7 @@ spec:html; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn; text: currently focused area; url: interaction.html#currently-focused-area-of-a-top-level-browsing-context
     type: dfn; text: responsible; url: webappapis.html#responsible-document
     type: dfn; text: rendering opportunity; url: webappapis.html#rendering-opportunity
+    type: dfn; text: current realm; url: webappapis.html#current
     type: dfn; text: same origin-domain; url: origin.html#same-origin-domain
     type: dfn; text: active document; url: browsers.html#active-document
     type: dfn; text: browsing context; url: browsers.html#browsing-context
@@ -87,6 +90,7 @@ spec: PageVisibility; urlPrefix: https://www.w3.org/TR/page-visibility-2/#
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: method; text: IsDetachedBuffer; url: sec-isdetachedbuffer
     type: dfn; text: ?; url: sec-returnifabrupt-shorthands
+    type: dfn; text: Realm; url: realm
     type: abstract-op; text: ToString; url: sec-tostring
 spec: dom; urlPrefix: https://dom.spec.whatwg.org/#
     type:algorithm; text:fire an event; url: concept-event-fire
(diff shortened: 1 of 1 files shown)


asajeffrey:
Fixes part of https://github.com/immersive-web/webxr/issues/497, in that every XR object now has a well-defined relevant realm.

asajeffrey:
Mostly this was routine, but there were a few odd cases:

- When creating a `XRWebGLLayer` there's a corner case where the GL context comes from a different realm than the XR session.
- When a user calls the `XRRigidTransform` constructor, we don't have an XR session, so we just have to use the current realm, which might not be associated with a session.

index.bs
```diff
@@ -629,7 +633,7 @@ When an {{XRSession}} is shut down the following steps are run:
 
 The <dfn method for="XRSession">end()</dfn> method provides a way to manually shut down a session. When invoked, it MUST run the following steps:
 
-  1. Let |promise| be [=a new Promise=].
+  1. Let |promise| be [=a new Promise=] in the [=relevant realm=] of this {{XRSystem}}.
```
Manishearth, line 636:
```suggestion
  1. Let |promise| be [=a new Promise=] in the [=relevant realm=] of this {{XRSession}}.
```

index.bs
```diff
@@ -1419,18 +1423,13 @@ interface XRRigidTransform {
 
 The <dfn constructor for="XRRigidTransform">XRRigidTransform(|position|, |orientation|)</dfn> constructor MUST perform the following steps when invoked:
 
-  1. Let |transform| be a new {{XRRigidTransform}}.
-  1. If |position| is not a {{DOMPointInit}} initialize |transform|'s {{XRRigidTransform/position}} to <code>{ x: 0.0, y: 0.0, z: 0.0, w: 1.0 }</code>.
+  1. Let |transform| be a [=new=] {{XRRigidTransform}} in the [=current realm=].
+  1. Let |transform|'s {{XRRigidTransform/position}} be a [=new=] {{DOMPointReadOnly}} in the [=current realm=].
```
toji, line 1427: Was worried that we lost something in this section, but since the constructor defaults line up with what we had originally specified this is a nice cleanup. Thanks!

index.bs
```diff
@@ -1456,25 +1455,25 @@ To <dfn for="XRRigidTransform">obtain the matrix</dfn> for a given {{XRRigidTran
...
 
-To <dfn lt="multiply transforms">multiply two {{XRRigidTransform}}s</dfn>, |B| and |A|, the UA MUST perform the following steps:
+To <dfn lt="multiply transforms">multiply two {{XRRigidTransform}}s</dfn>, |B| and |A| in a [=Realm=] |realm|, the UA MUST perform the following steps:
 
-  1. Let |result| be a new {{XRRigidTransform}} object.
-  1. Set |result|'s {{XRRigidTransform/matrix}} to the result of premultiplying |B|'s {{XRRigidTransform/matrix}} from the left onto |A|'s {{XRRigidTransform/matrix}}.
-  1. Set |result|'s {{XRRigidTransform/orientation}} to the quaternion that describes the rotation indicated by the top left 3x3 sub-matrix of |result|'s {{XRRigidTransform/matrix}}.
-  1. Set |result|'s {{XRRigidTransform/position}} to the vector given by the fourth column of |result|'s {{XRRigidTransform/matrix}}.
+  1. Let |result| be a [=new=] {{XRRigidTransform}} object in |realm|.
+  1. Set |result|'s {{XRRigidTransform/matrix}} to a [=new=] {{Float32Array}} in |realm|, the result of premultiplying |B|'s {{XRRigidTransform/matrix}} from the left onto |A|'s {{XRRigidTransform/matrix}}.
```
Manishearth, line 1474: tbh it feels like this step should instead use an abstract `[=matrix=]` variable since XRRigidTransform lazy-inits matrices

index.bs
toji: Oh, good catch! I just merged without noticing this, so I'll go apply the change myself.

index.bs
Manishearth: Oh, not really worth it.

I wish I had the time to do the matrix infra spec so that this stuff wouldn't be necessary. We'd also be able to directly spec rigid transforms so there's no decomp-recomp going on

index.bs
toji: Fixed in 8408e8824708f76ea9193d34b1791d7748fc7a93

PR #1028, opened 2020-05-11
Use `this` instead of "target {{foo}}"

Manishearth:
`[=this=]` is only defined for methods/attributes/etc, so in the case of algorithms I just gave the object a name in the algorithm definition.

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/1028.html" title="Last updated on May 11, 2020, 9:19 PM UTC (42e1281)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1028/8408e88...Manishearth:42e1281.html" title="Last updated on May 11, 2020, 9:19 PM UTC (42e1281)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -614,9 +614,8 @@ A number of different circumstances may <dfn>shut down the session</dfn>, which 
 
 <div class="algorithm" data-algorithm="shut-down-session">
 
-When an {{XRSession}} is shut down the following steps are run:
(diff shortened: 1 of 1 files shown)


toji:
Awesome, thanks! That "target {{XRWhatever}}" verbiage always felt weird to me.

PR #1030, opened 2020-05-11
Removed use of responsible and active and focused documents

### LINKED ISSUE #873: "responsible" is not an adjective that can be applied to documents
There are a few places in https://immersive-web.github.io/webxr/ that talk about "a responsible document" or "document is not responsible" with links to https://html.spec.whatwg.org/multipage/webappapis.html#responsible-document

That linked-to thing is a definition of a property of script settings object, and the property's name is "responsible document".   There is no separate concept of "responsible" that applies to documents in the specs right now.

Maybe the spec means "is the responsible document of some settings object"?  Does the spec really mean that, as opposed to, say "a document that has a browsing context" in the sense of <https://html.spec.whatwg.org/multipage/browsers.html#concept-document-bc>?  (Those concepts are not identical, but they _are_ for documents that are the active document of their browsing context, and doing the browsing context check would lead to a somewhat-natural fix to https://github.com/immersive-web/webxr/issues/872 as well, since you would then have a browsing context.)

asajeffrey:
The definition of "transitory consent" already requires that the UI control that triggered the consent event must either be a descendant of the browsing context requesting the session, or a same-origin ancestor. So it's not obvious that the extra security checks are buying anything.

Fixes #873.

***
<a href="https://pr-preview.s3.amazonaws.com/asajeffrey/webxr-spec/pull/1030.html" title="Last updated on May 15, 2020, 4:56 PM UTC (1532f45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1030/26ffc6e...asajeffrey:1532f45.html" title="Last updated on May 15, 2020, 4:56 PM UTC (1532f45)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2203,6 +2203,10 @@ User intention {#user-intention}
 
 It is often necessary to be sure of <dfn>user intent</dfn> before exposing sensitive information or allowing actions with a significant effect on the user's experience. This intent may be communicated or observed in a number of ways.
 
+Note: A common way of determining user intent is by [=transient activation=] of a UI control, typically an "enter VR" button. Since activation is transient,
+the [=browsing context=] requesting an XR session must be an [=ancestor=] or a [=same origin-domain=] [=descendant=] of the context containing the UI control, and must recently
+have been the [=active document=] of the browsing context.
+
 ### User activation ### {#user-activation}
 [=Transient activation=] MAY serve as an indication of [=user intent=] in some scenarios.
 
--- a/index.bs
+++ b/index.bs
@@ -78,7 +78,6 @@ spec:html; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn; text: rendering opportunity; url: webappapis.html#rendering-opportunity
     type: dfn; text: current realm; url: webappapis.html#current
     type: dfn; text: same origin-domain; url: origin.html#same-origin-domain
-    type: dfn; text: active document; url: browsers.html#active-document
     type: dfn; text: browsing context; url: browsers.html#browsing-context
 spec: SecureContexts; urlPrefix: https://w3c.github.io/webappsec-secure-contexts/#
     type: dfn; text: secure context; url: secure-contexts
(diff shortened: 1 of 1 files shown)


asajeffrey:
cc @avadacatvera

asajeffrey:
I can't spell @avadacatavra

Manishearth:
> The definition of "transitory consent" already requires that the UI control that triggered the consent event must either be a descendant of the browsing context requesting the session, or a same-origin ancestor. 

Can we perhaps have a note somewhere that  underscores this? Also, we don't seem to link to "transitory consent" anywhere?

asajeffrey:
Sorry, I meant "transient activation" not "transitory consent". It's in  the defn of immersive-session-allowed.

Good idea re a note.

index.bs
```diff
@@ -2307,7 +2283,6 @@ To determine if <dfn>poses may be reported</dfn> to an {{XRSession}} |session|,
 
   1. Let |document| be the document that owns |session|.
   1. If the request does not originate from |document|, return <code>false</code>.
-  1. If |document| is not [=active and focused=], return <code>false</code>.
```
Manishearth, line 2310: Hmm, isn't removing this a behavior change?

Manishearth: Actually, we handle this in the "currently focused area" section.

asajeffrey: Yes, this is a behavior change, in that in the old spec there was a race condition, if a user gave consent (e.g. by clicking on en "enter VR" button) then quickly changed the focus to an iframe, the old check would fail but the new check succeeds.

Manishearth: I don't understand: this is a per-frame thing about when poses can be reported. What per-frame behavior is here now?

Manishearth: Shouldn't tabbing out of a tethered session pause the session?

Manishearth: It is weird that tabbing out would cause an exception to be thrown, though. In this case it might be better to report a null pose (something I plan to bring up later)

asajeffrey: I still think we shouldn't base pose data on the state of the originating document, since the user can't see the originating document while they've got a headset on

I added back text requiring document visibility, we should discuss this in a separate issue.

PR #1031, opened 2020-05-11
Short circuit updateRenderState() if no render state is passed

### LINKED ISSUE #880: Corrections and clarifications for description/steps for updateRenderState()
A couple of things about the information provided for [`updateRenderState()`](https://immersive-web.github.io/webxr/#dom-xrsession-updaterenderstate):

First, item 3 has a minor typo: "If newState's baseLayer's was created..." The "'s" should be removed from "baseLayer's", leaving the text as "If newState's baseLayer was created..."

More importantly, it would be useful to explicitly state what this method does if you do not provide the optional parameter. I know it's optional because WebIDL requires that dictionary parameters with all optional members must also be optional, but given that, the spec should be clear on what the method does when this is the case. Does the method have no effect, or does it still cause some secondary effect to occur? Presumably it's essentially a no-op but it would be wise to be specific either way.

Manishearth:
Fixes #880

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/1031.html" title="Last updated on May 11, 2020, 10:03 PM UTC (fa56fc2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1031/f22797e...Manishearth:fa56fc2.html" title="Last updated on May 11, 2020, 10:03 PM UTC (fa56fc2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -663,11 +663,11 @@ NOTE: The <a href="https://immersive-web.github.io/layers">WebXR layers module</
 The <dfn method for="XRSession">updateRenderState(|newState|)</dfn> method queues an update to the [=active render state=] to be applied on the next frame. Unset fields of the {{XRRenderStateInit}} |newState| passed to this method will not be changed.
 
 When this method is invoked, the user agent MUST run the following steps:
-
   1. Let |session| be [=this=].
   1. If |session|'s [=ended=] value is <code>true</code>, throw an {{InvalidStateError}} and abort these steps.
   1. If |newState|'s {{XRRenderStateInit/baseLayer}} was created with an {{XRSession}} other than |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |newState|'s {{XRRenderStateInit/inlineVerticalFieldOfView}} is set and |session| is an [=immersive session=], throw an {{InvalidStateError}} and abort these steps.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -664,11 +664,11 @@ NOTE: The <a href="https://immersive-web.github.io/layers">WebXR layers module</
...
 
 When this method is invoked, the user agent MUST run the following steps:
-
   1. Let |session| be the target {{XRSession}}.
   1. If |session|'s [=ended=] value is <code>true</code>, throw an {{InvalidStateError}} and abort these steps.
-  1. If |newState|'s {{XRRenderStateInit/baseLayer}}'s was created with an {{XRSession}} other than |session|, throw an {{InvalidStateError}} and abort these steps.
+  1. If |newState|'s {{XRRenderStateInit/baseLayer}} was created with an {{XRSession}} other than |session|, throw an {{InvalidStateError}} and abort these steps.
   1. If |newState|'s {{XRRenderStateInit/inlineVerticalFieldOfView}} is set and |session| is an [=immersive session=], throw an {{InvalidStateError}} and abort these steps.
+  1. If none of |newState|'s {{XRRenderStateInit/depthNear}}, {{XRRenderStateInit/depthFar}}, {{XRRenderStateInit/inlineVerticalFieldOfView}}, {{XRRenderStateInit/baseLayer}}, {{XRRenderStateInit/layers}} are set, abort these steps.
```
toji, line 671: I'm slightly concerned that this lends itself to accidentally missing future additions to the dictionary, but given that partial dictionaries are apparently verboten I guess that's not as bad as it could be.

Manishearth: This was my concern as well, unfortunately since dictionaries aren't real you can't just "check if it's empty"

Manishearth:
Rebased

Issue #1022, opened 2020-05-11
All tasks should be queued with a task source


Manishearth:
The "queue a task" algorithm must be spawned with a [task source](https://html.spec.whatwg.org/multipage/webappapis.html#task-source). We should probably have our own, since that gives UAs the freedom to schedule accordingly.

Originally brought up by @annevk in https://github.com/immersive-web/webxr/issues/1020

Issue #1020, opened 2020-05-11
Queue a task before resolving a promise


annevk:
You can no longer resolve a promise from "in parallel" directly. It requires a task that ideally identifies the task source and document it's associated with.

Manishearth:
Makes sense. We don't have a task source we use right now: should we just make our own? Should we use the DOM task source for most things, and the user input one for input-related stuff.

annevk:
Your own task source might make sense if in principle the tasks don't need to be ordered relative to other things. Gives user agents more freedom. (I suspect there might be more changes to this setup going forward, but this should at least set you up a bit better.)

Manishearth:
Yeah there's very little dependent ordering

Manishearth:
Filed a separate issue for the task source: https://github.com/immersive-web/webxr/issues/1022

PR #1032, opened 2020-05-11
Cleanups on how we do tasks and promises

Manishearth:
fixes https://github.com/immersive-web/webxr/issues/1020, fixes https://github.com/immersive-web/webxr/issues/1022

I didn't update every instance of "queue a task" because both WebGL and HTML's Websocket section specify the task source used once in the document.

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/1032.html" title="Last updated on May 11, 2020, 10:13 PM UTC (17301e4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1032/f22797e...Manishearth:17301e4.html" title="Last updated on May 11, 2020, 10:13 PM UTC (17301e4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -313,8 +313,8 @@ When this method is invoked, it MUST run the following steps:
   1. Run the following steps [=in parallel=]:
     1. [=ensures an immersive XR device is selected|Ensure an immersive XR device is selected=].
     1. If the [=XRSystem/immersive XR device=] is null, [=/resolve=] |promise| with <code>false</code> and abort these steps.
-    1. If the [=XRSystem/immersive XR device=]'s [=list of supported modes=] does not [=list/contain=] |mode|, [=/resolve=] |promise| with <code>false</code> and abort these steps.
-    1. [=/Resolve=] |promise| with <code>true</code>.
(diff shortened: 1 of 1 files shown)


Manishearth:
cc @annevk

index.bs
```diff
@@ -313,8 +313,8 @@ When this method is invoked, it MUST run the following steps:
   1. Run the following steps [=in parallel=]:
     1. [=ensures an immersive XR device is selected|Ensure an immersive XR device is selected=].
```
annevk, line 314: This step is running in a parallel thread but then you invoke something that grabs objects from the main thread. I think you need to make that more clear and separate what happens on the main thread and what happens in parallel.

Also, the selection that happens here presumably updates the state on the main thread. That should all happen in the same task that eventually resolves the promise. This also goes for the step below here where you resolve with false.

Does that makes sense?

(I suspect something similar applies to the other in parallel algorithm, but I did not look in detail.)

Manishearth: Hmm, so that's a little bit tricky: there are some parts of the ensure selected algorithm that need to run on the main thread, and some that don't. There doesn't seem to be an easy way to chain async algorithms in spec text.

Manishearth: I'm going to merge this PR because this is an unrelated issue, but filing a new one.

Issue #854, opened 2019-09-21
Result of calling checkFramebufferStatus on an opaque framebuffer


asajeffrey:
https://immersive-web.github.io/webxr/#opaque-framebuffer doesb't say what the result of calling `checkFramebufferStatus` on an opaque framebuffer should be.

asajeffrey:
Should it be `FRAMEBUFFER_COMPLETE` during an rAF and `FRAMEBUFFER_UNSUPPORTED` otherwise?

toji:
We do have text that says:

> When not in a `requestAnimationFrame()` callback calls to `checkFramebufferStatus` outside of a `requestAnimationFrame()` callback MUST generate a `FRAMEBUFFER_UNSUPPORTED` error

But that doesn't explicitly cover what the status should be inside a rAF, and we should add that. It should indeed be `FRAMEBUFFER_COMPLETE`, but I'm not sure if we can say that it MUST be that. Need to give some thought to if there's situations where it would be necessary or appropriate for the framebuffer to report something other than complete inside a rAF.

asajeffrey:
Yes, that's what I was wondering, is there ever a time when the session is in an rAF and it returns anything other than `FRAMEBUFFER_COMPLETE`?

toji:
Thinking out loud: Most common scenario I can think of is context loss but that behavior is already described in the WebGL spec (returns `FRAMEBUFFER_UNSUPPORTED`) and we shouldn't override it. Could possibly run into an `OUT_OF_MEMORY` situation as well, but in that case your whole context is pretty effectively screwed and reporting complete vs. unsupported is the least of your problems. 

Beyond that most of the things that would potentially cause an incomplete framebuffer are state management issues, and we should be able to trust that the browser is capable of handling that properly. So yeah, I'm gonna say that we should always consider it complete in the rAF callback (barring the existing behavior for lost contexts).

rcabanier:
Do we know what happens today?

Manishearth:
Closing, this is already technically specified and specifying it gets too redundant: https://github.com/immersive-web/webxr/pull/1035

PR #1035, opened 2020-05-11
Explicitly specify that checkFramebufferStatus() returns a normal code during rAF

Manishearth:
Fixes https://github.com/immersive-web/webxr/issues/854

This text really feels redundant to me, however, perhaps we should just close the original issue? The definition of an opaque framebuffer is that it behaves normally except when otherwise specified.

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/1035.html" title="Last updated on May 11, 2020, 10:59 PM UTC (e6a3579)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1035/fb2ff08...Manishearth:e6a3579.html" title="Last updated on May 11, 2020, 10:59 PM UTC (e6a3579)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -44,6 +44,7 @@ spec: WebGL; urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/1.0/
     type: typedef; text: INVALID_OPERATION; url: WebGLRenderingContextBase
     type: typedef; text: INVALID_FRAMEBUFFER_OPERATION; url: WebGLRenderingContextBase
     type: typedef; text: FRAMEBUFFER_UNSUPPORTED; url: WebGLRenderingContextBase
+    type: typedef; text: FRAMEBUFFER_COMPLETE; url: WebGLRenderingContextBase
     type: method; text: clear; url: 5.14.11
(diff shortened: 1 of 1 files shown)


Manishearth:
cc @asajeffrey

toji:
This feels redundant to me as well, though I'd be curious to hear if @asajeffrey feels differently since the request was originally from him?

The updated linking for `checkFrameBufferStatus()` is good. I'd like to keep that even if the new bullet point is discarded.

asajeffrey:
This bullet point is kind of redundant. We only need the extra bullet point if we want to sat that inside an rAF, an opaque framebuffer is always complete.

Manishearth:
Should we then just close this and the linked issue?

asajeffrey:
Fine, but this is a different decision to the one in https://github.com/immersive-web/webxr/issues/854#issuecomment-534287001 (which was that opaque framebuffer are always complete inside an rAF).

Manishearth:
@asajeffrey well, that also included OUT_OF_MEMORY

PR #1037, opened 2020-05-12
Minor change to when empty input profile arrays are appropriate.

### LINKED ISSUE #781: Additional situations when XRInputSource.profiles should be empty
In the spec, "For example, if the gamepad attribute is not null the profile identifies the mapping of the buttons and axes of the attribute of the XRInputSource." Does this imply that the profiles list should be empty for an XRInputSource that does not have a gamepad? If not, then some text clarifying that it's still useful even without a gamepad would be helpful.

Also, should the profiles list be empty for transient input sources? For example, WMR voice commands are transient input sources that can occur in immersive sessions.

toji:
Fixes #781.

Turns out most of the text was already there, I just added one more scenario and made sure they were clearly seen as examples rather than the only applicable situations.

***
<a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/pull/1037.html" title="Last updated on May 12, 2020, 10:42 PM UTC (b99430f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1037/b3a95eb...b99430f.html" title="Last updated on May 12, 2020, 10:42 PM UTC (b99430f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1570,7 +1570,7 @@ Profiles are given in descending order of specificity. Any [=input profile name=
 
 If the {{XRSession}}'s [=mode=] is {{XRSessionMode/"inline"}}, {{XRInputSource/profiles}} MUST be an empty list.
 
-The user agent MAY choose to only report an appropriate generic [=input profile name=]s or an empty list at it's discretion. Some scenarios where this would be appropriate are if the input device cannot be reliably identified, no known input profiles match the input device, or the user agent wishes to mask the input device being used.
+The user agent MAY choose to only report an appropriate generic [=input profile name=]s or an empty list at its discretion. Some scenarios where this would be appropriate are if the input device cannot be reliably identified, no known input profiles match the input device, or the user agent wishes to mask the input device being used.
 
 <p class="note">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1570,7 +1570,7 @@ Profiles are given in descending order of specificity. Any [=input profile name=
 
 If the {{XRSession}}'s [=mode=] is {{XRSessionMode/"inline"}}, {{XRInputSource/profiles}} MUST be an empty list.
 
-If the input device cannot be reliably identified, or the user agent wishes to mask the input device being used, it MAY choose to only report generic [=input profile name=]s or an empty list.
+The user agent MAY choose to only report an appropriate generic [=input profile name=]s or an empty list at it's discretion. Some scenarios where this would be appropriate are if the input device cannot be reliably identified, no known input profiles match the input device, or the user agent wishes to mask the input device being used.
```
Manishearth, line 1573:
```suggestion
The user agent MAY choose to only report an appropriate generic [=input profile name=]s or an empty list at its discretion. Some scenarios where this would be appropriate are if the input device cannot be reliably identified, no known input profiles match the input device, or the user agent wishes to mask the input device being used.
```

Issue #1042, opened 2020-05-15
Allow poses to be null in more situations


Manishearth:
https://immersive-web.github.io/webxr/#populate-the-pose

Currently `pose` is only ever set to `null` when `space’s pose relative to baseSpace has never been determined`, otherwise it's returned with the last known pose with `emulatedPosition=true`.

As we add more and more spaces, I think it's worthwhile to allow poses to be null in more situations. For example, this behavior isn't great for joint spaces: if my hand is behind me and I turn around, there will be this weird floating claw with the current spec behavior. This also applies somewhat to general input spaces. Emulation is sometimes possible, but not always, and it might be worth allowing user choice here.

Furthermore, currently we throw a `SecurityError` when [poses may not be reported](https://immersive-web.github.io/webxr/#poses-may-be-reported). The current conditions for that are about:

 - where the request comes from
 - document focus
 - user intent
 - whether we "know the pose is not exposing fingerprintable data"

It seems incorrect to throw an _error_ for situations that can unpredictably change, in particular the "document focus" and "do we know if the pose is exposing fingerprintable data" should probably cause `populate the pose` to return `null`, not throw an error.

This also handles the case of input poses during visible-blurred sessions (which is discussed in https://github.com/immersive-web/webxr/issues/1008). We should just null them out.

It seems like overall we should be allowing poses to be null when:

 - the pose cannot be reported for temporary security reasons (visible-blurred, fingerprintable, document focus)
 - the pose is unknown and cannot be emulated, where "can I emulate this" is up to the user agent. "Report the last known value of the pose" is a valid form of emulation so user agents can continue with the current specced behavior.

Applications need to deal with null poses _anyway_, so it seems like this change should be fine.

Manishearth:
I also believe that sometimes returning null poses mid-session is the current behavior in Chromium (see [XRSpace::getPose](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/xr/xr_space.cc;drc=2f11470d7ad8963a9add116df64d2edd1b85d3a4;l=71?originalUrl=https:%2F%2Fcs.chromium.org%2F) returning null), and it is the current behavior in Servo.

Manishearth:
> Furthermore, currently we throw a SecurityError when poses may not be reported. The current conditions for that are about:

Actually, for the "document not in focus" one, we can simply _not_ trigger an XR frame instead.

Manishearth:
[...]

Issue #1041, opened 2020-05-15
Reporting pose data depends on the state of the document?


asajeffrey:
In https://immersive-web.github.io/webxr/#protect-poses there's a per-frame check based on the document associated with the session. This seems like it may have unexpected UX consequences, as the user is no longer interacting with that document, since they have a headset on.

asajeffrey:
This came up in #1030, in conversation with @Manishearth

asajeffrey:
As an example, `magazine.com` might have an advert from `car-manufacturer.com` with VR content (which the magazine cam bless as a feature on the `iframe` containing the ad). If the user enters the ad, but while they're in the headset, accidentally clicks or scrolls away from the advert, the pose data will stop being reported.

Manishearth:
cc @avadacatavra

/agenda for discussions of security implications

Manishearth:
I'm also a bit uneasy about this _throwing a SecurityError_, i'd rather it make the pose null. See https://github.com/immersive-web/webxr/issues/1042 for more discussion about that.

Manishearth:
Related: https://github.com/immersive-web/webxr/issues/1042

avadacatavra:
@Manishearth can you elaborate on why you're uneasy about throwing a Security Error?

Manishearth:
@avadacatavra I'm totally for throwing the security error in _some_ of the cases there.

In general for web specs errors should be thrown in cases where the application _did something wrong_ (and thus, should only be thrown in cases which are preventable). In other words, if an error is being thrown, the user should be able to restructure their code to avoid it.

This is possible for the case when realms don't match ("don't call `iframe.contentWindow.currentFrame.getPose()`"), but the other cases are unpredictable. It's possible to have a race condition with the "hidden" state, and it's not possible at all to handle this focus condition well. Throwing a security error when the user switches focus seems wrong because we're telling the author that "they did something wrong", but the user switching focus is not something the author can easily detect so there is no better thing to do here.

This is why I think a null pose instead of an error makes sense. Null pose means "no pose for you", an error means "you did something wrong, please fix it".

Manishearth:
This is fixed in https://github.com/immersive-web/webxr/pull/1058 : the document being hidden causes webxr to report null or emulated (cached) poses.

asajeffrey:
Where "this" is returning null rather than throwing an exception, not the original issue that I don't think we should be doing this check at all!

Manishearth:
[...]

Issue #674, opened 2019-05-31
Should requestReferenceSpace() wait until getViewerPose() won't be null?


thetuvix:
@NellWaliczek had an interesting idea for `requestReferenceSpace()` to wait until the viewer pose won't be null.  I do wonder, however, if that won't cause other problems in apps that were tested on systems where this continues immediately.

Thoughts?

_Originally posted by @thetuvix in https://github.com/immersive-web/webxr/pull/559/review_comment/create_

cwilso:
Blocked on finishing #243.

klausw:
This is important for smartphone AR, where there may be several seconds delay between session start and establishing tracking for valid poses. If the `requestReferenceSpace()` response is delayed, the application would need to restructure the main drawing loop to either work without a reference space, or use a separate `viewer` reference space, if it wants to draw before getting a pose.

On the flip side, a delayed response could potentially mean that poses are effectively never null on some platforms, so apps assuming that may no longer cope well with tracking loss or transient null poses.

(Overall, I think either approach would work, it's mainly a matter of clearly documenting and establishing best practices for apps.)

thetuvix:
Alternatively, for smartphone AR, the origin of a `local` reference space is arbitrary anyway, and would likely just end up at the position of the first 6DoF pose generated by the device.  Could a `local` reference space just start tracking right away with live orientation but an `emulated` position at the origin, and then set `emulated` to false once 6DoF tracking is established?  Or is even orientation tracking not available for those few seconds?

klausw:
@thetuvix wrote:
> Could a local reference space just start tracking right away with live orientation but an emulated position at the origin

Interesting idea, but this seems to add a fair amount of complexity to the implementation, especially if it would require using a separate source of orientation data if the underlying AR platform doesn't deliver any pose information at all before 6DoF tracking is established.

A typical use case is that an AR app may want to start drawing instructions or a non-world-attached intro banner on startup, then waiting for a signal that tracking got established. Currently, that signal is generally poses becoming non-null, and in the alternate proposal it would be getting a reference space. A switch from emulated to non-emulated position seems a bit odd to me, but I don't have a strong opinion as long as the behavior is sufficiently clearly documented to avoid implementation differences.

NellWaliczek:
[...]

PR #1058, opened 2020-05-22
Various changes around null and emulated poses

Manishearth:
This contains changes to:

 - Allow poses to be null in more situations (fixes https://github.com/immersive-web/webxr/issues/1042)
 - <s>Force emulation in `getViewerPose()`, making it not possible to return null poses, and delay `requestReferenceSpace()` until tracking has been established.</s> (https://github.com/immersive-web/webxr/issues/674)
 - Reintroduce check for current global object in pose security check (removed in #1030)
 - Make it so that the document visibility check can suppress poses without throwing an exception. Exceptions are for cases where the user's code is incorrect, not when stuff unpredictably changed from underneath the user. (partially addresses https://github.com/immersive-web/webxr/issues/1041)
 - Some spec language cleanups around reference space creation

For users, the main change here is that `requestReferenceSpace()` might have some delay involved in resolving the promise, but on the plus side `getViewerPose()` will never be null.

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/1058.html" title="Last updated on May 29, 2020, 5:12 AM UTC (6758a5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1058/58b5731...Manishearth:6758a5b.html" title="Last updated on May 29, 2020, 5:12 AM UTC (6758a5b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2287,6 +2287,7 @@ When based on sensor data, {{XRPose}} and {{XRViewerPose}} will expose [=sensiti
 
 To determine if <dfn>poses may be reported</dfn> to an {{XRSession}} |session|, the user agent MUST run the following steps:
 
+  1. If |session|'s [=relevant global object=] is not the [=current global object=], return <code>false</code>.
   1. If |session|'s {{XRSession/visibilityState}} in not {{XRVisibilityState/"visible"}}, return <code>false</code>.
   1. Determine if the pose data can be returned as follows:
     <dl class="switch">
--- a/index.bs
+++ b/index.bs
@@ -760,6 +760,8 @@ When any previously added <dfn for="XRSession" lt="remove input source">[=XR inp
     1. [=list/Remove=] each {{XRInputSource}} in |removed| from |session|'s [=list of active XR input sources=].
     1. Fire an {{XRInputSourcesChangeEvent}} named {{inputsourceschange!!event}} on |session| with {{XRInputSourcesChangeEvent/removed}} set to |removed|.
 
+Note: The user agent MAY fire this event when an input source temporarily loses both position and orientation tracking. It is recommended that this only be done for physical handheld controller input sources. It is not recommended that this event be fired when this happens for tracked hand input sources, because this will happen often, nor is it recommended when this happens for tracker object input sources, since this makes it harder for the application to maintain a notion of identity.
+
 </div>
 
 <div class="algorithm" data-algorithm="on-input-source-change">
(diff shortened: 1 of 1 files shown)


Manishearth:
/agenda to discuss the changes being made here

Manishearth:
In the call today there seemed to be general consensus around allowing `getPose()` to be null for arbitrary spaces.

It was pointed out that for handheld input controllers we may wish to encourage them to never return `null` when the input source is active and instead trigger `oninputsourceschange` events. (Worth noting, you can still hold on to a reference to an old input source's space, and that should definitely return `null` in `getPose()`). I could potentially add language to that effect somewhere, though I'd rather wait for @toji to get in an initial review.

The point about `getViewerPose()` did not get discussed as much, however this PR fixes the main issues that caused contention in #674, so it feels like that change is a good one.

alcooper91:
I thought in a previous call the resolution for #674 that was discussed was to not enforce any particular behavior and to leave that to be a UA choice? (Relevant Minutes: https://www.w3.org/2020/05/12-immersive-web-minutes.html)

Manishearth:
Hmm, you're right. However, I think neither I nor @toji had a strong opinion on that, and _my_ opinion has shifted towards Alex's original one of "`getViewerPose()` should never be `null`" now that `getPose()` is much more free to be `null`. The reason I felt that `getViewerPose()` should be allowed to be `null` was really that I felt that "populate a pose" should be null in more situations so that `getPose()` can be null.

cc @thetuvix 

If you or others feel strongly I can revert that bit.

toji:
Sorry it's taken me a while to comment on this. Wanted to hear the call discussion and then come back. I'm OK with everything here with the exception of blocking `requestReferenceSpace()` resolves so that `getViewerPose()` is never null. (That said, I do like the forced emulation bit once tracking have been established.)

My biggest concern is that a very common pattern in many apps is to wait for a reference space before spinning up the rAF loop, but in many cases the device will have already switched over to presenting the session's content as soon as the session is created. If there's a need for some action on the user's part to help establish tracking (as is commonly the case for mobile AR, where you want to wiggle your phone around) then having that rAF loop available as a point to ping the tracking state is useful, as well as offering a logical place to draw any helper dialogs with a more reliable (viewer?) space.

Manishearth:
Oh, that's a fair point. I'll remove those changes and fix up the rest of the PR. You should leave a comment to that effect on the other issue, and perhaps we should reagenda it.

Manishearth:
@toji oh, I also added a note about hand controllers losing tracking, as discussed in the call. want to quickly check that as well?

toji:
Controller comment LGTM too!

Manishearth:
Going to merge, the only part of this PR that was not discussed fully in the call was the one I removed.

PR #1068, opened 2020-05-29
Wait till poses are non-null for `requestReferenceSpace()`

### LINKED ISSUE #674: Should requestReferenceSpace() wait until getViewerPose() won't be null?

Manishearth:
This implements the change proposed in #674. There does not seem to be general consensus around that change at the moment, and over the past few meetings it seems like opinion has shifted away from this, so the correct path forward is likely to close that issue and this PR.

Fixes #674

***
<a href="https://pr-preview.s3.amazonaws.com/Manishearth/webxr/pull/1068.html" title="Last updated on May 29, 2020, 3:57 PM UTC (d506182)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1068/403f2e5...Manishearth:d506182.html" title="Last updated on May 29, 2020, 3:57 PM UTC (d506182)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -715,6 +715,8 @@ When this method is invoked, the user agent MUST run the following steps:
   1. Run the following steps [=in parallel=]:
     1. If the result of running [=reference space is supported=] for |type| and |session| is <code>false</code>, [=queue a task=] to [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
     1. Set up any platform resources required to track reference spaces of type |type|.
+    1. Wait until a time when the [=native origin=] of an {{XRReferenceSpace}} of type |type| will be tracked or has been tracked at any previous point in the lifecycle of |session|.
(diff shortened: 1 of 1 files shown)


Manishearth:
There weren't any strong reactions during the call, so I think we're going to move ahead with merging this one. If folks end up having any issues, please let us know!

r? @toji

Manishearth:
There weren't any strong reactions during the call, so I think we're going to move ahead with merging #1070. If folks end up having any issues, please let us know!

PR #1076, opened 2020-06-05
Add CG specs links

Yonet:
Added the CG spec repos to readme. Let me know if I am missing anything or want to add it somewhere else instead.


himorin:
I'm a fan to have link in `[xx](url)` format, but are we better to align with others in the file (having references by numbers, like `[xx][number]` and `[number]: url`)?
Even if we set to have links in reference format, I'd push to have them with short name or something rather than numbers. (it's hard to copy and paste to other REAMDEs with reference by numbers)

PR #1108, opened 2020-08-05
Update CODE_OF_CONDUCT.md

AdaRoseCannon:
This PR is to update the CoC file to only use the W3C one. The W3C code of conduct has been updated and since it is very similar to our existing code of conduct it is redundant to use both.


toji:
Awesome, thank you!

PR #1111, opened 2020-08-07
Change timing of when updateRenderState changes apply

### LINKED ISSUE #1051: Difficulty with multiprocess implementation of updateRenderState
The spec for `updateRenderState` (https://immersive-web.github.io/webxr/#dom-xrsession-updaterenderstate) is that it synchronously updates the pending render state, which is then applied (synchronously) to be the new active render state just before rAF callbacks are executed (https://immersive-web.github.io/webxr/#xr-animation-frame).

The end result of this is that it is very difficult to provide a mutiprocess implementation of `updateRenderState`, where the script is run in one process and the webxr device driver is run in another. Consider:
```js
   flag = false;
   session.updateRenderState(A);
   button.onclick = function () { flag = true; session.updateRenderState(B); }
   session.requestAnimationFrame(function f() { 
      session.requestAnimationFrame(f);
      if (flag) { assert(render state is B); }
   });
```
According to the spec, the assertion can never fail, but it is difficult to implement this without shared memory.

toji:
Fixes #1051. Would love @asajeffrey's input on this!

Apologies that this has taken as long as it has to get a PR up. Doing some GH and meeting minutes archeology, the primary concerns about the algorithm as it currently is written seem to be that Alan needs it to allow for non-blocking communication to the XR hardware and otherwise myself and a couple of others just want to ensure that the timing of when the changes apply is reasonably predictable.

This change moves the  `apply the pending render state` execution to immediately after the frame loop executes, and queues a task to do it. Then it also moves the execution of the frame loop into a task on the same queue, which _should_ implicitly ensure that the render state update must finish before the next frame executes. (I believe that the frame loop execution was effectively taking place in a task before anyway, so this just makes that explicit. Please correct me if I'm wrong!)

This gives us the behavior that any changes will apply some time after the frame finishes executing, but the only guarantee is that they will be finished applying by the time the next frame begins executing. This means the most predictable way to interact with the render state is to inspect and update it within a frame, with the expectation that any changes will take effect in the next frame.

This is technically a backwards-compatibility break, but I have a hard time imagining that any developers are strongly dependent on the current timing.

***
<a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/pull/1111.html" title="Last updated on Aug 7, 2020, 7:38 PM UTC (df78902)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1111/127a84f...df78902.html" title="Last updated on Aug 7, 2020, 7:38 PM UTC (df78902)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -707,10 +707,11 @@ When requested, the {{XRSession}} |session| MUST <dfn>apply the pending render s
   1. Let |activeState| be |session|'s [=active render state=].
   1. Let |newState| be |session|'s [=pending render state=].
   1. Set |session|'s [=pending render state=] to <code>null</code>.
-  1. Let |oldLayer| be |activeState|'s {{XRRenderState/baseLayer}}.
+  1. Let |oldBaseLayer| be |activeState|'s {{XRRenderState/baseLayer}}.
+  1. Let |oldLayers| be |activeState|'s {{XRRenderState/layers}}.
   1. [=Queue a task=] to perform the following steps:
     1. Set |activeState| to |newState|.
-    1. If |oldLayer| is not equal to |activeState|'s {{XRRenderState/baseLayer}}, or the dimensions of |oldLayer| are different from |baseLayer|, [=update the viewports=] for |session|.
+    1. If |oldBaseLayer| is not equal to |activeState|'s {{XRRenderState/baseLayer}}, |oldLayers| is not equal to |activeState|'s {{XRRenderState/layers}}, or the dimensions of any of the layers have changed, [=update the viewports=] for |session|.
     1. If |activeState|'s {{XRRenderState/inlineVerticalFieldOfView}} is less than |session|'s [=minimum inline field of view=] set |activeState|'s {{XRRenderState/inlineVerticalFieldOfView}} to |session|'s [=minimum inline field of view=].
     1. If |activeState|'s {{XRRenderState/inlineVerticalFieldOfView}} is greater than |session|'s [=maximum inline field of view=] set |activeState|'s {{XRRenderState/inlineVerticalFieldOfView}} to |session|'s [=maximum inline field of view=].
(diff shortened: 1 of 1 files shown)


toji:
Preemptively /agenda -ing this.

index.bs
```diff
@@ -708,22 +708,23 @@ When requested, the {{XRSession}} |session| MUST <dfn>apply the pending render s
...
-      <dt> If |session|'s [=XRSession/mode=] is {{XRSessionMode/"inline"}} and |baseLayer| is an instance of an {{XRWebGLLayer}} with [=XRWebGLLayer/composition disabled=] set to <code>true</code>
-      <dd> Set |activeState|'s [=XRRenderState/composition disabled=] boolean to <code>true</code>.
-      <dd> Set |activeState|'s [=XRRenderState/output canvas=] to |baseLayer|'s [=XRWebGLLayer/context=]'s {{WebGLRenderingContext|canvas}}.
-      <dt> Otherwise
-      <dd> Set |activeState|'s [=XRRenderState/composition disabled=] boolean to <code>false</code>.
-      <dd> Set |activeState|'s [=XRRenderState/output canvas=] to <code>null</code>.
-    </dl>
+  1. [=Queue a task=] to perform the following steps:
+    1. Set |activeState| to |newState|.
+    1. If |oldLayer| is not equal to |activeState|'s {{XRRenderState/baseLayer}}, or the dimensions of |oldLayer| are different from |baseLayer|, [=update the viewports=] for |session|.
```
asajeffrey, line 713: Might be worth making sure that we've put the right hooks in for the layers spec to make changes? E.g. in this case, the layers spec is going to want to do this when the layers array has changed.

toji: Good point! In this case it seems like, since the `layers` attribute is defined in this spec, we can check for differences without doing the same sort of "create an algorithm to be overridden" pattern. Gave it a pass with some fairly broad language around detecting changes, let me know what you think!

toji:
Merging this now, given that Manish and Alan have both signed off on it. I think I'll leave it on the agenda to mention as a "heads up" for other implementers.

PR #1134, opened 2020-09-29
Clarify that minimum viewport scale may change

klausw:
The minimum viewport scale must ensure that the viewport has a nonzero
size, and this may in theory require an adjustment if the session's
baseLayer is changed during a session to use a very small
framebufferScaleFactor.

Change the wording to clarify that it's not necessarily constant for an
entire session.

***
<a href="https://pr-preview.s3.amazonaws.com/klausw/webxr/pull/1134.html" title="Last updated on Sep 30, 2020, 1:22 AM UTC (919e516)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1134/132fa32...klausw:919e516.html" title="Last updated on Sep 30, 2020, 1:22 AM UTC (919e516)">Diff</a>


klausw:
On second thought, I think it would be easier to enforce the nonzero size
in getViewport, then this change isn't needed and there's less ambiguity
about baseLayer changes.

cabanier:
> On second thought, I think it would be easier to enforce the nonzero size in getViewport, then this change isn't needed and there's less ambiguity about baseLayer changes.

Yes, you can update `getViewPort` to make that adjustment. 
Are there UA that require a certain minimum size for the viewport? If so, you could make the adjustment in that function as well.

klausw:
> > On second thought, I think it would be easier to enforce the nonzero size in getViewport, then this change isn't needed and there's less ambiguity about baseLayer changes.
> 
> Yes, you can update `getViewPort` to make that adjustment.

I've moved the logic there.

> Are there UA that require a certain minimum size for the viewport? If so, you could make the adjustment in that function as well.

I'm not aware of a specific UA requirement (beyond ensuring nonzero size), but I added an optional adjustment there where the UA could clamp to a minimum viewport scale.

Manishearth (approved):

toji (approved):

PR #1143, opened 2020-10-26
Fixes in webvr-migration.md

ericmick:
Just fixing a couple of mistakes that might trip up anyone using the webvr migration guide code snippets.


dontcallmedom:

toji:
Thanks!

PR #1152, opened 2020-12-16
Create dynamic-viewport-scaling-privacy-security.md

klausw:
This is intended as supporting material for a TAG review of the dynamic viewport scaling addition to the WebXR API.

TAG review request: https://github.com/w3ctag/design-reviews/issues/588

(I'd normally keep this in a repository related to a specific feature being incubated, but in this case the change didn't have a separate incubation repository. Please let me know in case you'd prefer a different location or procedure.)

### DIFF
--- a/designdocs/dynamic-viewport-scaling-privacy-security.md
+++ b/designdocs/dynamic-viewport-scaling-privacy-security.md
@@ -1,0 +1,78 @@
+# Security and Privacy Questionnaire
+
+This document answers the [W3C Security and Privacy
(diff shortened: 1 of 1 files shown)


Manishearth (approved):
Yeah, I'm not really sure if this is the best place for it, perhaps it should be a PR to the main questionnaire? But i'm fine with this too.

toji (approved):
Agreed that:

 - This seems like a weird place for this
 - I have no idea what a better place would be
 - It seems like overkill to need to do this deep of a dive for an API modification
 
That said, LGTM.

PR #1165, opened 2021-01-26
Add associated XRFrame to XRView, remove frame time

bialpio:
Other specs (depth sensing API, layers) rely or will rely on being able
to obtain XRFrame instance from an XRView. At a first glance, this
change should be safe to make, but please let me know if I am missing
something here and there was a reason not to put XRFrame on an XRView
other than "no-one needed it before".

One algorithm was also changed to compare XRFrame instances - this is
not required and we could keep comparing frame times there if people
prefer that.

Misc. change: fix one linking error for getNativeFramebufferScaleFactor.

***
<a href="https://pr-preview.s3.amazonaws.com/bialpio/webxr/pull/1165.html" title="Last updated on Jan 26, 2021, 9:56 PM UTC (c0addfc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1165/022c516...bialpio:c0addfc.html" title="Last updated on Jan 26, 2021, 9:56 PM UTC (c0addfc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1104,7 +1104,7 @@ When this method is invoked, the user agent MUST run the following steps:
       1. Let |xrview| be a new {{XRView}} object in the [=relevant realm=] of |session|.
       1. Initialize |xrview|'s [=XRView/underlying view=] to |view|.
       1. Initialize |xrview|'s {{XRView/eye}} to |view|'s [=view/eye=].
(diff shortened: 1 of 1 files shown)


bialpio:
@toji, please take a look.

Manishearth (approved):

toji:
This looks good, thanks!

PR #1182, opened 2021-03-12
Added support for angular and linear velocity

### LINKED ISSUE #619: Add Linear/Angular Acceleration/Velocity for XRInputSource

cabanier:
Closes #619

Also silences a couple of bikeshed warnings.

***
<a href="https://pr-preview.s3.amazonaws.com/cabanier/webxr/pull/1182.html" title="Last updated on Mar 13, 2021, 4:42 AM UTC (fb091a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1182/39aad3f...cabanier:fb091a2.html" title="Last updated on Mar 13, 2021, 4:42 AM UTC (fb091a2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1186,16 +1186,22 @@ To <dfn>populate the pose</dfn> of an {{XRSpace}} |space| in an {{XRSpace}} |bas
       : If |limit| is `false` and the tracking system provides a [=6DoF=] pose whose position is actively tracked or statically known for |space|'s pose relative to |baseSpace|:
       :: Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
       :: Set |transform|'s {{XRRigidTransform/position}} to the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
+      :: Set |pose|'s {{XRPose/linearVelocity}} to the linear velocity of |space|'s [=effective origin=] compared to |baseSpace|'s [=coordinate system=].
+      :: Set |pose|'s {{XRPose/angularVelocity}} to the angular velocity of |space|'s [=effective origin=] compared to |baseSpace|'s [=coordinate system=].
       :: Set |pose|'s {{XRPose/emulatedPosition}} to `false`.
 
       : Else if |limit| is `false` and the tracking system provides a [=3DoF=] pose or a [=6DoF=] pose whose position is neither actively tracked nor statically known for |space|'s pose relative to |baseSpace|:
       :: Set |transform|'s {{XRRigidTransform/orientation}} to the orientation of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=].
       :: Set |transform|'s {{XRRigidTransform/position}} to the tracking system's best estimate of the position of |space|'s [=effective origin=] in |baseSpace|'s [=coordinate system=]. This MAY include a computed offset such as a neck or arm model. If a position estimate is not available, the last known position MUST be used.
+      :: Set |pose|'s {{XRPose/linearVelocity}} to `null`.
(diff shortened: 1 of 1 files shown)


cabanier:
@Manishearth why can I not add reviewers?

index.bs
```diff
@@ -1689,12 +1695,19 @@ An {{XRPose}} describes a position and orientation in space relative to an {{XRS
...
+
   readonly attribute boolean emulatedPosition;
 };
 </pre>
 
 The <dfn attribute for="XRPose">transform</dfn> attribute describes the position and orientation relative to the base {{XRSpace}}.
 
+The <dfn attribute for="XRPose">linearVelocity</dfn> attribute describes the linear velocity in meters per second relative to the base {{XRSpace}}. If the user agent can't populate this, it's allowed to return <code>null</code>.
+
+The <dfn attribute for="XRPose">angularVelocity</dfn> attribute describes the angular velocity in meters per second relative to the base {{XRSpace}}. If the user agent can't populate this, it's allowed to return <code>null</code>.
```
toji, line 1709: Angular velocity isn't measured in meters per second. It seems like OpenXR reports this in radians per second, which seems fairly sensible.

toji (changes requested):
LGTM after one correction.

toji:
GitHub had a hangover this morning, which is why you couldn't add reviewers. I think it's generally better now.

cabanier:
> GitHub had a hangover this morning, which is why you couldn't add reviewers. I think it's generally better now.

I still can't add reviewers.

cabanier:
I was looking into a possible implementation and there are some spaces where velocity will never be reported.
For instance XRJointSpaces and XRTargetRaySpace with a XRTargetRayMode different than POINTING.

Should we add those exceptions to the spec, or leave it up to the UA?

toji:
I think it's best to leave that up to the UA. For example, I can see that OpenXR has an extension for [hand joint velocity](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrhandjointvelocityext3) and I suspect that the tracking method (ie: optical vs sensors) will make a big difference in what velocities are available. I wouldn't want to prematurely prevent future advances in tracking.

If we're concerned about developers needing to know what spaces can and can't report velocities, it would probably be best to report that as a property of the space (`hasLinearVelocity`/`hasAngularVelocity` booleans, maybe?) but I don't know that we can do that consistently and reliably on all platforms, and I believe even spaces that CAN report velocities may have moments where it lapses, such after the first frame or two of tracking (depending on the tracking method). As a result, developers will need to be able to handle `null` velocities no matter what.

We should probably just leave it as "If the user agent can't populate this, it's allowed to return `null`."

toji (approved):
Thanks for fixing the angular velocity definition, by the way!

cabanier:
@toji thanks for approving! Can you merge it for me? I'm not able to.

toji:
If you don't mind I'd like to make sure that @Manishearth has had a chance to look before I do? That should be Tues at the latest, since we have our editors sync then. Normally I wouldn't be too concerned but given this is a change to a largely "shipped" spec I feel a smidge of extra caution is warranted.

I doubt there's going to be any concerns, though.

cabanier:
I see. That sounds good! There's no hurry to get it in.

Manishearth (approved):
Seems good now.

One note: angular velocity can be obtained from the other attributes; and we tend towards not having APIs that expose information that can be calculated by the author, but I think it's fine in this case.

PR #1241, opened 2021-11-30
replaced list of specs in IWWG/IWCG to a link to list_spec page

himorin:
(wondered whether to keep current incomplete? list of specs, but considering maintainability, just replaced - as an idea at the first moment)

Replaced list of specs in README, with a link to list_spec page.


AdaRoseCannon:
That seems like a good way to ensure it is up-to-date 👍

PR #1242, opened 2021-12-01
Update spatial-tracking-explainer.md

song-fangzhen:
Fix the [Flow chart] image (line 432) that cannot be showed, which is due to the name suffix should be capitalized.


toji (approved):
Thank you!

PR #1249, opened 2022-01-27
[WIP] changes section

himorin:
changes section with github PR title and link

how about this format? (but seems no link is preferred,,,? like [woff2](https://www.w3.org/TR/WOFF2/#changes) or [css](https://www.w3.org/TR/css-conditional-4/#changes))

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2907,6 +2907,43 @@ To <dfn>resolve the requested features</dfn> given |requiredFeatures| and |optio
 
 </div>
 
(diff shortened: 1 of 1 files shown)


toji:
Seems reasonable to me, and if we're listing the commits made for a given range I see no reason why why shouldn't link them.

How are you generating these?

himorin:
by hand, one by one (w/ checking PR's targets)

toji (approved):

PR #1332, opened 2023-06-21
Clarify getPose behavior with visibile-blurred

toji:
The "poses may be reported" algorithm previously indicated that poses could only be reported if the visibility state was "visible", but that appears to be a mistake originating from the fact that "visible-blurred" was added later. The description of "visible-blurred" mentioned throttled head poses, so obviously the intention was to allow some poses in that state, just not input poses.

This change alters the "poses may be reported" algorithm to only suppress poses when the visiblity state is "hidden", and adds an explicit step to the "populate the pose" algorithm that prevents input poses from being returned if the visibility state is "visible-blurred"

Thanks @AdaRoseCannon for pointing this out!

***
<a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/pull/1332.html" title="Last updated on Jun 21, 2023, 4:03 PM UTC (4adf91e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1332/98e6946...4adf91e.html" title="Last updated on Jun 21, 2023, 4:03 PM UTC (4adf91e)">Diff</a>


AdaRoseCannon:
Thank you so much for putting this together!

cabanier (approved):
LGTM

Manishearth (approved):

PR #1369, opened 2024-04-09
Replace unnecessary use of FrozenArray in dictionaries

### LINKED ISSUE #1368: Use of FrozenArray<>s in dictionaries is unnecessary and bad
The spec has one "init" dictionary which accepts `FrozenArray<XRInputSource>`.

It's not clear what was intended here. But [per the Web IDL spec](https://webidl.spec.whatwg.org/#js-frozen-array), almost the exact same set of values are accepted compared to accepting the more conventional `sequence<XRInputSource>`.

(The difference is that certain proxies for arrays, which define throwing `getOwnPropertyDescriptor()` traps, would be rejected, because in theory the implementation is supposed to freeze the incoming array.)

We'd like to prohibit the usage of `FrozenArray<T>` as a dictionary member in Web IDL, per https://github.com/whatwg/webidl/issues/1399. It would be lovely if you could move to using `sequence<>`s.

toji:
Fixes #1368. 

Also fixes up a couple of minor markdown formatting issues.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2470,8 +2470,8 @@ interface XRInputSourcesChangeEvent : Event {
 
 dictionary XRInputSourcesChangeEventInit : EventInit {
   required XRSession session;
-  required FrozenArray&lt;XRInputSource&gt; added;
-  required FrozenArray&lt;XRInputSource&gt; removed;
+  required sequence&lt;XRInputSource&gt; added;
+  required sequence&lt;XRInputSource&gt; removed;
 
 };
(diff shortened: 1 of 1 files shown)


Manishearth (approved):

cabanier (approved):

toji:
Going to merge this now, and we can communicate that it happened as an FYI on the next call. This seems like clearly the right thing to do and I can't imagine there being any pushback against it.

PR #1376, opened 2024-07-17
Fix: incorrect type name `XRInputSourceChangeEvent` renamed `XRInputSourcesChangeEvent` as per spec

hybridherbst:
This is a bit inconsistent right now, fixing here since I assume the spec is up-to-date:
https://www.w3.org/TR/webxr/#xrinputsourceschangeevent-interface


toji (approved):
Thanks for catching and fixing the inconsistency!

PR #1404, opened 2025-03-24
Add support for pasthroughFullyObscured + small bikeshed fixes

cabanier:


index.bs
```diff
@@ -1042,6 +1046,10 @@ The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the dis
 
 Note: Typically when constructing a perspective projection matrix for rendering the developer specifies the viewing frustum and the near and far clip planes. When displaying to an [=immersive XR device=] the correct viewing frustum is determined by some combination of the optics, displays, and cameras being used. The near and far clip planes, however, may be modified by the application since the appropriate values depend on the type of content being rendered.
 
+The <dfn attribute for="XRRenderState">pasthroughFullyObscured</dfn> attribute is a hint to the {{XRSystem}} from the author to indicate that the pixels of the immersive experience completely cover the viewport.
```
toji, line 1049: Very minor nits here: The way this is stated makes it sound a bit like a guarantee that the viewport WILL be completely covered, but we don't want to imply any validation of that. I think a minor re-wording helps disambiguate that.

```suggestion
The <dfn attribute for="XRRenderState">pasthroughFullyObscured</dfn> attribute is a hint to the {{XRSystem}} from the author to indicate that they intend to completely cover the viewport with virtual content.
```

index.bs
```diff
@@ -1042,6 +1046,10 @@ The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the dis
...
+
+NOTE: the {{XRSystem}} MAY use this to temporarily disable passthrough to save system resources.
```
toji, line 1051: I think we should add more non-normative text here to explain that this flag does not guarantee that the user's environment is not visible, especially in the case of optically transparent displays.

toji (commented):
LGTM with a couple of tweaks.

I don't recall if we decided to definitely add this at the F2F or if we wanted to have more discussion once the proposed spec text was available?

index.bs
cabanier: I think the "MAY" already indicates this but I added some more prose.

index.bs
```diff
@@ -1046,9 +1046,9 @@ The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the dis
...
 
-The <dfn attribute for="XRRenderState">pasthroughFullyObscured</dfn> attribute is a hint to the {{XRSystem}} from the author to indicate that the pixels of the immersive experience completely cover the viewport.
+The <dfn attribute for="XRRenderState">pasthroughFullyObscured</dfn> attribute is a hint to the {{XRSystem}} from the author to indicate that they intend to completely cover the viewport with virtual content. The author SHOULD set this flag back to <code>false</code> once the viewport is no longer covered by opaque pixels.
 
-NOTE: the {{XRSystem}} MAY use this to temporarily disable passthrough to save system resources.
+NOTE: the {{XRSystem}} MAY use this a hint to temporarily disable passthrough. On devices with see-through optics, the user will continue to see their environment and this flag will have no effect.
```
toji, line 1051:
```suggestion
NOTE: the {{XRSystem}} MAY use this as a hint to temporarily disable passthrough. On devices with see-through optics, the user will continue to see their environment and this flag will have no effect.
```

toji (approved):
LGTM, though again I don't recall if the decision was to land this now or bring it back to the group. If the former feel free to merge.

cabanier:
> LGTM, though again I don't recall if the decision was to land this now or bring it back to the group. If the former feel free to merge.

I believe we had agreement. If not, I can revert.

PR #1405, opened 2025-03-24
fix workflow

cabanier:


toji:
Thank you!

cabanier:
Something is still wrong. The published editor's spec is from 2022

himorin:
/TR/ material is published using status overrited material https://labs.w3.org/echidna/api/status?id=55a9c3d4-7225-4df2-99ae-b515a1d11779, but github.io seems to be published using default written parameter - checked via htmldiff
https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2Fwebxr%2F&doc2=https%3A%2F%2Fimmersive-web.github.io%2Fwebxr%2F

from GHAction log, bikeshed runs twice one using configOverride parameter (which does not have favicon in material) and one without (which does have favicon in material), but pathinfo is not recorded - so it is unclear from only GHAction log that github.io or echidna is using which. (four files are commonly picked: `rigid_matrix.svg`, rotation_matrix.svg`, `spec-logo.png`, `translation_matrix.svg`)
/TR/ material does not contain favicon as link in head, but github.io one has.
[commit history of index.html in gh-pages branch](https://github.com/immersive-web/webxr/commits/gh-pages/index.html) has simple history all co-authored by github-actions, so no external interference found...

so, GHAction seems picking w/configoverride to echidna, but picking w/o configoverride to GH_PAGES_BRANCH??? (not sure whether there is some another configuration parameter or not.)

himorin:
@cabanier it seems override will not affect to artifacts to gh-pages, per https://github.com/w3c/spec-prod/pull/25
so, we may have several options

1. add configuration option`GH_PAGES_BUILD_OVERRIDE` also
2. set spec source configuration in main branch to ED without `Date` (nor `Deadline`?), and make github.io one as ED following the default way

I'll file an issue to,,, hrm,, administrivia repository.

PR #1408, opened 2025-04-16
Introduce XRViewGeometry mixin concept

alcooper91:
@toji suggested using a [mixin](https://webidl.spec.whatwg.org/#idl-interface-mixins) as the editorial concept we discussed in the latest IWWG call. I quickly tested this in Chrome and it doesn't seem to trigger any unexpected web-observable changes. 

However, when I run the spec through https://api.csswg.org/bikeshed/ after my changes, I see:
```
WARNING: Skipped generating some MDN panels, because the following IDs weren't present in the document. Use `Ignore MDN Failure` if this is expected.
  #dom-xrview-projectionmatrix
  #dom-xrview-transform
```
Not sure what if anything needs to be done to resolve this, but I welcome any insights

***
<a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/pull/1408.html" title="Last updated on Apr 17, 2025, 6:24 PM UTC (289445c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/immersive-web/webxr/1408/823070b...289445c.html" title="Last updated on Apr 17, 2025, 6:24 PM UTC (289445c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1042,7 +1042,7 @@ When an {{XRRenderState}} object is created for an {{XRSession}} |session|, the 
 
 The <dfn attribute for="XRRenderState">depthNear</dfn> attribute defines the distance, in meters, of the near clip plane from the [=viewer=]. The <dfn attribute for="XRRenderState">depthFar</dfn> attribute defines the distance, in meters, of the far clip plane from the [=viewer=].
 
-{{XRRenderState/depthNear}} and {{XRRenderState/depthFar}} are used in the computation of the {{XRViewCalibration/projectionMatrix}} of {{XRView}}s. When the {{XRViewCalibration/projectionMatrix}} is used during rendering, only geometry with a distance to the [=viewer=] that falls between {{XRRenderState/depthNear}} and {{XRRenderState/depthFar}} will be drawn. They also determine how the values of an {{XRWebGLLayer}} depth buffer are interpreted. {{XRRenderState/depthNear}} MAY be greater than {{XRRenderState/depthFar}}.
+{{XRRenderState/depthNear}} and {{XRRenderState/depthFar}} are used in the computation of the {{XRViewGeometry/projectionMatrix}} of {{XRView}}s. When the {{XRViewGeometry/projectionMatrix}} is used during rendering, only geometry with a distance to the [=viewer=] that falls between {{XRRenderState/depthNear}} and {{XRRenderState/depthFar}} will be drawn. They also determine how the values of an {{XRWebGLLayer}} depth buffer are interpreted. {{XRRenderState/depthNear}} MAY be greater than {{XRRenderState/depthFar}}.
 
 Note: Typically when constructing a perspective projection matrix for rendering the developer specifies the viewing frustum and the near and far clip planes. When displaying to an [=immersive XR device=] the correct viewing frustum is determined by some combination of the optics, displays, and cameras being used. The near and far clip planes, however, may be modified by the application since the appropriate values depend on the type of content being rendered.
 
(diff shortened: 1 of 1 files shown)


alcooper91:
Required for immersive-web/depth-sensing#59

alcooper91:
@toji PTAL (and LMK if you have a better name suggestion), and then I'll send to Rik for review.

toji:
`XRViewCalibration` isn't my favorite name, though it doesn't matter _too_ much since it'll never be developer-visible. I don't have anything I feel is a really strong replacement, but to throw out some possible alternatives:
 
 - XRViewBase
 - XRViewTransforms (Projection _is_ a transform, just not a RigidTransform. :shrug:)
 - XRViewMatrices (Transform has a matrix. :shrug:)
 - XRViewProjectionTransform
 - XRViewIntrinsics
 - XRAbstractView

index.bs
```diff
@@ -1496,18 +1497,55 @@ Note: Content generally should not provide a visualization of the {{boundsGeomet
...
 
-A [=view=] has an associated <dfn for="view">projection matrix</dfn> which is a [=matrix=] describing the projection to be used when rendering the [=view=], provided by the underlying XR device. The [=view/projection matrix=] MAY include transformations such as shearing that prevent the projection from being accurately described by a simple frustum.
+A [=view calibration=] has an associated <dfn for="view calibration">projection matrix</dfn> which is a [=matrix=] describing the projection to be used when rendering to the [=view calibration/containing object=] (or the inverse of the projection to use when reading from the object), provided by the underlying XR device. The [=view calibration/projection matrix=] MAY include transformations such as shearing that prevent the projection from being accurately described by a simple frustum.
```
toji, line 1515: "(or the inverse of the projection to use when reading from the object)"

Wait, when we return the projection for a depth view, is it already inverted? If so, it feels a little weird to label in `projectionMatrix`. I kinda wonder if we shouldn't just have a wholly separate `XRDepthView` type at that point that has an explicit `inverseProjectionMatrix`.

If, on the other hand, we return the `projectionMatrix` for the lens that captured the depth map and the developer is intended to invert it I think this approach is still appropriate but the phrasing of this sentence needs to be updated.

alcooper91:
I *think* XRViewIntrinsics is incomplete as I think technically the `transform` attribute is an *ex*trinsic (the combination of which seemingly often being referred to as a "calibration" was how I arrived at my current name).

Thinking "aloud" here, the prose about this object representing some piece of physical hardware, and this being a "subset" of data for that camera/sensor was kind of how I was trying to go through this. `XRViewBase` doesn't seem *terrible*; but it does make it feel more like a base class setup.

Then again, the other item we're attaching this to is something like a sensor, so maybe dropping "View" out of this name entirely would help clarify some of it. Really everything is kind of like a Camera, and that makes it unfortunate that XRCamera is already used :(.

I think from what you've proposed "XRViewTransforms" is maybe my favorite (or sticking with XRViewCalibration), but I don't have strong thoughts. It'd be interesting if we could remove "View" from the name.

index.bs
alcooper91: I was basing this off of my implementation to map between the XRView and "sampling" a point from the depth buffer, where one of these was inverted: https://source.chromium.org/chromium/chromium/src/+/main:device/vr/openxr/android/openxr_depth_sensor_android.cc;drc=a3a53e2b858715bb6f8ea1adbfda034f3fedc0de;l=184

I *think* the `projectionMatrix` works if you are going from screen space to camera space (which might be called "world space")

alcooper91: Essentially, I'm pretty sure it's the latter, we provide the matrix to map from the world to the screen space, and if you want to take e.g. a point from the screen and map it back to the world you need to invert the matrix.

Maybe instead of a parenthetical a non-normative note that the inverse of this matrix can be used to "read" values from the screen-space to the "real world"?

alcooper91:
Another name suggestion that I think I really like is `XRViewGeometry`

index.bs
toji: Proposed note verbiage SGTM!

toji:
I'm not against dropping "View" from the name. I guess that the awkward part is that in most cases these values will be handed off to the GPU as a projection and view matrix. While that view isn't exactly the same as WebXR's view concept they are related. As such it feels like it's difficult to get away from the term. If you wanted to, though, I'd see something like `XRProjectionTransform` to be the most straightforward. (Hm... I don't mind that, actually.)

`XRViewGeometry` would be fine? I think it describes the projection matrix more so than the transform, but that's OK. I don't see it as strictly better than something like `XRViewTransforms` but I do have a mild preference for it over `XRViewCalibration`.

alcooper91:
Going with XRViewGeometry for now. I don't think any of the definitions or embedded terms need changing, though I'm maybe less sure of that with the "containing object" definition.

@cabanier PTAL, and let me know if you have thoughts on a better name for the mixin. Note that this name isn't exposed to developers at all; and is purely an in-spec concept.

toji (approved):

cabanier:
> @cabanier PTAL, and let me know if you have thoughts on a better name for the mixin. Note that this name isn't exposed to developers at all; and is purely an in-spec concept.

LGTM, since it's not exposed, we can always change it later.

cabanier (approved):

alcooper91:
My hesitation with something like "ProjectionTransform" was that it felt a bit too much like we were just appending the names of the attributes, rather than coming up with a name that represented what it was meant to be, which could lead to less of a standard name in the future.

We can certainly still change this at any time in the future since it *is* only a spec concept (though it'll have a correlation in code); but the easiest time is certainly now before it's implemented anywhere. Would it be worth chatting about next week?

PR #1443, opened 2026-08-27
turned privacy-security-explainer into bikeshed

himorin:
related to administrivia issue https://github.com/immersive-web/administrivia/issues/249

Turning privacy-security-explainer into bikeshed source with minimum modification (incl. anchor changes).
temporary compiled draft at https://immersive-web.w3c.himor.in/privacy-security-explainer.html

ToDo:

- [x] add editor (co-chairs? core spec editor?)
- [x] add abstract (could be from leading text in section 1.?)
- [ ] define shortname (is current ok?)


simoneonofri (approved):
LGTM to have the document.

I don't know if you want a review of the content now, during chartering, as I think this should flow in the normal review cycle :)

toji (approved):

himorin:
Added @toji as Editor, and moved the first paragraph (overview-ish text in original markdown) into Abstract.
so, ready to go. (we shall define shortname before asking transition into Note Draft...)

himorin:
> I don't know if you want a review of the content now, during chartering, as I think this should flow in the normal review cycle :)

at some point, we (IWWG) shall consult with you for document text as well as coverage, but we could do during next security review of WebXR Core or other modules, I suppose..
also WG may need to collect text from modules (e.g. [analysis for questionnaire](https://github.com/immersive-web/lighting-estimation/blob/main/security-privacy-questionnaire.md)), but I haven't checked all yet..

PR #1446, opened 2026-09-09
cleaning up support files

himorin:
updated GHAction and removing old TravisCI files

see https://github.com/immersive-web/administrivia/issues/242


toji (approved):
LGTM

